PR #5, opened 2017-02-09
[teamshow]: Add teamsohw script to dump all team devices

stcheng:


bin/teamshow
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/python
+
+"""
+    Script to show LAG and LAG member status in a summary view
+    Example of the output:
```
lguohan: there is no output, suggest to remove.

bin/teamshow
```diff
@@ -0,0 +1,91 @@
...
+            result = re.match(re_ip, line)
+            if result is not None:
+                self.teams.append(result.group(1))
+
+    def get_teamdctl(self):
+        """
+            Command: 'teamdctl <teamdevname> state dump'
+        """
+        for team in self.teams:
+            teamdctl_cmd = 'teamdctl ' + team + ' state dump'
```
lguohan: is there command available outside the teamd docker?

bin/teamshow
```diff
@@ -0,0 +1,91 @@
...
+        output = []
+        for team in self.summary:
+            output.append([team, self.summary[team]['protocol'], self.summary[team]['ports']])
+        print tabulate(output, header)
+
+def main():
+    try:
+        team = Teamshow()
+        team.get_ip()
+        team.parse_ip()
```
lguohan: instead of ip link, should we get this from the minigraph? what if teamd failed to create a port channel nedev, then ip link show will not show it.

bin/teamshow
stcheng: there're some modificaions from the user. i'll update this output.

bin/teamshow
stcheng: yes. teamdctl has a mock command outside the teamd docker.

bin/teamshow
stcheng: i'll update this to use minigraph.

stcheng: in order to use minigraph, i'll also need this https://github.com/Azure/sonic-utilities/pull/6 to get the alias_map.json file.

stcheng:
Updated

bin/teamshow
```diff
@@ -0,0 +1,112 @@
...
+        for child in tree.findall("{0}DpgDec/{0}DeviceDataPlaneInfo/{0}PortChannelInterfaces/{0}PortChannel".format(ns)):
+            for x in child:
+                if x.tag == "{0}Name".format(ns):
+                    # Skip the 'PortChannel' prefix and store the port channel ID
+                    self.teams.append(x.text[11:])
+
+    def get_alias_map(self):
+        """
+        """
+        json_file = open("/etc/ssw/" + self.hwsku + "/alias_map.json").read()
```
jleveque: alias_map.json has been deprecated; aliases now reside in port_config.ini.

bin/teamshow
```diff
@@ -0,0 +1,112 @@
...
+class Teamshow(object):
+    def __init__(self):
+        self.hwsku = None
+        self.alias_map = {}
+        self.teams = []
+        self.teamsraw = {}
+        self.summary = {}
+        self.err = None
+
+    def get_minigraph(self):
```
lguohan: there is better way to get the port channel names from minigraph.

https://github.com/Azure/sonic-buildimage/blob/master/dockers/docker-teamd/config.sh

bin/teamshow
jleveque: Taoyu just added a new flag to sonic-cfggen in order to obtain the alias map in JSON format. You can just use this line to output a temporary file and reference that:

`sonic-cfggen -m /etc/sonic/minigraph.xml -s > /my/temp/location/alias_map.json`

stcheng:
for the output of the command, do we want to display the 'sonic' name e.g. EthernetX, or the alias name e.g. fortyGig0/X?

PR #6, opened 2017-02-10
[platform-detect]: Add script to print out HwSku

stcheng:


lguohan:
platform detect is probably not needed, why do we still need this?

stcheng:
It is deprecated. I'll remove it.

PR #8, opened 2017-02-23
Add support for Dell Z9100

nirenjan:


lguohan:
we are consolidating all the platform config files, please check https://github.com/Azure/sonic-buildimage/pull/316? Please submit a new one once we have merged that.

lguohan:
portmap.ini is no longer needed, please change to port_config.ini

ssw/ACS-Z9100/portmap.ini
```diff
@@ -0,0 +1,32 @@
+1=1,1
```
lguohan: do we need this portmap.ini? where is it used?

nirenjan: This was needed during our earlier testing of sfputil, however, I believe that with the changes in Azure/sonic-buildimage#316, this can be removed.

PR #12, opened 2017-02-28
[sonic_cli]: Fix bash completion for 'show' command

jleveque:
Fix bash completion for 'show' command


data/etc/bash.bashrc
```diff
@@ -0,0 +1,57 @@
+# System-wide .bashrc file for interactive bash(1) shells.
```
lguohan: this seems to be the same as the one in the bash package. Can you remove it?

jleveque: It's different. The bash-completion section is uncommented.

jleveque:
Removed bash.bashrc so there is no need to install by passing the non-standard "--force-overwrite" flag to dpkg. bash.bashrc file will instead by copied into place in sonic-buildimage.

PR #14, opened 2017-03-03
[teamshow]: Add teamshow script to display the LAG status

sihuihan88:

### DIFF
--- a/scripts/teamshow
+++ b/scripts/teamshow
@@ -80,8 +80,8 @@ class Teamshow(object):
         print "Flags: A - active, I - inactive, N/A - Not Available, S - selected, D - deselected"
         header = ['No.', 'Team Dev', 'Protocol', 'Ports']
         output = []
-        for team in self.summary:
-            output.append([team, 'PortChannel'+team, self.summary[team]['protocol'], self.summary[team]['ports']])
+        for team_id in self.summary:
+            output.append([team_id, 'PortChannel'+team_id, self.summary[team_id]['protocol'], self.summary[team_id]['ports']])
         print tabulate(output, header)
 
 def main():
--- a/scripts/teamshow
+++ b/scripts/teamshow
@@ -1,0 +1,98 @@
+#!/usr/bin/python
+
+"""
+    Script to show LAG and LAG member status in a summary view
+    Example of the output:
+    acsadmin@CO4SCH03010AALF:~$ teamshow
+    Flags: A - active, I - inactive, N/A - Not Available, S - selected, D - deselected
+     No.  Team Dev       Protocol    Ports
+    -----  -------------  ----------  ---------------------------
+       24  PortChannel24  LACP(A)     Ethernet28(S) Ethernet24(S)
+       48  PortChannel48  LACP(A)     Ethernet52(S) Ethernet48(S)
+       16  PortChannel16  LACP(A)     Ethernet20(S) Ethernet16(S)
+       32  PortChannel32  LACP(A)     Ethernet32(S) Ethernet36(S)
+       56  PortChannel56  LACP(A)     Ethernet60(S) Ethernet56(S)
+       40  PortChannel40  LACP(A)     Ethernet44(S) Ethernet40(S)
+        0  PortChannel0   LACP(A)     Ethernet0(S) Ethernet4(S)
+        8  PortChannel8   LACP(A)     Ethernet8(S) Ethernet12(S)
+"""
+
+import json
+import re
+import subprocess
+import sys
+from tabulate import tabulate
+from minigraph import parse_xml
+
+class Teamshow(object):
+    def __init__(self):
+        self.teams = []
+        self.teamsraw = {}
+        self.summary = {}
+        self.err = None
+
+    def get_minigraph(self):
+        """
+        """
+        minigraph_path = "/etc/ssw/minigraph.xml"
+
+        # Store port channel IDs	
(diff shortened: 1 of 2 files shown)


scripts/teamshow
```diff
@@ -0,0 +1,98 @@
...
+
+    def get_minigraph(self):
+        """
+        """
+        minigraph_path = "/etc/ssw/minigraph.xml"
+
+        # Store port channel IDs	
+        for port in parse_xml(minigraph_path)['minigraph_portchannel_interfaces'].keys():
+            # Skip the 'PortChannel' prefix and store the port channel ID
+            self.teams.append(port[11:])
```
lguohan: what is port[11:]?

scripts/teamshow
```diff
@@ -0,0 +1,98 @@
...
+    def __init__(self):
+        self.teams = []
+        self.teamsraw = {}
+        self.summary = {}
+        self.err = None
+
+    def get_minigraph(self):
+        """
+        """
+        minigraph_path = "/etc/ssw/minigraph.xml"
```
lguohan: this should be /etc/sonic/minigraph.xml in the one image

scripts/teamshow
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/python
+
+"""
+    Script to show LAG and LAG member status in a summary view
+    Example of the output:
+    acsadmin@CO4SCH03010AALF:~$ teamshow
```
lguohan: CO4SCH03010AALF -> sonic

scripts/teamshow
lguohan: I do not quite understand the logic here, in line42, you extract the team id, and on line 51, you contruct the team name using "PortChannel" + team. I think the minigraph portchannel is same as the team name, so we should use the minigraph name directly at line 51.

lguohan: teams = parse_xml(minigraph_path)['minigraph_portchannel_interfaces'].keys()

scripts/teamshow
```diff
@@ -0,0 +1,98 @@
...
+            # Skip the 'PortChannel' prefix and store the port channel ID
+            self.teams.append(port[11:])
+
+
+		
+    def get_teamdctl(self):
+        """
+            Command: 'teamdctl <teamdevname> state dump'
+        """
+        for team in self.teams:
```
lguohan: here the team should be the portchannel name

scripts/teamshow
```diff
@@ -0,0 +1,98 @@
...
+                info['ports'] += '(S)' if json_info['ports'][port]['runner']['selected'] else '(D)'
+                info['ports'] += ' '
+            self.summary[team] = info
+
+    def display_summary(self):
+        print "Flags: A - active, I - inactive, N/A - Not Available, S - selected, D - deselected"
+        header = ['No.', 'Team Dev', 'Protocol', 'Ports']
+        output = []
+        for team in self.summary:
+            output.append([team, 'PortChannel'+team, self.summary[team]['protocol'], self.summary[team]['ports']])
```
lguohan: since we need to output team id here, can you add a function get_team_id() to extract the id from the portchannel name?

scripts/teamshow
```diff
@@ -0,0 +1,98 @@
...
+            self.teams.append(port[11:])
+
+
+		
+    def get_teamdctl(self):
+        """
+            Command: 'teamdctl <teamdevname> state dump'
+        """
+        for team in self.teams:
+            teamdctl_cmd = 'teamdctl PortChannel' + team + ' state dump'
```
lguohan: what if PortChannelX is not configured? What will be the output?

root@str-s6000-acs-7:/# teamdctl PortChannel2 state dump
Device "PortChannel2" does not exist

In this case follwing should be output

2  PortChannel2   LACP(N/A)

scripts/teamshow
```diff
@@ -0,0 +1,98 @@
...
+from minigraph import parse_xml
+
+class Teamshow(object):
+    def __init__(self):
+        self.teams = []
+        self.teamsraw = {}
+        self.summary = {}
+        self.err = None
+
+    def get_minigraph(self):
```
lguohan: rename -> get_portchannel_names

scripts/teamshow
sihuihan88: Here I used the same logic as the previous pr to extract the team ID and only change the way to parse minigraph. I will change the logic here.

scripts/teamshow
```diff
@@ -0,0 +1,99 @@
...
+                info['ports'] += port
+                info['ports'] += '(S)' if json_info['ports'][port]['runner']['selected'] else '(D)'
+                info['ports'] += ' '
+            self.summary[team_id] = info
+
+    def display_summary(self):
+        print "Flags: A - active, I - inactive, N/A - Not Available, S - selected, D - deselected"
+        header = ['No.', 'Team Dev', 'Protocol', 'Ports']
+        output = []
+        for team in self.summary:
```
lguohan: rename team to team_id

scripts/teamshow
```diff
@@ -0,0 +1,99 @@
...
+
+class Teamshow(object):
+    def __init__(self):
+        self.teams = []
+        self.teamsraw = {}
+        self.summary = {}
+        self.err = None
+
+    def get_portchannel_names(self):
+        """
```
stcheng: add some descriptions here? and also below some other functions?

PR #18, opened 2017-03-23
[aclshow]: Add aclshow utility to view ACL info and counters

andriymoroz-mlnx:

PR #19, opened 2017-03-23
[Portstat]: Read counters from database

sihuihan88:


scripts/portstat
```diff
@@ -16,304 +16,200 @@ import re
...
+    'SAI_PORT_STAT_IF_IN_OCTETS': 8,
+    'SAI_PORT_STAT_IF_OUT_OCTETS': 9
 }
 
+COUNTER_TABLE_PREFIX = "COUNTERS:"
+
+def natural_sort(l):
+    convert = lambda text: int(text) if text.isdigit() else text.lower()
+    alphanum_key = lambda key: [ convert(c) for c in re.split('([0-9]+)', key) ]
+    return sorted(l, key = alphanum_key)
```
jleveque: This is not very portable (I know because I wrote a similar function in minigraph_facts.py. Might be better to install the natsort library as a dependency and use natsort.natsorted(), which is what I wound up doing instead.

scripts/sonic_py.py
```diff
@@ -0,0 +1,36 @@
...
+"""
+
+import redis
+
+
+REDIS_UNIX_SOCKET = "/var/run/redis/redis.sock"
+CNT_DB = 2
+COUNTER_PORT_NAME_MAP = "COUNTERS_PORT_NAME_MAP"
+
+class sonic_py(object):
```
jleveque: This module name should be more descriptive. Maybe sonic_db.py?

pavel-shirshov: https://github.com/Azure/sonic-py-swsssdk - Python utility library for sonic/switch state service database access. Should we use them?

scripts/portstat
pavel-shirshov: Do we need to make it portable?
All we're sorting here is name of the interface which is in our case most time 'EthernetX'

jleveque: If we move up to Python 3 we might have issues.

scripts/sonic_py.py
jleveque: Thanks, Pavel. I was unaware that library existed. if it works, we should use it and not reinvent the wheel. Sihui, please take a look.

stcheng: I think we should use that. It seems that the library is a little bit out of date, but we could still pick it up. For the other tool, the 'aclshow' in https://github.com/Azure/sonic-utilities/pull/18, I think we also need to use that.

sihuihan88: Thanks for all the info. I will take a look and try to use this library

scripts/portstat
paulnice: I don't like natsort library because:
1. debian has outdated version of the library
2. this library break an api

Also I think this function is so small so it's ok to have it in our code.

pavel-shirshov: I don't like natsort library because:
1. debian has outdated version of the library
2. this library break an api
Also I think this function is so small so it's ok to have it in our code.

jleveque: @pavel-shirshov: I agree with #1. Could you expound on #2? Which API does it break?

pavel-shirshov: natsort() in debian repo vs natsorted() in pypi.
See:
https://github.com/Azure/sonic-mgmt/commit/957220e0631ccb9d1e1258742ab8878f84f2c225

jleveque: Gotcha . You're just referring the change in their own API.

sihuihan88: If the lib keeps changing API name, I am not sure whether it's a good idea to use it in our project.

jleveque: As far as I can tell, the API changed only once, when the package changed its name from `naturalsort` to `natsort.`

sihuihan88: Gotcha. Great I will use this lib then :)

setup.py
```diff
@@ -34,6 +34,7 @@
     install_requires=[
         'click',
         'click-default-group',
+        'natsort',
         'tabulate'
```
jleveque: We should probably list sswsdk as a dependency here, also.

scripts/portstat
```diff
@@ -31,289 +30,180 @@ NStats = namedtuple("NStats", "rx_ok, rx_err, rx_drop, rx_ovr, tx_ok,\
...
-    "xe28": "Ethernet112",
-    "xe29": "Ethernet116",
-    "xe30": "Ethernet120",
-    "xe31": "Ethernet124"
+counter_bucket_dict = {
+    'SAI_PORT_STAT_IF_IN_UCAST_PKTS': 0,
+    'SAI_PORT_STAT_IF_IN_NON_UCAST_PKTS': 0,
+    'SAI_PORT_STAT_IF_IN_ERRORS': 1,
+    'SAI_PORT_STAT_IF_IN_DISCARDS': 2,
+    'SAI_PORT_STAT_ETHER_TX_OVERSIZE_PKTS': 3,
```
lguohan: -> RX_OVERSIZE_PKTS

scripts/portstat
```diff
@@ -31,289 +30,180 @@ NStats = namedtuple("NStats", "rx_ok, rx_err, rx_drop, rx_ovr, tx_ok,\
...
+    'SAI_PORT_STAT_IF_OUT_UCAST_PKTS': 4,
+    'SAI_PORT_STAT_IF_OUT_NON_UCAST_PKTS': 4,
+    'SAI_PORT_STAT_IF_OUT_ERRORS': 5,
+    'SAI_PORT_STAT_IF_OUT_DISCARDS': 6,
+    'SAI_PORT_STAT_ETHER_RX_OVERSIZE_PKTS': 7,
```
lguohan: -> TX_OVERSIZE_PKTS

scripts/portstat
```diff
@@ -31,289 +30,180 @@ NStats = namedtuple("NStats", "rx_ok, rx_err, rx_drop, rx_ovr, tx_ok,\
...
+            Get the counters info from database.
+        """
+        def get_counters(table_id):
+            """
+                Get the counters from specific table.
+            """
+            fields = [0,0,0,0,0,0,0,0,0,0]
+            for counter_name, pos in counter_bucket_dict.iteritems():
+                full_table_id = COUNTER_TABLE_PREFIX + table_id
+                fields[pos] += int(self.db.get(self.db.COUNTERS_DB, full_table_id, counter_name))
```
lguohan: can we use hmget to get all interested fields for a key in one shot?

https://redis.io/commands/hmget

scripts/portstat
```diff
@@ -31,289 +30,180 @@ NStats = namedtuple("NStats", "rx_ok, rx_err, rx_drop, rx_ovr, tx_ok,\
...
+counter_bucket_dict = {
+    'SAI_PORT_STAT_IF_IN_UCAST_PKTS': 0,
+    'SAI_PORT_STAT_IF_IN_NON_UCAST_PKTS': 0,
+    'SAI_PORT_STAT_IF_IN_ERRORS': 1,
+    'SAI_PORT_STAT_IF_IN_DISCARDS': 2,
+    'SAI_PORT_STAT_ETHER_TX_OVERSIZE_PKTS': 3,
+    'SAI_PORT_STAT_IF_OUT_UCAST_PKTS': 4,
+    'SAI_PORT_STAT_IF_OUT_NON_UCAST_PKTS': 4,
+    'SAI_PORT_STAT_IF_OUT_ERRORS': 5,
+    'SAI_PORT_STAT_IF_OUT_DISCARDS': 6,
```
lguohan: do we know if OUT_DISCARDS map to all the counters we used to get from bshell? TDBGC0, 3, 5, 8 and the UCQ_DROP_PKT counters?

setup.py
jleveque: Unfortunately, we can't list sswsdk as a dependency here. Since we compile sonic-utilities into a .deb package, stdeb converts all Python dependencies here into Debian dependencies, thus it looks for installed .deb packages named "python-<python_package_name>. However, since we install sswsdk directly as a Python Wheel, it does not show up as an installed .deb package, so the  sonic-utilities install will fail. We may want to consider unifying our package creation process.

scripts/portstat
stcheng: @lguohan ask broadcom about this?

scripts/portstat
sihuihan88: We are using swssdk lib which hasn't implemented the method similar as hmget in redis-py.  If we want to use the hmget, we need to update the swssdk first.

scripts/portstat
```diff
@@ -31,289 +30,180 @@ NStats = namedtuple("NStats", "rx_ok, rx_err, rx_drop, rx_ovr, tx_ok,\
...
-
-def ns_diff(newstr, oldstr):
-    new, old = int(newstr), int(oldstr)
-    return '{:,}'.format(new - old)
+COUNTER_TABLE_PREFIX = "COUNTERS:"
+COUNTERS_PORT_NAME_MAP = "COUNTERS_PORT_NAME_MAP"
+
+class Portstat(object):
+    def __init__(self):
+        self.db = sswsdk.SonicV2Connector()
```
stcheng: https://github.com/Azure/sonic-py-swsssdk/pull/4 once this is merged. be sure to update all the related names.

PR #21, opened 2017-03-28
[teamshow]: Update teamshow according to the update of minigraph.py

stcheng:
Signed-off-by: Shuotian Cheng <shuche@microsoft.com>


sihuihan88:
It might be better that we call config-gen directly to get the minigraph info in case minigraph.py keeps changing.

stcheng:
@sihuihan88 i would vote to include minigraph.py library here. If we're using sonic-cfggen, this change is still needed as the interface is different.

PR #24, opened 2017-03-29
[scripts]: Update sonic-cfggen path to /usr/local/bin

jleveque:
- Also unrelated update of .gitignore


stcheng:
I think in the future it is better not to scatter around all the files around many places. Good for us to maintain.

PR #25, opened 2017-03-31
Add fast-reboot

pavel-shirshov:
Add fast-reboot utility to reboot SONiC box using kexec-tools.
Add fast-reboot to kernel parameters to indicate that kernel was loaded after fast reboot

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -8,7 +8,7 @@ then
 fi
 
 # Unload previous kernel if any loaded
-if [ "x`cat /sys/kernel/kexec_loaded`y" = "x1y" ]
+if [ $(cat /sys/kernel/kexec_loaded) -eq 1 ]
 then
   /sbin/kexec -u
 fi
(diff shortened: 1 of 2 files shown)


scripts/fast-reboot
```diff
@@ -0,0 +1,45 @@
...
+# Stop docker container engine. Otherwise will have broken docker storage
+systemctl stop docker.service
+
+# Wait until all buffers synced with disk
+sync
+sleep 1
+sync
+sleep 1
+sync
+sleep 1
```
jleveque: Is there not a more foolproof method? Possibly waiting for the "Dirty" value in /proc/meminfo to reach 0?

pavel-shirshov: I think the best method is sleeping here. Because 'Dirty pages' parameter indicates only about the Linux page cache usage. But I'd like to be sure the HDD writes all data back to disk from it's own caches. So it's better just to wait.

P.S.: Another method to sync is to remount the partition.

jleveque: Understood. Thanks for your explanation.

PR #28, opened 2017-04-05
[CLI]: Fix broken commands

jleveque:
Fixed the following commands which were broken:
- `show version`
- `show platform`
- `show lldp table`
- `show techsupport`

Also some other cleanup.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,6 +15,7 @@ setup(
         'sonic_cli': ['aliases.ini']
     },
     scripts=[
+        'scripts/aclshow',
         'scripts/boot_part',
         'scripts/coredump-compress',
         'scripts/decode-syseeprom',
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -15,15 +15,16 @@
...
         'scripts/boot_part',
         'scripts/coredump-compress',
         'scripts/decode-syseeprom',
         'scripts/fast-reboot',
         'scripts/generate_dump',
+        'scripts/lldpshow',
         'scripts/portstat',
         'scripts/sfputil',
         'scripts/teamshow', 
-        'scripts/aclshow',
```
stcheng: is it okay to have a comma at the last entry?

jleveque: Yes, setuptools allows commas after the last entry. It was news to me when I started on the project, too. I'm not the biggest fan of it, but it makes adding and removing entries easier, so I've stuck with the convention.

PR #30, opened 2017-04-08
[Teamshow]: get platform info before parsing minigraph

sihuihan88:


stcheng:
hold this pull request before https://github.com/Azure/sonic-buildimage/pull/489 is approved

sihuihan88:
Since we plan to add port channel status from database in teamshow, I think it also makes senses that we get the portchannel list from database directly at the same time. This helps us get rid of parsing the minigraph logic and keep the logic simple. 
I would prefer to obtain portchannel list and status from database directly rather than parsing minigraph.

stcheng:
the reason is that it is possible that the port channels are probably not up or initialized due to some unknown reasons/bugs. that is why we prefer to load all the information from minigraph so that to make sure that the desired state could be shown.

sihuihan88:
Just curious, where the database gets the port channel info? It also shows some port channel are down. I was thinking database will check minigraph or somewhere first to get the port channel list and then determine the status of each.

stcheng:
the database get the port channel information when the teamsyncd gets the notification about the initialization of team instances in docker-teamd, and then the swss gets the message from teamsyncd. thus there will be several steps before the database finally gets the information.

in this case, either way makes logical sense, it just depends on the requirements of this utility by users. if users want to know the status of the port channels defined in the minigraph, then we need to read the minigraph. if the user only cares about the current status of the team instances on the box, then we will only need to read the database.

thus it is the users of this utility that decide which approach to choose.

sihuihan88:
Got it. I understand it much better now. Thanks!

PR #31, opened 2017-04-11
[Teamshow]: Add port channel status to the teamshow result

sihuihan88:

### DIFF
--- a/scripts/teamshow
+++ b/scripts/teamshow
@@ -4,27 +4,33 @@
     Script to show LAG and LAG member status in a summary view
     Example of the output:
     acsadmin@sonic:~$ teamshow
-    Flags: A - active, I - inactive, N/A - Not Available, S - selected, D - deselected
+    Flags: A - active, I - inactive, Up - up, Dw - down, N/A - Not Available, S - selected, D - deselected
      No.  Team Dev       Protocol    Ports
     -----  -------------  ----------  ---------------------------
-       24  PortChannel24  LACP(A)     Ethernet28(S) Ethernet24(S)
-       48  PortChannel48  LACP(A)     Ethernet52(S) Ethernet48(S)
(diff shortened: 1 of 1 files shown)


scripts/teamshow
```diff
@@ -4,34 +4,43 @@
...
-        0  PortChannel0   LACP(A)     Ethernet0(S) Ethernet4(S)
-        8  PortChannel8   LACP(A)     Ethernet8(S) Ethernet12(S)
+        0  PortChannel0   LACP(A)(Up)     Ethernet0(D) Ethernet4(S)
+        8  PortChannel8   LACP(A)(Up)     Ethernet8(S) Ethernet12(S)
+       16  PortChannel16  LACP(A)(Up)     Ethernet20(S) Ethernet16(S)
+       24  PortChannel24  LACP(A)(Dw)     Ethernet28(S) Ethernet24(S)
+       32  PortChannel32  LACP(A)(Up)     Ethernet32(S) Ethernet36(S)
+       40  PortChannel40  LACP(A)(Dw)     Ethernet44(S) Ethernet40(S)
+       48  PortChannel48  LACP(A)(Up)     Ethernet52(S) Ethernet48(S)
+       56  PortChannel56  LACP(A)(Dw)     Ethernet60(S) Ethernet56(S)
```
stcheng: i think you could combine the status together? similar to other commercial switch output format:
LACP(AU) or LACP(AD)? I am open to other formats.

sihuihan88: Since there are cases where the up and down info are not available, it will show N/A. I was thinking (AN/A) is somewhat misleading. If you feel comfortable with cases such as (AN/A), I could combine the state. Another thing is that "D" is already used as short for "Deselected". So I used "Up" and "Dw" here.

stcheng: agree with your idea. i will approve it.

Issue #27, opened 2017-04-05
[CLI]: 'Default' commands prevent autocomplete from displaying subcommands


jleveque:
For example, typing `show interfaces` and pressing Tab will not display available subcommands `counters` or `portchannel`.

When redesigning the CLI command set, we should eliminate 'default' commands.

jleveque:
Newer versions of Click support this concept natively (see `default_if_no_args`), so DefaultGroup is no longer necessary.

Look into inheriting `click.Group` in place of `DefaultGroup` everywhere. Then we can completely remove dependencies on click_default_group and no longer install that package.

jleveque:
An example of using `invoke_without_command` and accepting parameters can be found here: https://github.com/pallets/click/issues/430#issuecomment-282015177. I have not verified that it works, but we could tru to utilize this if we want. Personally, I prefer to not allow groups to be invoked without subcommands, as it makes the behavior vary between different groups and isn't intuitive to the user.

PR #49, opened 2017-05-11
[show]: Remove default commands from groups as they are confusing to users

jleveque:
Default commands in a group prevent the use of autocomplete, which can prevent a user from knowing all of the available commands within that group.

See issue https://github.com/Azure/sonic-utilities/issues/27

TODO: If we agree not to use "default groups" in the future (I vote "no") we can also remove the installation and importing of "click_default_group" altogether.


jleveque:
TODO: If we decide not to use "default groups" we can also remove the installation and importing of "click_default_group."

PR #53, opened 2017-05-16
Enhance 'show' command

jleveque:
- Add 'show interfaces sfp' subcommand to wrap sfputil
 - Remove 'sfp' arg from 'show interfaces' command
 - Add '-p/--period' argument and '-c/--clear' flag to 'show interfaces
   counters'
 - Add missing docstrings for Click
- Add sudo to all 'vtysh', 'docker', 'lldpctl', 'lldpshow' and 'sensors' calls

### DIFF
--- a/show/main.py
+++ b/show/main.py
@@ -331,7 +331,7 @@ def version():
     click.echo(p.stdout.read())
 
     click.echo("Docker images:")
-    command = 'docker images --format "table {{.Repository}}\\t{{.Tag}}\\t{{.ID}}\\t{{.Size}}"'
+    command = 'sudo docker images --format "table {{.Repository}}\\t{{.Tag}}\\t{{.ID}}\\t{{.Size}}"'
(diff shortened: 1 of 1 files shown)


show/main.py
```diff
@@ -156,23 +150,46 @@ def summary(interfacename, sfp):
 
 # 'counters' subcommand
 @interfaces.command()
-def counters():
+@click.option('-p', '--period')
+@click.option('-c', '--clear', is_flag=True)
+def counters(period, clear):
```
sihuihan88: Can you also provide a -a --printall option to print all the information?

jleveque: Good catch, thanks! I meant to do that initially, but forgot.

sihuihan88: Shall we also provide -a as an argument option as -p and -c such that we don't have to print all the info by default?

jleveque: Done.

PR #57, opened 2017-05-23
[Fast-reboot]: Keep fdb and arp tables state between fast reboots

pavel-shirshov:
1. Save fdb and arp entries inside of json files to allow consume them by swssconfig later
2. Explicitly kill -9 bgpd process to start graceful restart BGP feature.

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -30,6 +30,16 @@ esac
 # Load kernel into memory
 /sbin/kexec -l "$KERNEL_IMAGE" --initrd="$INITRD" --append="$BOOT_OPTIONS"
 
+# Save ARP and FDB table to files
+/usr/bin/fast-reboot-dump.py
+docker cp /tmp/fdb.json swss:/
+docker cp /tmp/arp.json swss:/
+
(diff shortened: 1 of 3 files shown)


scripts/fast-reboot
```diff
@@ -30,6 +30,16 @@ esac
...
 
+# Save ARP and FDB table to files
+/usr/bin/fast-reboot-dump.py
+docker cp /tmp/fdb.json swss:/
+docker cp /tmp/arp.json swss:/
+
+# Kill bgpd to enable graceful restart of BGP
+docker exec -ti bgp killall -9 watchquagga
+docker exec -ti bgp killall -9 zebra
+docker exec -ti bgp killall -9 bgpd
```
paulnice: Yes. When we're doing docker stop bgp, quagga is stopped gracefully. This means that bgpd sends bgp notifications to all its neighbors about its stopping intentions. This will prevent the bgpd neighbors to start bgp graceful restart helper mode.
When we're killing quagga, OS just closes all sockets, and all bgp neighbors start bgp graceful restart helper mode. Something like this. You can read it further here: https://tools.ietf.org/html/rfc4724

PR #60, opened 2017-06-05
[show]: Fix 'broken pipe' errors when piping output of 'show' to other commands

jleveque:


show/main.py
```diff
@@ -81,14 +83,23 @@ def get_command(self, ctx, cmd_name):
...
+            click.echo_via_pager(p.stdout.read())
+        else:
+            click.echo(p.stdout.read())
+    except IOError as e:
+        # In our version of Click, click.echo() and click.echo_via_pager() do not properly handle
+        # SIGPIPE, and if a pipe is broken before all output is processed (e.g., pipe output to 'head'),
+        # it will result in a stack trace. This is apparently fixed upstream, but for now, we silently
+        # ignore SIGPIPE here.
+        if e.errno == errno.EPIPE:
+            sys.exit(1)
```
qiluo-msft: >1 [](start = 21, length = 1)

return 0, since there is no error.

jleveque: Done.

PR #62, opened 2017-06-08
[dropcheck]: Add the dropcheck as a script to sum up drop counters in a duration

stcheng:
- This tool could be used in monit to check periodically and exit 1
  when the drop exceeds certain threshold for monit to alert in syslog.


jleveque:
You'll also need to add this script to setup.py to make it a part of the package.

stcheng:
@jleveque updated

PR #66, opened 2017-06-19
[config and sonic-installer] Support bgp admin state preserving

taoyl-ms:
BGP admin state will be recorded when calling `config bgp neighbor shutdown` and `config bgp neighbor startup`. Those states will be preserved upon bgp service restart and firmware version upgrade.


jleveque:
Could you please add comments describing these changes?

PR #77, opened 2017-06-26
[sfputil]: Use front panel index in port_config.ini to locate the front panel port

lguohan:
allow flexible mapping between front panel port name and index


lguohan:
resolve issue: https://github.com/Azure/sonic-platform-modules-dell/issues/7

lguohan:
check port_config.init format.

https://github.com/Azure/sonic-buildimage/pull/752

PR #85, opened 2017-07-18
[config] Adapt to config DB

taoyl-ms:
For BGP admin status, CLI will now only write config into DB, and config daemon in bgp docker will do actual config job.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -5,13 +5,11 @@ import os
 import click
 import json
 import subprocess
+from swsssdk import ConfigDBConnector
 
 SONIC_CFGGEN_PATH = "/usr/local/bin/sonic-cfggen"
 MINIGRAPH_PATH = "/etc/sonic/minigraph.xml"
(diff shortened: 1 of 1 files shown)


config/main.py
```diff
@@ -35,10 +33,9 @@ def run_command(command, pager=False, display_cmd=False):
     if p.returncode != 0:
         sys.exit(p.returncode)
 
-# Returns BGP ASN as a string
-def _get_bgp_asn_from_minigraph():
-    # Get BGP ASN from minigraph
-    proc = subprocess.Popen([SONIC_CFGGEN_PATH, '-m', MINIGRAPH_PATH, '--var-json', MINIGRAPH_BGP_ASN_KEY],
+# Returns BGP neighbor dict from minigraph
```
lguohan: better to move the comments as python doc.

lguohan:
:shipit:

PR #86, opened 2017-07-19
[config]: Add ANYDROP configuration enable/disable command

stcheng:
- This command is right now explicitly used on devices that have
  ACL DEFAULT_RULE which denies all the unmatched packets.

- Command: config everflow acldrop <ip_addr>
  This command will create a mirror session to the destination IP
  and a corresponding ACL rule that overrides the current default
  deny rule so that to mirror all the ought-to-denies packets to
  the destination IP for further analysis.

- Command: config everflow no acldrop
  This command will revert the previous command and remove the
  corresponding ACL rule and mirror session when they exist

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1,10 +1,11 @@
 #!/usr/sbin/env python
 
-import sys
-import os
 import click
 import json
+import os
+import re
 import subprocess
+import sys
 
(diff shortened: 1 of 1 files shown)


lguohan:
acl table is in config db, this pr needs to adapt to new config db design. new pr is needed.

PR #87, opened 2017-07-19
Add a script to config/show the HWSKU

zhenggen-xu:
This is the general utility script for change the HWSKUs of one platform.


lguohan:
I should be more specific, I mean to add the config hwsku to this file. https://github.com/Azure/sonic-utilities/blob/master/config/main.py

it has a framework do autocompletion.

PR #89, opened 2017-07-21
Added syncd SAI dump to sysdump script

andrii-savka:
It should merged only after following request:
https://github.com/Azure/sonic-sairedis/pull/199

### DIFF
--- a/scripts/generate_dump
+++ b/scripts/generate_dump
@@ -285,6 +285,14 @@ main() {
     save_redis "2" "COUNTERS_DB"
 
     save_cmd "docker exec -it syncd saidump" "saidump"
+    
+    local platform="$(/usr/local/bin/sonic-cfggen -v platform)"
+    if [[ $platform == *"mlnx"* ]]; then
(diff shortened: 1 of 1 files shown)


scripts/generate_dump
```diff
@@ -285,6 +285,11 @@ main() {
     save_redis "2" "COUNTERS_DB"
 
     save_cmd "docker exec -it syncd saidump" "saidump"
+    
+    local sai_dump_filename="/tmp/syncd_sai_dump"
```
stcheng: add timestamp?

lguohan: should be only on mellanox platform since it is the only platform that supports this.

andrii-savka:
Fixed above remarks. 
Thanks.

andrii-savka:
The saisdkdump tool is already merged. 
All remarks are fixed. 
So we can merge this request.

lguohan:
cannot merge this one, since the change is checked into 1.0.3 branch. It can only be merged after 1.0.3 branch is merged into master.

PR #93, opened 2017-08-03
Add support to reload mgmt conf file and minigraph

taoyl-ms:
`config load_mgmt_config [FILENAME]` to reconfigure mgmt interface based on device description file.
`config load_minigraph` to reload config from `/etc/sonic/minigraph.xml`

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -83,6 +83,14 @@ def _switch_bgp_session_status(ipaddr_or_hostname, status, verbose):
         raise click.Abort
     _switch_bgp_session_status_by_addr(ipaddress, status, verbose)
 
+def _change_hostname(hostname):
+    current_hostname = os.uname()[1]
+    if current_hostname != hostname:
+        run_command('echo {} > /etc/hostname'.format(hostname), display_cmd=True)
+        run_command('hostname -F /etc/hostname', display_cmd=True)
+        run_command('sed -i "/\s{}$/d" /etc/hosts'.format(current_hostname), display_cmd=True)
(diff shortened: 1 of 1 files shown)


config/main.py
```diff
@@ -112,6 +113,40 @@ def load(filename):
...
+@click.argument('filename', default='/etc/sonic/device_desc.xml', type=click.Path(exists=True))
+def load_mgmt_config(filename):
+    """Reconfigure mgmt interface based on device description file."""
+    command = "{} -M {} --write-to-db".format(SONIC_CFGGEN_PATH, filename)
+    run_command(command, display_cmd=True)
+    #FIXME: After config DB daemon for mgmt interface is implemented, we'll no longer need to manually config mgmt interface here
+    mgmt_conf = parse_device_desc_xml(filename)['minigraph_mgmt_interface']
+    command = "ifconfig eth0 {} netmask {}".format(str(mgmt_conf['addr']), str(mgmt_conf['mask']))
+    run_command(command, display_cmd=True)
+    command = "[ -f /var/run/dhclient.eth0.pid ] && kill `cat /var/run/dhclient.eth0.pid` && rm -f /var/run/dhclient.eth0.pid"
```
lguohan: can we do interfaces-config restart here like below?

config/main.py
```diff
@@ -112,6 +113,40 @@ def load(filename):
...
+    run_command(command, display_cmd=True)
```
lguohan: can we do interfaces-config restart here?

taoyl-ms: No. Because interfaces-config service will read data from minigraph but not device description file.

PR #95, opened 2017-08-11
[FastReboot]: Update FR to make it working with 1.0.3

pavel-shirshov:


pavel-shirshov:
Any comment why it was closed?

scripts/fast-reboot-dump.py
```diff
@@ -49,24 +53,56 @@ def get_vlan_ifaces():
 
     return vlans
 
+def get_bridge_port_id_2_port_id(db):
+    bridge_port_id_2_port_id = {}
+    keys = db.keys(db.ASIC_DB, 'ASIC_STATE:SAI_OBJECT_TYPE_BRIDGE_PORT:oid:*')
```
lguohan: how to make sure the ASIC_STATE:SAI_OBJECT_TYPE_BRIDGE_PORT table is available be the time you run this function?

lguohan: is it possible the asic db is not completely initialized and you do not have all bridge_port id to port id mappings?

pavel-shirshov: I run this command before the switch goes into the FR. I think external tools should check a state of a switch before the initiate FR.

pavel-shirshov: It's possible, but if the device is not initialized we shouldn't reboot a device.

PR #101, opened 2017-09-11
[portstat]: Show port speed accordingly

stcheng:
This pull request depends on https://github.com/Azure/sonic-swss/pull/287/files. After that pull request is merged, we could add port speed value column in the current `port_config.ini` files and display the speed via this script.


sihuihan88:
Also set port_rate based on port speed, instead of hardcoding it to 40?

stcheng:
@sihuihan88 good point, i'll update this pull request. thanks.

lguohan:
too old pr.

PR #102, opened 2017-09-11
[show]: Add 'ipv6' group along with 'bgp' and 'route' subcommands; Remove duplicate commands

jleveque:
- Add 'ipv6' group along with 'bgp' and 'route' subcommands
- Remove duplicate commands:
  - Remove 'arp' from under 'ip' group, now only under root group
  - Remove 'bgp' and 'route' commands from under root group (now different commands reside under 'ip' & 'ipv6')
  - Remove 'interfaces' from under 'ip' group, now only under root group

### DIFF
--- a/show/main.py
+++ b/show/main.py
@@ -108,10 +108,10 @@ CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help', '?'])
 
 
 #
-# 'cli' group (root group) ###
+# 'cli' group (root group)
 #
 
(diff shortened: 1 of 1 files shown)


nikos-li:
The bgp related changes are incorrect and will need to be reverted.

stcheng:
Hi @nikos-li could you create a separate issue? Since this pull request is already closed, the comment is hard to track.

PR #106, opened 2017-09-25
CLI support for Layer 2 MAC/FDB show 

prsunny:
Support to show Layer 2 MAC entries learnt in Hardware.  The entries can also be shown per Port/Vlan 

    usage: fdbshow [-p PORT] [-v VLAN]
    optional arguments:
      -p,  --port              FDB learned on specific port: Ethernet0
      -v,  --vlan              FDB learned on specific Vlan: 1000
  
    Example of the output:
    admin@str~$ fdbshow
      No.    Vlan  MacAddress         Port
    -----  ------  -----------------  ----------
        1    1000  7C:FE:90:80:9F:05  Ethernet20
        2    1000  7C:FE:90:80:9F:10  Ethernet40
        3    1000  7C:FE:90:80:9F:01  Ethernet4
        4    1000  7C:FE:90:80:9F:02  Ethernet8
    Total number of entries 4

### DIFF
--- a/scripts/fdbshow
+++ b/scripts/fdbshow
@@ -123,10 +123,7 @@ class FdbShow(object):
             return 0
 
         keys = [r[pos] for r in self.bridge_mac_list]
-        try:
-            return keys.index(key_value)
-        except ValueError as e:
-            print e.message
+        return keys.index(key_value)
 
 
     def display(self, vlan, port):
--- a/scripts/fdbshow
+++ b/scripts/fdbshow
@@ -1,0 +1,192 @@
+#!/usr/bin/python
+"""
+    Script to show MAC/FDB entries learnt in Hardware
+    
+    usage: fdbshow [-p PORT] [-v VLAN]
+    optional arguments:
+      -p,  --port              FDB learned on specific port: Ethernet0
+      -v,  --vlan              FDB learned on specific Vlan: 1000
+  
+    Example of the output:
+    admin@str~$ fdbshow
+      No.    Vlan  MacAddress         Port
+    -----  ------  -----------------  ----------
+        0    1000  7C:FE:90:80:9F:05  Ethernet20
+        1    1000  7C:FE:90:80:9F:10  Ethernet40
+        2    1000  7C:FE:90:80:9F:01  Ethernet4
+        3    1000  7C:FE:90:80:9F:02  Ethernet8
+
+    admin@str:~$ fdbshow -p Ethernet4
+      No.    Vlan  MacAddress         Port
+    -----  ------  -----------------  ---------
+        0    1000  7C:FE:90:80:9F:01  Ethernet4
+
+    admin@str:~$ fdbshow -v 1001
+    1001 is not in list
+
+"""
+
+import swsssdk
+import argparse
+import json
+import sys
+
+from tabulate import tabulate
+from natsort import natsorted
+
+class FdbShow(object):
+
+    HEADER = ['No.', 'Vlan', 'MacAddress', 'Port']
+    FDB_COUNT = 0
+
+    def __init__(self):
+        super(FdbShow,self).__init__()
+        self.db = swsssdk.SonicV2Connector(host="127.0.0.1")
+        self.get_interface_oid_map()
+        self.get_bridge_port_oid_map()
+        self.fetch_fdb_data()
+        return
+
+    
+    def get_interface_oid_map(self):
+        """
+            Get the Interface names from Counters DB
+        """
+        self.db.connect('COUNTERS_DB')
+        self.if_name_map = self.db.get_all('COUNTERS_DB', 'COUNTERS_PORT_NAME_MAP', blocking=True)
+        self.if_oid_map = {sai_oid: if_name for if_name, sai_oid in self.if_name_map.items()}
+
+
+    def get_bridge_port_oid_map(self):
+        """
+            Get the Bridge port names from ASIC DB
+        """
+        self.db.connect(self.db.ASIC_DB)
+        self.if_br_oid_map = {} 
+        br_port_str = self.db.keys('ASIC_DB', "ASIC_STATE:SAI_OBJECT_TYPE_BRIDGE_PORT:*")
+        if not br_port_str:
+            return
+        
+        offset = len("ASIC_STATE:SAI_OBJECT_TYPE_BRIDGE_PORT:")
+        oid_pfx = len("oid:0x")
+        for s in br_port_str:
+            br_port_id = s[(offset + oid_pfx):]
+            ent = self.db.get_all('ASIC_DB', s, blocking=True)
+            port_id = ent[b"SAI_BRIDGE_PORT_ATTR_PORT_ID"][oid_pfx:]
+            self.if_br_oid_map[br_port_id] = port_id 
+
+
+    def fetch_fdb_data(self):
+        """
+            Fetch FDB entries from ASIC DB. 
+            FDB entries are sorted on "VlanID" and stored as a list of tuples
+        """
+        self.db.connect(self.db.ASIC_DB)
+        self.bridge_mac_list = []
+        
+        fdb_str = self.db.keys('ASIC_DB', "ASIC_STATE:SAI_OBJECT_TYPE_FDB_ENTRY:*")
+        if not fdb_str:
+            return
+
+        oid_pfx = len("oid:0x")
+        for s in fdb_str:
+            fdb_entry = s.decode()
+            try:
+                fdb = json.loads(fdb_entry .split(":", 2)[-1])
+            except ValueError as e: 
+                print e.message
+                continue
+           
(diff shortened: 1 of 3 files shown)


scripts/fdbshow
```diff
@@ -0,0 +1,192 @@
...
+from natsort import natsorted
+
+class FdbShow(object):
+
+    HEADER = ['No.', 'Vlan', 'MacAddress', 'Port']
+    FDB_COUNT = 0
+
+    def __init__(self):
+        super(FdbShow,self).__init__()
+        self.db = swsssdk.SonicV2Connector(host="127.0.0.1")
```
qiluo-msft: >host [](start = 43, length = 4)

We prefer using unix socket.

scripts/fdbshow
```diff
@@ -0,0 +1,192 @@
...
+        offset = len("ASIC_STATE:SAI_OBJECT_TYPE_BRIDGE_PORT:")
+        oid_pfx = len("oid:0x")
+        for s in br_port_str:
+            br_port_id = s[(offset + oid_pfx):]
+            ent = self.db.get_all('ASIC_DB', s, blocking=True)
+            port_id = ent[b"SAI_BRIDGE_PORT_ATTR_PORT_ID"][oid_pfx:]
+            self.if_br_oid_map[br_port_id] = port_id 
+
+
+    def fetch_fdb_data(self):
```
qiluo-msft: >fetch_fdb_data [](start = 8, length = 14)

I notice code duplication with sonic-snmpagent. Could you reuse code?

scripts/fdbshow
```diff
@@ -0,0 +1,192 @@
...
+
+        for fdb in fdb_list:
+            self.FDB_COUNT += 1
+            output.append([self.FDB_COUNT, fdb[0], fdb[1], fdb[2]])
+
+        print tabulate(output, self.HEADER)
+        print "Total number of entries {0} ".format(self.FDB_COUNT)
+
+
+    def display_summary(self):
```
qiluo-msft: >display_summary [](start = 8, length = 15)

I guess you don't need one more function for no filters. Just reuse display_specific().

scripts/fdbshow
```diff
@@ -0,0 +1,192 @@
...
+        elif vlan is None and port is not None:
+            self.bridge_mac_list = natsorted(self.bridge_mac_list, key = lambda x: x[2])
+            s_index = self.get_iter_index(port, 2)
+            fdb_list = [fdb for fdb in self.bridge_mac_list[s_index:] if fdb[2] == port]
+
+        else:
+            vlan = int(vlan)
+            s_index = self.get_iter_index(vlan)
+            fdb_list = [fdb for fdb in self.bridge_mac_list[s_index:]
+                        if fdb[0] == vlan and fdb[2] == port]
```
qiluo-msft: Simply the conditions into one?

scripts/fdbshow
```diff
@@ -0,0 +1,192 @@
...
+    args = parser.parse_args()
+
+    try:
+        fdb = FdbShow()
+        if args.port is None and args.vlan is None:
+            fdb.display_summary()
+        else:
+            fdb.display_specific(args.port, args.vlan)
+        
+    except Exception as e:
```
qiluo-msft: >Exception [](start = 11, length = 9)

No need to catch all exceptions.

scripts/fdbshow
prsunny: As discussed, this format is used by other standalone scripts which also handles the permission issue.

scripts/fdbshow
prsunny: I checked this in detail to see if we can reuse some functionality but there is tight coupling with MIB and SNMP functionality. Since this is a standalone functionality, thought of keeping it separate.

scripts/fdbshow
prsunny: Addressed.

scripts/fdbshow
prsunny: Addressed

scripts/fdbshow
prsunny: As we discussed, I removed this and checked. But observed that in case of invalid inputs, in addition to displaying the error message, it also gives the trace back which is not expected. For e.g: 

admin@str-s6000-acs-7:~$ ./fdbshow -v 1001     
  File "./fdbshow", line 182, in <module>
    main()
  File "./fdbshow", line 179, in main
    fdb.display(args.vlan, args.port)
  File "./fdbshow", line 158, in display
    s_index = self.get_iter_index(arg_list[m_index], m_index)
  File "./fdbshow", line 127, in get_iter_index
    return keys.index(key_value)
ValueError: 1001 is not in list

Whereas, with the exception is would be:

admin@str-s6000-acs-7:~$ ./fdbshow -v 1001
1001 is not in list

scripts/fdbshow
prsunny: Modified to use common library API

scripts/fdbshow
qiluo-msft: If you are arguing about this case, the solution should be catch specific exception, instead of catch all.

---
In reply to: [140927960](https://github.com/Azure/sonic-utilities/pull/106#discussion_r140927960) [](ancestors = 140927960)

qiluo-msft:
>     install_requires=[

Should we add swsssdk as a dependency here?

---
Refers to: setup.py:45 in 25f704d. [](commit_id = 25f704d9e11202b52b17eaee69ca93577ddd34ef, deletion_comment = False)

prsunny:
I don't think swsssdk  is required in setup.py. It says it's mainly used to install dependencies when the project is installed using 'pip'. Also swsssdk is used by other scripts files in the same repo.

scripts/fdbshow
qiluo-msft: It is not simplified, and become more complex...

---
In reply to: [140849540](https://github.com/Azure/sonic-utilities/pull/106#discussion_r140849540) [](ancestors = 140849540)

prsunny: I think it now handles the three conditions -  
1 - All arguments None
2 - Some arguments None
3 - None of the arguments are None. 
Could you suggest a different way?

jleveque:
I don't think it would hurt to add swsssdk in the list of dependencies. I think it was overlooked in the past.

scripts/fdbshow
```diff
@@ -0,0 +1,157 @@
...
+    1001 is not in list
+
+"""
+
+import swsssdk
+import argparse
+import json
+import sys
+
+from swsssdk import port_util
```
jleveque: This line is redundant with the `import swsssdk` line above. Remove whichever is appropriate.

scripts/fdbshow
```diff
@@ -0,0 +1,157 @@
...
+from tabulate import tabulate
+from natsort import natsorted
```
jleveque: Please reorder so that `import ...` statements are ordered alphabetically followed by `from ... import ...` statements ordered alphabetically. This makes it easier to locate used packages/modules.

scripts/fdbshow
prsunny: removed

scripts/fdbshow
prsunny: done

scripts/fdbshow
```diff
@@ -0,0 +1,155 @@
...
+            fdb_entry = s.decode()
+            try:
+                fdb = json.loads(fdb_entry .split(":", 2)[-1])
+            except ValueError as e: 
+                print e.message
+                continue
+           
+            ent = self.db.get_all('ASIC_DB', s, blocking=True)
+            br_port_id = ent[b"SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID"][oid_pfx:]
+            try:
```
stcheng: use `if` instead of `try/except`?

lguohan:
somehow, I did not see my previous comments.

can we add hook to show command

show macs
show macs [-v vlan]
show macs [-p port]

scripts/fdbshow
prsunny: Agree its slightly complex since I accounted for adding future options. Revisited the logic and made it simple. Thanks for pointing out.

scripts/fdbshow
prsunny: Prefer to keep it same as other scripts

prsunny:
Added hooks to CLI command.

scripts/fdbshow
qiluo-msft: If you are arguing about this case, the solution should be catch specific exception, instead of catch all.

show/main.py
```diff
@@ -226,6 +226,25 @@ def sfp(interfacename):
 
     run_command(cmd)
 
+#
+# 'mac' command ("show mac ...")
```
qiluo-msft: >mac [](start = 23, length = 3)

I am not sure about other vendors' CLI. Is it more accurate to call it 'show fdb'?

prsunny: @lguohan , can you suggest?

PR #109, opened 2017-09-27
[acl-loader] Add acl-loader utility.

oleksandrivantsiv:
- Add acl-loader implementation based on translate-acl utility.
- Add shortcuts to show command.
- Add shortcuts to config command.

### DIFF
--- a/acl_loader/main.py
+++ b/acl_loader/main.py
@@ -1,0 +1,585 @@
+
+#!/usr/bin/env python
+
+import click
+import sys
+import os.path
+import json
+import argparse
+import tabulate
+
+import openconfig_acl
+import pyangbind.lib.pybindJSON as pybindJSON
+from swsssdk import ConfigDBConnector
+
+
+def info(msg):
+    click.echo(click.style("Info: ", fg='cyan') + click.style(str(msg), fg='green'))
+
+
+def warning(msg):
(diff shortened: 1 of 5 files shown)


acl_loader/main.py
```diff
@@ -0,0 +1,569 @@
...
+    """
+    acl_loader = ctx.obj["acl_loader"]
+    acl_loader.show_session(session_name)
+
+
+@show.command()
+@click.argument('table_name', type=click.STRING, required=False)
+@click.argument('rule_id', type=click.STRING, required=False)
+@click.pass_context
+def rule(ctx, table_name, rule_id):
```
taoyl-ms: Can we add an option to specify max_priority explicitly instead of using default value?

oleksandrivantsiv: Do you want to make it mandatory? Or we still should use a default value if priority is not provided by user?

stcheng:
resolve the conflicts

acl_loader/main.py
taoyl-ms: It should be optional and we should fall back to default value if not specified by user.

oleksandrivantsiv: done

oleksandrivantsiv:
@stcheng conflict resolved

PR #112, opened 2017-09-28
Update fast-reboot script to boot next selected image

andriymoroz-mlnx:
**- What I did**
Updated fast-reboot script to boot image which would boot after regular reboot

**- How I did it**
When deciding which kernel to kexec script now checks what is actually scheduled to boot in GRUB

**- How to verify it**
Install new image using SONiC-to-SONiC method (not from ONIE) and execute fast-reboot.
Make sure newly installed version booted

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -15,9 +15,10 @@ then
 fi
 
 # Kernel and initrd image
-KERNEL_IMAGE="/vmlinuz"
-INITRD="/initrd.img"
-BOOT_OPTIONS=$(cat /proc/cmdline)
+NEXT_SONIC_IMAGE=$(sonic_installer list | grep "Next: " | cut -d ' ' -f 2)
+BOOT_OPTIONS=$(cat /host/grub/grub.cfg | sed "/$NEXT_SONIC_IMAGE'/,/}/"'!'"g" | grep linux)
+KERNEL_IMAGE="/host$(echo $BOOT_OPTIONS | cut -d ' ' -f 2)"
+INITRD=$(echo $KERNEL_IMAGE | sed 's/vmlinuz/initrd.img/g')
 
 case "$BOOT_OPTIONS" in
   *fast-reboot*)
@@ -67,4 +68,5 @@ sleep 1
 sync
 
 # Reboot
+echo "Rebooting to $NEXT_SONIC_IMAGE..."
 reboot
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -16,9 +16,9 @@ fi
 
 # Kernel and initrd image
 NEXT_SONIC_IMAGE=$(sonic_installer list | grep "Next: " | cut -d ' ' -f 2)
-BOOT_OPTIONS=$(cat /host/grub/grub.cfg | sed "/$NEXT_SONIC_IMAGE'/,/}/"'!'"g" | grep linux)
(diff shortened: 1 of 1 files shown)


scripts/fast-reboot
```diff
@@ -15,9 +15,10 @@ then
 fi
 
 # Kernel and initrd image
-KERNEL_IMAGE="/vmlinuz"
-INITRD="/initrd.img"
-BOOT_OPTIONS=$(cat /proc/cmdline)
+NEXT_SONIC_IMAGE=$(sonic_installer list | grep "Next: " | cut -d ' ' -f 2)
+BOOT_OPTIONS=$(cat /host/grub/grub.cfg | sed "/$NEXT_SONIC_IMAGE'/,/}/"'!'"g" | grep linux)
```
pavel-shirshov: On my test machine previous version of BOOT_OPTIONS was
`BOOT_IMAGE=/image-HEAD.230-cce2a36/boot/vmlinuz-3.16.0-4-amd64 root=/dev/sda4 rw console=tty0 console=ttyS1,9600n8 quiet loop=image-HEAD.230-cce2a36/fs.squashfs loopfstype=squashfs apparmor=1 security=apparmor varlog_size=4096 fast-reboot`

Your version of the same variable is
`linux /image-HEAD.230-cce2a36/boot/vmlinuz-3.16.0-4-amd64 root=/dev/sda4 rw console=tty0 console=ttyS1,9600n8 quiet loop=image-HEAD.230-cce2a36/fs.squashfs loopfstype=squashfs apparmor=1 security=apparmor varlog_size=4096`

So difference is s/BOOT_IMAGE=/linux     /

I've checked your script with my fast-reboot test. The test passes.

So honestly, I don't know this "BOOT_IMAGE" is critical or not.
So I'll approve.
But it would be nice of you to convert BOOT_OPTIONS to "BOOT_IMAGE=" format.

scripts/fast-reboot
```diff
@@ -15,9 +15,10 @@ then
...
+KERNEL_IMAGE="/host$(echo $BOOT_OPTIONS | cut -d ' ' -f 2)"
+INITRD=$(echo $KERNEL_IMAGE | sed 's/vmlinuz/initrd.img/g')
 
 case "$BOOT_OPTIONS" in
```
pavel-shirshov: You can remove this block and append fast-reboot option unconditionally. You read the command line from the file

scripts/fast-reboot
pavel-shirshov: Please convert your BOOT_OPTIONS to "BOOT_IMAGE" format. You can do this by appending 
`| sed -e 's/\s*linux\s*/BOOT_IMAGE=/'` into your BOOT_OPTIONS. (after grep linux).
Also you need to adjust your KERNEL_IMAGE option

andriymoroz-mlnx: it worked... but you're right need to fix this

scripts/fast-reboot
andriymoroz-mlnx: agree

lguohan:
we also need aboot version of it, but let's do that in a separate pr.

pavel-shirshov:
The last change looks ok, except reusing of BOOT_OPTIONS twice.
I'll recheck everything and let you know (currently my testbed is not available)

lguohan:
retest this please

PR #121, opened 2017-10-05
[show]: Remove default commands; add missing docstrings

jleveque:


jleveque:
@rodnymolina: No. The tab fix will be a different commit. This won't actually fix the problem because it's broken elsewhere due the .deb -> .whl change, but you forgot to add the bash completion files for your new utilities to setup.py, so I'm trying to copy all of them all as a glob here to prevent this in the future.

As to the actual tab problem, I am going to try to enhance our build system in order to install sonic-utilities as a .deb package again. The .whl package has more limitations than the .deb package and copying files outside of the dist-packages directory is impossible.

jleveque:
@rodnymolina: With regards to AliasedGroup: We should no longer rely on DefaultGroup -- AliasedGroup should extend Click.group directly, OR we need to fix/write our own DefaultGroup that doesn't break tab completion.

PR #123, opened 2017-10-05
[config load_mgmt_config] add default route

taoyl-ms:
Add default route through mgmt when load device description files.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -145,7 +145,12 @@ def load_mgmt_config(filename):
     hostname = config_data['DEVICE_METADATA']['localhost']['hostname']
     _change_hostname(hostname)
     mgmt_conf = netaddr.IPNetwork(config_data['MGMT_INTERFACE'].keys()[0][1])
+    gw_addr = config_data['MGMT_INTERFACE'].values()[0]['gwaddr']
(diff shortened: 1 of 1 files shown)


config/main.py
```diff
@@ -145,8 +145,13 @@ def load_mgmt_config(filename):
     hostname = config_data['DEVICE_METADATA']['localhost']['hostname']
     _change_hostname(hostname)
     mgmt_conf = netaddr.IPNetwork(config_data['MGMT_INTERFACE'].keys()[0][1])
+    gw_addr = config_data['MGMT_INTERFACE'].values()[0]['gwaddr']
     command = "ifconfig eth0 {} netmask {}".format(str(mgmt_conf.ip), str(mgmt_conf.netmask))
     run_command(command, display_cmd=True)
+    command = "ip route add default via {} dev eth0 table default".format(gw_addr)
```
lguohan: is this management address ipv4? In our template, we support ipv6 mgmt port.

taoyl-ms: Current device_desc file parser enforces that mgmt address is ipv4.

lguohan: ok.

Issue #99, opened 2017-08-25
handle notimplementederror exception for the plugins


lguohan:
in sfputil, gracefully catch notimplementederror  exception throw by plugins and print out error instead of throw traceback stack.

```
admin@str-msn2700-20:/usr/share/sonic/device/x86_64-mlnx_msn2700-r0/plugins$ sudo sfputil show lpmode
Traceback (most recent call last):
  File "/usr/bin/sfputil", line 9, in <module>
    load_entry_point('sonic-utilities==1.1', 'console_scripts', 'sfputil')()
  File "/usr/lib/python2.7/dist-packages/click/core.py", line 716, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python2.7/dist-packages/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python2.7/dist-packages/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python2.7/dist-packages/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python2.7/dist-packages/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/sfputil/main.py", line 514, in lpmode
    lpmode = platform_sfputil.get_low_power_mode(physical_port)
  File "/usr/share/sonic/device/x86_64-mlnx_msn2700-r0/plugins/sfputil.py", line 54, in get_low_power_mode
    raise NotImplementedError
NotImplementedError
```

PR #127, opened 2017-10-06
[sfputil]: Catch NotImplementedError, print message and exit gracefully

jleveque:
Resolves https://github.com/Azure/sonic-utilities/issues/99

### DIFF
--- a/sfputil/main.py
+++ b/sfputil/main.py
@@ -448,7 +448,7 @@ def presence(port):
         if platform_sfputil.is_valid_sfputil_port(port) == 0:
             print "Error: invalid port '%s'\n" % port
             print_all_valid_port_values()
-            sys.exit(5)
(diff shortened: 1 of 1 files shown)


sfputil/main.py
```diff
@@ -466,7 +466,13 @@ def presence(port):
 
         for physical_port in physical_port_list:
             port_name = get_physical_port_name(logical_port_name, i, ganged)
-            presence = platform_sfputil.get_presence(physical_port)
+
+            try:
+                presence = platform_sfputil.get_presence(physical_port)
+            except NotImplementedError:
+                click.echo("This functionality is currently not implemented for this platform")
+                sys.exit(6)
```
lguohan: all not implemented should return same exit code, easily for user to know the error.

PR #134, opened 2017-10-21
[sonic_installer]: add command to cleanup installed images

lguohan:


sonic_installer/main.py
```diff
@@ -291,5 +295,24 @@ def binary_version(binary_image_path):
...
+        expose_value=False, prompt='Cleanup images which are not current and next, continue?')
+def cleanup():
+    """ Cleanup installed images which are not current and next """
+    images = get_installed_images()
+    curimage = get_current_image()
+    nextimage = get_next_image()
+    image_removed = 0
+    for image in get_installed_images():
+        if image != curimage and image != nextimage:
+            click.echo("Cleanup image %s" % image)
```
jleveque: Suggest to change from `Cleanup image %s` to `Removing image %s`

sonic_installer/main.py
```diff
@@ -291,5 +295,24 @@ def binary_version(binary_image_path):
...
+            remove_image(image)
+            image_removed += 1
+
+    if image_removed == 0:
+        click.echo("No image to cleanup")
```
jleveque: Suggest to change from `No image to cleanup` to `No image(s) to remove`

sonic_installer/main.py
lguohan: done

sonic_installer/main.py
lguohan: done

Issue #137, opened 2017-10-26
decode-syseeprom cannot save cache


stcheng:
Although we have the `/var/cache/ssw/decode-syseeprom` folder for saving the eeprom cache, this functionality never works. It seems that there's a bug in the `/usr/bin/decode-syseeprom` script line 115. Need to fix this issue so as to speed up the script read time.

jleveque:
@stcheng: You're correct, the bug is on line 115. `opts.target` does not exist. I looked back through all commits, and it appears that `target` was never a member of `opts`.

Do you know what the cache file name was originally intended to be when this was first designed?

stcheng:
I think we could name it as `syseeprom` or `eeprom`.

PR #140, opened 2017-10-26
[decode-syseeprom]: Fix caching

jleveque:
- `opts.target` did not exist, so a cache file was never specified. Now specifying cache file name explicitly.
- Also changed cache file path from `/var/cache/ssw/` to `/var/cache/sonic/`
- Resolves https://github.com/Azure/sonic-utilities/issues/137

### DIFF
--- a/scripts/decode-syseeprom
+++ b/scripts/decode-syseeprom
@@ -26,7 +26,8 @@ PLATFORM_KEY = 'platform'
 
 PLATFORM_ROOT = '/usr/share/sonic/device'
 
-cache_root = '/var/cache/ssw/decode-syseeprom'
+CACHE_ROOT = '/var/cache/sonic/decode-syseeprom'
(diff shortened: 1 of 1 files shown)


scripts/decode-syseeprom
```diff
@@ -98,21 +99,22 @@ def run(target, opts, args):
...
     if opts.init:
-        for file in glob.glob(os.path.join(cache_root, '*')):
+        for file in glob.glob(os.path.join(CACHE_ROOT, '*')):
             os.remove(file)
 
     #
     # only the eeprom classes that inherit from eeprom_base
     # support caching. Others will work normally
     #
+    print repr(opts)
```
stcheng: remove this debugging line?

jleveque: Removed. Thanks for spotting that.

PR #147, opened 2017-11-07
[pfcwd]: Configuration commands

marian-pritsak:
Add 'start/stop' commands
Create 'show' group
Move 'stats' to 'show' and add 'show config'


marian-pritsak:
Depends on: https://github.com/Azure/sonic-swss/pull/374/files

PR #152, opened 2017-11-20
Added PSU CLI

andrii-savka:


psuutil/main.py
```diff
@@ -0,0 +1,192 @@
...
+    import traceback
+    from tabulate import tabulate
+except ImportError as e:
+    raise ImportError("%s - required module not found" % str(e))
+
+VERSION = '1.0'
+
+SYSLOG_IDENTIFIER = "psuutil"
+PLATFORM_SPECIFIC_MODULE_NAME = "psuutil"
+PLATFORM_SPECIFIC_CLASS_NAME = "PSUutil"
```
jleveque: Suggest changing case of class name to `PsuUtil` to match CamelCase of `SfpUtil`

psuutil/main.py
```diff
@@ -0,0 +1,192 @@
...
+# 'version' subcommand
+@cli.command()
+def version():
+    """Display version info"""
+    click.echo("psuutil version {0}".format(VERSION))
+
+# 'numpsus' subcommand
+@cli.command()
+def numpsus():
+    "number of supported PSU in the device"
```
jleveque: Suggest `"Display number of supported PSU in the device"`

sonic_psu/psu_base.py
```diff
@@ -26,10 +26,22 @@ def get_num_psus(self):
     @abc.abstractmethod
     def get_psu_status(self, index):
         """
-        Retrieves the oprational status of power supply unit (PSU) defined
+        Retrieves the operational status of power supply unit (PSU) defined
                 by index <index>
 
         :param index: An integer, index of the PSU of which to query status
         :return: Boolean, True if PSU is operating properly, False if PSU is faulty
```
jleveque: Could you also please add more detail to my description here to reflect what we have discussed as "faulty" (e.g., if the PSU is present but powered off, it is considered a fault, and should return false.

psuutil/main.py
andrii-savka: Done

sonic_psu/psu_base.py
andrii-savka: More details has been added

sonic_psu/psu_base.py
```diff
@@ -26,10 +26,22 @@ def get_num_psus(self):
     @abc.abstractmethod
     def get_psu_status(self, index):
         """
-        Retrieves the oprational status of power supply unit (PSU) defined
+        Retrieves the operational status of power supply unit (PSU) defined
                 by index <index>
 
         :param index: An integer, index of the PSU of which to query status
```
yurypm: Is it zero-based index? I think this info should be here.

andrii-savka: Added more details

PR #153, opened 2017-11-21
Add utility to configure ECN WRED parameters

andriymoroz-mlnx:
Signed-off-by: Andriy Moroz <c_andriym@mellanox.com>

Should be merged AFTER https://github.com/Azure/sonic-buildimage/pull/1178

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -331,6 +331,30 @@ def incremental(file_name):
     command = "acl-loader update incremental {}".format(file_name)
     run_command(command)
 
+#
+# 'ecn' command
+#
+@cli.command()
+@click.option('-profile', metavar='<profile_name>', type=str, required=True, help="Profile name")
+@click.option('-rmax', metavar='<red threshold max>', type=int, help="Set red max threshold")
+@click.option('-rmin', metavar='<red threshold min>', type=int, help="Set red min threshold")
+@click.option('-ymax', metavar='<yellow threshold max>', type=int, help="Set yellow max threshold")
+@click.option('-ymin', metavar='<yellow threshold min>', type=int, help="Set yellow min threshold")
+@click.option('-gmax', metavar='<green threshold max>', type=int, help="Set green max threshold")
+@click.option('-gmin', metavar='<green threshold min>', type=int, help="Set green min threshold")
+@click.option('-v', '--verbose', is_flag=True, help="Enable verbose output")
+def ecn(profile, rmax, rmin, ymax, ymin, gmax, gmin, verbose):
+    """ECN-related configuration tasks"""
+    command = "ecnconfig -p %s" % profile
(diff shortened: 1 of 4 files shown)


lguohan:
@yxieca , to review

config/main.py
```diff
@@ -331,6 +331,30 @@ def incremental(file_name):
...
+@click.option('-v', '--verbose', is_flag=True, help="Enable verbose output")
+def ecn(profile, rmax, rmin, ymax, ymin, gmax, gmin, verbose):
+    """ECN-related configuration tasks"""
+    command = "ecnconfig -p %s" % profile
+    if rmax: command += " -rmax %d" % rmax
+    if rmin: command += " -rmin %d" % rmin
+    if ymax: command += " -ymax %d" % ymax
+    if ymin: command += " -ymin %d" % ymin
+    if gmax: command += " -gmax %d" % gmax
+    if gmin: command += " -gmin %d" % gmin
```
yxieca: The way you tested these parameters will ignore a parameter when it is 0 or None. Is this what you intended?

config/main.py
```diff
@@ -331,6 +331,30 @@ def incremental(file_name):
...
+    if verbose:
+        click.echo(command)
```
yxieca: Do you think you want to further pass verbose option to ecnconfig script?

scripts/ecnconfig
```diff
@@ -0,0 +1,121 @@
...
+                ecn.set_wred_threshold(args.profile, "gmax", args.green_max)
+            if args.green_min:
+                ecn.set_wred_threshold(args.profile, "gmin", args.green_min)
+            if args.yellow_max:
+                ecn.set_wred_threshold(args.profile, "ymax", args.yellow_max)
+            if args.yellow_min:
+                ecn.set_wred_threshold(args.profile, "ymin", args.yellow_min)
+            if args.red_max:
+                ecn.set_wred_threshold(args.profile, "rmax", args.red_max)
+            if args.red_min:
```
yxieca: All above if statements in try block will ignore value 0 and None, is that intentional?

yxieca: Maybe deleting comments is not the best thing.

I figured out that args are returning values in strings, so we don't have a risk of ignoring integer value 0 here.

config/main.py
andriymoroz-mlnx: yes, min/max parameters are optional

config/main.py
andriymoroz-mlnx: right, will add

config/main.py
yxieca: Sorry for the late reply. Andriy,

My question is more about if you intentionally ignores the value 0? I understand you would like to ignore when it is None.

Regards,
Ying

PR #160, opened 2017-12-06
Align PSU CLI output with SONiC CLI output template

andrii-savka:
Change the format of PSU CLI to align it with SONiC CLI output template

### DIFF
--- a/psuutil/main.py
+++ b/psuutil/main.py
@@ -172,7 +172,8 @@ def status(index, textonly):
         msg = ""
         psu_name = "PSU {}".format(psu)
         if psu not in supported_psu:
-            status_table.append([psu_name, "NOT SUPPORTED"])
+            print ("Error! The {} is not available on the platform.\n" \
+            "Number of supported PSU - {}.".format(psu_name, platform_psuutil.get_num_psus()))
             continue
         presence = platform_psuutil.get_psu_presence(psu)
         if presence:
@@ -182,11 +183,14 @@ def status(index, textonly):
             msg = 'NOT PRESENT'
         status_table.append([psu_name, msg])
 
+    if not status_table:
+        return
+
     if textonly:
         for status in status_table:
             print status[0] + ':' + status[1]
     else:
-        print tabulate(status_table, header, tablefmt="grid")
+        print tabulate(status_table, header, tablefmt="simple")
 
 if __name__ == '__main__':
     cli()
--- a/psuutil/main.py
+++ b/psuutil/main.py
@@ -44,7 +44,7 @@ def log_info(msg, also_print_to_console=False):
     syslog.closelog()
 
     if also_print_to_console:
-        print msg
+        click.echo(msg)
 
 
 def log_warning(msg, also_print_to_console=False):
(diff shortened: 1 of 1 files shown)


andrii-savka:
The CLI output:

Previous output command:
+-------+----------+
| PSU     | Status     |
+====+======+
| PSU 1  | OK          |
+-------+----------+
| PSU 2 | NOT OK   |
+-------+----------+

Actual updated output:
(aligned to other SONiC CLI utilities)
_PSU    Status
-----  --------
PSU 1  OK
PSU 2  NOT OK

psuutil/main.py
```diff
@@ -172,7 +172,8 @@ def status(index, textonly):
         msg = ""
         psu_name = "PSU {}".format(psu)
         if psu not in supported_psu:
-            status_table.append([psu_name, "NOT SUPPORTED"])
+            print ("Error! The {} is not available on the platform.\n" \
```
jleveque: Use `click.echo` instead of `print`

psuutil/main.py
```diff
@@ -182,11 +183,14 @@ def status(index, textonly):
...
 
+    if not status_table:
+        return
+
     if textonly:
         for status in status_table:
             print status[0] + ':' + status[1]
     else:
-        print tabulate(status_table, header, tablefmt="grid")
+        print tabulate(status_table, header, tablefmt="simple")
```
jleveque: Use `click.echo` instead of `print`

Also please do the same on lines 135 and 153.

psuutil/main.py
```diff
@@ -182,11 +183,14 @@ def status(index, textonly):
             msg = 'NOT PRESENT'
         status_table.append([psu_name, msg])
 
+    if not status_table:
+        return
+
     if textonly:
```
jleveque: Now that table format is simplified, do we need a "textonly mode"?

psuutil/main.py
andrii-savka: Replaced all "print" to "click.echo"

psuutil/main.py
andrii-savka: No, it doesn't needed because of SNMP uses PSU plugin directly. 
Removed "textonly mode

psuutil/main.py
andrii-savka: Replaced all "print" to "click.echo"

psuutil/main.py
```diff
@@ -172,7 +171,8 @@ def status(index, textonly):
         msg = ""
         psu_name = "PSU {}".format(psu)
         if psu not in supported_psu:
-            status_table.append([psu_name, "NOT SUPPORTED"])
+            click.echo("Error! The {} is not available on the platform.\n" \
+            "Number of supported PSU - {}.".format(psu_name, platform_psuutil.get_num_psus()))
```
jleveque: Change to `"Number of supported PSU: {}"`

Replacing hyphen and preceding space with a colon and removing the trailing period will make parsing the output easier, if need be.

jleveque: Too late; already merged. Can modify later.

andrii-savka: Thanks.
It does not needed now.

andrii-savka:
The changes contains an issue which was fixes in following PR:
https://github.com/Azure/sonic-utilities/pull/164

PR #162, opened 2017-12-06
Build a python3 wheel with sonic_psu

qiluo-msft:
So we can install the wheel in a python3 environment
```
root@8b17db2d733c:/# python3.6
Python 3.6.0 (default, Dec 29 2016, 04:29:02)
[GCC 4.9.2] on linux
>>> from sonic_psu.psu_base import PsuBase
>>>
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,11 +1,17 @@
 import glob
+import sys
 from setuptools import setup
 import unittest
 
+# For python3 wheel, we currently don't need test
+# TODO: build python3 wheel of all the dependencies, and remove conditional test
 def get_test_suite():
-    test_loader = unittest.TestLoader()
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -1,11 +1,17 @@
 import glob
+import sys
 from setuptools import setup
 import unittest
 
+# For python3 wheel, we currently don't need test
+# TODO: build python3 wheel of all the dependencies, and remove conditional test
```
jleveque: Do we really plan to do this in the future?

It is my understanding that this is a temporary solution until we break the sonic Python modules out into their own package, separate from sonic-utilities, then we will build the modules package for both Python 2 and 3, correct?

qiluo-msft: OK. This is an alternative, and we can remove TODO if either is achieved.

PR #177, opened 2017-12-19
[acl_loader]: Add status column to session table

marian-pritsak:


qiluo-msft:
This commit will break below command. It hangs and never exits.
```
/usr/bin/python /usr/bin/acl-loader update full /tmp/acltb_test_rules_allow_all.json
```

Further more, it will break ACL Test playbook. It also hangs.

PR #181, opened 2017-12-21
[sonic-installer]: Align command name style

marian-pritsak:
Other commands also use dash instead of underscore


marian-pritsak:
Will it be ok to create symlink with old name?

lguohan:
symbol link is ok.

PR #188, opened 2018-01-18
[reboot] While rebooting device, execute platform specific reboot tool when available

yxieca:
**- What I did**
Reboot device with platform specific tool when available.

**- How I did it**
- Added a script under /usr/bin/reboot, this script will be executed when reboot is invoked without absolute path.
- The script checks if <device path>/platform_reboot tool is available and executable. If so, use this tool to reboot device.
- Update fast-reboot to call native Linux reboot with absolute path.

**- How to verify it**
- Executed the script on device with platform_reboot available, and the tool was called.
- Executed the script on device without platform_reboot tool or not with execution permission, Linux native reboot was called.

### DIFF
--- a/scripts/reboot
+++ b/scripts/reboot
@@ -1,0 +1,25 @@
+#! /bin/bash
+
+function stop_sonic_services()
+{
+    echo "Stopping sonic services..."
+    systemctl stop swss
+    systemctl stop teamd
+    systemctl stop bgp
+    systemctl stop lldp
+}
+
+# Obtain our platform as we will mount directories with these names in each docker
+PLATFORM=`sonic-cfggen -v platform`
+
+DEVPATH="/usr/share/sonic/device"
+REBOOT="platform_reboot"
+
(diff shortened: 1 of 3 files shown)


pavel-shirshov:
About sync. The sync command doesn't guarantee that you got your disk in sync state when the command returns. You need to wait until all disk are synced. Usually 'sleep 3' works in 90% of time.
But you can also read information about the sync operation progress here:
https://unix.stackexchange.com/questions/48235/can-i-watch-the-progress-of-a-sync-operation

scripts/reboot
```diff
@@ -0,0 +1,26 @@
+#! /bin/bash
+
+function stop_sonic_services()
+{
+    echo "Stopping sonic services..."
+    systemctl stop swss
+    systemctl stop teamd
+    systemctl stop bgp
+    systemctl stop lldp
```
lguohan: why only four dockers?

lguohan: snmp docker?

yxieca: Added stop snmp service to the list.

Thanks,
Ying

PR #189, opened 2018-01-18
[sonic_install] sync filesystem after file copy is done

yxieca:
**- What I did**
sync file system after all copies and installations are done.

**- How to verify it**
With platform specific reboot tool power cycling devices, it is highly repeatable that some file would be missing after sonic_install.

The fix is verified under the same circumstances.

The issue turns out not related to this sync. But putting the sync at the end of the execution list by itself is a good practice anyways.

### DIFF
--- a/sonic_installer/main.py
+++ b/sonic_installer/main.py
@@ -189,9 +189,11 @@ def install(url):
         os.chmod(image_path, stat.S_IXUSR)
         run_command(image_path)
         run_command('grub-set-default --boot-directory=' + HOST_PATH + ' 0')
-    run_command("sync")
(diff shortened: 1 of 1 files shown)


pavel-shirshov:
About sync. The sync command doesn't guarantee that you got your disk in sync state when the command returns. You need to wait until all disk are synced. Usually 'sleep 3' works in 90% of time.
But you can also read information about the sync operation progress here:
https://unix.stackexchange.com/questions/48235/can-i-watch-the-progress-of-a-sync-operation

yxieca:
Thanks for pointing that out. Pavel! I added a 3 second delay after sync.

Regards,
Ying

PR #192, opened 2018-01-25
[pfcwd]: add command to set pfcwd polling interval

sihuihan88:

**- What I did**
Add command to set pfcwd polling interval
**- How I did it**
Add set poll interval command
**- How to verify it**
Test on DUT
```
root@sonic:/home/admin# pfcwd show config                              
       PORT    ACTION    DETECTION TIME    RESTORATION TIME
-----------  --------  ----------------  ------------------
Ethernet112      drop               300                3000
```
```
root@sonic:/home/admin# pfcwd show config        
Changed polling interval to 100ms                           
       PORT    ACTION    DETECTION TIME    RESTORATION TIME 
-----------  --------  ----------------  ------------------ 
Ethernet112      drop               300                3000 
```
-->

### DIFF
--- a/pfcwd/main.py
+++ b/pfcwd/main.py
@@ -92,7 +92,9 @@ def config(ports):
             line = config_entry.get(config[1], config[2])
             config_list.append(line)
         table.append([port] + config_list)
-
+    poll_interval = configdb.get_entry( 'PFC_WD_TABLE', 'GLOBAL').get('POLL_INTERVAL')
+    if poll_interval is not None:
+        click.echo("Changed polling interval to " + poll_interval + "ms")
     click.echo(tabulate(table, CONFIG_HEADER, stralign='right', numalign='right', tablefmt='simple'))
 
 # Start WD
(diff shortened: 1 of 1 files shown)


sihuihan88:
depends on https://github.com/Azure/sonic-swss/pull/435

pfcwd/main.py
```diff
@@ -92,7 +92,8 @@ def config(ports):
             line = config_entry.get(config[1], config[2])
             config_list.append(line)
         table.append([port] + config_list)
-
+    poll_interval = configdb.get_entry( 'PFC_WD_TABLE', 'GLOBAL').get('POLL_INTERVAL')
```
marian-pritsak: Does someone write default value there or should we check if it's not empty?

pfcwd/main.py
```diff
@@ -127,6 +128,21 @@ def start(action, restoration_time, ports, detection_time):
...
 
+# Set WD poll interval
+@cli.command()
+@click.option('--poll_interval', '-p', type=click.IntRange(100, 3000))
+def set(poll_interval):
+    """ Set PFC watchdog counter polling interval """
+    configdb = swsssdk.ConfigDBConnector()
+    configdb.connect()
+    countersdb = swsssdk.SonicV2Connector(host='127.0.0.1')
+    countersdb.connect(countersdb.COUNTERS_DB)
```
marian-pritsak: No need to connect to COUNTERS_DB in this function

pfcwd/main.py
```diff
@@ -127,6 +128,21 @@ def start(action, restoration_time, ports, detection_time):
         configdb.mod_entry("PFC_WD_TABLE", port, None)
         configdb.mod_entry("PFC_WD_TABLE", port, pfcwd_info)
 
+# Set WD poll interval
+@cli.command()
+@click.option('--poll_interval', '-p', type=click.IntRange(100, 3000))
```
marian-pritsak: Argument instead of an option?
Options are usually optional, and it's not the case for poll interval,

pfcwd/main.py
sihuihan88: changed to display only when the value is changed.

PR #212, opened 2018-02-27
[Fast-reboot]: Gracefully shutdown syncd in fast-reboot

pavel-shirshov:
**- What I did**
Add a feature to gracefully shutdown syncd

**- How I did it**
Using syncd_request_shutdown utility

**- How to verify it**
Run fast-reboot test

-->

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -41,9 +41,6 @@ docker kill lldp > /dev/null
 # Kill teamd, otherwise it gets down all LAGs
 docker kill teamd > /dev/null
 
-# Kill swss before stopping syncd
-docker kill swss > /dev/null
-
 # Gracefully stop syncd
 docker exec -ti syncd /usr/bin/syncd_request_shutdown --cold > /dev/null
 
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -41,14 +41,18 @@ docker kill lldp > /dev/null
 # Kill teamd, otherwise it gets down all LAGs
 docker kill teamd > /dev/null
 
-# Gracefully stop syncd
-docker exec -ti syncd /usr/bin/syncd_request_shutdown --cold > /dev/null
+# syncd graceful stop is supported only for Broadcoms platforms only for now
+if [[ "$sonic_asic_type" = 'broadcom' ]];
+then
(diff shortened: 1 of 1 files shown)


scripts/fast-reboot
```diff
@@ -41,6 +41,18 @@ docker kill lldp > /dev/null
 # Kill teamd, otherwise it gets down all LAGs
 docker kill teamd > /dev/null
 
+# Kill swss before stopping syncd
+docker kill swss > /dev/null
+
+# Gracefully stop syncd
+docker exec -ti syncd /usr/bin/syncd_request_shutdown --cold > /dev/null
```
lguohan: will this break the mellanox platform?

scripts/fast-reboot
```diff
@@ -41,6 +41,15 @@ docker kill lldp > /dev/null
 # Kill teamd, otherwise it gets down all LAGs
 docker kill teamd > /dev/null
 
+# Gracefully stop syncd
+docker exec -ti syncd /usr/bin/syncd_request_shutdown --cold > /dev/null
+
+# Check it was stopped
+until [[ $(docker top syncd | grep -q /usr/bin/syncd) -eq 0 ]]
```
yxieca: I think you need to s/until/while/

jleveque: Both `grep -q` and `grep -c` will return 0 if any matches are found, and will return 1 if no matches are found.

Therefore, either are applicable, however, the condition needs to be changed to either

`until [[ $(docker top syncd | grep -q /usr/bin/syncd) -eq 1 ]]`

or 

`while [[ $(docker top syncd | grep -q /usr/bin/syncd) -eq 0 ]]`

yxieca: $(...) is getting the output of the execution. Therefore, with -q it is always ''.

With -c, the output would be how many lines find. We want to wait until there is 0 lines.

jleveque: You're right, Ying. For some reason I was stuck on the return value, not the stdout output.

scripts/fast-reboot
yxieca: I see. I missed the other change. Sorry.

Regards,
Ying

PR #213, opened 2018-02-28
[PFCWD]: set default configuration when enabled by default

sihuihan88:

**- What I did**
Set default configuration when enabled by default
**- How I did it**
Implement the pfcwd start_default function to set default configuration.
**- How to verify it**
Test on DUT

-->

### DIFF
--- a/pfcwd/main.py
+++ b/pfcwd/main.py
@@ -42,6 +42,17 @@ def get_all_ports(db):
     port_names = db.get_all(db.COUNTERS_DB, 'COUNTERS_PORT_NAME_MAP')
     return natsorted(port_names.keys())
 
+def get_server_facing_ports(db):
+    candidates = db.get_table('DEVICE_NEIGHBOR')
+    server_facing_ports = []
+    for port in candidates.keys():
+        neighbor = db.get_entry('DEVICE_NEIGHBOR_METADATA', candidates[port]['name'])
+        if neighbor and neighbor['type'].lower() == 'server':
+            server_facing_ports.append(port)
+    if not server_facing_ports:
+        server_facing_ports = [p[1] for p in db.get_table('VLAN_MEMBER').keys()]
+    return server_facing_ports
+
 # Show commands
 @cli.group()
 def show():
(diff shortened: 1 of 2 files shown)


pfcwd/main.py
```diff
@@ -162,5 +169,34 @@ def stop(ports):
...
+
+    pfcwd_info = {
+        'detection_time': DEFAULT_DETECTION_TIME * max(port_num/DEFAULT_PORT_NUM, 1),
+        'restoration_time': DEFAULT_RESTORATION_TIME * max(port_num/DEFAULT_PORT_NUM, 1),
+        'action': DEFAULT_ACTION
+    }
+
+    for port in vlan_members:
+        configdb.mod_entry("PFC_WD_TABLE", port, None)
+        configdb.mod_entry("PFC_WD_TABLE", port, pfcwd_info)
```
lguohan: set_entry will remove extra fields. use set_entry instead of mod_entry.

pfcwd/main.py
```diff
@@ -162,5 +169,34 @@ def stop(ports):
...
+       return
+    device_type = configdb.get_entry('DEVICE_METADATA', 'localhost').get('type')
+    if device_type != "ToRRouter":
+        return
+    port_num = len(configdb.get_table('PORT').keys())
+    vlan_members = [p[1] for p in configdb.get_table('VLAN_MEMBER').keys()]
+
+    pfcwd_info = {
+        'detection_time': DEFAULT_DETECTION_TIME * max(port_num/DEFAULT_PORT_NUM, 1),
+        'restoration_time': DEFAULT_RESTORATION_TIME * max(port_num/DEFAULT_PORT_NUM, 1),
```
lguohan: let's not do the calculation here, just use the DEFAULT_DETECTION_TIME. The reason is that the detection time is based on external network requirement, it should not be set based on the port num in the system.

pfcwd/main.py
```diff
@@ -162,5 +169,34 @@ def stop(ports):
...
+        'action': DEFAULT_ACTION
+    }
+
+    for port in vlan_members:
```
lguohan: using vlan to determine whether the port is connected to server is the last resort, should use neighbor table first. check this commit.

https://github.com/Azure/sonic-buildimage/commit/720c71399c9b4d7d0e5a6083b3f63b887f3f6201

pfcwd/main.py
```diff
@@ -5,6 +5,13 @@
 from tabulate import tabulate
 from natsort import natsorted
 
+# Default configuration
+DEFAULT_DETECTION_TIME = 200
+DEFAULT_RESTORATION_TIME = 200
+DEFAULT_POLL_INTERVAL = 200
+DEFAULT_PORT_NUM = 32
```
lguohan: we do not need this.

pfcwd/main.py
sihuihan88: I checked the above commit and I believe it also uses vlan to determine the server facing port as we don't have neighbor table for servers.

PR #214, opened 2018-03-01
Add fast-reboot support for Aboot based images

Staphylo:
**- What I did**

Added support of fast-reboot for Arista platform by adding support of Aboot in fast-reboot.
Also changed sonic_installer to use `install=y` instead of `sonic_upgrade=1`.
The boot0 will stay backward compatible with previous `sonic_installer` versions.

**- How to verify it**

Running `sonic_installer install` and then `fast-reboot` should show kexec working.
It is assumed that all ONIE based platform define onie_platform= in /host/machine.conf.

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -16,9 +16,18 @@ fi
 
 # Kernel and initrd image
 NEXT_SONIC_IMAGE=$(sonic_installer list | grep "Next: " | cut -d ' ' -f 2)
-KERNEL_OPTIONS=$(cat /host/grub/grub.cfg | sed "/$NEXT_SONIC_IMAGE'/,/}/"'!'"g" | grep linux)
-KERNEL_IMAGE="/host$(echo $KERNEL_OPTIONS | cut -d ' ' -f 2)"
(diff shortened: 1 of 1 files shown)


sonic_installer/main.py
```diff
@@ -184,7 +184,7 @@ def install(url):
 
     if get_image_type() == IMAGE_TYPE_ABOOT:
         run_command("/usr/bin/unzip -od /tmp %s boot0" % image_path)
-        run_command("swipath=%s target_path=/host sonic_upgrade=1 . /tmp/boot0" % image_path)
+        run_command("swipath=%s target_path=/host install=y . /tmp/boot0" % image_path)
```
lguohan: are you going to change boot0.j2 in sonic-buildimage to support install option? current I only see sonic_upgrade there.

lguohan: you do not really need to change to install=y. keep it as it is is better, so that we can merge your PR without the boot0.j2 change. there is no coordination, otherwise, we need to coordinate this PR with the boot0.j2 PR.

Staphylo: I was planning on the pull request on sonic-buildimage that bumps the sonic-utilities and adds the boot0.j2 change as soon as this one merged.
But you are correct I can leave the sonic_upgrade=1 as it is. Will update this PR.

PR #217, opened 2018-03-01
[AAA] Support login(ascii) authentication type

taoyl-ms:
**- What I did**
Add support to login=login (ascii) authentication type to tacacs+


taoyl-ms:
@liuqu, could you help review?

PR #218, opened 2018-03-06
[config bgp] Convert user input ipv6 addr to lower case before comparing

taoyl-ms:
**- What I did**
1. Do case conversion only for user input but not for keys queried from configDB. 
1. Use lower case instead of upper case.
We will make a change in minigraph parser as well to make sure all bgp session ipv6 addresses are written into configDB in lower case.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -77,8 +77,8 @@ def _change_bgp_session_status(ipaddr_or_hostname, status, verbose):
     """
     ip_addrs = []
 
-    # Convert the IP address to lowercase because IPv6 addresses will be stored
-    # in ConfigDB with all lowercase alphabet characters during minigraph parsing
+    # If we were passed an IP address, convert it to lowercase because IPv6 addresses were
+    # stored in ConfigDB with all lowercase alphabet characters during minigraph parsing
     if _is_neighbor_ipaddress(ipaddr_or_hostname.lower()):
         ip_addrs.append(ipaddr_or_hostname.lower())
     else:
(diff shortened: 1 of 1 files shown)


config/main.py
```diff
@@ -73,16 +70,17 @@ def _change_bgp_session_status_by_addr(ipaddress, status, verbose):
     config_db = ConfigDBConnector()
     config_db.connect()
 
-    # Convert the IP address to uppercase because IPv6 addresses are stored
-    # in ConfigDB with all uppercase alphabet characters
-    config_db.mod_entry('bgp_neighbor', ipaddress.upper(), {'admin_status': status})
+    config_db.mod_entry('bgp_neighbor', ipaddress, {'admin_status': status})
```
jleveque: You will either need to add a `.lower()` call on ipaddress here or on line 95 (`_change_bgp_session_status_by_addr(ip_addr.lower(), status, verbose)`)

However, since we will only need to covert the IP address to lowercase when getting or modifying ConfigDB entries, I feel it is best to add the `.lower()` calls when passing into the config_db methods (i.e., simply change my two `.upper()` calls to `.lower()` calls).

config/main.py
```diff
@@ -73,16 +70,17 @@ def _change_bgp_session_status_by_addr(ipaddress, status, verbose):
...
 
 def _change_bgp_session_status(ipaddr_or_hostname, status, verbose):
     """Start up or shut down BGP session by IP address or hostname
     """
     ip_addrs = []
-    if _is_neighbor_ipaddress(ipaddr_or_hostname):
-        ip_addrs.append(ipaddr_or_hostname)
+
+    # Convert the IP address to lowercase because IPv6 addresses will be stored
```
jleveque: Maybe change comment to `If we were passed an IP address, convert it to lowercase because...`

PR #220, opened 2018-03-16
add fast-reboot support for nephos platform by stop kernel modules

simonJi2018:
**- What I did**
add fast-reboot support for nephos platform 
**- How I did it**
Stop kernel modules for Nephos platform
**- How to verify it**
Passed test.

-->

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -63,6 +63,12 @@ then
   systemctl stop "$service_name"
 fi
 
+# Stop kernel modules for Nephos platform
+if [ "$sonic_asic_type" = 'nephos' ];
+then
+  systemctl stop nps-modules-`uname -r`.service
+fi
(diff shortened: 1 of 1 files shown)


simonJi2018:
add fast-reboot support for nephos platform

scripts/fast-reboot
```diff
@@ -63,6 +63,12 @@ then
   systemctl stop "$service_name"
 fi
 
+# Stop kernel modules for Nephos platform
+if [ "$sonic_asic_type" = 'nephos' ];
```
lguohan: can you use [[  ]] instead of [ ]?

https://stackoverflow.com/questions/3427872/whats-the-difference-between-and-in-bash

simonJi2018:
hi，all
what should I do next ？

PR #221, opened 2018-03-21
[crm]: Fix failures in CLI show commands

volodymyrsamotiy:
Signed-off-by: Volodymyr Samotiy <volodymyrs@mellanox.com>

**- What I did**
Fixed issue "CRM module throws error" (https://github.com/Azure/SONiC/issues/159)
**- How I did it**
Changed output to empty line instead of throwing error when retrieved nothing from config DB
**- How to verify it**
Execute CRM show commands and observe that error is not shown

### DIFF
--- a/crm/main.py
+++ b/crm/main.py
@@ -32,6 +32,8 @@ class Crm:
 
         if crm_info:
             print '\nPolling Interval: ' + crm_info['polling_interval'] + ' second(s)\n'
+        else:
+            print '\nError! Could not get CRM configuration.\n'
 
     def show_thresholds(self, resource):
         """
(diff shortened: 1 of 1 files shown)


crm/main.py
```diff
@@ -30,7 +30,8 @@ def show_summary(self):
 
         crm_info = configdb.get_entry('CRM', 'Config')
 
-        print '\nPolling Interval: ' + crm_info['polling_interval'] + ' second(s)\n'
+        if crm_info:
+            print '\nPolling Interval: ' + crm_info['polling_interval'] + ' second(s)\n'
```
prsunny: We would need to print some failure message if crm_info is _null_ for some reason!

crm/main.py
```diff
@@ -68,12 +70,13 @@ def show_resources(self, resource):
         header = ("Resource Name", "Used Count", "Available Count")
         data = []
 
-        if resource == 'all':
-            for res in ["ipv4_route", "ipv6_route", "ipv4_nexthop", "ipv6_nexthop", "ipv4_neighbor", "ipv6_neighbor",
-                        "nexthop_group_member", "nexthop_group", "fdb_entry"]:
-                data.append([res, crm_stats['crm_stats_' + res + "_used"], crm_stats['crm_stats_' + res + "_available"]])
-        else:
-            data.append([resource, crm_stats['crm_stats_' + resource + "_used"], crm_stats['crm_stats_' + resource + "_available"]])
+        if crm_stats:
```
prsunny: The _else_ part has to display some message stating crm_stats is _not_ready_ since it would be populated only after the polling interval

volodymyrsamotiy:
Added printing of appropriate messages when get empty entry from DB

PR #223, opened 2018-03-28
[reboot] reduce stop service to only stop syncd

yxieca:
**- What I did**
When the service dockers are in some rare failure situations, stopping
service might get stuck. To make sure reboot always move ahead, we
could request syncd to stop, wait some time, then proceed with reboot.

**- How to verify it**
I executed a few hundreds of reboots with the new script on my dut.

### DIFF
--- a/scripts/reboot
+++ b/scripts/reboot
@@ -2,12 +2,9 @@
 
 function stop_sonic_services()
 {
-    echo "Stopping sonic services..."
-    systemctl stop swss
-    systemctl stop teamd
-    systemctl stop bgp
-    systemctl stop lldp
-    systemctl stop snmp
+    echo "Stopping syncd..."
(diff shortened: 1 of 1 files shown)


yxieca:
Just added 200+ reboots to the test record for this change.

scripts/reboot
```diff
@@ -2,12 +2,9 @@
...
 {
-    echo "Stopping sonic services..."
-    systemctl stop swss
-    systemctl stop teamd
-    systemctl stop bgp
-    systemctl stop lldp
-    systemctl stop snmp
+    echo "Stopping syncd..."
+    docker exec -it syncd /usr/bin/syncd_request_shutdown --cold > /dev/null
+    sleep 3
```
lguohan: does this work on Mellanox platform?

yxieca: It works on all platforms. Fast-reboot is also calling this to terminate syncd.

Regards,
Ying

PR #225, opened 2018-03-29
[Fast-Reboot]: Check user id for fast-reboot-dump.py script

pavel-shirshov:
**- What I did**
Add an uid check for fast-reboot-dump.py script

**- How I did it**
Using standard python modules

**- How to verify it**
Try to run fast-reboot-dump.py with regular user

-->

### DIFF
--- a/scripts/fast-reboot-dump.py
+++ b/scripts/fast-reboot-dump.py
@@ -261,6 +261,9 @@ def generate_default_route_entries(filename):
 
 
 def main():
+    if os.geteuid() != 0:
+        print >> sys.stderr, 'Please run as root'
+        sys.exit(0)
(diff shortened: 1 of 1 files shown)


scripts/fast-reboot-dump.py
```diff
@@ -261,6 +261,9 @@ def generate_default_route_entries(filename):
 
 
 def main():
+    if os.geteuid() != 0:
+        print >> sys.stderr, 'Please run as root'
+        sys.exit(0)
```
lguohan: should exit 1?

pavel-shirshov: Fixed

PR #239, opened 2018-04-12
Workaround rmtree bug on non-ascii filenames in python2

qiluo-msft:


keboliu:
it works well on mlnx platform with this fix, rpc image can be successfully removed.

PR #242, opened 2018-04-19
[acl-loader] Fix and refactor incremental update

jleveque:
- Incremental update was not working as intended. `cmp(self.rules_info[key], self.rules_db_info[key])` would consistently fail because some string elements in `rules_info` were encoded in Unicode when loading the ACLs from a file, whereas all elements in `rules_db_info`, pulled from ConfigDB, were encoded in ASCII.
- Now, string elements in `rules_db_info` are converted to ASCII, which causes the compare operation to work properly.
- Also modified incremental update to only perform an incremental update on control plane ACLs. A full update is still performed on dataplane ACLs temporarily, until we fully test that we can insert ACLs (and shift existing ACLs) in the ASICs.

### DIFF
--- a/acl_loader/main.py
+++ b/acl_loader/main.py
@@ -248,10 +248,10 @@ class AclLoader(object):
                 rule_props["IP_PROTOCOL"] = rule.ip.config.protocol
 
         if rule.ip.config.source_ip_address:
-            rule_props["SRC_IP"] = rule.ip.config.source_ip_address
+            rule_props["SRC_IP"] = rule.ip.config.source_ip_address.encode("ascii")
 
         if rule.ip.config.destination_ip_address:
-            rule_props["DST_IP"] = rule.ip.config.destination_ip_address
+            rule_props["DST_IP"] = rule.ip.config.destination_ip_address.encode("ascii")
 
         # NOTE: DSCP is available only for MIRROR table
         if self.is_table_mirror(table_name):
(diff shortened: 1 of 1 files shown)


acl_loader/main.py
```diff
@@ -385,21 +385,53 @@ def incremental_update(self):
...
-        for key in added_rules:
+        for key in added_controlplane_rules:
             self.configdb.mod_entry(self.ACL_RULE, key, self.rules_info[key])
 
-        for key in existing_rules:
-            if cmp(self.rules_info[key], self.rules_db_info[key]):
+        for key in existing_controlplane_rules:
+            if cmp(self.rules_info[key], self.rules_db_info[key]) != 0:
                 self.configdb.mod_entry(self.ACL_RULE, key, None)
                 self.configdb.mod_entry(self.ACL_RULE, key, self.rules_info[key])
```
lguohan: self.configdb.set_entry(self.ACL_RULE, key, self.rules_info[key])

acl_loader/main.py
```diff
@@ -385,21 +385,53 @@ def incremental_update(self):
...
+        removed_controlplane_rules = current_controlplane_rules.difference(new_controlplane_rules)
+        existing_controlplane_rules = new_rules.intersection(current_controlplane_rules)
 
-        for key in removed_rules:
+        for key in removed_controlplane_rules:
             self.configdb.mod_entry(self.ACL_RULE, key, None)
 
-        for key in added_rules:
+        for key in added_controlplane_rules:
             self.configdb.mod_entry(self.ACL_RULE, key, self.rules_info[key])
```
lguohan: move added rules above removed rules

PR #244, opened 2018-04-20
[show] Now displaying output while command is running (e.g., 'tail'); Remove 'it' flags from 'docker exec' calls

jleveque:
- Previously, command output wasn't being displayed until the command finished running. This would cause programs like 'tail' (e.g., `sudo tail -F /var/log/syslog`) to not display any output.

- For all `docker exec ...` calls, I removed the `i` and `t` flags, as the `t` flag would create an unnecessary terminal with a narrow width which would truncate log lines. The `i` flag is not necessary as we are simply running non-interactive commands.

- I also have an updated sudoers file which I will add in sonic-buildimage when updating this submodule.

### DIFF
--- a/show/main.py
+++ b/show/main.py
@@ -117,22 +117,28 @@ def run_command(command, display_cmd=False):
         click.echo(click.style("Command: ", fg='cyan') + click.style(command, fg='green'))
 
     proc = subprocess.Popen(command, shell=True, stdout=subprocess.PIPE)
-    (out, err) = proc.communicate()
 
-    try:
-        click.echo(out)
-    except IOError as e:
-        # In our version of Click (v6.6), click.echo() and click.echo_via_pager() do not properly handle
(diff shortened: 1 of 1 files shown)


show/main.py
```diff
@@ -117,22 +117,28 @@ def run_command(command, display_cmd=False):
...
+        if output:
+            try:
+                click.echo(output.rstrip('\n'))
+            except IOError as e:
+                # In our version of Click (v6.6), click.echo() and click.echo_via_pager() do not properly handle
+                # SIGPIPE, and if a pipe is broken before all output is processed (e.g., pipe output to 'head'),
+                # it will result in a stack trace. This is apparently fixed upstream, but for now, we silently
+                # ignore SIGPIPE here.
+                if e.errno == errno.EPIPE:
+                    print("Got here")
```
lguohan: can we get a more descriptive message here?

jleveque: That was a debug message I missed. It is now removed.

PR #256, opened 2018-05-16
Fix teamshow issues due to dependencies on minigraph

yangbashuang:
**- What I did**
Remove the dependencies on minigraph for certain commands.
**- How I did it**
Use configDB values instead.
**- How to verify it**
show interfaces portchannel
root@switch1:/home/admin# show interfaces portchannel
Command: teamshow
  No.  Team Dev       Protocol     Ports
-----  -------------  -----------  ---------------------------

root@switch1:/home/admin#

no portchannel displayed, because no configuration in minigraph but in config_db.json
root@switch1:/home/admin# show interfaces portchannel
Command: teamshow
  No.  Team Dev       Protocol     Ports
-----  -------------  -----------  ---------------------------
   01  PortChannel01  LACP(A)(Dw)  Ethernet11(D) Ethernet12(D)
   02  PortChannel02  LACP(A)(Dw)  Ethernet14(D) Ethernet13(D)

root@switch1:/home/admin#
-->

### DIFF
--- a/scripts/teamshow
+++ b/scripts/teamshow
@@ -46,10 +46,8 @@ class Teamshow(object):
         """
             Get the portchannel names from minigraph.
         """
-        minigraph_path = '/etc/sonic/minigraph.xml'
-        machine_info = get_machine_info()
-        platform_info = get_platform_info(machine_info)
-        self.teams = parse_xml(minigraph_path, platform = platform_info)['PORTCHANNEL'].keys();
+        team_keys = self.db.keys(self.db.APPL_DB, PORT_CHANNEL_TABLE_PREFIX+"*")
+        self.teams = [key[len(PORT_CHANNEL_TABLE_PREFIX):] for key in team_keys]
 
     def get_portchannel_status(self, port_channel_name):
         """


scripts/teamshow
```diff
@@ -46,10 +46,8 @@ class Teamshow(object):
         """
             Get the portchannel names from minigraph.
         """
-        minigraph_path = '/etc/sonic/minigraph.xml'
-        machine_info = get_machine_info()
-        platform_info = get_platform_info(machine_info)
-        self.teams = parse_xml(minigraph_path, platform = platform_info)['PORTCHANNEL'].keys();
+        team_keys = self.db.keys(self.db.APPL_DB, PORT_CHANNEL_TABLE_PREFIX+"*")
```
taoyl-ms: I think we should get those team_keys from CFG db instead of APP db, as we need the **expected** port channel list instead of actual list here.

scripts/teamshow
```diff
@@ -46,10 +46,8 @@ class Teamshow(object):
         """
             Get the portchannel names from minigraph.
         """
```
sihuihan88: Can you update the comment of "Get the portchannel names from minigraph" accordingly?

yangbashuang: hi, please review again, thanks

scripts/teamshow
```diff
@@ -44,12 +43,10 @@ class Teamshow(object):
...
-            Get the portchannel names from minigraph.
+            Get the portchannel names from database.
         """
-        minigraph_path = '/etc/sonic/minigraph.xml'
-        machine_info = get_machine_info()
-        platform_info = get_platform_info(machine_info)
-        self.teams = parse_xml(minigraph_path, platform = platform_info)['PORTCHANNEL'].keys();
+        team_keys = self.db.keys(self.db.APPL_DB, PORT_CHANNEL_TABLE_PREFIX+"*")
+        self.teams = [key[len(PORT_CHANNEL_TABLE_PREFIX):] for key in team_keys]
```
taoyl-ms: I think `team_keys` here should be read from CFG_DB instead of APPL_DB, as it should presents expected port channels instead of actual running ones. If a portchannel is defined in CFG_DB but not available in APPL_DB, there should be a line with status N/A.

yangbashuang: indeed, It is more reasonable to read from CONFIG_DB, thanks very much

yangbashuang:
@jleveque please review

scripts/teamshow
```diff
@@ -24,12 +24,12 @@ import swsssdk
 import subprocess
 import sys
 from tabulate import tabulate
-from minigraph import parse_xml
 from natsort import natsorted
 from sonic_platform import get_machine_info
 from sonic_platform import get_platform_info
 
 PORT_CHANNEL_TABLE_PREFIX = "LAG_TABLE:"
```
jleveque: Now that you've added `PORT_CHANNEL_CFG_TABLE_PREFIX` below, I believe this constant should be renamed to `PORT_CHANNEL_APPL_TABLE_PREFIX` for consistency and to eliminate confusion.

yangbashuang: thanks, changes are done, please review

PR #259, opened 2018-05-25
[config] Log invocation of config commands to syslog

jleveque:
- Also add a few missing docstrings

These log message can be useful in debugging, to understand if issues may have been caused by a user modifying the device configuration.


jleveque:
Retest this please

PR #260, opened 2018-05-30
[config] Add support for accepting port alias in lieu of port name

paavaanan:
[portalias-logs.txt](https://github.com/Azure/sonic-utilities/files/2053930/portalias-logs.txt)

**- What I did**
Implemented portalias feature for two basic commands "config interface shutdown/startup"

-Introduced new CLI to configure the interface mode: 
  -config interface_mode "alias" 
  -config interface_mode "defaut"

User can stick to any mode. Mode selection is unique for each sessions. Default mode stick to sonic ethernet interface style and Alias mode stick to individual switch vendor interface style. User selection  sticks to the current logged session. If multiple login session do exists it maintains the user selection seperately.  

**- How I did it**

1. Introduced convert_interface_name() function to toggle interfaces between alias/default mode.
2. Included a termporary file for each individual session to maintain the mode selection.
3. interface_mode.sh script is included in /etc/profile.d to create a unique file for each session at the time of login.

**- How to verify it**
1. config interface_mode  alias 
2. config interface shutdown fortyGigE1/1/7
3. config interface_mode  default 
4. config interface startup Ethernet6    

--> interface_mode.sh is raised in https://github.com/Azure/sonic-buildimage/pull/1751


config/main.py
```diff
@@ -33,6 +35,64 @@ def run_command(command, display_cmd=False, ignore_error=False):
     if proc.returncode != 0 and not ignore_error:
         sys.exit(proc.returncode)
 
+def convert_interface_name(interface_name):
```
jleveque: This function will blindly convert a port name to a port alias and vice-versa. This could pose issues if a device shares the same string for both a port name and a port alias, as it will always find a match on line 49 and return that port's alias.

I think it is better to have two functions with distinct functionality: `interface_name_to_alias(interface_name)` and `interface_alias_to_name(interface_alias)`.

config/main.py
```diff
@@ -33,6 +35,64 @@ def run_command(command, display_cmd=False, ignore_error=False):
...
+        for port_name in natsorted(port_dict.keys()):
+            if interface_name == port_name:
+                return port_dict[port_name]['alias']
+            elif interface_name == port_dict[port_name]['alias']:
+                return port_name
+        print "Invalid interface {}".format(interface_name)
+
+def set_interface_mode(mode):
+    tmp = "/tmp/"
+    output = subprocess.check_output(['tty']).strip('\n')
```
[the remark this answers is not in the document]
paavaanan: Agreed. Initially we approched the issue in the above context. Since python is the child process of parent bash (as like for any sessios) we can inherit the status of enviroment variable say SONIC_CLI_ALIAS_MODE to python but we can't revert back the state to TRUE/FALSE back to bash environment variable. .  The change do exists until the script terminates. SONIC_CLI_ALIAS_MODE value cannot be exported to parent bash.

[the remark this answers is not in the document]
jleveque: Ah, yes ... you can't modify the environment of the parent shell. That's a problem. If we wanted to proceed with this approach, I guess we would need to prompt the user to log out and log back in again for the change to take effect; I think that's quite an inconvenience.

@lguohan: Do you have any alternative suggestions? Otherwise, I believe the tty-based approach may be our best bet.

config/main.py
paavaanan: Introduced interface_alias_to_name() to convert the alias to default mode. In the current code change, there is no any need for interface_name_to_alias() conversion. While implementing the feature for other subset of commands if needed will implement on need basis.

config/main.py
```diff
@@ -33,6 +35,64 @@ def run_command(command, display_cmd=False, ignore_error=False):
     if proc.returncode != 0 and not ignore_error:
         sys.exit(proc.returncode)
 
+def interface_alias_to_name(interface_name):
```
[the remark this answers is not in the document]
paavaanan: Okay

paavaanan:
•	Modified the design to user based approach.   Harnessed the power of /etc/sudoers and /etc/environment file to retain the environment variable value even in sudo sessions. So, when the user request the change from default mode to alias mode, the action is completed in two phases. 
•	config/main.py script fetch the current value of IFMODE and modify as user requested alias mode and write in bashrc.
•	When the user logout and login the session bashrc update the environment variable.
•	This works for both root, and non-root sessions.

paavaanan:
- In this latest commit, addressed review comments.
- Maintained uniformity in naming convention.
- Environement variable IFMODE  alias/default switch code is refined.

[port_alias_config.log](https://github.com/Azure/sonic-utilities/files/2174959/port_alias_config.log)

config/main.py
```diff
@@ -33,6 +35,70 @@ def run_command(command, display_cmd=False, ignore_error=False):
...
+
+def set_interface_mode(mode):
+    """Modify IFMODE env variable in user .bashrc
+    """
+    sudo_user = os.getenv('SUDO_USER')
+    user = os.getenv('USER')
+    set_mode = "IFMODE={}".format(mode)
+
+    if not sudo_user:
+        bashrc = "/{}/.bashrc".format(user)
```
jleveque: The .bashrc path should always be `"/home/{}/.bashrc"`, whether or not the user has called the command with sudo.

I also believe you can simplify lines 75-82 as follows:

```
bashrc_ifmode_line = "IFMODE={}".format(mode)

user = os.getenv('SUDO_USER')

if not user:
    user = os.getenv('USER')

bashrc = "/home/{}/.bashrc".format(user)
...
```

config/main.py
```diff
@@ -33,6 +35,70 @@ def run_command(command, display_cmd=False, ignore_error=False):
...
+        sys.exit(0)
+
+
+
+def set_interface_mode(mode):
+    """Modify IFMODE env variable in user .bashrc
+    """
+    sudo_user = os.getenv('SUDO_USER')
+    user = os.getenv('USER')
+    set_mode = "IFMODE={}".format(mode)
```
jleveque: - The environment variable should have a more descriptive name, like `SONIC_CLI_IFACE_MODE`

- `set_mode` should also have a more descriptive name, like `bashrc_ifacemode_line`.

- Also, you should append a newline at the end to ensure the file ends with a newline.

config/main.py
```diff
@@ -33,6 +35,70 @@ def run_command(command, display_cmd=False, ignore_error=False):
...
+        bashrc = "/{}/.bashrc".format(user)
+    else:
+        bashrc = "/home/{}/.bashrc".format(sudo_user)
+
+    f = open(bashrc,'r')
+    filedata = f.read()
+    f.close()
+
+    if "IFMODE" not in filedata:
+        newdata = filedata + set_mode
```
jleveque: You should probably prepend a newline here to be safe. If the file happens to end with no newline, the `IFMODE` line will be concatenated onto the existing final line.

config/main.py
```diff
@@ -33,6 +35,70 @@ def run_command(command, display_cmd=False, ignore_error=False):
...
+    else:
+        newdata = re.sub(r"IFMODE=\w+",set_mode,filedata)
```
jleveque: Please add spaces after commas.

config/main.py
```diff
@@ -33,6 +35,70 @@ def run_command(command, display_cmd=False, ignore_error=False):
...
+
+    f = open(bashrc,'w')
```
jleveque: Please add space after comma.

config/main.py
```diff
@@ -33,6 +35,70 @@ def run_command(command, display_cmd=False, ignore_error=False):
...
+    f.write(newdata)
+    f.close()
+    print "Please logout and login again!"
+    
+def get_interface_mode():
+    mode = os.getenv('IFMODE')
```
jleveque: What if the environment variable is not set? You should probably return "default" in this case so that this function consistently returns a value.

config/main.py
```diff
@@ -556,5 +647,26 @@ def platform():
 platform.add_command(mlnx.mlnx)
 
 
+#
+# 'interface_mode' group
+#
+
+@cli.group()
+def interface_mode():
+    """Interface mode change tasks"""
```
jleveque: Suggest `"""Modify interface naming mode for interacting with SONiC CLI"""`

config/main.py
```diff
@@ -556,5 +647,26 @@ def platform():
...
+    pass
+
+@interface_mode.command('default')
+def interface_mode_default():
+    """Set interface mode to DEFAULT"""
```
jleveque: Suggest `"""Set CLI interface naming mode to DEFAULT (SONiC port name)"""`

config/main.py
```diff
@@ -556,5 +647,26 @@ def platform():
...
+    alias_mode = "default"
+    set_interface_mode(alias_mode)
+
+@interface_mode.command('alias')
+def interface_mode_alias():
+    """Set interface mode to ALIAS"""
```
jleveque: Suggest `"""Set CLI interface naming mode to ALIAS (Vendor port alias)"""`

config/main.py
```diff
@@ -556,5 +647,26 @@ def platform():
 platform.add_command(mlnx.mlnx)
 
 
+#
+# 'interface_mode' group
+#
+
+@cli.group()
+def interface_mode():
```
jleveque: Suggest change to `interface_naming_mode()`

config/main.py
```diff
@@ -33,6 +35,70 @@ def run_command(command, display_cmd=False, ignore_error=False):
...
+
+    if "IFMODE" not in filedata:
+        newdata = filedata + set_mode
+    else:
+        newdata = re.sub(r"IFMODE=\w+",set_mode,filedata)
+
+    f = open(bashrc,'w')
+    f.write(newdata)
+    f.close()
+    print "Please logout and login again!"
```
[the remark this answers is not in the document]
paavaanan: Agreed

config/main.py
```diff
@@ -534,9 +532,9 @@ def interface():
 @click.option('-v', '--verbose', is_flag=True, help="Enable verbose output")
 def shutdown(interface_name, verbose):
     """Shut down interface"""
-    if get_interface_mode() == "ALIAS": 
-        interface_name = interface_alias_to_name(interface_name)
-    """Shut down interface"""
+    if get_interface_mode() == "alias": 
+        interface_name = interface_alias(interface_name)
```
paavaanan: If interface_alias() returns None. It tries to execute the "ip link <interface_name> down". This is totally invalid.
Why do this needs to be executed with in-appropriate name.

config/main.py
```diff
@@ -33,6 +35,70 @@ def run_command(command, display_cmd=False, ignore_error=False):
...
+    p = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE)
+
+    port_dict = json.loads(p.stdout.read())
+
+    if interface_name is not None:
+        for port_name in natsorted(port_dict.keys()):
+            if interface_name == port_name:
+                return port_dict[port_name]['alias']
+        print "Invalid interface {}".format(interface_name)
+        sys.exit(0)
```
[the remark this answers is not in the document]
paavaanan:
```
def shutdown(interface_name, verbose):
    """Shut down interface"""
    print "lias"
    if get_interface_mode() == "alias":
        interface_name = interface_alias(interface_name)
```
- If we fail to return sys.exit(0) in function interface_alias() it propogates to next line and execute 
"ip link set <interface_nam> down. Where interface_name is invalid.
- This is unnecessary and tends to double opertaion. To avoid this sys.exit(0) introduced. and also it alerts the user about invalid interface.
- If you are skeptical to the exit in  utiltiy function we can shift the exit in parent shutdown().

```
    command = "ip link set {} down".format(interface_name)
    run_command(command, display_cmd=verbose)

```

#

config/main.py
```diff
@@ -33,6 +35,70 @@ def run_command(command, display_cmd=False, ignore_error=False):
     if proc.returncode != 0 and not ignore_error:
         sys.exit(proc.returncode)
 
+def interface_alias(interface_name):
```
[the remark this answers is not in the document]
paavaanan: Thanks for the clarification

config/main.py
```diff
@@ -33,6 +35,70 @@ def run_command(command, display_cmd=False, ignore_error=False):
...
+    p = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE)
+
+    port_dict = json.loads(p.stdout.read())
+
+    if interface_name is not None:
+        for port_name in natsorted(port_dict.keys()):
+            if interface_name == port_dict[port_name]['alias']:
+                return port_name
+        print "Invalid interface {}".format(interface_name)
+        sys.exit(0)
```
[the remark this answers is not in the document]
paavaanan: comment addressed below

config/main.py
```diff
@@ -33,6 +35,70 @@ def run_command(command, display_cmd=False, ignore_error=False):
...
+def interface_name_to_alias(interface_name):
+    """Return alias interface name if default name is given as argument
+    """
+
+    cmd = 'sonic-cfggen -d --var-json "PORT"'
+    p = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE)
+
+    port_dict = json.loads(p.stdout.read())
+
+    if interface_name is not None:
```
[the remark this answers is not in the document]
paavaanan: - The interface_name validation is done by pthon click package at the initial phase of the function it throws missing argument error.

```
admin@sonic:~$ sudo config interface shutdown None
Invalid interface None
````

```
admin@sonic:~$ sudo config interface shutdown
Usage: config interface shutdown [OPTIONS] <interface_name>
Error: Missing argument "interface_name".
```

File: config/main.py (hunk 9 above)
Comment (paavaanan) [reply config/main.py]: Will address this

File: config/main.py (hunk 4 above)
Comment (paavaanan) [reply config/main.py]: - What if root user try to do the change the .bashrc will be in /root/.bashrc. 
- Do we need to adress this scenario?

File: config/main.py (hunk 16 above)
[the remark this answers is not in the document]
Comment (jleveque) [reply config/main.py]: Yes. I think the caller functions should check for `None`, print an error and exit.

File: config/main.py (hunk 4 above)
Comment (jleveque) [reply config/main.py]: We never want to modify the .bashrc of the root user. This is the entire point of first checking to see if `SUDO_USER` is not empty. If SUDO_USER is not empty, we want to modify *that* user's .bashrc. Similarly, we want to check the environment variables of *that* user to determine the mode to use.

File: config/main.py (hunk 16 above)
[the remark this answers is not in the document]
Comment (paavaanan) [reply config/main.py]: Will addres the changes

File: config/main.py (hunk 19 above)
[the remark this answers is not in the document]
Comment (jleveque) [reply config/main.py]: I wasn't asking what happens if a user provides the string "None" as the interface name.

My question is regarding the fact that you have an `if` statement (`if interface_name is not None:`) which returns a value but no corresponding `else` statement. Thus, if `interface_name` is of type None (not the string "None"), this function will not return anything. I think this was overlooked, so I was asking for an explanation to see if this is the behavior you expect.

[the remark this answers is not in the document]
Comment (jleveque) [reply config/main.py]: I wasn't asking what happens if a user provides the string "None" as the interface name.

My question is regarding the fact that you have an `if` statement (`if interface_name is not None:`) which returns a value but no corresponding `else` statement. Thus, if `interface_name` is of type None (not the string "None"), this function will not return anything. I think this was overlooked, so I was asking for an explanation to see if this is the behavior you expect.

File: config/main.py [hunk 20]
```diff
@@ -64,39 +63,42 @@ def interface_name_to_alias(interface_name):
...
+    """Modify SONIC_CLI_IFACE_MODE env variable in user .bashrc
     """
-    sudo_user = os.getenv('SUDO_USER')
-    user = os.getenv('USER')
-    set_mode = "IFMODE={}".format(mode)
-
-    if not sudo_user:
-        bashrc = "/{}/.bashrc".format(user)
+    user = os.getenv('SUDO_USER')
+    bashrc_ifacemode_line = "SONIC_CLI_IFACE_MODE={}".format(mode)
```
jleveque: I think this line is best moved down just above line 87, right before it is used. That will also keep all of the user name determination code together.

config/main.py
```diff
@@ -64,39 +63,42 @@ def interface_name_to_alias(interface_name):
...
-    if "IFMODE" not in filedata:
-        newdata = filedata + set_mode
+    if "SONIC_CLI_IFACE_MODE" not in filedata:
+        newdata = filedata + bashrc_ifacemode_line
+        newdata += "\n"
     else:
-        newdata = re.sub(r"IFMODE=\w+",set_mode,filedata)
-
+        newdata = re.sub(r"SONIC_CLI_IFACE_MODE=\w+", bashrc_ifacemode_line, filedata)
+
```
jleveque: Delete whitespace.

config/main.py
```diff
@@ -406,7 +408,12 @@ def add_vlan_member(ctx, vid, interface_name, untagged):
     vlan = db.get_entry('VLAN', vlan_name)
     
     if get_interface_mode() == "alias": 
-        interface_name = interface_alias(interface_name)
+        default_name = interface_alias_to_name(interface_name)
```
jleveque: The introduction of the variable `default_name` is unnecessary. Simply assign directly to `interface_name` and compare it to None. You can then also eliminate the `else` clause:

```
interface_name = interface_alias_to_name(interface_name)
if interface_name is None:
    print "Invalid interface {}".format(interface_name)
    raise click.Abort()
```

config/main.py
```diff
@@ -406,7 +408,12 @@ def add_vlan_member(ctx, vid, interface_name, untagged):
...
+        if default_name is None:
+            print "Invalid interface {}".format(interface_name)
```
jleveque: `interface_name` is not defined. This will crash. However, instead of replacing it with `default_name`, please refactor the code as I suggested in my comment on line 411.

config/main.py
```diff
@@ -415,6 +422,11 @@ def add_vlan_member(ctx, vid, interface_name, untagged):
     if interface_name in members:
         if get_interface_mode() == "alias": 
             alias_name = interface_name_to_alias(interface_name)
+            if alias_name is None:
+                print "Invalid interface {}".format(interface_name)
+                sys.exit(0)
```
jleveque: Please replace `sys.exit(0)` `raise click.Abort()`, which will cause the program to exit with code 1.

config/main.py
```diff
@@ -435,15 +447,25 @@ def del_vlan_member(ctx, vid, interface_name):
...
         print "{} doesn't exist".format(vlan_name)
         raise click.Abort
     members = vlan.get('members', [])
     if interface_name not in members:
-        alias_name = interface_name_to_alias(interface_name)
         if get_interface_mode() == "alias": 
+            alias_name = interface_name_to_alias(interface_name)
+            if alias_name is None:
+                print "Invalid interface {}".format(interface_name)
+                sys.exit(0)
```
jleveque: Please replace `sys.exit(0)` `raise click.Abort()`, which will cause the program to exit with code 1.

config/main.py
```diff
@@ -435,15 +447,25 @@ def del_vlan_member(ctx, vid, interface_name):
     vlan = db.get_entry('VLAN', vlan_name)
 
     if get_interface_mode() == "alias": 
-        interface_name = interface_alias(interface_name)
+        default_name = interface_alias_to_name(interface_name)
```
jleveque: Please refactor this code in the same manner as I suggested in my comment on line 411.

config/main.py
```diff
@@ -533,7 +555,12 @@ def interface():
 def shutdown(interface_name, verbose):
     """Shut down interface"""
     if get_interface_mode() == "alias": 
-        interface_name = interface_alias(interface_name)
+        default_name = interface_alias_to_name(interface_name)
```
jleveque: Please refactor this code in the same manner as I suggested in my comment on line 411.

config/main.py
```diff
@@ -548,7 +575,12 @@ def shutdown(interface_name, verbose):
 def startup(interface_name, verbose):
     """Start up interface"""
     if get_interface_mode() == "alias": 
-        interface_name = interface_alias(interface_name)
+        default_name = interface_alias_to_name(interface_name)
```
jleveque: Please refactor this code in the same manner as I suggested in my comment on line 411.

config/main.py
```diff
@@ -564,7 +596,12 @@ def startup(interface_name, verbose):
 def speed(interface_name, interface_speed, verbose):
     """Set interface speed"""
     if get_interface_mode() == "alias": 
-        interface_name = interface_alias(interface_name)
+        default_name = interface_alias_to_name(interface_name)
```
jleveque: Please refactor this code in the same manner as I suggested in my comment on line 411.

portalias/PORTALIAS.md
```diff
@@ -0,0 +1,75 @@
+#Portalias
```
jleveque: This file is now very out-of-date, since we are no longer using a flag to determine the mode. Also, a design doc of this sort doesn't belong in this repo alongside the code; if anything it should be placed in the GitHub wiki. Please delete this file.

config/main.py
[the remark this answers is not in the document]
paavaanan: Joe I got your point.  When the interface_name variable type is None, even though else is not defined inside the corresponding function the return type of interface_alias_to_name() will be None.  All parent functions  are equipped to handle this case. So, Invalid Interface message will be raised to the user. Even if we add an else the behavior will be the same.

```def shutdown(interface_name, verbose):
    """Shut down interface"""
    if get_interface_mode() == "alias":
        interface_name = interface_alias_to_name(interface_name)
        if interface_name is None:
            print "Invalid interface {}".format(interface_name)<--------
            raise click.Abort()
)
```

```
admin@sonic:~$ sudo config interface shutdown hundredGigE1/1
Invalid interface None
Aborted!
admin@sonic:~$
```

config/main.py
paavaanan: How do this crash? interface_name is the incoming argument for add_vlan_member() So, argument do have a value when invoked. 
```
def add_vlan_member(ctx, vid, interface_name, untagged):
```

jleveque: You're correct. Technically, Click should prevent `interface_name` from being `None`. I think I got a bit lost due the introduction of `default_interface`. Ignore this comment. #Closed

paavaanan:
@jleveque addressed all review comments and raised 2 commits with all changes. 
Please, have a look on this.

config/main.py
```diff
@@ -83,7 +83,7 @@ def set_interface_mode(mode):
     if user != "root":
         bashrc = "/home/{}/.bashrc".format(user)
     else:
-        sys.exit(0)
+        raise click.Abort
```
jleveque: Add parentheses: `raise click.Abort()`

config/main.py
```diff
@@ -418,7 +418,7 @@ def add_vlan_member(ctx, vid, interface_name, untagged):
     if get_interface_mode() == "alias":
         interface_name = interface_alias_to_name(interface_name)
         if interface_name is None:
-            sys.exit(0)
+            raise click.Abort
```
jleveque: Add parentheses: `raise click.Abort()`

config/main.py
```diff
@@ -453,7 +453,7 @@ def del_vlan_member(ctx, vid, interface_name):
...
```
jleveque: Add parentheses: `raise click.Abort()`

config/main.py
```diff
@@ -555,7 +555,7 @@ def shutdown(interface_name, verbose):
...
```
jleveque: Add parentheses: `raise click.Abort()`

config/main.py
```diff
@@ -572,7 +572,7 @@ def startup(interface_name, verbose):
...
```
jleveque: Add parentheses: `raise click.Abort()`

config/main.py
```diff
@@ -591,7 +591,7 @@ def speed(interface_name, interface_speed, verbose):
...
```
jleveque: Add parentheses: `raise click.Abort()`

config/main.py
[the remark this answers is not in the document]
jleveque: I still believe that every path should explicitly return a value. Otherwise the behavior of the function isn't easily determined to the reader. I think you can easily fix this as follows:

```
    if interface_alias is not None:
        for port_name in natsorted(port_dict.keys()):
            if interface_alias == port_dict[port_name]['alias']:
                return port_name
        print "Invalid interface {}".format(interface_alias)
       
    return None
```

config/main.py
```diff
@@ -33,6 +36,77 @@ def run_command(command, display_cmd=False, ignore_error=False):
...
+    else:
+        raise click.Abort()
+
+    f = open(bashrc, 'r')
+    filedata = f.read()
+    f.close()
+
+    if "SONIC_CLI_IFACE_MODE" not in filedata:
+        newdata = filedata + bashrc_ifacemode_line
+        newdata += "\n"
```
jleveque: I suggest adding the newline directly to `bashrc_ifacemode_line` on line 78 as follows:

`bashrc_ifacemode_line = "SONIC_CLI_IFACE_MODE={}\n".format(mode)`

PR #268, opened 2018-06-14
[sonic-utilities/scripts] Fixing FP ports issue

vharish02:
This commit fixes the Front Panel Ports not coming up during
cold reboot and traffic being sent at line rate. The root cause
was the PCIe bus error which was caused because of the host
side not doing a gracefull shutdown calling bcm_shutdown and
soc_shutdown. The fix will call the gracefull shutdown of syncd
during reboot which in turns calls the bcm_shutdown and soc_shutdown.
Additionally added the check to make sure the script will run only when
it has root privileges otherwise it will quit.

Verified the fix by running traffic at line rate 40G from Ixia and
rebooted using script multiple times and traffic was successfully
forwarded after every reboot.

Signed-off-by: Harish Venkatraman <Harish_Venkatraman@dell.com>

**- What I did**

**- How I did it**

**- How to verify it**

-->

### DIFF
--- a/scripts/reboot
+++ b/scripts/reboot
@@ -1,5 +1,12 @@
 #! /bin/bash
 
+# Check root privileges
+if [[ "$EUID" -ne 0 ]]
+then
+  echo "Please run as root" >&2
+  exit 1
+fi
+
 function stop_sonic_services()
 {
     echo "Stopping syncd..."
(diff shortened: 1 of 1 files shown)


vharish02:
@yxieca Can you please review this PR

scripts/reboot
```diff
@@ -1,5 +1,12 @@
 #! /bin/bash
 
+# Check root privileges
+if [[ "$EUID" -ne 0 ]]
+then
+  echo "Please run as root"
+  exit
```
qiluo-msft: >exit [](start = 2, length = 4)

exit with error code

scripts/reboot
```diff
@@ -1,5 +1,12 @@
 #! /bin/bash
 
+# Check root privileges
+if [[ "$EUID" -ne 0 ]]
+then
+  echo "Please run as root"
```
qiluo-msft: >echo [](start = 2, length = 4)

output to stderr?

PR #271, opened 2018-06-23
 Add support for reboot cause and 'show reboot-cause' command for display

jleveque:
Example output:

```
admin@sonic:~$ show reboot-cause 
User issued reboot command [User: admin, Time: Sat Jun 23 01:30:43 UTC 2018]
```

```
admin@sonic:~$ show reboot-cause 
Unexpected reboot
```

### DIFF
--- a/show/main.py
+++ b/show/main.py
@@ -1064,6 +1064,8 @@ def ecn():
 def reboot_cause():
     """Show cause of most recent reboot"""
     PREVIOUS_REBOOT_CAUSE_FILE = "/var/cache/sonic/previous-reboot-cause.txt"
+
+    # File should always be created at boot, but check just in case it's not
     if not os.path.isfile(PREVIOUS_REBOOT_CAUSE_FILE):
         click.echo("Unable to determine cause of previous reboot\n")
     else:
--- a/show/main.py
+++ b/show/main.py
@@ -1065,7 +1065,10 @@ def reboot_cause():
     """Show cause of most recent reboot"""
     PREVIOUS_REBOOT_CAUSE_FILE = "/var/cache/sonic/previous-reboot-cause.txt"
 
-    # File should always be created at boot, but check just in case it's not
+    # At boot time, PREVIOUS_REBOOT_CAUSE_FILE is generated based on
+    # the contents of the 'reboot cause' file as it was left when the device
+    # went down for reboot. This file should always be created at boot,
+    # but check first just in case it's not present.
     if not os.path.isfile(PREVIOUS_REBOOT_CAUSE_FILE):
         click.echo("Unable to determine cause of previous reboot\n")
     else:
--- a/show/main.py
+++ b/show/main.py
@@ -1057,5 +1057,19 @@ def ecn():
     proc = subprocess.Popen(cmd, stdout=subprocess.PIPE, shell=True)
     click.echo(proc.stdout.read())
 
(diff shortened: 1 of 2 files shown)


show/main.py
```diff
@@ -1057,5 +1057,19 @@ def ecn():
...
 
+#
+# 'reboot-cause' command ("show reboot-cause")
+#
+@cli.command('reboot-cause')
+def reboot_cause():
+    """Show cause of most recent reboot"""
+    PREVIOUS_REBOOT_CAUSE_FILE = "/var/cache/sonic/previous-reboot-cause.txt"
+    if not os.path.isfile(PREVIOUS_REBOOT_CAUSE_FILE):
+        click.echo("Unable to determine cause of previous reboot\n")
```
qiluo-msft: >Unable to determine cause of previous reboot [](start = 20, length = 44)

You mentioned "Unexpected reboot" in the sample in PR comment.

show/main.py
```diff
@@ -1057,5 +1057,19 @@ def ecn():
...
     click.echo(proc.stdout.read())
 
+#
+# 'reboot-cause' command ("show reboot-cause")
+#
+@cli.command('reboot-cause')
+def reboot_cause():
+    """Show cause of most recent reboot"""
+    PREVIOUS_REBOOT_CAUSE_FILE = "/var/cache/sonic/previous-reboot-cause.txt"
+    if not os.path.isfile(PREVIOUS_REBOOT_CAUSE_FILE):
```
qiluo-msft: >isfile [](start = 19, length = 6)

So you mean after each unexpected reboot, the file will be missing? Could you explain the reason as code comment?

show/main.py
```diff
@@ -1057,5 +1057,19 @@ def ecn():
     proc = subprocess.Popen(cmd, stdout=subprocess.PIPE, shell=True)
     click.echo(proc.stdout.read())
 
+#
+# 'reboot-cause' command ("show reboot-cause")
+#
+@cli.command('reboot-cause')
+def reboot_cause():
+    """Show cause of most recent reboot"""
+    PREVIOUS_REBOOT_CAUSE_FILE = "/var/cache/sonic/previous-reboot-cause.txt"
```
qiluo-msft: >previous- [](start = 51, length = 9)

Who prefix "previous-" to the file name?

show/main.py
jleveque: I have written another part of the mechanism in the base image (rc.local) which I intend to check into the buildimage repo along with this submodule update once this PR merges.

show/main.py
jleveque: Added comment.

show/main.py
jleveque: The mechanism uses two files: "reboot-cause.txt" which is used to writing the cause of the *next* reboot, and "previous-reboot-cause.txt" which is used for reading the cause of the previous reboot. "reboot-cause.txt" initially gets created with the the text "Unexpected reboot" which will become the reboot reason in the event of power loss. This is done in rc.local.

qiluo-msft: By reading your reply, I understand the design fully. Could you move them to code comment so other developer could easily maintain based on your thought?

---
In reply to: [197602125](https://github.com/Azure/sonic-utilities/pull/271#discussion_r197602125) [](ancestors = 197602125)

PR #277, opened 2018-06-30
[fast-reboot] Update reboot reason file to indicate fast-reboot was called

jleveque:


scripts/fast-reboot
```diff
@@ -82,11 +86,16 @@ then
...
+
 # Wait until all buffers synced with disk
 sync
 sleep 1
 sync
 
 # Reboot: explicity call Linux native reboot under sbin
 echo "Rebooting to $NEXT_SONIC_IMAGE..."
-/sbin/reboot
+exec /sbin/reboot
```
pavel-shirshov: I'm not sure for this exec. Did you test it?

jleveque: Yes. I made the same change recently in the `scripts/reboot` script and tested that thoroughly.

PR #279, opened 2018-07-06
Add empty case

tientienle:
**- What I did**
 - Handle the case when there is no port channel in the device

**- How I did it** 
 - Check when the team_keys == None, return the function instead of continue to assign a none object to the team

**- How to verify it**
 - Run teamshow when there is no port channel in the device

'NoneType' object is not iterable

No.    Team Dev    Protocol    Ports
-----  ----------  ----------  -------

### DIFF
--- a/scripts/teamshow
+++ b/scripts/teamshow
@@ -49,7 +49,7 @@ class Teamshow(object):
         """
         team_keys = self.db.keys(self.db.CONFIG_DB, PORT_CHANNEL_CFG_TABLE_PREFIX+"*")
         if team_keys == None:
-                return
+            return
         self.teams = [key[len(PORT_CHANNEL_CFG_TABLE_PREFIX):] for key in team_keys]
 
     def get_portchannel_status(self, port_channel_name):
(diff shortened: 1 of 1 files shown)


scripts/teamshow
```diff
@@ -48,6 +48,9 @@ class Teamshow(object):
             Get the portchannel names from database.
         """
         team_keys = self.db.keys(self.db.CONFIG_DB, PORT_CHANNEL_CFG_TABLE_PREFIX+"*")
+        if team_keys == None:
+                #print "None"
```
jleveque: Please remove commented print statement.

scripts/teamshow
```diff
@@ -48,6 +48,8 @@ class Teamshow(object):
             Get the portchannel names from database.
         """
         team_keys = self.db.keys(self.db.CONFIG_DB, PORT_CHANNEL_CFG_TABLE_PREFIX+"*")
+        if team_keys == None:
+                return
```
qiluo-msft: Indent 4x spaces?

PR #280, opened 2018-07-10
Add a stub for arp-helper-start'

pavel-shirshov:
**- What I did**
Added a stub for arp-helper-start script

**- How I did it**
use my editor

**- How to verify it**
Build an image with this script and then run the script
'arp-helper-start 10.0.0.1'

Currently the output would be OK
-->

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,6 +37,7 @@ setup(
     },
     scripts=[
         'scripts/aclshow',
+        'scripts/arp-helper-start',
         'scripts/boot_part',
         'scripts/coredump-compress',
         'scripts/decode-syseeprom',
--- a/scripts/arp-helper-start
+++ b/scripts/arp-helper-start
@@ -1,0 +1,21 @@
+#!/usr/bin/env python
+
+import argparse
+from pprint import pprint
+
+def request_arp_helper(ip):
+    # FIXME: make request to an ARP Helper
+    return "OK"
+
+def main():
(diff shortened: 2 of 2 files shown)


setup.py
```diff
@@ -37,6 +37,7 @@ def get_test_suite():
     },
     scripts=[
         'scripts/aclshow',
+        'scripts/arp-helper-start',
```
lguohan: rename to arp-helper is better, we could have arp-helper start ip address or arp-helper stop later.

scripts/arp-helper
```diff
@@ -0,0 +1,21 @@
...
+import argparse
+from pprint import pprint
+
+def request_arp_helper(ip):
+    # FIXME: make request to an ARP Helper
+    return "OK"
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('arp_helper_ip', help="ip address of an ARP Helper")
```
lguohan: for completeness, we should have --start and --stop to start of stop arphelper.

scripts/arp-helper
```diff
@@ -0,0 +1,21 @@
...
+    argv = parser.parse_args()
+    arp_helper_ip = argv.arp_helper_ip
+    result = request_arp_helper(arp_helper_ip)
+    print result
+
+    return
+
+if __name__ == '__main__':
+    main()
```
jleveque: Add newline at end of file.

PR #284, opened 2018-07-16
Clear line implementation

cawand:
Adds `clear line <linenum>` implementation
`sonic-clear line <linenum>` will call clear/main.py line function, which calls `consutil clear <linenum>`. The consutil clear line function checks if a preexisting process is connecting to the specified line, and if so kills that process.
This is called by running `sonic-clear line <linenum>`
It will output one of these messages:

- `Sending SIGTERM to process <pid>`
- `No process is connecting to line <linenum>`

### DIFF
--- a/consutil/main.py
+++ b/consutil/main.py
@@ -1,0 +1,44 @@
+#!/usr/bin/env python
+#
+# main.py
+#
+# Command-line utility for interacting with switches over serial via console device
+#
+
+try:
+    import click
+    import re
+    import subprocess
+except ImportError as e: 
+    raise ImportError("%s - required module not found" % str(e))
+
+@click.group()
+def consutil():
+    """consutil - Command-line utility for interacting with switchs via console device"""
+
+    if os.geteuid() != ""
+        print "Root privileges are required for this operation"
+        sys.exit(1)
+
+# 'show' subcommand
+@consutil.command()
+def line():
+    """Show all /dev/ttyUSB lines"""
+    click.echo("show line")
+
+# 'clear' subcommand
+@consutil.command()
+@click.argument('linenum')
+def clear(linenum):
+    """Clear preexisting connection to line"""
+    click.echo("clear line linenum")
+
+# 'connect' subcommand
+@consutil.command()
+@click.argument('linenum')
+def connect(linenum):
+    """Connect to switch via console device"""
+    click.echo("connect linenum")
+
+if __name__ == '__main__':
+    consutil()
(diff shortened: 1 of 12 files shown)


consutil/main.py
```diff
@@ -31,7 +31,17 @@ def line():
...
-    click.echo("clear line linenum")
+    checkDevice(linenum)
+    linenum = str(linenum)
+
+    busyDevices = getBusyDevices()
+    if linenum in busyDevices:
+        pid, _ = busyDevices[linenum]
+        cmd = "sudo kill -15 " + pid
+        click.echo("Sending SIGTERM to process " + pid)
+        popenWrapper(cmd)
```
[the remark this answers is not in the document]
cawand: popenWrapper() is defined in consutil/lib.py, which isn't it this PR. It's not a very descriptive name - I could change it to run_command() or something else?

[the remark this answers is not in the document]
jleveque: Yeah, you should probably rename to `run_command()` to align with the other utilities.

At this point, we now have multiple utilities containing duplicate implementations of `run_command()`. I think we should plan to implement a separate common library that all the utilities can share. Not a request for you, just thinking out loud on the record :)

consutil/main.py
```diff
@@ -31,7 +31,17 @@ def line():
...
 def clear(linenum):
     """Clear preexisting connection to line"""
-    click.echo("clear line linenum")
+    checkDevice(linenum)
+    linenum = str(linenum)
+
+    busyDevices = getBusyDevices()
+    if linenum in busyDevices:
+        pid, _ = busyDevices[linenum]
+        cmd = "sudo kill -15 " + pid
```
pavel-shirshov: I'd suggest to use `kill -SIGTERM $pid` or `kill -TERM $pid` here. More readable than `-15`

consutil/main.py
```diff
@@ -31,7 +31,17 @@ def line():
 @click.argument('linenum')
 def clear(linenum):
     """Clear preexisting connection to line"""
-    click.echo("clear line linenum")
```
pavel-shirshov: Is there any input parameter linenum validation?
what if we run `clear line swefswef`, what it will return?

cawand: Yes - `checkDevice(linenum)` (which is in `consutil/lib.py`) checks that there's actually a device for that line number. If there's not, it prints out an error. Right now it also calls `sys.exit()` which might not be ideal - maybe `checkDevice()` should just return an error, which `clear()` can catch and return from instead of exiting?

pavel-shirshov: Thank you for the detailed explanations. I'd say it's good enough to exit in checkDevice() function.

PR #295, opened 2018-08-08
Connect line implementation (and show line output change)

cawand:
Adds `connect line <target>` implementation
`connect line <target>` will call connect/main.py line function, which calls `consutil connect <target>`. `connect line -d <target>` or `connect line --devicename <target>` are also available to connect by device name instead of by line number.
This will start a serial communication session with the specified device by using picocom running in a pty. 

Example output of `connect line 7`:
```
Successful connection to line 7
Press ^A ^X to disconnect


Debian GNU/Linux 8 str-dx010-acs-2 ttyS0

str-dx010-acs-2 login:
```
And upon quitting:
```
Terminating...
Thanks for using picocom
```
Picocom exit message can be removed by setting QUIET var in consutil/main.py `connect()`

This PR also adds reading connection info from the config db.
There is a change in `show line` output to reflect this - now instead of just displaying a baud rate, `show line` displays the baud rate actually used to connect as well as the baud rate in config db:
```
  Line    Actual/Configured Baud    PID                Start Time
------  ------------------------  -----  ------------------------
     0      9600/9600
     1      9600/9600
    *2      9600/9600             23285  Wed Aug  8 00:49:02 2018
     3      9600/-
```
Previously, `show line` displayed:
```
  Line    Baud    PID                Start Time
------  ------  -----  ------------------------
     0    9600
     1    9600
    *2    9600  23285  Wed Aug  8 00:49:02 2018
     3    9600
```

### DIFF
--- a/connect/main.py
+++ b/connect/main.py
@@ -3,6 +3,7 @@
 import click
 import errno
 import os
+import pexpect
 import subprocess
 import sys
 from click_default_group import DefaultGroup
@@ -85,18 +86,8 @@ def run_command(command, display_cmd=False):
     if display_cmd:
         click.echo(click.style("Command: ", fg='cyan') + click.style(command, fg='green'))
 
-    proc = subprocess.Popen(command, shell=True, stdout=subprocess.PIPE)
-
-    while True:
-        output = proc.stdout.readline()
-        if output == "" and proc.poll() is not None:
-            break
-        elif output:
-            click.echo(output.rstrip('\n'))
-
-    rc = proc.poll()
-    if rc != 0:
-        sys.exit(rc)
+    proc = pexpect.spawn(command)
+    proc.interact()
 
 CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help', '-?'])
 
@@ -114,13 +105,21 @@ def connect():
 # 'line' command ("connect line")
 #
 @connect.command('line')
-@click.argument('target')
-@click.option('--devicename', '-d', is_flag=True, help="connect by name - if flag is set, interpret target as device name")
-def line(target, devicename):
-    """Connect to TARGET line number or device name via serial connection"""
-    deviceFlag = "-d " if devicename else ""
-    cmd = "consutil connect " + deviceFlag + target
-    run_command(cmd, display_cmd=verbose)
+@click.argument('linenum')
+def line(linenum):
+    """Connect to line LINENUM via serial connection"""
+    cmd = "consutil connect " + linenum
+    run_command(cmd)
+
+#
+# 'device' command ("connect device")
+#
+@connect.command('device')
+@click.argument('devicename')
+def device(devicename):
+    """Connect to device DEVICENAME via serial connection"""
+    cmd = "consutil connect -d " + devicename
+    run_command(cmd)
 
 if __name__ == '__main__':
     connect()
--- a/connect/main.py
+++ b/connect/main.py
@@ -126,6 +126,8 @@ def connect():
 @connect.command('line')
 @click.argument('linenum')
 def line(linenum):
+    """Connect to line via serial connection"""
+    # TODO: Stub
     return
 
 if __name__ == '__main__':
--- a/connect/main.py
+++ b/connect/main.py
@@ -1,0 +1,132 @@
+#! /usr/bin/python -u
+
+import click
+import errno
+import os
+import subprocess
+import sys
+from click_default_group import DefaultGroup
+
+try:
+    # noinspection PyPep8Naming
+    import ConfigParser as configparser
+except ImportError:
+    # noinspection PyUnresolvedReferences
+    import configparser
+
+
+# This is from the aliases example:
+# https://github.com/pallets/click/blob/57c6f09611fc47ca80db0bd010f05998b3c0aa95/examples/aliases/aliases.py
+class Config(object):
+    """Object to hold CLI config"""
+
+    def __init__(self):
+        self.path = os.getcwd()
+        self.aliases = {}
+
+    def read_config(self, filename):
+        parser = configparser.RawConfigParser()
+        parser.read([filename])
+        try:
+            self.aliases.update(parser.items('aliases'))
+        except configparser.NoSectionError:
+            pass
+
+
+# Global Config object
+_config = None
+
+
+# This aliased group has been modified from click examples to inherit from DefaultGroup instead of click.Group.
+# DefaultGroup is a superclass of click.Group which calls a default subcommand instead of showing
+# a help message if no subcommand is passed
+class AliasedGroup(DefaultGroup):
+    """This subclass of a DefaultGroup supports looking up aliases in a config
+    file and with a bit of magic.
+    """
+
+    def get_command(self, ctx, cmd_name):
+        global _config
+
+        # If we haven't instantiated our global config, do it now and load current config
+        if _config is None:
+            _config = Config()
+
+            # Load our config file
+            cfg_file = os.path.join(os.path.dirname(__file__), 'aliases.ini')
+            _config.read_config(cfg_file)
+
+        # Try to get builtin commands as normal
+        rv = click.Group.get_command(self, ctx, cmd_name)
+        if rv is not None:
+            return rv
+
+        # No builtin found. Look up an explicit command alias in the config
+        if cmd_name in _config.aliases:
+            actual_cmd = _config.aliases[cmd_name]
+            return click.Group.get_command(self, ctx, actual_cmd)
+
+        # Alternative option: if we did not find an explicit alias we
+        # allow automatic abbreviation of the command.  "status" for
+        # instance will match "st".  We only allow that however if
+        # there is only one command.
+        matches = [x for x in self.list_commands(ctx)
+                   if x.lower().startswith(cmd_name.lower())]
+        if not matches:
+            # No command name matched. Issue Default command.
+            ctx.arg0 = cmd_name
+            cmd_name = self.default_cmd_name
+            return DefaultGroup.get_command(self, ctx, cmd_name)
+        elif len(matches) == 1:
+            return DefaultGroup.get_command(self, ctx, matches[0])
+        ctx.fail('Too many matches: %s' % ', '.join(sorted(matches)))
+
+def run_command(command, display_cmd=False):
+    if display_cmd:
+        click.echo(click.style("Command: ", fg='cyan') + click.style(command, fg='green'))
+
+    proc = subprocess.Popen(command, shell=True, stdout=subprocess.PIPE)
+
+    while True:
+        output = proc.stdout.readline()
+        if output == "" and proc.poll() is not None:
+            break
+        if output:
+            try:
+                click.echo(output.rstrip('\n'))
+            except IOError as e:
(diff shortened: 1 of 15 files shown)


consutil/main.py
```diff
@@ -28,7 +29,7 @@ def show():
     devices = getAllDevices()
     busyDevices = getBusyDevices()
 
-    header = ["Line", "Baud", "PID", "Start Time"]
+    header = ["Line", "Actual/Configured Baud", "PID", "Start Time"]
```
SAI-shared-user: what else can this baud rate be? what is the reason to add it?

SAI-shared-user: can they be different?

consutil/main.py
```diff
@@ -62,10 +67,56 @@ def clear(linenum):
...
+    cmd = "sudo picocom -b {} -f {} {} {}{}".format(baud, flowCmd, quietCmd, DEVICE_PREFIX, lineNumber)
+    proc = pexpect.spawn(cmd)
+    proc.send("\n")
+
+    if QUIET:
+        READY_MSG = r"([Ll]ogin:|[$>#])" # login prompt or command line prompt
+    else:
+        READY_MSG = "Terminal ready" # picocom ready message
+    BUSY_MSG = "Resource temporarily unavailable" # picocom busy message
+    TIMEOUT_SEC = 0.2
```
SAI-shared-user: why not set it larger like 1 second? what is the downside to have larger timeout value?

consutil/lib.py
```diff
@@ -75,16 +80,32 @@ def getBusyDevices():
...
+# returns baud rate and flow control settings of device corresponding to line number
 # input: linenum (str)
-def getBaud(linenum):
-    checkDevice(linenum)
-    cmd = "sudo stty -F " + DEVICE_PREFIX + str(linenum)
-    output = run_command(cmd)
+def getConnectionInfo(linenum):
+    config_db = ConfigDBConnector()
+    config_db.connect()
+    entry = config_db.get_entry(TABLE_NAME, str(linenum))
```
SAI-shared-user: is it possible that we cannot get an entry here?

consutil/lib.py
```diff
@@ -75,16 +80,32 @@ def getBusyDevices():
...
+    baud_rate = "" if BAUD_KEY not in entry else entry[BAUD_KEY]
+    flow_control = False
+    if FLOW_KEY in entry and entry[FLOW_KEY] == "1":
+        flow_control = True
+
+    return (baud_rate, flow_control)
+
+# returns the line number of a given device name
+# input: devicename (str)
+def getLineNumber(devicename):
```
SAI-shared-user: is linenumber an integer, if yes, then return integer and if you cannot find a line number return None

consutil/main.py
cawand: Actual baud is the baud rate that would actually be used to connect to the device, and configured baud is the baud rate that is pulled from config db. If the baud rate is successfully pulled from config db then the two are the same, but if a baud rate can't be pulled from config db, this is indicated by actual baud being "9600" (i.e. default value) and configured baud being "-"

connect/main.py
```diff
@@ -114,11 +114,13 @@ def connect():
 # 'line' command ("connect line")
 #
 @connect.command('line')
-@click.argument('linenum')
-def line(linenum):
-    """Connect to line via serial connection"""
-    # TODO: Stub
-    return
+@click.argument('target')
+@click.option('--devicename', '-d', is_flag=True, help="connect by name - if flag is set, interpret target as device name")
```
jleveque: I'm wondering whether we should use the "-d" flag to denote device hostname, or if we should create a separate subcommand, something like the following:

`connect line <line_number>`
`connect device <device_name>`

Thoughts?

consutil/lib.py
```diff
@@ -18,6 +18,11 @@
 ERR_CMD = 1
 ERR_DEV = 2
 
+TABLE_NAME = "CONSOLE_PORT"
```
jleveque: Add more description to constant name (e.g., `CONSOLE_PORT_TABLE_NAME`).

consutil/main.py
cawand: This number is pretty arbitrary, it could be 1 second. I chose 0.2 seconds because it's long enough for a login/command line prompt to output, but short enough that to a human the delay isn't very noticeable.

consutil/main.py
jleveque: Maybe split Actual Baud and Configured Baud into separate columns?

consutil/lib.py
cawand: get_entry() will return an empty dictionary if the table or entry doesn't exist.

consutil/lib.py
cawand: linenum is a string and if the line number isn't found, getLineNumber() returns an empty string. This is because in connect() the line number is only ever used as a string

cawand: The comments for getLineNumber() and getConnectionInfo() should include the output type, I'll add it

connect/main.py
cawand: Either makes sense to me!
@yxieca Thoughts?

consutil/main.py
yxieca: We have current show line output combine them in a same column.

yxieca: emm. the current show line shows tx/rx in the same column. I guess having 2 columns for configured would be ok then.

connect/main.py
yxieca: I think the decision factor is more with the underline code in soncutil, right?

From what I see below. we are calling a same function with a flag, then it makes more sense to keep -d. However, if we have 2 separate command in underline console until, then it makes more sense to have 2 commands.

consutil/main.py
```diff
@@ -39,6 +40,10 @@ def show():
             pid, date = busyDevices[lineNum]
             busy = "*"
         baud = getBaud(lineNum)
+        if baud == "":
+            baud = "9600/-"
```
yxieca: I have a small concern here. We are hard coding 9600 as default baud rate. But the actual default value is either decided by piccocom or some other code in console util.

This creates a possibility for inconsistent default value in the future. Is it possible to unify the code where default value was chosen?

consutil/main.py
```diff
@@ -62,10 +67,56 @@ def clear(linenum):
...
+    if devicename:
+        lineNumber = getLineNumber(target)
+        if lineNumber == "":
+            click.echo("Device {} does not exist".format(target))
+            return
+    else:
+        lineNumber = target
+    checkDevice(lineNumber)
+    lineNumber = str(lineNumber)
+
```
yxieca: Code below this comment can probably go to a helper function like connecToLine or similar. So that we don't have a long function. Also it makes easier if we decide to have 2 command for connecting to line/device.

consutil/main.py
jleveque: I'm OK with whatever is decided to be most legible. I just thought it was worth a discussion.

consutil/main.py
yxieca: yeah. 1 second is notice-able by human.

consutil/main.py
```diff
@@ -62,10 +67,56 @@ def clear(linenum):
...
+    if QUIET:
+        READY_MSG = r"([Ll]ogin:|[$>#])" # login prompt or command line prompt
+    else:
+        READY_MSG = "Terminal ready" # picocom ready message
+    BUSY_MSG = "Resource temporarily unavailable" # picocom busy message
+    TIMEOUT_SEC = 0.2
+
+    index = proc.expect([READY_MSG, BUSY_MSG, pexpect.EOF, pexpect.TIMEOUT], timeout=TIMEOUT_SEC)
+    if index == 0: # terminal ready
```
yxieca: I am wondering if TIMEOUT_SEC, DEF_BAUD, DEFAULT_PREFIX should be defined as class attributes and initialized in a init function? So that we know that these are our global choices and available to see/change at a same location. QUIET could fall into this category, I think.

And READY_MSG, BUSY_MSG actually are local variables might be better to follow local variable naming convention. like readyMsg and busyMsg instead?

consutil/main.py
yxieca: maybe shorten the title to baud/conf? it seems that the title is taking lots of horizontal space.

jleveque: I think "baud/conf" isn't very descriptive and can be confusing.

consutil/main.py
cawand: Yeah, right now connect() also sets the default baud to 9600, so that default value is hardcoded in two different places. I can change getConnectionInfo() to return (actual_baud, conf_baud, flow_control) instead of just (baud_rate, flow_control)

consutil/main.py
cawand: Here's how two columns looks:
```
  Line    Actual Baud    Configured Baud    PID    Start Time
------  -------------  -----------------  -----  ------------
     0           9600               9600
     1           9600               9600
     2           9600               9600  23285  Wed Aug  8 00:49:02 2018
     3           9600                  -
```

I could also keep the current headers but align the dash:
```
  Line    Actual/Configured Baud    PID    Start Time
------  ------------------------  -----  ------------
     0      9600/9600
     1      9600/9600
     2      9600/9600             23285  Wed Aug  8 00:49:02 2018
     3      9600/-
```

consutil/main.py
cawand: I like that idea - how would the structure of that look? What class would they be attributes of?

consutil/main.py
cawand: How about moving the code above the comment into getLineNumber()? So that all the connect logic is in connect(), and all the get line logic is in getLineNumber()

connect/main.py
jleveque: I prefer two separate subcommands, as this fits more in the CLI structure used by most switches (many nested subcommands, very few flags), and this is what most NEs are familiar with. This would help them realize they have two completely different methods of connecting to a device. I think it is fine if the underlying consutil uses the '-d' flag, but I think the 'connect' command should have separate subcommands.

cawand:
All comments should be addressed now - let me know if there's more I should change

consutil/main.py
```diff
@@ -38,10 +40,13 @@ def show():
         if lineNum in busyDevices:
             pid, date = busyDevices[lineNum]
             busy = "*"
-        baud = getBaud(lineNum)
+        actBaud, confBaud, _ = getConnectionInfo(lineNum)
+        # repeated "~" will be replaced by spaces - hacky way to align the "/"s
+        baud = "{}/{}{}".format(actBaud, confBaud, "~"*(15-len(confBaud)))
```
jleveque: Will this solution keep the `/` aligned in cases where we have baud rates of varying lengths (i.e. 9600, 115200)?

cawand: Yes - "~" is multiplied by (15 - len(confBaud)), so the total number of characters to the right of the "/" stays the same

cawand: The entries of the columns are all right-aligned, so the placement of the "/" only depends on how many chars are to the right of it.
Here's the output I get with varying baud rates:
```
  Line    Actual/Configured Baud    PID                Start Time
------  ------------------------  -----  ------------------------
     0      9600/-
     1    115200/115200
    *2      9600/9600              5718  Thu Aug  9 18:35:35 2018
     3      9600/-
```

PR #300, opened 2018-08-16
Update consutil for picocom v1.7

cawand:
**- What I did**
Updated consutil for picocom v1.7 by removing QUIET flag and accounting for /var/lock files
**- How I did it**
Removed any code that used QUIET flag. Added code in `clear()` that removes /var/lock/LCK..ttyUSB<linenumber> file
**- How to verify it**
Run `connect line <linenumber>` or `connect device <devicename>` and `clear line <linenumber>` on device with picocom v1.7. Ensure that you can successfully connect to devices, that you get an error for trying to connect to a busy device, and that `clear` successfully clears a busy line.

-->

### DIFF
--- a/consutil/main.py
+++ b/consutil/main.py
@@ -1,0 +1,44 @@
+#!/usr/bin/env python
+#
+# main.py
+#
+# Command-line utility for interacting with switches over serial via console device
+#
+
+try:
+    import click
+    import re
+    import subprocess
+except ImportError as e: 
+    raise ImportError("%s - required module not found" % str(e))
+
+@click.group()
+def consutil():
+    """consutil - Command-line utility for interacting with switchs via console device"""
+
+    if os.geteuid() != ""
+        print "Root privileges are required for this operation"
+        sys.exit(1)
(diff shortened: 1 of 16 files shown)


cawand:
Yes - originally I developed consutil using picocom v3, but now we're using picocom v1.7, which doesn't have the -q flag and uses /var/lock files

consutil/main.py
```diff
@@ -62,6 +62,8 @@ def clear(linenum):
         cmd = "sudo kill -SIGTERM " + pid
         click.echo("Sending SIGTERM to process " + pid)
         run_command(cmd)
+        cmd = "sudo rm /var/lock/LCK..ttyUSB" + linenum
```
jleveque: Suggest to add `-f` flag to `rm` command to help ensure file is deleted.

lguohan:
https://github.com/Azure/sonic-buildimage/pull/1943

cawand:
We can delete this PR since the base image has picocom v3.1

PR #301, opened 2018-08-17
ecnconfig: add show and change ECN on/off status on queues 

wendani:
Signed-off-by: Wenda Ni <wenni@microsoft.com>

**- What I did**

**- How I did it**

**- How to verify it**
Tested on DUT with https://github.com/Azure/sonic-swss/pull/577

-->

### DIFF
--- a/scripts/ecnconfig
+++ b/scripts/ecnconfig
@@ -14,7 +14,7 @@ optional arguments:
   -v     --version             show program's version number and exit
   -vv    --verbose             verbose output
   -l     --list                show ECN WRED configuration
-  -p'    --profile             specify WRED profile name
+  -p     --profile             specify WRED profile name
   -gmin  --green-min           set min threshold for packets marked green
   -gmax  --green-max           set max threshold for packets marked green
   -ymin  --yellow-min          set min threshold for packets marked yellow
@@ -27,16 +27,24 @@ optional arguments:
 
 2) show and change ECN on/off status on queues
 
-usage: ecn [-h] [-v] [{on,off}] [-q QUEUE_INDEX] [-vv]
+usage: ecnconfig [-h] [-v] [-q QUEUE_INDEX] [{on,off}] [-vv]
 
 positional arguments:
   {on,off}                  turn on/off ecn
 
 optional arguments:
-  -h     --help             show this help message and exit
-  -v     --version          show program's version number and exit
-  -vv    --verbose          verbose output
-  -q     --queue            specify queue index list
+  -h     --help                show this help message and exit
+  -v     --version             show program's version number and exit
+  -vv    --verbose             verbose output
+  -q     --queue               specify queue index list
+
+Sample outputs:
+$ecnconfig -q 3 on -vv
+Enable ECN on Ethernet0,Ethernet4,Ethernet8,Ethernet12,Ethernet16,Ethernet20,Ethernet24,Ethernet28,Ethernet32,Ethernet36,Ethernet40,Ethernet44,Ethernet48,Ethernet52,Ethernet56,Ethernet60,Ethernet64,Ethernet68,Ethernet72,Ethernet76,Ethernet80,Ethernet84,Ethernet88,Ethernet92,Ethernet96,Ethernet100,Ethernet104,Ethernet108,Ethernet112,Ethernet116,Ethernet120,Ethernet124 queue 3
+
+$ecnconfig -q 3
+ECN status:
+queue 3: on
 """
 from __future__ import print_function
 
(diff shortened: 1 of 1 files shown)


scripts/ecn
```diff
@@ -0,0 +1,118 @@
...
+            for port in port_table:
+                ports.append(port)
+
+            self.ports_key = ','.join(sorted(ports, key = lambda k: int(k[8:])))
+
+    def set(self, enable):
+        for queue in self.queues:
+            if self.verbose:
+                print("%s ECN on %s queue %s" % ("Enable" if enable else "Disable", self.ports_key, queue))
+            key = '|'.join([QUEUE_TABLE_NAME, self.ports_key, queue])
```
lguohan: does the order matter here?

wendani: Order of what?

scripts/ecnconfig
```diff
@@ -22,6 +24,19 @@ optional arguments:
   -gdrop --green-drop-prob     set drop probability for packets marked green
   -ydrop --yellow-drop-prob    set drop probability for packets marked yellow
   -rdrop --red-drop-prob       set drop probability for packets marked red
+
+2) show and change ECN on/off status on queues
+
+usage: ecn [-h] [-v] [{on,off}] [-q QUEUE_INDEX] [-vv]
```
lguohan: ecnconfig [-h] [-v] [-q QUEUE_INDEX] [{on,off,set_profile]

wendani:
This is no output for ecnconfig -q on/off, just like when you set ecn/wred profile through ecnconfig -p  AZURE_LOSSLESS -gmax 555555. There is no output.

https://github.com/Azure/sonic-utilities/pull/301#issuecomment-414033237

qiluo-msft:
>         field = WRED_CONFIG_FIELDS[threshold]

what if thread is not a key?

---
Refers to: scripts/ecnconfig:107 in 9205d30. [](commit_id = 9205d307470d491456c3ebe659b681f7dda048bd, deletion_comment = False)

qiluo-msft:
>         wred_tables = self.db.get_table(WRED_PROFILE_TABLE_NAME)

'tables' is not a good name. It is a table in concept, including many rows (redis hashes).
I propose wred_profiles.

---
Refers to: scripts/ecnconfig:93 in 9205d30. [](commit_id = 9205d307470d491456c3ebe659b681f7dda048bd, deletion_comment = False)

qiluo-msft:
>         for table in wred_tables:

python2: you could get key and value at the same time
```python
for key, value in d.iteritems():
```

---
Refers to: scripts/ecnconfig:94 in 9205d30. [](commit_id = 9205d307470d491456c3ebe659b681f7dda048bd, deletion_comment = False)

scripts/ecnconfig
```diff
@@ -84,11 +115,73 @@ class EcnConfig(object):
...
+        self.ports_key = ''
+        self.queues = queues.split(',')
+        self.validate_queues()
+        self.verbose = verbose
+
+        # Set up db connections
+        self.config_db = swsssdk.ConfigDBConnector()
+        self.config_db.connect()
+
+        self.db = swsssdk.SonicV2Connector(host="127.0.0.1")
```
qiluo-msft: >host [](start = 43, length = 4)

We should use unix socket to improve performance. I notice many cases in this repo.

scripts/ecnconfig
```diff
@@ -101,36 +194,50 @@ def main():
...
+                    ecn.set_wred_prob(args.profile, "gdrop", args.green_drop_prob)
+                if args.yellow_drop_prob:
+                    ecn.set_wred_prob(args.profile, "ydrop", args.yellow_drop_prob)
+                if args.red_drop_prob:
+                    ecn.set_wred_prob(args.profile, "rdrop", args.red_drop_prob)
+        elif args.queue:
+            if len(sys.argv) < (4 if args.verbose else 3):
+                raise Exception("Input arguments error. Specify at least one queue by index")
+
+            ecn = EcnQ(args.queue, args.verbose)
```
qiluo-msft: >ecn [](start = 12, length = 3)

Not to reuse var names for totally different things.

scripts/ecnconfig
```diff
@@ -22,6 +24,27 @@ optional arguments:
   -gdrop --green-drop-prob     set drop probability for packets marked green
   -ydrop --yellow-drop-prob    set drop probability for packets marked yellow
   -rdrop --red-drop-prob       set drop probability for packets marked red
+
+2) show and change ECN on/off status on queues
```
qiluo-msft: >2) [](start = 0, length = 2)

It is weird that one script has 2 usages. How about combine them into 1 usage or separate into 2 scripts?

wendani: The original idea was 2 scripts. Guohan wants to put them together to avoid confusing the users. Then we have two streams of usage (i.e., two usages) here

wendani:
-p' is a typo in the original script before adding the max drop probability and the ecn on/off here

https://github.com/Azure/sonic-utilities/pull/301#issuecomment-414077182

wendani:
A KeyError exception is raised, and wil be caught in https://github.com/Azure/sonic-utilities/pull/301/files#diff-37a0a0daebb36e3825ebc34f6cf101f5R246, I have added "Exception caught: " to the print message

https://github.com/Azure/sonic-utilities/pull/301#issuecomment-414077487

scripts/ecnconfig
wendani: Is there a way to connect to unix socket via SonicV2Connector?  I need the get() member function to get the value of a field. ConfigDBConnector does not provide an equivalent call; its get_entry() returns a list of field-value tuples under a key, not a value of a particular field. I replace the set() call by ConfigDBConnector's mod_entry() to minimize the usage through ConfigDBConnector

PR #306, opened 2018-08-21
[config]: Add file check before load_minigraph command

stcheng:
This will prevent config load_minigraph fail during the
middle of the command execution.


prsunny (dismissed):
Looks ok. To me it appears more of a defensive check. I'm just thinking the use-case between load_minigraph vs user could execute a config reload.

stcheng:
yes. I think the use-case for load_minigraph is that the config_db.json file is not easy to modify while right now the minigraph is the source of the 'original configuration' for the switch.

config/main.py
```diff
@@ -321,7 +322,11 @@ def load_mgmt_config(filename):
                 expose_value=False, prompt='Reload config from minigraph?')
 def load_minigraph():
     """Reconfigure based on minigraph."""
-    #Stop services before config push
+    # Check /etc/sonic/minigraph.xml file exists
+    if not os.path.isfile(SONIC_MINIGRAPH_PATH):
```
qiluo-msft: >if not os.path.isfile(SONIC_MINIGRAPH_PATH): [](start = 4, length = 44)

I suggest test it by command ```sonic-cfggen -m```

zhenggen-xu: Agree here, sonic-cfggen might even do some sanity check here.

PR #308, opened 2018-08-27
[config]: Add IP subgroup commands

stcheng:
config interface <interface_name> add <ip_addr>
config interface <interface_name> remove <ip_addr>

These two commands are used to configure IP addresses

config interface <interface_name> startup
config interface <interface_name> shutdown

These two commands are moved under interface subgroup, and
the order of the arguments are swapped to match the rest of
the commands

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -645,64 +645,112 @@ def neighbor(ipaddr_or_hostname, verbose):
 #
 
 @cli.group()
-def interface():
-    """Interface-related configuration tasks"""
-    pass
-
-#
-# 'shutdown' subcommand
-#
-
-@interface.command()
 @click.argument('interface_name', metavar='<interface_name>', required=True)
-@click.option('-v', '--verbose', is_flag=True, help="Enable verbose output")
-def shutdown(interface_name, verbose):
-    """Shut down interface"""
+@click.pass_context
+def interface(ctx, interface_name):
+    """Interface-related configuration tasks"""
+    config_db = ConfigDBConnector()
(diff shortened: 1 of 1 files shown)


prsunny:
How about changing ip address? Does it require remove/add or just add will overwrite?

stcheng:
@prsunny It requires remove/add. Since we support multiple IP addresses.

config/main.py
```diff
@@ -601,65 +601,109 @@ def neighbor(ipaddr_or_hostname, verbose):
...
-        if interface_name is None:
-            raise click.Abort()
-
-    command = "ip link set {} down".format(interface_name)
-    run_command(command, display_cmd=verbose)
+@click.pass_context
+def startup(ctx):
+    """Start up interface"""
+    config_db = ctx.obj['config_db']
+    interface_name = ctx.obj['interface_name']
```
jleveque: You removed the recent `interface_alias_to_name()` alias-to-name translation addition here and below. Just wanted to check if this was overlooked because it's new. Do we want to maintain the ability to bring ports up/down by their vendor alias? @lguohan: do you have an opinion on this?

stcheng: Thanks for the feedback. I believe it is better to retain the ability of using vendor alias as an option. I will see how to update the pull request to fix this issue.

stcheng: @jleveque I have added the support for the `get_interface_mode() == "alias"`

stcheng:
thanks!

PR #314, opened 2018-09-05
[config]: Move setting GRE/queue for mirror session as optional

stcheng:
GRE type and queue number are set internally and shall not be set by users.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -353,11 +353,11 @@ def mirror_session():
 @click.argument('session_name', metavar='<session_name>', required=True)
 @click.argument('src_ip', metavar='<src_ip>', required=True)
 @click.argument('dst_ip', metavar='<dst_ip>', required=True)
-@click.argument('gre_type', metavar='<gre_type>', required=True)
 @click.argument('dscp', metavar='<dscp>', required=True)
 @click.argument('ttl', metavar='<ttl>', required=True)
-@click.argument('queue', metavar='<queue>', required=True)
-def add(session_name, src_ip, dst_ip, gre_type, dscp, ttl, queue):
+@click.argument('gre_type', metavar='[gre_type]', required=False)
+@click.argument('queue', metavar='[queue]', required=False)
+def add(session_name, src_ip, dst_ip, dscp, ttl, gre_type, queue):
     """
     Add mirror session
     """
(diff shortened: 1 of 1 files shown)


config/main.py
```diff
@@ -353,11 +353,9 @@ def mirror_session():
 @click.argument('session_name', metavar='<session_name>', required=True)
 @click.argument('src_ip', metavar='<src_ip>', required=True)
 @click.argument('dst_ip', metavar='<dst_ip>', required=True)
-@click.argument('gre_type', metavar='<gre_type>', required=True)
```
lguohan: can you change to not required, if not specified by the user, you do not need that field.

config/main.py
```diff
@@ -350,14 +350,14 @@ def mirror_session():
...
-@click.argument('dscp', metavar='<dscp>', required=True)
-@click.argument('ttl', metavar='<ttl>', required=True)
-@click.argument('queue', metavar='<queue>', required=True)
-def add(session_name, src_ip, dst_ip, gre_type, dscp, ttl, queue):
+@click.argument('session_name', metavar='session_name', required=True)
+@click.argument('src_ip', metavar='src_ip', required=True)
+@click.argument('dst_ip', metavar='dst_ip', required=True)
+@click.argument('dscp', metavar='dscp', required=True)
+@click.argument('ttl', metavar='ttl', required=True)
+@click.argument('gre_type', metavar='<gre_type>', required=False)
```
jleveque: For help/usage messages, python-click will display optional arguments in square brackets (`[...]`). However, it doesn't put any type of brackets around required arguments, which I believe is why the original author added the angle brackets. I think we should either never add brackets, or only add angle brackets around required arguments.

Also, FYI, `metavar` can be omitted if it is exactly the same as the argument name -- the argument name will be used unless metavar is specified.

stcheng: thanks for the info. so i think is it better to use square brackets here for the non-required arguments?

jleveque: Click will automatically add square brackets around optional arguments when displaying usage/help messages.

stcheng:
```
root@str-s6100-acs-2:/# config mirror_session add
Usage: config mirror_session add [OPTIONS] <session_name> <src_ip> <dst_ip>
                                 <dscp> <ttl> [GRE_TYPE] [QUEUE]
```
Click will automatically capitalize the options. Do we want that?

jleveque: By default, Click capitalizes all options and arguments. It's not my preference, but it looks OK because it's consistent. Mixing the two doesn't look good. The alternative, which would require more diligence when creating/editing commands is to enforce the use of metavars for all options and arguments. Examples in the Click documentation below:

http://click.pocoo.org/6/documentation/#meta-variables

PR #328, opened 2018-09-27
[config]: Add port channel related commands

stcheng:
config port_channel add <port_channel_name>
config port_channel del <port_channel_name>
config port_channel member add <port_channel_name> <port_name>
config port_channel member del <port_channel_name> <port_name>

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -342,6 +342,68 @@ def load_minigraph():
     _restart_services()
     print "Please note setting loaded from minigraph will be lost after system reboot. To preserve setting, run `config save`."
 
+
+#
+# 'portchannel' group
+#
+@cli.group()
+@click.pass_context
+def portchannel(ctx):
+    config_db = ConfigDBConnector()
+    config_db.connect()
+    ctx.obj = {'db': config_db}
+    pass
+
+@portchannel.command('add')
+@click.argument('portchannel_name', metavar='<portchannel_name>', required=True)
+@click.option('--min-links', default=0, type=int)
+@click.option('--fallback', default='false')
+@click.pass_context
+def add_portchannel(ctx, portchannel_name, min_links, fallback):
+    """Add port channel"""
+    db = ctx.obj['db']
+    fvs = {'admin_status': 'up',
+           'mtu': '9100'}
+    if min_links != 0:
+        fvs['min_links'] = str(min_links)
+    if fallback != 'false':
+        fvs['fallback'] = 'true'
+    db.set_entry('PORTCHANNEL', portchannel_name, fvs)
+
+@portchannel.command('del')
+@click.argument('portchannel_name', metavar='<portchannel_name>', required=True)
+@click.pass_context
+def remove_portchannel(ctx, portchannel_name):
+    """Remove port channel"""
+    db = ctx.obj['db']
+    db.set_entry('PORTCHANNEL', portchannel_name, None)
+
+@portchannel.group('member')
+@click.pass_context
+def portchannel_member(ctx):
+    pass
+
+@portchannel_member.command('add')
+@click.argument('portchannel_name', metavar='<portchannel_name>', required=True)
+@click.argument('port_name', metavar='<port_name>', required=True)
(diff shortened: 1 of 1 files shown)


config/main.py
```diff
@@ -342,6 +342,70 @@ def load_minigraph():
     _restart_services()
     print "Please note setting loaded from minigraph will be lost after system reboot. To preserve setting, run `config save`."
 
+
+#
+# 'port_channel' group
+#
+@cli.group()
+@click.pass_context
+def port_channel(ctx):
```
jleveque: In the 'show' command, we use 'portchannel', one word, no underscores, as the subcommand string. We should choose one format and use it everywhere for consistency.

config/main.py
```diff
@@ -342,6 +342,70 @@ def load_minigraph():
...
+@port_channel_member.command('del')
+@click.argument('port_channel_name', metavar='<port_channel_name>', required=True)
+@click.argument('port_name', metavar='<port_name>', required=True)
+@click.pass_context
+def del_port_channel_member(ctx, port_channel_name, port_name):
+    """Remove member from port channel"""
+    db = ctx.obj['db']
+    db.set_entry('PORTCHANNEL_MEMBER', (port_channel_name, port_name), None)
+    db.set_entry('PORTCHANNEL_MEMBER', port_channel_name + '|' + port_name, None)
+    print port_channel_name
```
jleveque: Suggest to use `click.echo()` instead of `print` for consistency here and the line below. Also, this looks like the output will be confusing as to the result of the operation. Suggest to output a more descriptive message indicating success/failure like, "Member port <port_name> removed from portchannel <port_channel_name>".

config/main.py
```diff
@@ -342,6 +342,70 @@ def load_minigraph():
...
+def add_port_channel(ctx, port_channel_name, min_links, fallback):
+    """Add port channel"""
+    db = ctx.obj['db']
+    fvs = {'admin_status': 'up',
+           'mtu': '9100'}
+    if min_links != 0:
+        fvs['min_links'] = str(min_links)
+    if fallback != 'false':
+        fvs['fallback'] = 'true'
+    db.set_entry('PORTCHANNEL', port_channel_name, fvs)
```
jleveque: Suggest to output a message indicating success/failure.

config/main.py
```diff
@@ -342,6 +342,70 @@ def load_minigraph():
...
+
+@port_channel.command('del')
+@click.argument('port_channel_name', metavar='<port_channel_name>', required=True)
+@click.pass_context
+def remove_port_channel(ctx, port_channel_name):
+    """Remove port channel"""
+    db = ctx.obj['db']
+    db.set_entry('PORTCHANNEL', port_channel_name, None)
```
jleveque: Suggest to output a message indicating success/failure.

config/main.py
```diff
@@ -342,6 +342,70 @@ def load_minigraph():
...
+
+@port_channel_member.command('add')
+@click.argument('port_channel_name', metavar='<port_channel_name>', required=True)
+@click.argument('port_name', metavar='<port_name>', required=True)
+@click.pass_context
+def add_port_channel_member(ctx, port_channel_name, port_name):
+    """Add member to port channel"""
+    db = ctx.obj['db']
+    db.set_entry('PORTCHANNEL_MEMBER', (port_channel_name, port_name),
+            {'NULL': 'NULL'})
```
jleveque: Suggest to output a message indicating success/failure.

config/main.py
stcheng: this was accidentally added here. i just removed it.

config/main.py
stcheng: i will use portchannel

stcheng:
for indicating success or failure. i suggest this as a separate pull request in the future once we do have some indications about the return value. the current behavior is to write data into the database which can't be a failure unless the database is not there. for the other config commands, right now there's no indications towards success/failure. so i did the portchannel configurations the similar way.

PR #329, opened 2018-09-30
[acl_loader] fix show acl table 

richard28530:
**- What I did**
change show acl table ports parsing

**- How I did it**

**- How to verify it**
show acl table 

admin@Accton-AS7116-54X-Neph:~$ show acl table
Name    Type    Binding    Description
------  ------  ---------  -------------
1       L3      0          Counter
                1
                ,
                E
                E
                e
                e
                e
                e
                h
                h
                n
                n
                r
                r
                t
                t
                t
                t
3       L3      1          Counter
                6
                E
                e
                e
                h
                n
                r
                t
                t

admin@Accton-AS7116-54X-Neph:~$ show acl table
Name    Type    Binding     Description
------  ------  ----------  -------------
1       L3      Ethernet0   Counter
                Ethernet1
3       L3      Ethernet16  Counter

### DIFF
--- a/acl_loader/main.py
+++ b/acl_loader/main.py
@@ -471,7 +471,7 @@ class AclLoader(object):
                 if not val["ports"]:
                     data.append([key, val["type"], "", val["policy_desc"]])
                 else:
-                    ports = natsorted(val["ports"])
+                    ports = val["ports"].split(",")
                     data.append([key, val["type"], ports[0], val["policy_desc"]])
 
                     if len(ports) > 1:
--- a/show/main.py
+++ b/show/main.py
@@ -763,20 +763,29 @@ def config(redis_unix_socket_path):
     data = config_db.get_table('VLAN')
     keys = data.keys()
 
-    def mode(key, data):
-        info = []
-        for m in data.get('members', []):
-            entry = config_db.get_entry('VLAN_MEMBER', (key, m))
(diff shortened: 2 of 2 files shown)


acl_loader/main.py
```diff
@@ -471,7 +471,7 @@ def show_table(self, table_name):
                 if not val["ports"]:
                     data.append([key, val["type"], "", val["policy_desc"]])
                 else:
-                    ports = natsorted(val["ports"])
+                    ports = val["ports"].split(",")
```
lguohan: can you still sort after split?

kevinwangsk: From the configuration Wiki page https://github.com/Azure/SONiC/wiki/Configuration, the "ports" field in "ACL_TABLE" should be list rather than string. List doesn't have split function and it will crash here.  I believe the old code works well if the port is list.

qiluo-msft: @richard28530 Please check the code https://github.com/Azure/sonic-py-swsssdk/blob/7ce52d86101906e404af96fe76e04079f4afd863/src/swsssdk/configdb.py#L117-L120.
The var 'ports' is a list already. The original code is right.
Is it possible that you are testing with an old sonic-py-swsssdk?

richard28530: image

I think the documentation is wrong.

PR #335, opened 2018-10-05
[show] Eliminate need for writing template files for 'show platform summary' and 'show version'

jleveque:
- Eliminate the need to write template files to disk in both "show platform summary" and "show version"
- Now catching exception if we cannot retrieve HwSKU from ConfigDB and display "Unknown" as result

### DIFF
--- a/show/main.py
+++ b/show/main.py
@@ -2,17 +2,17 @@
 
 import click
 import errno
-import getpass
 import json
 import os
 import re
+import sonic_platform
 import subprocess
 import sys
 from click_default_group import DefaultGroup
 from natsort import natsorted
-from tabulate import tabulate
 from swsssdk import ConfigDBConnector
 from swsssdk import SonicV2Connector
+from tabulate import tabulate
 
 import mlnx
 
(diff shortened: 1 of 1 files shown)


show/main.py
```diff
@@ -878,25 +878,30 @@ def platform():
...
+    config_db = ConfigDBConnector()
+    config_db.connect()
+    data = config_db.get_table('DEVICE_METADATA')
 
-    # Create a temporary Jinja2 template file to use with sonic-cfggen
-    f = open(PLATFORM_TEMPLATE_FILE, 'w')
-    f.write(PLATFORM_TEMPLATE_CONTENTS)
-    f.close()
+    try:
+        platform = data['localhost']['platform']
```
qiluo-msft: I think the ground truth for 'platform' is in /host/machine.conf instead of ConfigDB.

jleveque: In latest commit, I now obtain information via sonic_platform module where possible, which has logic to directly parse /host/machine.conf.

show/main.py
```diff
@@ -2,17 +2,17 @@
...
 import re
+import sonic_platform
 import subprocess
 import sys
 from click_default_group import DefaultGroup
 from natsort import natsorted
-from tabulate import tabulate
 from swsssdk import ConfigDBConnector
 from swsssdk import SonicV2Connector
+from tabulate import tabulate
```
qiluo-msft: >tabulate [](start = 21, length = 8)

Just curios why it is moved?

jleveque: Alphabetical order by package name.

jleveque: In the latest commit I have reordered import statements into the following groups:

1. Standard library imports
2. Third-party imports
3. SONiC-specific imports
4. Vendor-specific imports

PR #341, opened 2018-10-11
[config] Use 'click.fail()' in place of 'raise click.Abort' in order to print error messages to user; Other minor tweaks

jleveque:
**- What I did**
- Replace calls to 'click.Abort()' with the Click context method 'fail()' which takes an error message as a parameter in order to ensure descriptive messages are printed rather than simply the message, "Aborted"

- Change entrypoint name from 'cli' to 'config'

- Rename a couple helper functions for clarity

- Replace 'print' calls with 'click.echo()' for consistency

- Enhance comments

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -87,7 +87,7 @@ def set_interface_naming_mode(mode):
     if user != "root":
         bashrc = "/home/{}/.bashrc".format(user)
     else:
-        click.get_current_context().fail("Error: Cannot set interface naming mode for root user!")
+        click.get_current_context().fail("Cannot set interface naming mode for root user!")
 
     f = open(bashrc, 'r')
     filedata = f.read()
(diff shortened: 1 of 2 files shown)


config/main.py
```diff
@@ -85,6 +85,7 @@ def set_interface_mode(mode):
     if user != "root":
         bashrc = "/home/{}/.bashrc".format(user)
     else:
+        click.echo("Error: Cannot set interface naming mode for root user!")
```
qiluo-msft: encapsulate 2 statement together into one function, so we can reuse every other place

jleveque: In my latest commit, I have replaced click.echo() + raise click.Abort() calls with one call to click.fail().

config/main.py
```diff
@@ -85,7 +85,7 @@ def set_interface_mode(mode):
     if user != "root":
         bashrc = "/home/{}/.bashrc".format(user)
     else:
-        raise click.Abort()
+        click_context.fail("Error: Cannot set interface naming mode for root user!")
```
qiluo-msft: >click_context [](start = 8, length = 13)

[https://click.palletsprojects.com/en/7.x/advanced/#global-context-access](https://click.palletsprojects.com/en/7.x/advanced/#global-context-access)
may help reduce click_context. so the change will be minimized.

PR #345, opened 2018-10-15
Some improvement for Mellanox sniffer CLI

kevinwangsk:
* Hide Mellanox sniffer CLI on other platforms
* Set enable/disable to be the argument of sniffer command rather
than the sub-command

**- What I did**
Make the CLI "config/show platform mlnx" just appear on Mellanox platform. 
**- How I did it**
Use sonic_platform.get_sonic_version_info()['asic_type'] to identify the running platform.
**- How to verify it**
Should not see "config/show platform mlnx" CLI on the other platforms.

-->

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -7,6 +7,8 @@ import json
 import subprocess
 import netaddr
 import re
+
+import sonic_platform
 from swsssdk import ConfigDBConnector
 from natsort import natsorted
 from minigraph import parse_device_desc_xml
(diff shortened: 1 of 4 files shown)


qiluo-msft:
The quick fix is reverting.
The root cause is missing /etc/sonic/sonic_version.yml in vs docker https://github.com/Azure/sonic-buildimage/issues/2134

PR #346, opened 2018-10-15
tacacs management vrf changes

kannankvs:
**- What I did**
Changes for accepting "vrfname" as an optional parameter for the command "config tacacs -v [vrfname] server_ip" and pass it to TACPLUS_SERVER in the config DB.

**- How I did it**
Added checks for argument "vrf" and passed it to config DB. Diff is at https://github.com/kannankvs/sonic-utilities/commit/bf51bb12b5c820fc40eff7d9b6d02405af2e68be  

**- How to verify it**
Tested it along with the command "config tacacs -v mgmt 10.11.55.40" and verified that the passed parameter "vrfname" is there in /etc/sonic/config_db.json file under TACPLUS_SERVER. Verified that "show tacacs" displays the newly configured vrfname for the server IP.

```
root@sonic:/etc/sonic# show tacacs
TACPLUS global auth_type login
TACPLUS global timeout 5 (default)
TACPLUS global passkey testing123

TACPLUS_SERVER address 10.11.55.40
               priority 1
               tcp_port 49
```

```
root@sonic:/etc/sonic# config tacacs add -v mgmt 10.11.55.40
root@sonic:/etc/sonic# show tacacs
TACPLUS global auth_type login
TACPLUS global timeout 5 (default)
TACPLUS global passkey testing123

TACPLUS_SERVER address 10.11.55.40
               priority 1
               tcp_port 49
               vrf mgmt

-->
```

### DIFF
--- a/config/aaa.py
+++ b/config/aaa.py
@@ -161,8 +161,8 @@ default.add_command(passkey)
 @click.option('-a', '--auth_type', help='Authentication type, default pap', type=click.Choice(["chap", "pap", "mschap", "login"]))
 @click.option('-o', '--port', help='TCP port range is 1 to 65535, default 49', type=click.IntRange(1, 65535), default=49)
 @click.option('-p', '--pri', help="Priority, default 1", type=click.IntRange(1, 64), default=1)
-@click.option('-v', '--vrf', help="Management vrf, default is no vrf")
-def add(address, timeout, key, auth_type, port, pri, vrf):
+@click.option('-m', '--use-mgmt-vrf', help="Management vrf, default is no vrf", is_flag=True)
+def add(address, timeout, key, auth_type, port, pri, use_mgmt_vrf):
     """Specify a TACACS+ server"""
     if not is_ipaddress(address):
         click.echo('Invalid ip address')
@@ -184,8 +184,8 @@ def add(address, timeout, key, auth_type, port, pri, vrf):
             data['timeout'] = str(timeout)
         if key is not None:
             data['passkey'] = key
-        if vrf is not None:
-            data['vrf'] = vrf
+        if use_mgmt_vrf :
+            data['vrf'] = "mgmt"
         config_db.set_entry('TACPLUS_SERVER', address, data)
 tacacs.add_command(add)
 
--- a/config/aaa.py
+++ b/config/aaa.py
@@ -161,7 +161,8 @@ default.add_command(passkey)
 @click.option('-a', '--auth_type', help='Authentication type, default pap', type=click.Choice(["chap", "pap", "mschap", "login"]))
 @click.option('-o', '--port', help='TCP port range is 1 to 65535, default 49', type=click.IntRange(1, 65535), default=49)
 @click.option('-p', '--pri', help="Priority, default 1", type=click.IntRange(1, 64), default=1)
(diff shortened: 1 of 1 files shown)


config/aaa.py
```diff
@@ -161,7 +161,8 @@ def passkey(ctx, secret):
 @click.option('-a', '--auth_type', help='Authentication type, default pap', type=click.Choice(["chap", "pap", "mschap", "login"]))
 @click.option('-o', '--port', help='TCP port range is 1 to 65535, default 49', type=click.IntRange(1, 65535), default=49)
 @click.option('-p', '--pri', help="Priority, default 1", type=click.IntRange(1, 64), default=1)
-def add(address, timeout, key, auth_type, port, pri):
+@click.option('-v', '--vrf', help="Management vrf, default is no vrf")
```
lguohan: prefer not to let user to specify the management vrf name, the management vrf is created internally. Here the command line option should be --use-mgmt-vrf. If option is here, then use management vrf for tacacs client. If option is not provided, then do not use management vrf for tacacs client.

nikos-github: It would be best to agree on Option 1.

lguohan: option is fine, but we need to make sure mgmt vrf is created before we configure tacacs client.

kannankvs: Got it. Can I assume that it would be merged? Accordingly I will modify the code related to parsing the parameter and using it in PAM & NSS and raise pull request for that.

nikos-github: The mgmt vrf configuration is outside the scope of tacacs. If "--use-mgmt-vrf" is passed but the bind to device with SO_BINDTODEVICE setsockopt fails, then it shouldn't operate on default vrf.

kannankvs: Agreed. I will be making changes in PAM/NSS accordingly to fail the authentication process if user had specified "--use-mgmt-vrf" and if user had NOT configured management vrf interface with fixed name "mgmt". Can I assume that the PR will be merged?

lguohan: @kannankvs, can you help to me to understand following flow. when there is ssh request, the tacacs client will be invoked to establish a new session for authentication. if --use-mgmt-vrf is applied and mgmt vrf is not there, you will fail the request. Then, mgmt vrf is created. another ssh request comes in, will tacacs client succeed this time?

Is this bind to device operation checked everytime for a new authentication or it is only check at the very first time?

kannankvs: Yes, tacacs client will try to bind every time when SSH tries to authenticate. Hence, if mgmt vrf is created later and if any new ssh request comes after that, tacacs client will succeed.

lguohan: i think we are in sync now.

kannankvs: @lguohan  - I assume you are fine with having "-v" as short form for "--use-mgmt-vrf". 
Ex: "config tacacs add -v 10.11.55.40" is same as "config tacacs add --use-mgmt-vrf 10.11.55.40". In both cases, it will set the data['vrf'] = "mgmt".

lguohan: use -m

-v is for version or verbose.

kannankvs:
@lguohan  - I did the required changes to use "-m, --use-mgmt-vrf" as discussed above. Tested the same. Requesting you to merge.

PR #351, opened 2018-10-19
neighbor advertiser script

trzhang-msft:
neighbor advertiser script
- initiate API call to Ferret server to set up mappings
- set up mirror session
- set up vxlan session

### DIFF
--- a/scripts/neighbor_advertiser
+++ b/scripts/neighbor_advertiser
@@ -1,0 +1,307 @@
+#!/usr/bin/env python
+
+import swsssdk
+import os
+import subprocess
+import json
+import requests
+import argparse
+
+
+DEFAULT_DURATION = 300
+NET_PATH = '/sys/class/net/'
+VLAN_PATH_PREFIX = 'Vlan'
+VLAN_ADDRESS_PATH = '/address'
+DEFAULT_DECODING = 'utf-8'
+TEMPLATE_PATH = '/usr/share/sonic/templates/neighbor_advertiser_addr.j2'
+VXLAN_CONFIG_PATH = '/tmp/neighbor_advertiser/neighbor_advertiser_vxlan_config.json'
+MIRRORING_CONFIG_NAME = 'neighbor_advertiser_mirroring_config.json'
+MIRRORING_CONFIG_PATH = '/tmp/neighbor_advertiser/' + MIRRORING_CONFIG_NAME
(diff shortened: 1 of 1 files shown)


scripts/neighbor_advertiser
```diff
@@ -0,0 +1,307 @@
...
+MIRRORING_CONFIG_PATH = '/tmp/neighbor_advertiser/' + MIRRORING_CONFIG_NAME
+FERRET_NEIGHBOR_ADVERTISER_API_PREFIX = '/Ferret/NeighborAdvertiser/Slices/'
+VXLAN_TUNNEL_NAME = 'neighbor_advertiser_tunnel'
+MIRROR_SESSION_NAME = 'neighbor_advertiser_session'
+MIRROR_TABLE_NAME = 'acl_table_mirror_na'
+MIRROR_RULE_NAME = 'acl_rule_mirror_na'
+
+
+def get_switch_name():
+    hostname = subprocess.check_output(['hostname'])
```
pavel-shirshov: No need to run `hostname`. You can find this information in the file system.
```
def read_line(filename):
    ret = ''
    with open(filename) as f:
        ret = f.read()
    return ret.strip()

hostname = read_line('/etc/hostname')
```

scripts/neighbor_advertiser
```diff
@@ -0,0 +1,307 @@
...
+    url = 'http://' + ferret_service_ip + ':85' + FERRET_NEIGHBOR_ADVERTISER_API_PREFIX + get_switch_name()
+    slice_obj = construct_neighbor_advertiser_slice()
+
+    response = requests.post(url, json = slice_obj)
+
+    ferret_ipv4_addr = ''
+
+    if response.ok:
+        na_configuration = json.loads(response.content)
+        ferret_ipv4_addr = na_configuration['ipv4Addr']
```
pavel-shirshov: if response is not ok?

scripts/neighbor_advertiser
```diff
@@ -0,0 +1,307 @@
...
+            'src_ip': lo_inet_addr,
+            'dst_ip': ferret_ip
+        }
+    }
+
+    vxlan_tunnel_maps = {}
+    vlan_interfaces = get_vlan_interfaces()
+    index = 0
+
+    for vlan_interface in vlan_interfaces:
```
pavel-shirshov: you could use `enumerate` here to have an index of the element and traversed object element

scripts/neighbor_advertiser
mslocrian: Or why not socket.gethostname()?

PR #362, opened 2018-10-31
Fix to get all port related attributes from config_db

paavaanan:
**- What I did**
- interface_naming_mode functions fetch the alias_interface_name via sonic-cfggen. But, with the recent change in config_db removes all port related entries when a sepecific interface is down. This affects the existing implementation of fetching port details to following fucntions [ interface_alias_to_name(), interface_name_to_allias(), InterfaceAliasConverter() ] and throws key error.

**- How I did it**

- To fix this, harnessed the existing port_config.ini file to fetch the interface details and populated a port_dict and this fetch the alias interface names quickly.

NOTE: config commands are entirely modified from linux commands to db format. This fix does not cover interface validation for those commands.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -12,19 +12,16 @@ import sonic_platform
 from swsssdk import ConfigDBConnector
 from natsort import natsorted
 from minigraph import parse_device_desc_xml
-from portconfig import get_port_config
 
 import aaa
 import mlnx
 
-PLATFORM_ROOT_PATH = '/usr/share/sonic/device'
 SONIC_CFGGEN_PATH = '/usr/local/bin/sonic-cfggen'
-HWSKU_KEY = 'DEVICE_METADATA.localhost.hwsku'
-PLATFORM_KEY = 'DEVICE_METADATA.localhost.platform'
 
 #
 # Helper functions
 #
+
 
 def run_command(command, display_cmd=False, ignore_error=False):
     """Run bash command and print output to stdout
--- a/config/main.py
+++ b/config/main.py
@@ -12,15 +12,49 @@ import sonic_platform
 from swsssdk import ConfigDBConnector
 from natsort import natsorted
 from minigraph import parse_device_desc_xml
+from portconfig import get_port_config
 
 import aaa
 import mlnx
 
-SONIC_CFGGEN_PATH = "sonic-cfggen"
+PLATFORM_ROOT_PATH = '/usr/share/sonic/device'
+SONIC_CFGGEN_PATH = '/usr/local/bin/sonic-cfggen'
+HWSKU_KEY = 'DEVICE_METADATA.localhost.hwsku'
+PLATFORM_KEY = 'DEVICE_METADATA.localhost.platform'
 
 #
 # Helper functions
 #
+
+# Returns platform and HW SKU
+def get_platform_and_hwsku():
+    try:
+        proc = subprocess.Popen([SONIC_CFGGEN_PATH, '-H', '-v', PLATFORM_KEY],
+                                stdout=subprocess.PIPE,
+                                shell=False,
+                                stderr=subprocess.STDOUT)
+        stdout = proc.communicate()[0]
+        proc.wait()
+        platform = stdout.rstrip('\n')
+
+        proc = subprocess.Popen([SONIC_CFGGEN_PATH, '-d', '-v', HWSKU_KEY],
+                                stdout=subprocess.PIPE,
+                                shell=False,
+                                stderr=subprocess.STDOUT)
+        stdout = proc.communicate()[0]
+        proc.wait()
+        hwsku = stdout.rstrip('\n')
+    except OSError, e:
+        raise OSError("Cannot detect platform")
+
+    return (platform, hwsku)
+
+
+# Get platform, hwsku and populate port_dict
+(platform, hwsku) = get_platform_and_hwsku()
+port_config = os.path.join("/usr/share/sonic/device", platform, hwsku,
+                            "port_config.ini")
+(port_dict, _) = get_port_config(platform, hwsku, port_config)
 
 
 def run_command(command, display_cmd=False, ignore_error=False):
@@ -43,12 +77,9 @@ def interface_alias_to_name(interface_alias):
     """Return default interface name if alias name is given as argument
     """
 
-    cmd = 'sonic-cfggen -d --var-json "PORT"'
-    p = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE)
-
-    port_dict = json.loads(p.stdout.read())
-
     if interface_alias is not None:
+        if not port_dict:
+            click.echo("Invalid alias interface")
         for port_name in natsorted(port_dict.keys()):
             if interface_alias == port_dict[port_name]['alias']:
                 return port_name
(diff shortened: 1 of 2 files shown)


paavaanan:
https://github.com/Azure/SONiC/issues/268

config/main.py
```diff
@@ -12,16 +12,50 @@
...
+SONIC_CFGGEN_PATH = '/usr/local/bin/sonic-cfggen'
+HWSKU_KEY = 'DEVICE_METADATA.localhost.hwsku'
+PLATFORM_KEY = 'DEVICE_METADATA.localhost.platform'
 
 #
 # Helper functions
 #
 
+# Returns platform and HW SKU
+def get_platform_and_hwsku():
```
stcheng: since this script talks directly with configuration database, it is better to retrieve the information e.g. HWSKU and platform from the database rather than calling commands

show/main.py
```diff
@@ -43,15 +49,43 @@ def read_config(self, filename):
         except configparser.NoSectionError:
             pass
 
+# Returns platform and HW SKU
+def get_platform_and_hwsku():
```
stcheng: same as above

show/main.py
```diff
@@ -43,15 +49,43 @@ def read_config(self, filename):
...
         self.alias_max_length = 0
-        cmd = 'sonic-cfggen -d --var-json "PORT"'
-        p = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE)
-        self.port_dict = json.loads(p.stdout.read())
+
+        # Get platform and hwsku
+        (platform, hwsku) = get_platform_and_hwsku()
+        port_config = os.path.join("/usr/share/sonic/device", platform, hwsku,
+                                    "port_config.ini")
+        (self.port_dict, _) = get_port_config(platform, hwsku, port_config)
```
stcheng: same as above

config/main.py
```diff
@@ -12,16 +12,50 @@
...
+        raise OSError("Cannot detect platform")
+
+    return (platform, hwsku)
+
+
+# Get platform, hwsku and populate port_dict
+(platform, hwsku) = get_platform_and_hwsku()
+port_config = os.path.join("/usr/share/sonic/device", platform, hwsku,
+                            "port_config.ini")
+(port_dict, _) = get_port_config(platform, hwsku, port_config)
```
stcheng: port name and alias mapping is also stored in the configuration database, it is better to query in that way.

config/main.py
paavaanan: 👍 I can use PORT attribute in config_db to fetch both default and alias interfaces of each port. So, _get_hwsku() and _get_platform() are not necessary.

show/main.py
paavaanan: Addressed in the new commit.

show/main.py
paavaanan: Addressd in the new commit

config/main.py
paavaanan: Agreed @stcheng. Addressed in the new commit.

stcheng:
@paavaanan this is a nice article about the git commit message: https://medium.com/@preslavrachev/what-s-with-the-50-72-rule-8a906f61f09c

config/main.py
```diff
@@ -42,13 +45,13 @@ def run_command(command, display_cmd=False, ignore_error=False):
...
-    p = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE)
-
-    port_dict = json.loads(p.stdout.read())
+    config_db = ConfigDBConnector()
+    config_db.connect()
+    port_dict = config_db.get_table('PORT')
 
     if interface_alias is not None:
+        if not port_dict:
+            click.echo("port_dict is None!")
```
jleveque: If port_dict is None, then the next line will crash. Should you do something else here in addition to printing a message in order to handle this gracefully?

config/main.py
```diff
@@ -60,13 +63,13 @@ def interface_alias_to_name(interface_alias):
...
-    p = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE)
-
-    port_dict = json.loads(p.stdout.read())
+    config_db = ConfigDBConnector()
+    config_db.connect()
+    port_dict = config_db.get_table('PORT')
 
     if interface_name is not None:
+        if not port_dict:
+            click.echo("port_dict is None!")
```
jleveque: Same concern as above.

show/main.py
```diff
@@ -43,15 +49,18 @@ def read_config(self, filename):
...
-        cmd = 'sonic-cfggen -d --var-json "PORT"'
-        p = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE)
-        self.port_dict = json.loads(p.stdout.read())
+
+        config_db = ConfigDBConnector()
+        config_db.connect()
+        self.port_dict = config_db.get_table('PORT')
+
+        if not self.port_dict:
+            click.echo("port_dict is None!")
```
jleveque: Same concern as above.

paavaanan:
> @paavaanan this is a nice article about the git commit message: https://medium.com/@preslavrachev/what-s-with-the-50-72-rule-8a906f61f09c

Thanks shoutian

config/main.py
```diff
@@ -12,17 +12,20 @@
...
+from portconfig import get_port_config
 
 import aaa
 import mlnx
 
-SONIC_CFGGEN_PATH = "sonic-cfggen"
+PLATFORM_ROOT_PATH = '/usr/share/sonic/device'
+SONIC_CFGGEN_PATH = '/usr/local/bin/sonic-cfggen'
+HWSKU_KEY = 'DEVICE_METADATA.localhost.hwsku'
+PLATFORM_KEY = 'DEVICE_METADATA.localhost.platform'
```
jleveque: Why did you add definitions for `PLATFORM_ROOT_PATH`, `HWSKU_KEY` and `PLATFORM_KEY`? They are not used anywhere in the file.

config/main.py
```diff
@@ -12,17 +12,20 @@
 from swsssdk import ConfigDBConnector
 from natsort import natsorted
 from minigraph import parse_device_desc_xml
+from portconfig import get_port_config
```
jleveque: Why add this import if you are not using it?

show/main.py
```diff
@@ -15,9 +15,15 @@
...
 from swsssdk import ConfigDBConnector
 from swsssdk import SonicV2Connector
+from portconfig import get_port_config
 
 import mlnx
 
+PLATFORM_ROOT_PATH = '/usr/share/sonic/device'
+SONIC_CFGGEN_PATH = '/usr/local/bin/sonic-cfggen'
+HWSKU_KEY = 'DEVICE_METADATA.localhost.hwsku'
+PLATFORM_KEY = 'DEVICE_METADATA.localhost.platform'
```
jleveque: Why add these four constants? They are not used anywhere in the file.

show/main.py
```diff
@@ -15,9 +15,15 @@
 import sonic_platform
 from swsssdk import ConfigDBConnector
 from swsssdk import SonicV2Connector
+from portconfig import get_port_config
```
jleveque: Why add this import if you are not using it?

config/main.py
paavaanan: Joe, it is added as part of the initiall pull request. This is not necessary. Will remove the lines

config/main.py
paavaanan: removed in the new commit

show/main.py
paavaanan: Other than SONIC_CFGGEN_PATH removed unused constants

show/main.py
paavaanan: removed in new commit

jleveque: ? This was not removed.

paavaanan: Addressed in new commit!

stcheng:
thanks for fixing that!

PR #372, opened 2018-11-08
[warm boot] move warm reboot/fast reboot knowledge to syncd service script

yxieca:
**- What I did**
This PR is enabled by sonic-buildimage PR 2238

In test, we found that when issue-ing "docker kill swss", it causes swss service to become FAILED. As result, systemctrl will execute swss stop script. Which will also stop syncd service if the warmboot flag is not set in warm boot case. Before sonic-buildimage PR 2238, this will cause syncd process being killed instead of shutting down properly in both warm/fast reboot cases. Most importantly switch_remove() is not called. For both warm boot and fast reboot, properly shutting down syncd process is very crucial to the operations.

- Setting warm boot flag before kill swss docker, so that it won't stop syncd at the same time. (syncd will be stopped explicitly later)
- In fast reboot case, syncd script has knowledge to stop syncd process before killing the docker, which satisfies the fast reboot requirement.

**- How to verify it**
Please see verification steps in sonic-buildimage PR 2238

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -147,22 +147,8 @@ if [[ "$REBOOT_TYPE" = "warm-reboot" ]]; then
     $DUMP_CMD -d 3 -o $WARM_DIR/loglevel_db.json
 fi
 
-if [[ "$REBOOT_TYPE" = "warm-reboot" ]]; then
-    # Gracefully stop syncd for warm-reboot
-    systemctl stop syncd
-elif [[ "$REBOOT_TYPE" = "fast-reboot" ]]; then
-    # syncd graceful stop for fast-reboot is supported only for Broadcom platforms only for now
-    if [[ "$sonic_asic_type" = 'broadcom' ]]; then
-        # Gracefully stop syncd
-        docker exec -i syncd /usr/bin/syncd_request_shutdown --cold > /dev/null
-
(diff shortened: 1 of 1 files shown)


qiluo-msft:
Let's wait https://github.com/Azure/sonic-buildimage/pull/2242 merged or abandoned

scripts/fast-reboot
```diff
@@ -109,15 +122,15 @@ docker kill teamd > /dev/null
 # Kill swss dockers
 docker kill swss
 
+# syncd service stop is capable of handling both warm/fast/cold shutdown
+systemctl stop syncd
```
qiluo-msft: Possible risk with these 2 lines.

`docker kill swss` will trigger swss service stop(), and will `docker stop syncd`. Then it will trigger syncd service stop().

But you manually `systemctl stop syncd` again.

yxieca: I tested with following code:

> #!/bin/bash
> 
> docker kill swss
> echo $?
> 
> sudo systemctl stop syncd
> echo $?

 sonic prompt # ./test.sh 
swss
0
0

Stopping a service after it has been stopped is no-op. Is this your concern or are you thinking about something else?

yxieca:
retest this please

yxieca:
Rebased to latest master

PR #373, opened 2018-11-08
[reboot] clear warm boot files when reboot is requested

yxieca:

**- What I did**
Both warm/fast reboot are invoking /sbin/reboot directly.

If user invoked 'reboot', one scenario could be that user tried to issue
either fast or warm reboot and failed to complete. User then tries to
recover the system with a cold reboot.

In this scenario, particularly after warm-reboot request failed, we should
remove db backup files, so that the system will come up and perform a cold
recovery.

**- How to verify it**
created a few .json files under /host/warmboot. issue reboot, and after device rebooted, these .json files are gone.

touch /host/warmboot/config_db.json
reboot
After device boot up, check /host/warmboot, config_db.json has been renamed to config_db-\<timestamp\>.json

### DIFF
--- a/scripts/reboot
+++ b/scripts/reboot
@@ -14,6 +14,16 @@ function stop_sonic_services()
     sleep 3
 }
 
+function clear_warm_boot()
+{
+    # If reboot is requested, make sure the outstanding warm-boot is cleared
+    # So the system will come up from a cold boot.
+    WB_DIR="/host/warmboot"
+    if [[ -d ${WB_DIR} ]]; then
+        rm -f ${WB_DIR}/*.json
+    fi
+}
+
 # Exit if not superuser
 if [[ "$EUID" -ne 0 ]]; then
     echo "This command must be run as root" >&2
@@ -22,6 +32,8 @@ fi
 
 # Stop SONiC services gracefully.
 stop_sonic_services
+
+clear_warm_boot
 
 # Update the reboot cause file to reflect that user issued 'reboot' command
 # Upon next boot, the contents of this file will be used to determine the
--- a/scripts/reboot
+++ b/scripts/reboot
@@ -18,9 +18,10 @@ function clear_warm_boot()
 {
     # If reboot is requested, make sure the outstanding warm-boot is cleared
     # So the system will come up from a cold boot.
-    WB_DIR="/host/warmboot"
-    if [[ -d ${WB_DIR} ]]; then
-        rm -f ${WB_DIR}/*.json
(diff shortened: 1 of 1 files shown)


scripts/reboot
```diff
@@ -14,6 +14,16 @@ function stop_sonic_services()
     sleep 3
 }
 
+function clear_warm_boot()
+{
+    # If reboot is requested, make sure the outstanding warm-boot is cleared
```
qiluo-msft: If we config warm_reboot flag in DB, deleting files are not needed.

scripts/reboot
```diff
@@ -14,6 +14,16 @@ function stop_sonic_services()
...
+    # So the system will come up from a cold boot.
+    WB_DIR="/host/warmboot"
+    if [[ -d ${WB_DIR} ]]; then
+        rm -f ${WB_DIR}/*.json
```
pavel-shirshov: I'd suggest to use `rm -fr ${WB_DIR}/*`

scripts/reboot
yxieca: We do need to delete these back up files, because database.sh would restore them, therefore restore the warm boot flag. Which is what we don't want during recovery. Particularly, removing all prevents a partial restoration.

scripts/reboot
yxieca: I thought about that. The reason I did this way was because removing the db backup files will make sure the next time o.s. boots up, database won't be restored with may or may not complete files.

I particularly didn't want to remove sai-warmboot.bin because that might be needed for investigate potential SAI warmboot bugs.

But if we backup these files before reboot, then rm -rf ${WB_DIR}/* would be the right choice.

I just thought about rename the folder, but then realized that this would rename the folder at every reboot.

I am looking at if I can trap on warm-reboot failures and rename the folder. So far I only found way to execute code on all exit codes. I'll explore this approach a bit more.

qiluo-msft: So follow your thought, the safest way is:
1. try 'config warm_reboot' to disable all the flags and save the ConfigDB
2. if config failed, remove ConfigDb json file
All the other files could be kept for troubleshooting purpose.

yxieca: We had offline discussion. For this clean up action. What we would like to achieve is:

- Do not have warm boot enabled flag persist across o.s. reboot.
- Leave any warm boot persistent store files behind for future analysis.

With that in mind, we decide to rename config_db.json as our clean up method. So all files except config_db.json would be persisted across the reboot as-is. config_db.json will be persisted with a new name.

PR #374, opened 2018-11-08
Stop teamd gracefully in both modes

pavel-shirshov:
**- What I did**
Change how we stop teamd.

**- How I did it**
In fast-reboot mode the teamd stopping method are the same.
In warm-reboot mode the teamd stopping method uses USR1 signal.

**- How to verify it**
1. Build an image with the changes and issue `sudo warm-reboot` command.
2. teamd must send lacp updates on all lagg member interfaces
3. After restart you should see lacp dump files in /host/warmboot/teamd/

-->

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -101,10 +101,20 @@ docker exec -i bgp pkill -9 bgpd
 # Kill lldp, otherwise it sends informotion about reboot
 docker kill lldp > /dev/null
 
-# Kill teamd, otherwise it gets down all LAGs
-# Note: teamd must be killed before syncd, because it will send the last packet through CPU port
-# TODO: stop teamd gracefully to allow teamd to send last valid update to be sure we'll have 90 seconds reboot time
-docker kill teamd > /dev/null
+# Stop teamd gracefully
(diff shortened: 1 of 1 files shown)


scripts/fast-reboot
```diff
@@ -101,10 +101,20 @@ docker exec -i bgp pkill -9 bgpd
...
-# Kill teamd, otherwise it gets down all LAGs
-# Note: teamd must be killed before syncd, because it will send the last packet through CPU port
-# TODO: stop teamd gracefully to allow teamd to send last valid update to be sure we'll have 90 seconds reboot time
-docker kill teamd > /dev/null
+# Stop teamd gracefully
+if [[ "$REBOOT_TYPE" = "warm-reboot" ]]; then
+    # Send USR1 signal to all teamd instances to stop them
+    # It will prepare teamd for warm-reboot
+    # Note: We must send USR1 signal before syncd, because it will send the last packet through CPU port
+    docker exec -ti teamd pkill -USR1 teamd > /dev/null
```
jipanyang: is -t necessary?   other docker exec cases use -i only.

PR #377, opened 2018-11-09
[acl_loader]: Add support for IPv6 field match

stcheng:
import ipaddr package and check if the ip address is ipv4 or ipv6
and set the corresponding rule properties


lguohan:
retest this please

lguohan:
retest this please

PR #378, opened 2018-11-12
Adding bgp's warmrestart timer and on-off knob

rodnymolina:
```
root@node1# config warm_restart bgp_timer 14

root@node1# show warm_restart config
name    enable    timer_name    timer_duration
------  --------  ------------  ----------------
bgp     false     bgp_timer     14
system  true      NULL          NULL


root@node1:/# config warm_restart enable bgp

root@node1:/# show warm_restart config
name    enable    timer_name    timer_duration
------  --------  ------------  ----------------
bgp     true      bgp_timer     14
system  false     NULL          NULL
```

Signed-off-by: Rodny Molina <rmolina@linkedin.com>

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -524,14 +524,14 @@ def warm_restart(ctx, redis_unix_socket_path):
     pass
 
 @warm_restart.command('enable')
-@click.argument('module', metavar='<module>', default='system', required=False, type=click.Choice(["system", "swss"]))
+@click.argument('module', metavar='<module>', default='system', required=False, type=click.Choice(["system", "swss", "bgp"]))
 @click.pass_context
 def warm_restart_enable(ctx, module):
     db = ctx.obj['db']
     db.mod_entry('WARM_RESTART', module, {'enable': 'true'})
 
 @warm_restart.command('disable')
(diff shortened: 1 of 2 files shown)


config/main.py
```diff
@@ -546,6 +546,15 @@ def warm_restart_neighsyncd_timer(ctx, seconds):
         ctx.fail("neighsyncd warm restart timer must be in range 1-9999")
     db.mod_entry('WARM_RESTART', 'swss', {'neighsyncd_timer': seconds})
 
+@warm_restart.command('bgp_timer')
+@click.argument('seconds', metavar='<seconds>', required=True, type=int)
+@click.pass_context
+def warm_restart_bgp_timer(ctx, seconds):
+    db = ctx.obj['db']
+    if seconds not in range(1,3600):
```
zhenggen-xu: Can you update https://github.com/Azure/sonic-swss/blob/master/doc/swss-schema.md file or use the value here that was specified in the doc?

rodnymolina: Yep, will update it. Btw, the reason i'm changing this value is to match FRR's graceful-restart range -- just for consistency.

rodnymolina: done

PR #380, opened 2018-11-16
Optimize warm-reboot speed on dumping and rebooting

qiluo-msft:
1. dump the redis parallelly
2. reboot: don't write wtmp record, don't send wall message

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -144,17 +144,25 @@ if [[ "$REBOOT_TYPE" = "warm-reboot" ]]; then
     mkdir -p $WARM_DIR
     # Note: requiring redis-dump-load
     # Save applDB in /host/warm-reboot/appl_db.json
-    $DUMP_CMD -d 0 -o $WARM_DIR/appl_db.json
+    local pids=()
+    $DUMP_CMD -d 0 -o $WARM_DIR/appl_db.json &
+    pids+=($!)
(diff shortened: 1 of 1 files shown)


pavel-shirshov:
redis has one thread for everything. So dump it parallelly could take it longer. I'd suggest to test it.

qiluo-msft:
Tested on Mellanox 2700 (2 processors)
9.10 s -> 5.69 s

scripts/fast-reboot
```diff
@@ -188,7 +196,7 @@ sync
 
 # Reboot: explicity call Linux native reboot under sbin
 echo "Rebooting to $NEXT_SONIC_IMAGE..."
-exec /sbin/reboot
+exec /sbin/reboot --no-wtmp --no-wall
```
jipanyang: With `--no-wtmp --no-wall`,  what is the reboot time improvement?   I observed reboot command taking around 15seconds to print out fist message, but Pavel mentioned he only saw ~1 second delay.

qiluo-msft: I believe it is a minor minor improvement. No harm.

qiluo-msft: @pavel-shirshov for @jipanyang 's question

lguohan: also we should send wall message, it should not be time consuming at all. Not necessary to strip down this linux feature.

PR #382, opened 2018-11-19
[reboot] update stop syncd process wording

yxieca:

**- What I did**
Change wording to match action taking in reboot: stop syncd process (not the service).

### DIFF
--- a/scripts/reboot
+++ b/scripts/reboot
@@ -9,7 +9,7 @@ REBOOT_CAUSE_FILE="/var/cache/sonic/reboot-cause.txt"
 
 function stop_sonic_services()
 {
-    echo "Stopping syncd..."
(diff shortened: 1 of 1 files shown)


scripts/reboot
```diff
@@ -9,7 +9,7 @@ REBOOT_CAUSE_FILE="/var/cache/sonic/reboot-cause.txt"
 
 function stop_sonic_services()
 {
-    echo "Stopping syncd..."
+    echo "Stopping syncd process..."
```
qiluo-msft:
```suggestion
    echo "Stopping syncd processing..."
```

jleveque: @qiluo-msft: `Stopping syncd process...` is more appropriate.

PR #383, opened 2018-11-19
[warm reboot] unload kexec before rebooting and when error exiting from warm boot

yxieca:

**- What I did**
unload kexec config before rebooting and/or during warm reboot failure clean up.


lguohan:
retest this please

PR #385, opened 2018-11-20
[show command] allow warm restart table contains empty restore_count

yxieca:

**- What I did**
In practice, a software bug could cause not having restore_count, and
we could have use case (e.g. warm-shutdown) has state without
restore_count.

It would be nice that the command show all these rows without exception.

### DIFF
--- a/show/main.py
+++ b/show/main.py
@@ -1624,12 +1624,12 @@ def state(redis_unix_socket_path):
         entry = db.get_all(db.STATE_DB, tk)
         r = []
         r.append(remove_prefix(tk, prefix))
-        if 'restore_count' not in  entry:
+        if 'restore_count' not in entry:
             r.append("")
         else:
             r.append(entry['restore_count'])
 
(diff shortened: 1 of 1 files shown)


show/main.py
```diff
@@ -1624,7 +1624,10 @@ def remove_prefix(text, prefix):
         entry = db.get_all(db.STATE_DB, tk)
         r = []
         r.append(remove_prefix(tk, prefix))
-        r.append(entry['restore_count'])
+        if 'restore_count' not in  entry:
```
stcheng: extra space before entry

yxieca: Copied from next if block :-) Will remove it.

yxieca: Done. Removed both.

yxieca:
yes. 2 stage warm shutdown.

I want to reuse the existing show command. But don't really want having to add an artificial restore_count. I use this table to sync up warm-reboot script and syncd pre-shutdown state.

PR #386, opened 2018-11-20
[acl_loader]: Add support for parsing input interface

stcheng:
If input_interface is set, it will be converted to IN_PORTS.
```
format:
"input-interface": {
  "interface-ref": {
    "config": {
      "interface": <interface_list>
    }
  }
}
```
It will be parsed and converted to:
"IN_PORTS": <interface_list>
as an attribute to the current ACL rule.


stcheng:
retest this please

PR #389, opened 2018-11-21
[warmboot] Use 'redis-cli save' to dump database to file

qiluo-msft:
**Don't merge this disruptive change!**

The 'redis-cli save' will save database content in a binary file, which is much fast than redis-dump.

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -5,6 +5,7 @@ REBOOT_TIME=$(date)
 REBOOT_CAUSE_FILE="/var/cache/sonic/reboot-cause.txt"
 REBOOT_TYPE=$(basename $0)
 WARM_DIR=/host/warmboot
+REDIS_FILE=dump.rdb
 
 function clear_warm_boot()
 {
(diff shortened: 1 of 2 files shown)


scripts/fast-reboot
```diff
@@ -139,31 +139,19 @@ systemctl stop syncd
...
+    # Delete keys in stateDB except FDB_TABLE|* and WARM_RESTA*
+    redis-cli -n 6 eval "
+        for _, k in ipairs(redis.call('keys', '*')) do
+            if not string.match(k, 'FDB_TABLE|') and not string.match(k, 'WARM_RESTART_TABLE|') then
+                redis.call('del', k)
+            end
+        end
+    " 0
+    redis-cli save
+    docker cp database:/var/lib/redis/$REDIS_FILE $WARM_DIR
```
zhenggen-xu: Can we read the location and file name from redis.conf instead off hard-code value here?

scripts/fast-reboot
```diff
@@ -34,14 +42,6 @@ case "$REBOOT_TYPE" in
...
 esac
 
-# Check root privileges
-if [[ "$EUID" -ne 0 ]]
-then
-    echo "This command must be run as root" >&2
-    exit 1
-fi
-
-
```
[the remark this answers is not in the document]
yxieca: oh, you moved up. never mind.

scripts/fast-reboot
qiluo-msft: /etc/redis/redis.conf may be not the ground truth. I can find it from `redis-cli config get '*'`.
To make life easier, I prefer keep it hard-coded now. I will revisit if later we have requirement to make it flexible.

PR #391, opened 2018-11-26
Fixing the delete vlan function

tientienle:
Base on this link https://github.com/Azure/SONiC/wiki/Configuration, there are three tables involve when configuring vlan:  VLAN, VLAN_MEMBER and VLAN_INTERFACE. 
However, in the delete vlan function in https://github.com/Azure/sonic-utilities/blob/master/config/main.py, just two table are deleted, and the VLAN_INTERFACE has been left. 

**- What I did**
    Deleting VLAN_INTERFACE table as well when deleting a vlan. 

**- How I did it**
Find the Vlan in VLAN_INTERFACE table and delete that table: 
    keys = [ (k, v) for k, v in db.get_table('VLAN_INTERFACE') if k == 'Vlan{}'.format(vid)]
    for k in keys:
        db.set_entry('VLAN_INTERFACE',k,None)

**- How to verify it** 
Config a vlan with 3 tables: VLAN, VLAN_MEMBER and VLAN_INTERFACE as in https://github.com/Azure/SONiC/wiki/Configuration
Checking information in redis before and after delete vlan as below: 

root@leaf2-10-58-0-22:/# redis-cli -n 4 keys \*Vlan5\*
1) "VLAN_INTERFACE|Vlan5|192.168.5.1/24"
2) "VLAN_MEMBER|Vlan5|Ethernet16"
3) "VLAN|Vlan5"

root@leaf2-10-58-0-22:/# config vlan del 5
root@leaf2-10-58-0-22:/# redis-cli -n 4 keys \*Vlan5\*
(empty list or set)
root@leaf2-10-58-0-22:/#

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -630,6 +630,9 @@ def del_vlan(ctx, vid):
     for k in keys:
         db.set_entry('VLAN_MEMBER', k, None)
     db.set_entry('VLAN', 'Vlan{}'.format(vid), None)
+    keys = [ (k, v) for k, v in db.get_table('VLAN_INTERFACE') if k == 'Vlan{}'.format(vid)]
+    for k in keys:
+        db.set_entry('VLAN_INTERFACE',k,None)
 
 
 #


tientienle:
Please let me know if anything needs to be more clarified.

jleveque:
@tientienle: You did not make changes to the PR title. You updated the description. Please make the title more descriptive.

stcheng:
@tientienle I believe right now the correct flow is to remove the VLAN interface IP first before removing the VLAN itself. What should be enhanced is to indicate the user that the current VLAN interface still has IP configured on it and shall not be removed directly.

tientienle:
@stcheng Thanks for your comment. I've just checked the remove ip address in interface function. That makes sense now. So this is something that users need to be aware of.

PR #399, opened 2018-12-04
[warm boot] introduce command line options to warm/fast reboot scripts

yxieca:
**- What I did**
- Add command line option parsing to warm/fast reboot script.
- Allow ignoring orchagent pausing failure with -f option.
- Add verbose output statements.
- Log progress to syslog

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -22,6 +22,7 @@ function debug()
     if [[ x"${VERBOSE}" == x"yes" ]]; then
         echo `date` $@
     fi
+    logger "$@"
 }
 
 function showHelpAndExit()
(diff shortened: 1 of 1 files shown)


scripts/fast-reboot
```diff
@@ -15,10 +17,48 @@ then
...
+                ;;
+        esac
+    done
+}
+
 sonic_asic_type=$(sonic-cfggen -y /etc/sonic/sonic_version.yml -v asic_type)
 
 function clear_warm_boot()
 {
+    debug "Failure ($?) recovering ..."
```
qiluo-msft: >recovering [](start = 24, length = 10)

cleanup

scripts/fast-reboot
```diff
@@ -15,10 +17,48 @@ then
     exit 1
 fi
 
+function debug()
+{
+    if [[ x"${VERBOSE}" == x"yes" ]]; then
+        echo `date` $@
```
qiluo-msft: >echo [](start = 8, length = 4)

This is a script for shutdown. Left message on stdout may be the best choice. How about syslog?

yxieca: Good idea. I also send these debug output to syslog.

I think you meant to say not leaving messages on stdout. But I think that is what verbose supposed to mean.

yxieca:
retest this please

Issue #403, opened 2018-12-05
show interface status, show platform summary and other cmds are broken


nikos-github:
The https://github.com/Azure/sonic-utilities/pull/298 commit is breaking show commands for any port_config.ini that doesn't have an alias in it.

Alias is optional in sonic.

This is a blocker since platform/OS stability is compromised and identification as well as status for deployment is not possible.

show interface status error:

root@sonic:/# show interface status
Traceback (most recent call last):
File "/usr/bin/show", line 9, in <module>
  load_entry_point('sonic-utilities==1.2', 'console_scripts', 'show')()
File "/usr/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 561, in load_entry_point
  return get_distribution(dist).load_entry_point(group, name)
File "/usr/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 2631, in load_entry_point
  return ep.load()
File "/usr/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 2291, in load
  return self.resolve()
File "/usr/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 2297, in resolve
  module = __import__(self.module_name, fromlist=['__name__'], level=0)
File "/usr/lib/python2.7/dist-packages/show/main.py", line 199, in <module>
  iface_alias_converter = InterfaceAliasConverter()
File "/usr/lib/python2.7/dist-packages/show/main.py", line 64, in __init__
  self.port_dict[port_name]['alias']):
KeyError: 'alias'

show platform summary syslog error:
Dec 05 05:08:31 spine-1 aos[5216]:     iface_alias_converter = InterfaceAliasConverter()
Dec 05 05:08:31 spine-1 aos[5216]:   File "/usr/lib/python2.7/dist-packages/show/main.py", line 64, in __init__
Dec 05 05:08:31 spine-1 aos[5216]:     self.port_dict[port_name]['alias']):
Dec 05 05:08:31 spine-1 aos[5216]: KeyError: 'alias'

jleveque:
@paavaanan, @padmanarayana: Please add logic to check for the existence of the "alias" field before all attempts to access it in order to prevent this crash.

Per our spec, https://github.com/Azure/SONiC/wiki/Configuration, the 'alias' field is optional in ConfigDB. Therefore, we must gracefully handle the case where it doesn't exist.

nikos-github:
[...]

PR #405, opened 2018-12-06
Fix to make alias 'key' as optional atrribute in config/show commands

paavaanan:
**- What I did**

- Relaxed the validation of alias 'key' attribute as optional one. Script won't crash in the absence of alias-names in config/show commands.

**- How I did it**
- In show/main.py we fetch the alias-interface-names at startup to calculate the maximum length. Fixed by raising exception for invalid KeyError and break the operation. So, other part of script-execution won't be affected by the absence.
- In config/main.py we need to force the user to avoid configuring the CLI ["sudo config interface_naming_mode alias" ] to alias_mode. Without any alias entries in config it is useless to proceed further. So, allow alias mode configuration only when all ports are configured with suitable alias entries.

- This fix address #403 

 **- How to verify it**
show interface status
show interface counters

```
root@sonic:/home/admin# show interface status
  Interface            Lanes    Speed    MTU            Alias    Oper    Admin
-----------  ---------------  -------  -----  ---------------  ------  -------
  Ethernet0      49,50,51,52      N/A   9100   hundredGigE1/1    down       up
  Ethernet4      53,54,55,56      N/A   9100   hundredGigE1/2    down       up
  Ethernet8      57,58,59,60      N/A   9100   hundredGigE1/3    down       up
 Ethernet12      61,62,63,64      N/A   9100   hundredGigE1/4    down       up
 Ethernet16      65,66,67,68      N/A   9100   hundredGigE1/5    down       up
 Ethernet20      69,70,71,72      N/A   9100   hundredGigE1/6    down       up
 Ethernet24      73,74,75,76      N/A   9100   hundredGigE1/7    down       up
...
..
```

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -110,7 +110,7 @@ def set_interface_naming_mode(mode):
     user = os.getenv('SUDO_USER')
     bashrc_ifacemode_line = "export SONIC_CLI_IFACE_MODE={}".format(mode)
 
-    # Check alias 'key' presence in PORT dict
+    # Ensure all interfaces have an 'alias' key in PORT dict
     config_db = ConfigDBConnector()
     config_db.connect()
     port_dict = config_db.get_table('PORT')
--- a/config/main.py
+++ b/config/main.py
@@ -124,7 +124,7 @@ def set_interface_naming_mode(mode):
             if port_dict[port_name]['alias']:
                 pass
         except KeyError:
-            click.echo("Cannot set alias mode!")
+            click.echo("Platform does not support alias mapping")
             raise click.Abort()
 
     if not user:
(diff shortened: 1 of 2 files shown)


paavaanan:
1. "You may also want to prevent the user from ever setting the naming mode to "alias" if aliases are not present." --- I already added support for this in config/main.py. With the new code change when a user tries to change the interface_naming_mode from default to "alias", it checks the presence of "alias" entry for all ports if it fails in any one of them it return "Platform does not support alias mapping error"

2. This change is included in set_interface_naming_mode() function

2. By this way we stop user from configuring alias mode at the early stage.  This approach avoid adding additional code in other part of the scripts for "alias" check. Also, minimises the possiblity of crashes in other part of code.

config/main.py
```diff
@@ -110,6 +110,23 @@ def set_interface_naming_mode(mode):
     user = os.getenv('SUDO_USER')
     bashrc_ifacemode_line = "export SONIC_CLI_IFACE_MODE={}".format(mode)
 
+    # Check alias 'key' presence in PORT dict
```
jleveque: Suggest change comment to: `# Ensure all interfaces have an 'alias' key in PORT dict`

PR #406, opened 2018-12-07
[warm reboot] provide option to reboot with "kexec -e"

yxieca:
**- What I did**
Provide option to reboot with "kexec -e".

There is a difference of 10-16 seconds between "kexec -e" and "reboot" methods. Adding the option to use either. So test would be easier.

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -40,7 +40,7 @@ function showHelpAndExit()
 
 function parseOptions()
 {
-    while getopts "vfh?" opt; do
+    while getopts "vfh?rk" opt; do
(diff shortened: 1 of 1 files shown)


yxieca:
retest this please

pavel-shirshov:
I suggest to make 1000 reboot with this option to find how often device started fsck to fix disk issues.
Also I found that docker images filesystem structure might be broken with kexec -e

yxieca:
Pavel,

I second your concern and that is the reason I am adding this option so that we could have a test script to run this test without changing installed code.

Regards,
Ying

PR #417, opened 2018-12-12
[Warm-reboot] Don't allow warm-reboot script to err exit when bgpd process doesn't exist(EXIT_CODE 1)

dzhangalibaba:
* when bgp is not configured, there is no bgpd process, we should ignore the err check.
* when bgpd does not exist, EXIT_CODE is 1, we allow it along with EXIT_CODE SUCCESS 0, for both, we make it pass the errexit check
  

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -281,7 +281,7 @@ fi
 # Kill bgpd to start the bgp graceful restart procedure
 debug "Stopping bgp ..."
 docker exec -i bgp pkill -9 zebra
-docker exec -i bgp pkill -9 bgpd
+docker exec -i bgp pkill -9 bgpd || [ $? == 1 ]
 debug "Stopped  bgp ..."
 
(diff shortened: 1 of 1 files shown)


scripts/fast-reboot
```diff
@@ -281,7 +281,7 @@ fi
 # Kill bgpd to start the bgp graceful restart procedure
 debug "Stopping bgp ..."
 docker exec -i bgp pkill -9 zebra
-docker exec -i bgp pkill -9 bgpd
+docker exec -i bgp pkill -9 bgpd || [ $? == 1 ]
```
pavel-shirshov: || true would be enough?

yxieca: || true will white list all errors. pkill return code 1 means process not found. We should be ok with this failure.

However, if the BGP is supposed to be configured, then we should not get into warm-reboot at the first place. This is for warm-reboot pre-check.

PR #419, opened 2018-12-21
[warm reboot] timeout waiting for database operation and add -x option

yxieca:
**- What I did**
- add -x option to enable script execute in -x mode for debugging purpose.
- timeout on database query so the script will fail after timeout instead of stuck.

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -34,13 +34,14 @@ function showHelpAndExit()
     echo "    -f    : force execution"
     echo "    -r    : reboot with /sbin/reboot [default]"
     echo "    -k    : reboot with /sbin/kexec -e"
+    echo "    -x    : execute script with -x flag"
 
     exit 0
 }
 
 function parseOptions()
 {
-    while getopts "vfh?rk" opt; do
+    while getopts "vfh?rkx" opt; do
         case ${opt} in
             h )
                 showHelpAndExit
@@ -59,6 +60,9 @@ function parseOptions()
                 ;;
             k )
                 REBOOT_METHOD="/sbin/kexec -e"
+                ;;
+            x )
+                set -x
                 ;;
         esac
     done
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -117,8 +117,10 @@ function wait_for_pre_shutdown_complete_or_fail()
     debug "Waiting for pre-shutdown ..."
     TABLE="WARM_RESTART_TABLE|warm-shutdown"
     STATE="requesting"
-    declare -i waitcount;
+    declare -i waitcount
+    declare -i retrycount
     waitcount=0
+    retrycount=0
     # Wait up to 60 seconds for pre-shutdown to complete
     while [[ ${waitcount} -lt 600 ]]; do
         # timeout doesn't work with -i option of "docker exec". Therefore we have
@@ -127,7 +129,11 @@ function wait_for_pre_shutdown_complete_or_fail()
 
         if [[ x"${STATE}" == x"timed out" ]]; then
             waitcount+=50
-            debug "Timed out getting pre-shutdown state (${waitcount}) ..."
+            retrycount+=1
+            debug "Timed out getting pre-shutdown state (${waitcount}) retry count ${retrycount} ..."
+            if [[ retrycount -gt 2 ]]; then
+                break
+            fi
         elif [[ x"${STATE}" != x"requesting" ]]; then
             break
         else
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -73,7 +73,7 @@ sonic_asic_type=$(sonic-cfggen -y /etc/sonic/sonic_version.yml -v asic_type)
 function clear_warm_boot()
 {
     debug "Failure ($?) cleanup ..."
-    config warm_restart disable || /bin/true
+    timeout 10 config warm_restart disable || /bin/true
     /sbin/kexec -u || /bin/true
 
     TIMESTAMP=`date +%Y%m%d-%H%M%S`
@@ -121,7 +121,7 @@ function wait_for_pre_shutdown_complete_or_fail()
     waitcount=0
     # Wait up to 60 seconds for pre-shutdown to complete
     while [[ ${waitcount} -lt 600 ]]; do
-        STATE=`/usr/bin/redis-cli -n 6 hget "${TABLE}" state`
+        STATE=`timeout 10 /usr/bin/redis-cli -n 6 hget "${TABLE}" state`
         if [[ x"${STATE}" != x"requesting" ]]; then
             break
         fi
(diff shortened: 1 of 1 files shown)


scripts/fast-reboot
```diff
@@ -117,7 +121,7 @@ function wait_for_pre_shutdown_complete_or_fail()
     waitcount=0
     # Wait up to 60 seconds for pre-shutdown to complete
     while [[ ${waitcount} -lt 600 ]]; do
-        STATE=`/usr/bin/redis-cli -n 6 hget "${TABLE}" state`
+        STATE=`timeout 10 /usr/bin/redis-cli -n 6 hget "${TABLE}" state`
```
zhenggen-xu: If timeout happened,  STATE might be in a unknown state? Might be better to check the timeout exist code before assigning the value.

scripts/fast-reboot
```diff
@@ -69,7 +73,7 @@ sonic_asic_type=$(sonic-cfggen -y /etc/sonic/sonic_version.yml -v asic_type)
 function clear_warm_boot()
 {
     debug "Failure ($?) cleanup ..."
-    config warm_restart disable || /bin/true
+    timeout 10 config warm_restart disable || /bin/true
```
zhenggen-xu: If timeout happens (check the exit status), should we bail out from the process?

scripts/fast-reboot
yxieca: STATE won't be unknown per se. It would previously be empty before 3rd commit. but that is hard to tell from if the database key was never set.

scripts/fast-reboot
yxieca: If we couldn't figure out the timeout, then we likely have to add more "timeout x" to all db calls, for now, I don't really want to litter the script with that.

zhenggen-xu: I think if we time-out for this configuration, we should not continue,  the script should exist with some reason code and print an error messages.  At this point, human intervention would be required.

scripts/fast-reboot
```diff
@@ -121,12 +121,19 @@ function wait_for_pre_shutdown_complete_or_fail()
...
+        # timeout doesn't work with -i option of "docker exec". Therefore we have
+        # to invoke docker exec directly below.
+        STATE=`timeout 5s docker exec database redis-cli -n 6 hget "${TABLE}" state; if [[ $? == 124 ]]; then echo "timed out"; fi`
+
+        if [[ x"${STATE}" == x"timed out" ]]; then
+            waitcount+=50
+            debug "Timed out getting pre-shutdown state (${waitcount}) ..."
+        elif [[ x"${STATE}" != x"requesting" ]]; then
             break
+        else
```
zhenggen-xu: I don't know how often we see the timeout or we see some significant time spent for CLI checking DB based on your experience.   In case the database check CLI took somewhere below 5seconds but much higher than 0.1 seconds,  this loop could end up getting much higher than 60seconds.

scripts/fast-reboot
zhenggen-xu: right, my point was that we should check the shell command (with timeout) exit code before we assign the value.  I saw you already add such logic in later commit.

scripts/fast-reboot
yxieca: I thought about breaking out. But I was a bit concerned about if the DB issue was transient. With this change. At least we guaranteed to end execution in 60 second. I will have more information to have decision when the DB issue is RCA'ed. For instance, we could have limited retry count, maybe as you said, no retry at all.

Because at this time, we have already killed bgp, pre shutdown syncd, if there is a chance to move forward, I think we should. With that said, I agree with you thought of ending soon to get next level intervention.

BTW, to answer your other concern of running longer than 60 seconds. I adjusted the counter increment, so the worst case we could execute 65 second.

scripts/fast-reboot
yxieca: You are absolutely right :-)

scripts/fast-reboot
yxieca: Looks like we hit a docker bug and killed the database docker. I found that people hitting this bug with docker 1.12.6. And we are still running 1.11.1.

With this information. I think limited retry would be ok. Are you ok with 3 retries and no more than 65 seconds?

yxieca: Found this discussion: https://github.com/coreos/bugs/issues/1809. The conclusion is that it is not centos bug, rather a docker bug.

scripts/fast-reboot
zhenggen-xu: Just one example so it is more clear:  If the redis-cli took some significant time say 1second.  but we got the STATE = "requesting" all the time, we are going to wait for 600 seconds in this loop.

yxieca: I was adding 50 to the loop counter when timeout happens to compensate for the 5 seconds timeout. :-)

scripts/fast-reboot
yxieca: As far as I can tell, the db container hang. The flag won't do anything at this point. I am still trying to track down how the container hang.

scripts/fast-reboot
yxieca: The normal redis-cli operations takes very little time to complete, with 600 retries, it is still roughly 60 seconds.

yxieca:
Even though this issue was triggered 1 in 3 times with syncd warm shutdown on 7260cx3 platform, warm restart syncd test won't trigger it. Which means that the syncd pre-shutdown is part of the trigger in the warm reboot sequence, but not the only trigger to this issue.

... ...
Sat Dec 22 01:02:07 UTC 2018 ==== Test iteration 29 ====
Sat Dec 22 01:02:07 UTC 2018 Initialize pre-shutdown ...
(integer) 0
Sat Dec 22 01:02:08 UTC 2018 Requesting pre-shutdown ...
requested PRE-SHUTDOWN shutdown
Sat Dec 22 01:02:08 UTC 2018 Waiting for pre-shutdown ...
Sat Dec 22 01:02:15 UTC 2018 Pre-shutdown succeeded ...
Sat Dec 22 01:02:15 UTC 2018 Warm restarting syncd ...
Sat Dec 22 01:03:25 UTC 2018 ==== Test iteration 30 ====
Sat Dec 22 01:03:25 UTC 2018 Initialize pre-shutdown ...
(integer) 0
Sat Dec 22 01:03:25 UTC 2018 Requesting pre-shutdown ...
requested PRE-SHUTDOWN shutdown
Sat Dec 22 01:03:25 UTC 2018 Waiting for pre-shutdown ...
Sat Dec 22 01:03:32 UTC 2018 Pre-shutdown succeeded ...
Sat Dec 22 01:03:32 UTC 2018 Warm restarting syncd ...

scripts/fast-reboot
yxieca: Limited retry to 3 times.

scripts/fast-reboot
```diff
@@ -69,7 +73,7 @@ sonic_asic_type=$(sonic-cfggen -y /etc/sonic/sonic_version.yml -v asic_type)
 function clear_warm_boot()
 {
     debug "Failure ($?) cleanup ..."
-    config warm_restart disable || /bin/true
+    timeout 10s config warm_restart disable || /bin/true
```
zhenggen-xu: Do you mind to add a check here and if timeout happened, it prints an error message?

yxieca: That wasn't easy. But I got it figured out :-)

yxieca:
retest this please

PR #422, opened 2018-12-31
mgmt_vrf_namespace: management vrf using namespace solution

kannankvs:
**- What I did**
Added support for management VRF using namespace solution.
Requirements that are covered are explained in the design document.

**- How I did it**
Added config commands and show commands to enable/disable the management VRF. On enabling, it creates the management namespace, attached eth0 to management namespace, creates the required iptables rules and restarts the networking service.
Detailed design is explained in the design document.

**- How to verify it**
Use the following commands to enable/disable mgmt vrf and test the features.
config vrf add mgmt
config vrf del mgmt
config interface eth0 ip add ip/mask gatewayIP
Ex: config interface eth0 ip add 10.16.206.11/24 10.16.206.1

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -6,6 +6,9 @@ import click
 import json
 import subprocess
 import netaddr
+import syslog
+import logging
+import logging.handlers
 import re
 from swsssdk import ConfigDBConnector
 from natsort import natsorted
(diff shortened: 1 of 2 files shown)


config/main.py
```diff
@@ -662,6 +665,96 @@ def del_vlan_member(ctx, vid, interface_name):
     db.set_entry('VLAN_MEMBER', (vlan_name, interface_name), None)
 
 
+def vrf_add_management_vrf():
+	"""Enable management vrf"""
+
+	cmd = 'sonic-cfggen -d --var-json "MGMT_VRF_CONFIG"'
```
pavel-shirshov: Is it possible to read this address from the config DB?
Why do you read using sonic-cfggen, but write directly to DB?

config/main.py
```diff
@@ -662,6 +665,96 @@ def del_vlan_member(ctx, vid, interface_name):
...
+	config_db.connect()
+	config_db.mod_entry('MGMT_VRF_CONFIG',"vrf_global",{"mgmtVrfEnabled": "true"})
+
+
+def vrf_delete_management_vrf():
+	"""Disable management vrf"""
+
+	cmd = 'sonic-cfggen -d --var-json "MGMT_VRF_CONFIG"'
+	p = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+	res = p.communicate()
```
pavel-shirshov: I'd extract subprocess.Popen chunk as a function, otherwise you have a lot of duplicate code

config/main.py
```diff
@@ -662,6 +665,96 @@ def del_vlan_member(ctx, vid, interface_name):
...
+    """VRF-related configuration tasks"""
+    pass
+
+
+@vrf.command('add')
+@click.argument('vrfname', metavar='<vrfname>. Type mgmt for management VRF', required=True)
+@click.pass_context
+def vrf_add (ctx, vrfname):
+    """VRF ADD"""
+    if vrfname == 'mgmt' or vrfname == 'management':
```
pavel-shirshov: why two VRF names? Do we need to choose one only?

config/main.py
```diff
@@ -787,6 +880,42 @@ def speed(ctx, verbose):
...
+        address = netaddr.IPAddress(address)
+        # There wont be any exception if the IPAddress is valid
+    except (netaddr.core.AddrFormatError, ValueError):
+        raise ValueError("Address (%s) is not in correct presentation format" %
+                         address)
+
+    if address in network:
+        return True
+    else:
+        return False
```
pavel-shirshov: `return address in network` ?

config/main.py
```diff
@@ -803,14 +932,42 @@ def ip(ctx):
 
 @ip.command()
 @click.argument("ip_addr", metavar="<ip_addr>", required=True)
+@click.argument('gw1', metavar='<default gateway IP address>', required=False)
```
pavel-shirshov: should we use `gw` here? Why `gw1`?

config/main.py
```diff
@@ -803,14 +932,42 @@ def ip(ctx):
...
 @click.pass_context
-def add(ctx, ip_addr):
+def add(ctx, ip_addr, gw1):
     """Add an IP address towards the interface"""
     config_db = ctx.obj["config_db"]
     interface_name = ctx.obj["interface_name"]
 
     if interface_name.startswith("Ethernet"):
         config_db.set_entry("INTERFACE", (interface_name, ip_addr), {"NULL": "NULL"})
+    elif interface_name.startswith("eth0"):
```
pavel-shirshov: It could match interface 'eth01'. I think it's better to compare to 'eth' here

config/main.py
```diff
@@ -828,6 +985,8 @@ def remove(ctx, ip_addr):
 
     if interface_name.startswith("Ethernet"):
         config_db.set_entry("INTERFACE", (interface_name, ip_addr), None)
+    elif interface_name.startswith("eth0"):
```
pavel-shirshov: `interface_name == `

kannankvs:
Since we faced some issues in raising pull requests on top of this pull request, we shall cancel this PR. New PR https://github.com/Azure/sonic-utilities/pull/431 is raised. Once if it is decided to use namespace solution for management VRF, all the comments given in this PR will be addressed as part of PR#431. This PR is cancelled.

PR #423, opened 2019-01-02
[reboot] stop docker service before rebooting

yxieca:
**- What I did**
After upgrading to docker-ce 18.09.0. Shutdown SONiC takes 5-10 minutes
waiting for all the running docker containers to quit. They all enventually
hit the timeout and caused the huge delay.

This is a work-around until proper solution is found.

**- How to verify it**
with the change, reboot takes little time after upgrading to docker-ce 18.09.0. Otherwise the reboot takes about 5-10 minutes on shutdown path.

### DIFF
--- a/scripts/reboot
+++ b/scripts/reboot
@@ -14,11 +14,11 @@ function stop_sonic_services()
         echo "Stopping syncd process..."
         docker exec -i syncd /usr/bin/syncd_request_shutdown --cold > /dev/null
         sleep 3
-
-        # Stop docker service to avoid 5-10 minutes delay in shutdown
-        # while waiting for existing docker containers to quit.
-        systemctl stop docker
     fi
+
+    # Stop docker service to avoid 5-10 minutes delay in shutdown
+    # while waiting for existing docker containers to quit.
+    timeout 30s systemctl stop docker
 }
 
 function clear_warm_boot()
--- a/scripts/reboot
+++ b/scripts/reboot
@@ -14,6 +14,10 @@ function stop_sonic_services()
         echo "Stopping syncd process..."
         docker exec -i syncd /usr/bin/syncd_request_shutdown --cold > /dev/null
         sleep 3
+
+        # Stop docker service to avoid 5-10 minutes delay in shutdown
+        # while waiting for existing docker containers to quit.
+        systemctl stop docker
     fi
 }
 


zhenggen-xu:
While we are figuring out the root cause, do you see needs to make any changes in fastboot script after docker upgrade? 

BTW:  "5-10 seconds" should be "5-10 minutes" in the description. :)

yxieca:
The fast reboot script is the same base as warm reboot, I suspect it would be fine but some test is needed.

Thanks for noticing the wording error :-)

yxieca:
Sorry, I guess I didn't answer your question in the way you expected.

fast/warm reboot script already stopped docker service before reboot. So they are fine.

yxieca:
Oops. The reboot issue wasn't fixed by the proper installation of docker engine. This change is still needed for now.

scripts/reboot
```diff
@@ -14,6 +14,10 @@ function stop_sonic_services()
         echo "Stopping syncd process..."
         docker exec -i syncd /usr/bin/syncd_request_shutdown --cold > /dev/null
         sleep 3
+
+        # Stop docker service to avoid 5-10 minutes delay in shutdown
+        # while waiting for existing docker containers to quit.
+        systemctl stop docker
```
zhenggen-xu: Should we put this "systemctl stop docker"  outside of the "if" statement so it is common for all platforms?

yxieca: Sharp eyes!

To add more protection, I also give this command a 30 seconds timeout.

zhenggen-xu: I assume 30 seconds would be more than enough for regular docker service stop.

yxieca: Correct. Normally stopping docker service takes about 15 seconds.

yxieca:
Retest this please

yxieca:
retest this please

yxieca:
The root cause is not clear at the moment. The problem doesn't happen when issuing "docker stop <any container" or "systemctl stop docker". But it is somehow 100% repeatable during regular reboot.

jipanyang:
Is it related to the docker corruption issue: https://github.com/Azure/sonic-buildimage/issues/2382?

yxieca:
retest this please

yxieca:
Jipan,

I think this issue is not the same as the issue you referenced.

The issue I am working on, is very particular to shutdown path of reboot. And I didn't see docker corruption afterwards, yet.

Regards,
Ying

yxieca:
retest this please

lguohan:
retest this please

PR #425, opened 2019-01-04
[sonic_instaler]: use triple sync to flush fs changes to disk

lguohan:
single sync is not enough to flush all changes to disk sometimes

**- What I did**

**- How I did it**

**- How to verify it**

-->


fyx:
Are we sure 3 syncs is enough?

PR #429, opened 2019-01-08
add show route-map and show ip prefix-list commands

tsvanduyn:
**- What I did**
I updated the /usr/lib/python2.7/dist-packages/show/main.py file to add new commands
show route-map 
show ip prefix-list 

**- How I did it**
I'm pulling the information from quagga (vtysh) into the main cli. 

**- How to verify it**
run commands
"show route-map"
and 
"show ip prefix-list" 

There was no previous command for this unless you went into vtysh 

admin@str-s6000-acs-11:/usr/lib/python2.7/dist-packages/show$ show route-map
ZEBRA:
route-map RM_SET_SRC, permit, sequence 10
  Match clauses:
  Set clauses:
    src 10.1.0.32
  Call clause:
  Action:
    Exit routemap
ZEBRA:
route-map RM_SET_SRC6, permit, sequence 10
  Match clauses:
  Set clauses:
    src fc00:1::32
  Call clause:
  Action:
    Exit routemap
BGP:
route-map FROM_TIER0_V4, permit, sequence 10
  Match clauses:
    community UPSTREAM_PREFIX
    ip address prefix-list DEFAULT_IPV4
  Set clauses:
  Call clause:
  Action:
    Exit routemap
route-map FROM_TIER0_V4, permit, sequence 20
  Match clauses:
    community LEAK_COMMUNITY
    ip address prefix-list IPV4_18_OR_LONGER
  Set clauses:
    comm-list LEAK_COMMUNITY delete
  Call clause:
  Action:
    Exit routemap
route-map FROM_TIER0_V4, permit, sequence 30
  Match clauses:
    ip address prefix-list IPV4_18_OR_LONGER
  Set clauses:
  Call clause:
  Action:
    Exit routemap
route-map FROM_TIER0_V4, deny, sequence 1000
  Match clauses:
  Set clauses:
  Call clause:
  Action:
    Exit routemap
BGP:
route-map FROM_TIER0_V6, permit, sequence 10
  Match clauses:
    community UPSTREAM_PREFIX
    ipv6 address prefix-list DEFAULT_IPV6
  Set clauses:
  Call clause:
  Action:
    Exit routemap
route-map FROM_TIER0_V6, permit, sequence 20
  Match clauses:
    community LEAK_COMMUNITY
    ipv6 address prefix-list IPV6_64_ONLY
  Set clauses:
    comm-list LEAK_COMMUNITY delete
  Call clause:
  Action:
    Exit routemap
route-map FROM_TIER0_V6, permit, sequence 30
  Match clauses:
    ipv6 address prefix-list IPV6_64_ONLY
  Set clauses:
  Call clause:
  Action:
    Exit routemap
route-map FROM_TIER0_V6, deny, sequence 1000
  Match clauses:
  Set clauses:
  Call clause:
  Action:
    Exit routemap
BGP:
route-map FROM_TIER2_V4, permit, sequence 10
  Match clauses:
  Set clauses:
    community 8075:54000 additive
  Call clause:
  Action:
    Exit routemap
BGP:
route-map FROM_TIER2_V6, permit, sequence 10
  Match clauses:
  Set clauses:
    community 8075:54000 additive
  Call clause:
  Action:
    Exit routemap
BGP:
route-map TO_TIER0_V4, permit, sequence 10
  Match clauses:
    community UPSTREAM_PREFIX
    ip address prefix-list DEFAULT_IPV4
  Set clauses:
  Call clause:
  Action:
    Exit routemap
route-map TO_TIER0_V4, permit, sequence 20
  Match clauses:
    community LEAK_COMMUNITY
  Set clauses:
  Call clause:
  Action:
    Exit routemap
route-map TO_TIER0_V4, permit, sequence 30
  Match clauses:
    community UPSTREAM_PREFIX
  Set clauses:
    community no-export additive
  Call clause:
  Action:
    Exit routemap
route-map TO_TIER0_V4, permit, sequence 1000
  Match clauses:
  Set clauses:
  Call clause:
  Action:
    Exit routemap
BGP:
route-map TO_TIER0_V6, permit, sequence 10
  Match clauses:
    community UPSTREAM_PREFIX
    ipv6 address prefix-list DEFAULT_IPV6
  Set clauses:
  Call clause:
  Action:
    Exit routemap
route-map TO_TIER0_V6, permit, sequence 20
  Match clauses:
    community LEAK_COMMUNITY
  Set clauses:
  Call clause:
  Action:
    Exit routemap
route-map TO_TIER0_V6, permit, sequence 30
  Match clauses:
    community UPSTREAM_PREFIX
  Set clauses:
    community no-export additive
  Call clause:
  Action:
    Exit routemap
route-map TO_TIER0_V6, permit, sequence 1000
  Match clauses:
  Set clauses:
  Call clause:
  Action:
    Exit routemap
BGP:
route-map TO_TIER2_V4, deny, sequence 5
  Match clauses:
    community UPSTREAM_PREFIX
  Set clauses:
  Call clause:
  Action:
    Exit routemap
route-map TO_TIER2_V4, permit, sequence 10
  Match clauses:
  Set clauses:
  Call clause:
  Action:
    Exit routemap
BGP:
route-map TO_TIER2_V6, deny, sequence 5
  Match clauses:
    community UPSTREAM_PREFIX
  Set clauses:
  Call clause:
  Action:
    Exit routemap
route-map TO_TIER2_V6, permit, sequence 10
  Match clauses:
  Set clauses:
  Call clause:
  Action:
    Exit routemap
BGP:
route-map TO_TYCOON_ER_V4, permit, sequence 100
  Match clauses:
    ip address prefix-list DEFAULT_IPV4
  Set clauses:
  Call clause:
  Action:
    Exit routemap
route-map TO_TYCOON_ER_V4, deny, sequence 10000
  Match clauses:
  Set clauses:
  Call clause:
  Action:
    Exit routemap
BGP:
route-map FROM_TYCOON_ER_V4, deny, sequence 100
  Match clauses:
    ip address prefix-list DEFAULT_IPV4
  Set clauses:
  Call clause:
  Action:
    Exit routemap
route-map FROM_TYCOON_ER_V4, permit, sequence 200
  Match clauses:
    community TYCOON_ER_COMMUNITY
    ip address prefix-list IPV4_30_OR_LONGER
  Set clauses:
  Call clause:
  Action:
    Exit routemap
route-map FROM_TYCOON_ER_V4, deny, sequence 10000
  Match clauses:
  Set clauses:
  Call clause:
  Action:
    Exit routemap
BGP:
route-map TO_TYCOON_ER_V6, permit, sequence 100
  Match clauses:
    ipv6 address prefix-list DEFAULT_IPV6
  Set clauses:
  Call clause:
  Action:
    Exit routemap
route-map TO_TYCOON_ER_V6, deny, sequence 10000
  Match clauses:
  Set clauses:
  Call clause:
  Action:
    Exit routemap
BGP:
route-map FROM_TYCOON_ER_V6, deny, sequence 100
  Match clauses:
    ipv6 address prefix-list DEFAULT_IPV6
  Set clauses:
  Call clause:
  Action:
    Exit routemap
route-map FROM_TYCOON_ER_V6, permit, sequence 200
  Match clauses:
    community TYCOON_ER_COMMUNITY
    ipv6 address prefix-list IPV6_64_ONLY
  Set clauses:
  Call clause:
  Action:
    Exit routemap
route-map FROM_TYCOON_ER_V6, deny, sequence 10000
  Match clauses:
  Set clauses:
  Call clause:
  Action:
    Exit routemap
BGP:
route-map ISOLATE, permit, sequence 10
  Match clauses:
  Set clauses:
    as-path prepend 65100
  Call clause:
  Action:
    Exit routemap

admin@str-s6000-acs-11:/usr/lib/python2.7/dist-packages/show$ show ip prefix-list 
BGP: ip prefix-list DEFAULT_IPV4: 1 entries
   seq 5 permit 0.0.0.0/0
BGP: ip prefix-list IPV4_18_OR_LONGER: 1 entries
   seq 5 permit 0.0.0.0/0 ge 18
BGP: ip prefix-list IPV4_30_OR_LONGER: 1 entries
   seq 5 permit 0.0.0.0/0 ge 30

-->

### DIFF
--- a/show/main.py
+++ b/show/main.py
@@ -2,8 +2,6 @@
 
 import errno
 import json
-import netaddr
-import netifaces
 import os
 import re
 import subprocess
@@ -62,13 +60,10 @@ class InterfaceAliasConverter(object):
             raise click.Abort()
 
         for port_name in self.port_dict.keys():
-            try:
-                if self.alias_max_length < len(
-                        self.port_dict[port_name]['alias']):
-                   self.alias_max_length = len(
-                        self.port_dict[port_name]['alias'])
-            except KeyError:
-                break
+            if self.alias_max_length < len(
+                    self.port_dict[port_name]['alias']):
+               self.alias_max_length = len(
+                    self.port_dict[port_name]['alias'])
 
     def name_to_alias(self, interface_name):
         """Return vendor interface alias if SONiC
@@ -410,6 +405,43 @@ def interfaces():
     """Show details of the network interfaces"""
     pass
 
+# 'alias' subcommand ("show interfaces alias")
+@interfaces.command()
+@click.argument('interfacename', required=False)
+def alias(interfacename):
+    """Show Interface Name/Alias Mapping"""
+
+    cmd = 'sonic-cfggen -d --var-json "PORT"'
+    p = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE)
+
+    port_dict = json.loads(p.stdout.read())
+
+    header = ['Name', 'Alias']
+    body = []
+
+    if interfacename is not None:
+        if get_interface_mode() == "alias":
+            interfacename = iface_alias_converter.alias_to_name(interfacename)
+
+        # If we're given an interface name, output name and alias for that interface only
+        if interfacename in port_dict:
+            if 'alias' in port_dict[interfacename]:
+                body.append([interfacename, port_dict[interfacename]['alias']])
+            else:
+                body.append([interfacename, interfacename])
+        else:
+            click.echo("Invalid interface name, '{0}'".format(interfacename))
+            return
+    else:
+        # Output name and alias for all interfaces
+        for port_name in natsorted(port_dict.keys()):
+            if 'alias' in port_dict[port_name]:
+                body.append([port_name, port_dict[port_name]['alias']])
+            else:
+                body.append([port_name, port_name])
+
+    click.echo(tabulate(body, header))
+
 #
 # 'neighbor' group ###
 #
@@ -457,6 +489,23 @@ def expected(interfacename):
                          neighbor_metadata_dict[device]['type']])
 
     click.echo(tabulate(body, header))
+
+# 'summary' subcommand ("show interfaces summary")
+@interfaces.command()
+@click.argument('interfacename', required=False)
+@click.option('--verbose', is_flag=True, help="Enable verbose output")
+def summary(interfacename, verbose):
+    """Show interface status and information"""
+
+    cmd = "/sbin/ifconfig"
+
+    if interfacename is not None:
+        if get_interface_mode() == "alias":
+            interfacename = iface_alias_converter.alias_to_name(interfacename)
+
+        cmd += " {}".format(interfacename)
+
+    run_command(cmd, display_cmd=verbose)
 
 
 @interfaces.group(cls=AliasedGroup, default_if_no_args=False)
@@ -792,95 +841,28 @@ def mac(vlan, port, verbose):
     run_command(cmd, display_cmd=verbose)
 
 #
+# 'show route-map' command ("show route-map") 
+#
+
+@cli.command('route-map')
+@click.argument('route_map_name', required=False)
+@click.option('--verbose', is_flag=True, help="Enable verbose output")
+def route_map(route_map_name, verbose):
+    """show route-map"""
+    cmd = 'sudo vtysh -c "show route-map'
+    if route_map_name is not None:
+        cmd += ' {}'.format(route_map_name)
+    cmd += '"'
+    run_command(cmd, display_cmd=verbose)
+#
 # 'ip' group ("show ip ...")
 #
 
 # This group houses IP (i.e., IPv4) commands and subgroups
-@cli.group(cls=AliasedGroup, default_if_no_args=False)
+@cli.group()
 def ip():
     """Show IP (IPv4) commands"""
     pass
-
-
-#
-# get_if_admin_state
-#
-# Given an interface name, return its admin state reported by the kernel.
-#
-def get_if_admin_state(iface):
-    admin_file = "/sys/class/net/{0}/flags"
-
-    try:
-        state_file = open(admin_file.format(iface), "r")
-    except IOError as e:
-        print "Error: unable to open file: %s" % str(e)
-        return "error"
-
-    content = state_file.readline().rstrip()
-    flags = int(content, 16)
-
-    if flags & 0x1:
-        return "up"
-    else:
-        return "down"
-
-
-#
-# get_if_oper_state
-#
-# Given an interface name, return its oper state reported by the kernel.
-#
-def get_if_oper_state(iface):
-    oper_file = "/sys/class/net/{0}/carrier"
-
-    try:
-        state_file = open(oper_file.format(iface), "r")
-    except IOError as e:
-        print "Error: unable to open file: %s" % str(e)
-        return "error"
-
-    oper_state = state_file.readline().rstrip()
-    if oper_state == "1":
-        return "up"
-    else:
-        return "down"
-
-
-#
-# 'show ip interfaces' command
-#
-# Display all interfaces with an IPv4 address and their admin/oper states.
-# Addresses from all scopes are included. Interfaces with no addresses are
-# excluded.
-#
-@ip.command()
-def interfaces():
-    """Show interfaces IPv4 address"""
-    header = ['Interface', 'IPv4 address/mask', 'Admin/Oper']
-    data = []
-
-    interfaces = natsorted(netifaces.interfaces())
-
-    for iface in interfaces:
-        ipaddresses = netifaces.ifaddresses(iface)
-
-        if netifaces.AF_INET in ipaddresses:
-            ifaddresses = []
-            for ipaddr in ipaddresses[netifaces.AF_INET]:
-                netmask = netaddr.IPAddress(ipaddr['netmask']).netmask_bits()
-                ifaddresses.append(["", str(ipaddr['addr']) + "/" + str(netmask)])
-
-            if len(ifaddresses) > 0:
-                admin = get_if_admin_state(iface)
-                if admin == "up":
-                    oper = get_if_oper_state(iface)
-                else:
-                    oper = "down"
-                data.append([iface, ifaddresses[0][1], admin + "/" + oper])
-            for ifaddr in ifaddresses[1:]:
-                data.append(["", ifaddr[1], ""])
-
-    print tabulate(data, header, tablefmt="simple", stralign='left', missingval="")
 
 
 #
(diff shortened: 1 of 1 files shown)


show/main.py
```diff
@@ -62,13 +60,10 @@ def __init__(self):
             raise click.Abort()
 
         for port_name in self.port_dict.keys():
-            try:
```
jleveque: Why are you removing this try/except block?

show/main.py
```diff
@@ -410,6 +405,43 @@ def interfaces():
     """Show details of the network interfaces"""
     pass
 
+# 'alias' subcommand ("show interfaces alias")
+@interfaces.command()
+@click.argument('interfacename', required=False)
+def alias(interfacename):
```
jleveque: This command was intentionally removed. Did you mean to add it back, or was your code out-of-date?

show/main.py
```diff
@@ -458,6 +490,23 @@ def expected(interfacename):
 
     click.echo(tabulate(body, header))
 
+# 'summary' subcommand ("show interfaces summary")
+@interfaces.command()
+@click.argument('interfacename', required=False)
+@click.option('--verbose', is_flag=True, help="Enable verbose output")
+def summary(interfacename, verbose):
```
jleveque: This command was intentionally removed. Did you mean to add it back, or was your code out-of-date?

show/main.py
```diff
@@ -792,97 +841,30 @@ def mac(vlan, port, verbose):
...
     """Show IP (IPv4) commands"""
     pass
 
 
-#
-# get_if_admin_state
-#
-# Given an interface name, return its admin state reported by the kernel.
-#
-def get_if_admin_state(iface):
```
jleveque: Why are you removing these? Was your code out-of-date?

show/main.py
```diff
@@ -915,49 +913,12 @@ def protocol(verbose):
...
 
-#
-# 'show ipv6 interfaces' command
-#
-# Display all interfaces with an IPv6 address and their admin/oper states.
-# Addresses from all scopes are included. Interfaces with no addresses are
-# excluded.
-#
-@ipv6.command()
-def interfaces():
```
jleveque: Again, why are you removing this? Was your code out-of-date?

show/main.py
```diff
@@ -1542,13 +1503,22 @@ def tacacs():
...
+    """Show mirroring (Everflow) information"""
+    pass
+
+
+# 'session' subcommand  ("show mirror session")
+@mirror.command()
 @click.argument('session_name', required=False)
 @click.option('--verbose', is_flag=True, help="Enable verbose output")
-def mirror_session(session_name, verbose):
+def session(session_name, verbose):
```
jleveque: Again, I don't think you meant to modify this. I believe your code was out-of-date.

show/main.py
tsvanduyn: My bad. I thought I had grabbed the most recent version of the file but it looks like I missed a step.   I've updated this now.

show/main.py
tsvanduyn: My bad. I thought I had grabbed the most recent version of the file but it looks like I missed a step.   I've updated this now.

show/main.py
tsvanduyn: My bad. I thought I had grabbed the most recent version of the file but it looks like I missed a step.   I've updated this now.

show/main.py
tsvanduyn: My bad. I thought I had grabbed the most recent version of the file but it looks like I missed a step.   I've updated this now.

show/main.py
```diff
@@ -791,6 +791,21 @@ def mac(vlan, port, verbose):
...
+@click.argument('route_map_name', required=False)
+@click.option('--verbose', is_flag=True, help="Enable verbose output")
+def route_map(route_map_name, verbose):
+    """show route-map"""
+    cmd = 'sudo vtysh -c "show route-map'
+    if route_map_name is not None:
+        cmd += ' {}'.format(route_map_name)
+    cmd += '"'
+    run_command(cmd, display_cmd=verbose)
+
```
nikos-github: Since both commands have options, I would suggest following what I have done for show bgp:

elif routing_stack == "frr":
    @cli.command()
    @click.argument('bgp_args', nargs = -1, required = False)
    def bgp(bgp_args, verbose):
        """BGP information"""
        bgp_cmd = "show bgp"
        for arg in bgp_args:
            bgp_cmd += " " + str(arg)
        cmd = 'sudo vtysh -c "{}"'.format(bgp_cmd)
        run_command(cmd, display_cmd=verbose)

That way the command you want will work along with all the options following it.

nikos-github: Something like so:
''''
def route_map()route_map_args, verbose):
    route_map_cmd = "show route-map"
    for arg in route_map_args:
        route_map_cmd += " " + str(arg)
    cmd = 'sudo vtysh -c "{}"'.format(route_map_cmd)
    run_command(cmd, display_cmd=verbose)
''''

jleveque: @nikos-github: I like your solution, as it is simple and allows the user to pass multiple parameters. However, without explicitly defining the available parameters in the SONiC CLI, how is the user to know which parameters are available? We can't expect the user to know what is being called under the hood. This is especially a concern with commands like the BGP commands where the underlying implementation is containerized, and can be different on different devices (Quagga, FRR, etc.). I prefer defining all parameters and flags explicitly in the SONiC CLI, that way they appear in the help message for that command.

nikos-github: @jleveque The options/parameters can be obtained by "?" as always (doesn't work now but @rodnymolina has it working with vtysh). If LI doesn't want to upstream to FRR I can redevelop and push. Regarding the different routing application, that should be achievable by having the specific .py files depending on the application. We do that today, right? The other advantage of the proposal is that it immediately unlocks the entire parse chain and options without having to redevelop and also shields us from having to react to parse chain changes in the routing application that could cause breakages outside.

jleveque: >The other advantage of the proposal is that it immediately unlocks the entire parse chain and options without having to redevelop and also shields us from having to react to parse chain changes in the routing application that could cause breakages outside.

I completely agree with this. Like I said earlier, I do like this idea ... a lot, actually. However,

>Regarding the different routing application, that should be achievable by having the specific .py files depending on the application. We do that today, right?

Yes, we do. That is not my concern. My concern is that if available options/arguments are not available via the command help, having multiple different underlying applications makes things even worse, as the available options/arguments may be different from platform to platform.

A user should be able to see all available arguments and options for *any and all* SONiC command by simply entering `show <command> --help`. If this is possible with your method, then I'm 100% for using that wherever possible. Can you confirm this?

nikos-github: As I mentioned, today that doesn't work because the work in vtysh hasn't been upstreamed. But  it exists and it's working and I can confirm that. Should be no problem to replace --help with "?" and pass it along so that it works too.

jleveque: >Should be no problem to replace --help with "?" and pass it along so that it works too.

The `--help`/`-h`/`-?` flags (whatever we define as a help flag) are processed internally by Click; if any of these are passed, the actual command function is not called; Click outputs a help message and exits. I'm not sure if we can override this behavior.

nikos-github: A couple of things:

- There was no problem with "?". We can check for the variations with "-" in front.
- I was actually wrong and the "?" support has already been upstreamed by Rodny (see https://github.com/FRRouting/frr/pull/2221). We will pick it up the moment we upgrade sonic to frr 6.0 (it's in my task list for this month). For the quagga users, we can use the same thing I did for bgp and checking the routing stack. It may not matter after February since the plan is for us to get the default routing stack to be frr in sonic (again, this is in my task list per schedule for this month).

jleveque: > There was no problem with "?". We can check for the variations with "-" in front.

There would be no problem with `?` because it is not a special flag in Click. It would be added to `bgp_args` and passed to vtysh which would produce the help message from vtysh.  However, the `-h/--help/-?` flags are registered as special help flags. These flags are processed by Click. If one of these flags is passed, the underlying command function is not called; Click prints a help message and exits. Again, I'm not sure how easy it is (or if it's even possible) to override behavior for help flags in Click.

I say we merge this PR as-is and wait for the transition to FRR as the default routing stack to implement your solution, hopefully also giving time for someone to research if we can override the behavior for help arguments on a command-by-command basis.

nikos-github: I just thought it's a good idea and we could do now. I'm fine either way.

PR #436, opened 2019-01-18
[teamshow]: Fixed teamshow output not displaying the custom portchann…

madhupalu:
…els #276

commit 4e80674b95b436a155bc4f0de3e158d757ec4fd2 (HEAD -> teamshow, origin/teamshow)
Author: madhu Pal <madhupa@aviznetworks.com>
Date:   Fri Jan 18 04:45:06 2019 +0000

    [teamshow]: Fixed teamshow output not displaying the custom portchannels #276

**- What I did**
Custom port channels created by user are displayed incorrectly in teamshow output. 
The root cause is - teamshow is expecting a "Portchannel" prefix in configured port channels. Hence it is incorrectly displays the port channels when the there is no prefix. Ex:  1) TestingChannel 2) Interface. 

**- How I did it**
Fixed code in teamshow to display the port channels currently irrespective of "Portchannel" Prefix.

**- How to verify it**
root@sonic-testing:/home/admin# config portchannel add TestingChannel
root@sonic-testing:/home/admin# teamshow
No.    Team Dev         Protocol     Ports
-----  ---------------  -----------  -------------------------
001    PortChannel001   LACP(A)(Dw)  Ethernet8(D) Ethernet4(D)
**nel    PortChannelnel**   LACP(A)(Dw)  N/A

**AFTER FIX**
root@sonic-testing:/home/admin# config portchannel add TestingChannel
root@sonic-testing:/home/admin# teamshow
No.    Team Dev         Protocol     Ports
________________________________________________________________________
001    PortChannel001   LACP(A)(Dw)  Ethernet8(D) Ethernet4(D)
**nel TestingChannel** LACP(A)(Dw) N/A

root@sonic:/home/admin# ifconfig TestingChannel
**TestingChannel:** flags=4099<UP,BROADCAST,MULTICAST>  mtu 9100
        ether 52:54:00:12:34:56  txqueuelen 1000  (Ethernet)
        RX packets 0  bytes 0 (0.0 B)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 0  bytes 0 (0.0 B)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0

### DIFF
--- a/scripts/teamshow
+++ b/scripts/teamshow
@@ -63,7 +63,7 @@ class Teamshow(object):
         """
             Skip the 'PortChannel' prefix and extract the team id.
         """
-        return team[11:]
+        return team[:]
 
     def get_teamdctl(self):
         """
@@ -123,7 +123,7 @@ class Teamshow(object):
         header = ['No.', 'Team Dev', 'Protocol', 'Ports']
         output = []
         for team_id in natsorted(self.summary):
-            output.append([team_id, 'PortChannel'+team_id, self.summary[team_id]['protocol'], self.summary[team_id]['ports']])
+            output.append([team_id[11:], team_id, self.summary[team_id]['protocol'], self.summary[team_id]['ports']])
         print tabulate(output, header)
 
 def main():


madhupalu:
In addition that, confirmed custom portchannels are operational with broadcom.

ramachandrareddygaddam:
In the display first 11 characters displayed under 'Team dev' and remaining characters displayed under 'No.'.   
Instead of that user input can be restricted to enter portchannel<id>.

madhupalu:
I fixed this code, as SONiC is allowed custom port channels in the hardware and it's operational,  so not restricting the user.  
For instance, user wanted to program portchannel name as Uplink/serverfacing user defined names, we should allow, I do agree that this is not the use case SONiC trying to solve.  
My opinion, user could provision to have port channel name and port channel id, this could be more involved.

ramachandrareddygaddam:
I agree that you fix is allowing to display the user entered custom port channels, instead of fixed name 'PortChannel' + team_id.

In your review the team_id value  is taken from team_id[11:].  What happens when users enters the port channel name enters less than 11 characters.

madhupalu:
(**No**)it displayed nothing, empty. Hope (**No**)it doesn't have significance in show commands.  Referring team show header below(**No**)
**No. Team Dev Protocol Ports**

ramachandrareddygaddam:
If (No) doesn't have significance in show commands, then request you to remove from display.  In that case  there is no restriction for Port Channel Name display.

stcheng:
Right now more than one place is using "PortChannel" + No as the logic for all the port channel interfaces. This is similar to the "Vlan" + No as the VLAN interfaces. It is better to restrict instead of loosing the logic here to prevent unknown issues.

madhupalu:
Hi All, 
I'll validate the portchannels as discussed will raise a PR shortly. Only concern is, we are blocking the user from programming the user defined port channels. 

-Madhu

madhupalu:
Shuotian, 
Added a validation to allow port channels with prefix 'PortChannel' and suffix 'xxxx'.  Though I suggested PortChannelNo should be 4 digits (4096), right now not validating NO should restrict to 4 digits, user allowed configure any number. 
Please review and let us know your comments.

Thanks,
-Madhu

PR #444, opened 2019-01-24
[acl_loader]: Prevent crash in the event ac ACL entry has no 'MATCH' fields

pyadvichuk:
Fix for crash while executing 'show acl rule' in case ACL entry doesn't
  have 'MATCH' fields (e.g. default action).

How to reproduce:
* apply config having ACL rule like

        "TESTACL|DEFAULT_RULE": {
            "PRIORITY": "1",
            "PACKET_ACTION": "DROP"
        }
* execute in console
        show acl rule

Fix details:
in case ACL entry doesn't have any MATCH field just emulate one by pushing
empty string into the array.

@lguohan @jipanyang @stcheng  Please review it

### DIFF
--- a/acl_loader/main.py
+++ b/acl_loader/main.py
@@ -580,7 +580,7 @@ class AclLoader(object):
             matches.sort()
 
             if len(matches) == 0:
-                matches.append("")
+                matches.append("N/A")
 
             rule_data = [[tname, rid, priority, action, matches[0]]]
             if len(matches) > 1:
--- a/acl_loader/main.py
+++ b/acl_loader/main.py
@@ -579,6 +579,9 @@ class AclLoader(object):
 
             matches.sort()
 
+            if len(matches) == 0:
+                matches.append("")
+
             rule_data = [[tname, rid, priority, action, matches[0]]]
             if len(matches) > 1:
                 for m in matches[1:]:


acl_loader/main.py
```diff
@@ -579,6 +579,9 @@ def show_rule(self, table_name, rule_id):
 
             matches.sort()
 
+            if len(matches) == 0:
+                matches.append("")
```
jleveque: If no matches are found, is it better to leave the column empty (this solution), or print `N/A` instead?

stcheng: `N/A` would be good

stcheng:
retest this please

acl_loader/main.py
pyadvichuk: example output:
```
admin@switch:~$ show acl rule 
Table    Rule          Priority    Action    Match
-------  ------------  ----------  --------  --------------------
TESTACL  RULE_1        9999        FORWARD   DST_IPV6: 2002::1/64
                                             SRC_IPV6: 2001::1/64
TESTACL  DEFAULT_RULE  1           DROP      N/A
```

pyadvichuk:
@jleveque pls merge it.

jleveque:
@pyadvichuk: FYI I changed the PR title to be more descriptive before merging.

PR #454, opened 2019-02-03
[config/main.py] Fixed - added validations such that portchannel and …

madhupalu:
…members add/delete

                     only when it is configured (#277)

Signed-off-by: madhu Pal <madhupa@aviznetworks.com>

**- What I did**

**- How I did it**

**- How to verify it**

-->

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -450,7 +450,10 @@ def add_portchannel(ctx, portchannel_name, min_links, fallback):
 def remove_portchannel(ctx, portchannel_name):
     """Remove port channel"""
     db = ctx.obj['db']
-    db.set_entry('PORTCHANNEL', portchannel_name, None)
+    if len(db.get_entry('PORTCHANNEL', portchannel_name)) != 0:
+        db.set_entry('PORTCHANNEL', portchannel_name, None)
+    else:
+        ctx.fail("{} is not configured".format(portchannel_name))
 
 @portchannel.group('member')
 @click.pass_context
@@ -464,8 +467,12 @@ def portchannel_member(ctx):
 def add_portchannel_member(ctx, portchannel_name, port_name):
     """Add member to port channel"""
     db = ctx.obj['db']
-    db.set_entry('PORTCHANNEL_MEMBER', (portchannel_name, port_name),
-            {'NULL': 'NULL'})
+    if len(db.get_entry('PORTCHANNEL', portchannel_name)) != 0:
+        db.set_entry('PORTCHANNEL_MEMBER', (portchannel_name, port_name),
+                {'NULL': 'NULL'})
+    else:
+        ctx.fail("{} is not configured to add {}".format(portchannel_name, port_name))
+
 
 @portchannel_member.command('del')
 @click.argument('portchannel_name', metavar='<portchannel_name>', required=True)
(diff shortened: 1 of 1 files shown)


madhupalu:
**Issue Link**- https://github.com/Azure/SONiC/issues/277
**Dependent on PR** - https://github.com/Azure/sonic-py-swsssdk/pull/47

madhupalu:
thanks for the comments, I'll add interface alias test cases to the code changes. 
-Madhu

madhupalu:
I've updated the PR with all switching modes [default | alias] cases for your review. 
Thanks,
-Madhu

madhupalu:
jleveque, Can you review and provide your comments. 
thanks,

jleveque:
@madhupalu: I updated the title of this PR for clarity.

jleveque:
Retest this please

madhupalu:
Joe, Shuotian,
thanks for review comments and approved my changes. I've covered few more corner cases, updated it. 
Thanks,
-Madhu

jleveque (dismissed):

config/main.py
```diff
@@ -464,6 +470,28 @@ def portchannel_member(ctx):
...
+        if port_name is None:
+            ctx.fail("'port_name' is None!")
+    
+    if len(port_channel) == 0:
+        ctx.fail("{} doesn't exists".format(portchannel_name))
+    
+    members = port_channel.get('members', [])
+    if port_name in members:
+        if get_interface_naming_mode() == "alias":
+            port_name = interface_name_to_alias(port_name)
```
stcheng: will this line duplicate with line 475?
if it is 'alias' mode and the port_name is already in members, it will try to convert the name again?

config/main.py
```diff
@@ -464,6 +470,28 @@ def portchannel_member(ctx):
...
+            if port_name is None:
+                ctx.fail("'port_name' is None!")
+            ctx.fail("{} is already a member of {}".format(port_name, portchannel_name))
+        else:
+            ctx.fail("{} is already a member of {}".format(port_name, portchannel_name))
+    
+    members.append(port_name)
+    port_channel['members'] = members
```
stcheng: the attribute 'members' is actually not used any more in the latest code base; i think this is not needed

config/main.py
```diff
@@ -474,9 +502,32 @@ def add_portchannel_member(ctx, portchannel_name, port_name):
...
+                ctx.fail("'port_name' is None!")
+            ctx.fail("{} is not a member of {}".format(port_name, portchannel_name))
+        else:
+            ctx.fail("{} is not a member of {}".format(port_name, portchannel_name))
+    
+    members.remove(port_name)
+    if len(members) == 0:
+        del port_channel['members']
+    else:
+        port_channel['members'] = members
```
stcheng: same as above

config/main.py
madhupalu: Thanks for review comments, I'll fix and run the test cases, will update the PR.

config/main.py
madhupalu: we need the `**members**` variable to delete the port channel when there are no members in it. AFAIK, we delete port channels when there are no members in it, not sure the SONiC use case. please advise. 

Thanks
Madhu

madhupalu: addressed the review comments and updated the PR.  Thanks.

jleveque:
@stcheng: Can you please re-review?

config/main.py
stcheng: there're alternative ways of checking if the port-channel has members or not. please do not use this approach. we are trying to make the code DRY by not having duplicated information on two places; this attribute won't be supported in the future release

PR #460, opened 2019-02-14
[sonic_installer] Improve error handling

jleveque:
**- What I did**
Improve error handling in sonic_installer. Output messages to console rather than crashing and displaying backtrace.

**- How I did it**
1. If attempting to install a local image, first check whether the specified image file exists and is a regular file. If not, output an error message to the console and abort the operation.
2. If provided a URL as an image location, check for an HTTP response from the remote machine, and verify that the response is not 4xx. If either of these conditions exist, output a message to the console  and abort the operation.

**- How to verify it**

1. Pass sonic-installer the name of an image file which doesn't exist on the local machine. Instead of a backtrace, you should see the following message:
```
Image file does not exist or is not a regular file. Aborting...
Aborted!
```

2. Pass sonic-installer a URL containing an unreachable remote machine. Instead of a backtrace, you should see the following message:
```
Did not receive a response from remote machine. Aborting...
Aborted!
```

3. Pass sonic-installer a URL containing a path to a nonexistent file on a reachable remote machine. Instead of a backtrace, you should see the following message:
```
Image file not found on remote machine. Aborting...
Aborted!
```

### DIFF
--- a/sonic_installer/main.py
+++ b/sonic_installer/main.py
@@ -188,6 +188,27 @@ def get_docker_tag_name(image):
         return "unknown"
     return tag
 
+# Function which validates whether a given URL specifies an existent file
+# on a reachable remote machine. Will abort the current operation if not
+def validate_url_or_abort(url):
+    # Attempt to retrieve HTTP response code
+    try:
+        urlfile = urllib.urlopen(url)
+        response_code = urlfile.getcode()
+        urlfile.close()
+    except IOError, err:
+        response_code = None
+
+    if not response_code:
+        click.echo("Did not receive a response from remote machine. Aborting...")
+        raise click.Abort()
+    else:
+        # Check for a 4xx response code which indicates a nonexistent URL
+        if response_code / 100 == 4:
+            click.echo("Image file not found on remote machine. Aborting...")
+            raise click.Abort()
+
+
 # Callback for confirmation prompt. Aborts if user enters "n"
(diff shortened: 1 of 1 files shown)


lguohan:
retest this please

jleveque:
@jipanyang: Thanks for the suggestion. I moved the remote file validation out to a separate function and now call that function from within both the `install` and `upgrade_docker` subcommands. Please review when you have a moment.

sonic_installer/main.py
```diff
@@ -388,11 +419,12 @@ def upgrade_docker(container_name, url, cleanup_image, enforce_check, tag):
...
+        validate_url_or_abort(url)
         try:
             urllib.urlretrieve(url, DEFAULT_IMAGE_PATH, reporthook)
         except Exception, e:
             click.echo("Download error", e)
-            return
+            raise click.Abort()
         image_path = DEFAULT_IMAGE_PATH
     else:
         image_path = os.path.join("./", url)
```
jipanyang: Could you  also put the same check after docker image_path is set? Thanks!

```
        if not os.path.isfile(image_path):
            click.echo("Image file does not exist or is not a regular file. Aborting...")
            raise click.Abort()

            # TODO: Validate file is a proper SONiC docker image file
```

jleveque: Done.

PR #464, opened 2019-02-19
Get PSU status from DB directly

kevinwangsk:
* Access DB directly to get the PSU status information
* Add test case for psu CLI

**- What I did**
Refactor the CLI "show platform psustatus". As we implement a daemon for PSU to collect the status/presence information and write to state db, the CLI could get that information from state DB directly. It could improve the performance a lot. From the test on Mellanox platform, the old CLI costs around 1.8s, the new CLI just costs 5ms. The performance is > 300 times improvement.
**- How I did it**
Implement a script "psushow" to read the information from state DB directly.
**- How to verify it**
Run "show platform psustatus".

-->

### DIFF
--- a/scripts/psushow
+++ b/scripts/psushow
@@ -1,0 +1,83 @@
+#! /usr/bin/python
+
+import argparse
+import sys
+import os
+from swsssdk import SonicV2Connector
+from tabulate import tabulate
+
+# mock the redis for unit test purposes #
+try:
+    if os.environ["UTILITIES_UNIT_TESTING"] == "1":
+        modules_path = os.path.join(os.path.dirname(__file__), "..")
+        test_path = os.path.join(modules_path, "sonic-utilities-tests")
+        sys.path.insert(0, modules_path)
+        sys.path.insert(0, test_path)
+        import mock_tables.dbconnector
+except KeyError:
+    pass
+
+def psu_status_show(index):
+    db = SonicV2Connector(host="127.0.0.1")
+    db.connect(db.STATE_DB)
+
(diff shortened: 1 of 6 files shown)


kevinwangsk:
> With this PR, is the current psuutil completely deprecated? If so, should it be removed? Also, do we foresee a possible future need to write to the PSU control registers? If so, this utility should be renamed "psuutil" instead of "psushow".

No, current per platform's psuutil is still used by PSU daemon. Sure, we can rename it to psuutil from psushow.

jleveque:
I'm not referring to the psuutil.py plugin, but rather the executable file which resides under the `psuutil` directory in this repo. Does this new application supplant that?

kevinwangsk:
> I'm not referring to the psuutil.py plugin, but rather the executable file which resides under the `psuutil` directory in this repo. Does this new application supplant that?

Yes, the script/psushow is to take replace of the psuutil package. So we need to remove it.

kevinwangsk:
Take another thought, shall we retain a way to get information directly rather than from DB?

jleveque:
I think we need to retain a way to get information directly from the hardware for debugging purposes. An earlier discussions, the thought was we could have separate, small scripts inside the pmon docker for this purpose.

kevinwangsk:
Agree. I would suggest split the task into two PRs.
The first PR should implement a new script to support get PSU infor from DB directly which is covered by this PR in the "script/psushow".
The second PR is to remove the psuutil package, and add another option in "script/psushow" to support get PSU from hardware. Also need to support option '-v' to show the version and option '-n' to show the number of PSUs like what current psuutil package did. May rename "script/psushow"to "script/psuutil".

PR #467, opened 2019-02-22
[decode-syseeprom] Add option '-d' for decode-syseeprom

kevinwangsk:
* Option '-d' means read the EEPROM information from DB directly
* CLI "show platform syseeprom" will invoke decode-syseeprom with
  '-d' option

**- What I did**
Add a new option '-d' for decode-syseeprom to read EEPROM data from DB directly rather than from cache file.
It will support on the platforms which are using the TLV format in EEPROM file. As Arista uses a different format, so doesn't support this function on Arista platform until the update_eeprom_db() and read_eeprom_db() are overridden on Arista platform.
This change depends on the parent PR https://github.com/Azure/sonic-platform-common/pull/21.
Should not be merged before the parent PR is merged.
At the init stage, need to invoke command"decode-syseeprom --init" to update the EEPROM information from cache file to the database. This command needs to be added in rc.local of sonic-buildimage repo after this PR is merged.
**- How I did it**
Change the CLI 'show platform syseeprom' to invoke decode-syseeprom with the new option '-d'
**- How to verify it**
show platform syseeprom

-->

### DIFF
--- a/scripts/decode-syseeprom
+++ b/scripts/decode-syseeprom
@@ -27,7 +27,7 @@ CACHE_ROOT = '/var/cache/sonic/decode-syseeprom'
 CACHE_FILE = 'syseeprom_cache'
 
 def main():
-
+    support_eeprom_db = True
     if not os.geteuid() == 0:
         raise RuntimeError("must be root to run")
 
@@ -36,6 +36,9 @@ def main():
 
     platform_path = '/'.join([PLATFORM_ROOT, platform])
 
+    # Currently, don't support eeprom db on Arista platform
+    if 'arista' in platform_path:
+        support_eeprom_db = False
     #
     # Currently we only support board eeprom decode.
     #
@@ -55,7 +58,7 @@ def main():
     #
     # execute the command
     #
-    run(t, opts, args)
+    run(t, opts, args, support_eeprom_db)
 
 #-------------------------------------------------------------------------------
 #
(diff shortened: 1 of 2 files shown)


scripts/decode-syseeprom
```diff
@@ -75,7 +80,15 @@ def get_cmdline_opts():
 #
 # Run
 #
-def run(target, opts, args):
+def run(target, opts, args, support_eeprom_db):
+
+    if support_eeprom_db and opts.db:
+        err = target.read_eeprom_db()
+        if err:
+            print "Failed to read EEPROM information from database. Read from cache file"
```
jleveque: This function now returns zero on success but returns nothing on failure. Return value type should be consistent. Either sero/non-zero or true/false. Also, nothing appears to be checking this return value.

scripts/decode-syseeprom
```diff
@@ -109,6 +122,9 @@ def run(target, opts, args):
         pass
 
     if opts.init:
+        err = target.update_eeprom_db(e)
+        if err:
+            print "Failed to update eeprom database"
```
jleveque: Same as above.

kevinwangsk: update_eeprom_db() will return 0 for success and -1 for error.

scripts/decode-syseeprom
kevinwangsk: If you see the latest version of https://github.com/Azure/sonic-platform-common/pull/21, read_eeprom_db() will return -1 for error.

kevinwangsk:
Any comments before we merge this PR?

keboliu:
retest this please

PR #474, opened 2019-02-28
Allow config shutdown and startup operations on valid PortChannel interface names

kirankella:
**- What I did**

**- How I did it**

**- How to verify it**
config interface PortChannelxxx shutdown
config interface PortChannelxxx startup

-->


jleveque:
It doesn't appear as though the `interface_name_is_valid()` function, nor the `config interface startup/shutdown" commands have support for the interface naming mode feature (naming mode can currently be switched between name (deafult) and alias using the `config interface_naming_mode` command. Can you please check/confirm?

qiluo-msft:
@kirankella Could you response to @jleveque 's request?

kirankella:
> Please add interface naming mode support per comment above.

@jleveque Added interface naming mode support to the config startup/shutdown commands.

PR #477, opened 2019-03-01
Skip INTERFACE entries w/o prefix

marian-pritsak:
The two possible key formats of INTERFACE table are:
INTERFACE|<NAME>|<PREFIX>
INTERFACE|<NAME>

The last one carries RIF information without the prefix so it has to be
omitted

Signed-off-by: Marian Pritsak <marianp@mellanox.com>


marian-pritsak:
Retest this please

prsunny:
@marian-pritsak , guess we've a problem. If there is only one key, then it is returned as a `string `instead of `tuple `and the `len(key) `cannot be 1. Can you fix this?

PR #489, opened 2019-03-22
[route_check] Move scripts under scripts/ directory, add to setup.py

jleveque:
- Move route_check_test.sh under scripts/ directory
- Add route_check_test.sh and route_check.py to setup.py so they are included in the package


lguohan:
i'll approve this PR, but the test is not good. It should be written in pytest and use mock. @renukamanavalan

PR #494, opened 2019-04-02
Fix TODO to get/set ecn status on active ports only

wendani:
BUG: on t0 topo, if Ethernet0 is disabled, ecnconfig -q 3 will return the ecn status to be off on active ports.

Signed-off-by: Wenda Ni <wenni@microsoft.com>

**- What I did**
Operate (get & set) ecn on/off status on active ports only. 
**- How I did it**

**- How to verify it**

on brcm dut

-->

### DIFF
--- a/scripts/ecnconfig
+++ b/scripts/ecnconfig
@@ -68,11 +68,12 @@ WRED_CONFIG_FIELDS = {
     "rdrop": "red_drop_probability"
 }
 
-PORT_TABLE_NAME         = "PORT"
-QUEUE_TABLE_NAME        = "QUEUE"
-FIELD                   = "wred_profile"
-ON                      = "[WRED_PROFILE|AZURE_LOSSLESS]"
-OFF                     = "[]"
+PORT_TABLE_NAME            = "PORT"
+QUEUE_TABLE_NAME           = "QUEUE"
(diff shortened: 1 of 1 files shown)


wendani:
If there is further 201803 versions, please include this fix as well.

wendani:
:bug:

yxieca:
It might be clear for people more familiar with this code. I have some questions:

- It is unclear why DEVICE_NEIGHBOR table would contain all 'active' ports
- If a port is enabled, but link is down or haven't discovered neighbor yet, will the port be included in the table?

If any 'enabled' table is not included in the list when the command is executed. When the port later becomes 'active', will it miss ecn configuration?

jleveque:
@wendani: Can you answer Ying's questions above? I am also unclear as to why the change from PORT table to DEVICE_NEIGHBOR table is better.

pavel-shirshov:
PORT table contains all ports of the device
DEVICE_NEIGHBOR table contains all ports of the device which have a change to be activated some time.
We don't need to spend any time on ports, which are going to be down all time.

wendani:
Thanks for the answer @pavel-shirshov. Ports not in the DEVICE_NEIGHBOR are not enabled. Its ps is !ena on brcm platform

PR #498, opened 2019-04-04
Update neighbor advertiser

trzhang-msft:
Update neighbor advertiser script:

1.  add the support of vip list, and try to iterate until setup getting successful
2.  add return value, 0 if no error, -1 otherwise
3.  change -s (vips) parameter to optional, only required in 'set' mode

### DIFF
--- a/scripts/neighbor_advertiser
+++ b/scripts/neighbor_advertiser
@@ -244,7 +244,7 @@ def post_neighbor_advertiser_slice(ferret_service_vip):
     url = 'http://{}:85{}{}'.format(ferret_service_vip, FERRET_NEIGHBOR_ADVERTISER_API_PREFIX, get_switch_name())
     response = requests.post(url, json = request_slice)
     
-    ferret_server_ipv4_addr = ''
+    ferret_server_ipv4_addr = None
 
     if response.ok:
         neighbor_advertiser_configuration = json.loads(response.content)
@@ -397,9 +397,8 @@ def main():
 
     if operation_mode == 'set':
         set_success = False
-        ferret_service_vip_list = ferret_service_vips.split(',')
 
-        for ferret_service_vip in ferret_service_vip_list:
+        for ferret_service_vip in ferret_service_vips.split(','):
             ferret_server_ip = post_neighbor_advertiser_slice(ferret_service_vip)
 
             if ferret_server_ip == None:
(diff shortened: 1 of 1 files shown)


scripts/neighbor_advertiser
```diff
@@ -383,30 +382,48 @@ def reset_vxlan_tunnel():
...
-    return true
+    return 0
 
 
 if __name__ == '__main__':
     try:
         main()
     except Exception as e:
         log_error('! [Failure] {} {}'.format(e, traceback.format_exc()))
+        return -1
```
yxieca: I am not sure this would be sufficient. Can you run a test to cause failure from bash, and check "echo $?" right after the failure and make sure $? is -1?

I thought you need to call sys.exit(-1) to set the return code.

scripts/neighbor_advertiser
```diff
@@ -383,30 +382,48 @@ def reset_vxlan_tunnel():
...
+
+            if ferret_server_ip == None:
+                continue
+            else:
+                set_vxlan_tunnel(ferret_server_ip)
+                set_mirror_tunnel(ferret_server_ip)
+                set_success = True
+                break
+
+        if set_success == False:
```
yxieca: I would write it as "if not set_success".

scripts/neighbor_advertiser
```diff
@@ -383,30 +382,48 @@ def reset_vxlan_tunnel():
...
-        set_vxlan_tunnel(ferret_server_ip)
-        set_mirror_tunnel(ferret_server_ip)
+        set_success = False
+
+        for ferret_service_vip in ferret_service_vips.split(','):
+            ferret_server_ip = post_neighbor_advertiser_slice(ferret_service_vip)
+
+            if ferret_server_ip == None:
+                continue
+            else:
```
yxieca: The 3 lines above can be condensed to "if ferret_server_ip is not None:

scripts/neighbor_advertiser
trzhang-msft: update, thx

scripts/neighbor_advertiser
trzhang-msft: updated, thx

scripts/neighbor_advertiser
trzhang-msft: you're right, updated, thx

PR #504, opened 2019-04-08
[config]: Change the order of interface commands

stcheng:
before:
config interface <interface_name> COMMANDS ...
after:
config interface COMMANDS <interface_name> ...

To make backwards compatible with 201803 branch and older,
change the order of the command arguments.


stcheng:
NOTE:
once this pull request is merged, some unit tests and integration tests will need corresponding changes to make them align with the new change.

PR #505, opened 2019-04-09
[neighbor advertiser] convert int to string before concatenating

yxieca:
**- What I did**
Fix a run time string exception

**- How to verify it**
Invoke the script to with set/reset operations.


trzhang-msft:
oh... i see... thx for the fix

PR #507, opened 2019-04-10
In sync with our latest change, where we default failthrough to be False

renukamanavalan:
Reference: Pull request #2746

**- What I did**

**- How I did it**

**- How to verify it**
Run "show aaa" command and you would see default value for failthrough as false.

-->


jleveque:
Looks good, but please edit the PR title (GitHub truncated it).

PR #511, opened 2019-04-12
load_minigraph: restart hostcfgd

renukamanavalan:
**- What I did**
Add hostcfgd as one of the services to restart upon load_minigraph, as load_minigraph does flush_db and creates the DB from scratch.
Even if we may get into flush_db to ensure  proper DB updates on 'deletes' to subscribers, when starting config from scratch it is more clean with less loop-holes if we could restart the subscribers.

**- How I did it**

**- How to verify it**
1) Setup TACACS auth and see it reflect in /etc/pam.d/common-auth-sonic
2) sudo config load_minigraph
3) View /etc/pam.d/common-auth-sonic to find those TACACS servers not there anymore, as load_minigraph should have removed them.

-->

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -294,6 +294,7 @@ def _stop_services():
         'pmon',
         'bgp',
         'teamd',
+        'hostcfgd',
     ]
     for service in services:
         try:
@@ -315,6 +316,7 @@ def _restart_services():
         'lldp',
         'snmp',
         'dhcp_relay',
+        'hostcfgd',
     ]
     for service in services:
         try:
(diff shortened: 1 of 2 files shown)


config/main.py
```diff
@@ -315,6 +316,7 @@ def _restart_services():
         'lldp',
         'snmp',
         'dhcp_relay',
+        'hostcfgd',
     ]
```
yxieca: You might have to put hostcfgd at the beginning of the list?

renukamanavalan: hostcfgd has no dependency relationship with any other daemon. Its only job is to make TACPLUS auth related updates (/etc/pam.d/*,  /etc/nsswitch.conf & "/etc/tacplus_nss.conf"). hostcfgd's only requirement is accessibility to redis.

So the order does not matter.

PR #515, opened 2019-04-17
If fast-reboot-dump gives an error, don't continue with fast-reboot

pavel-shirshov:
**- What I did**
1. Add exception handling in fast-reboot-dump.py. 
2. If we catch any exception in fast-reboot-dump, stop fast-reboot procedure

**- How I did it**
Modification of the scripts

**- How to verify it**
Copy changed files in your dut and run fast-reboot command

-->

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -341,8 +341,8 @@ if [[ "$REBOOT_TYPE" = "fast-reboot" ]]; then
     # Dump the ARP and FDB tables to files also as default routes for both IPv4 and IPv6
     # into /host/fast-reboot
     mkdir -p /host/fast-reboot
-    /usr/bin/fast-reboot-dump.py -t /host/fast-reboot
-    FAST_REBOOT_DUMP_RC=$?
+    FAST_REBOOT_DUMP_RC=0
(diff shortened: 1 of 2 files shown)


scripts/fast-reboot
```diff
@@ -337,6 +342,12 @@ if [[ "$REBOOT_TYPE" = "fast-reboot" ]]; then
     # into /host/fast-reboot
     mkdir -p /host/fast-reboot
     /usr/bin/fast-reboot-dump.py -t /host/fast-reboot
+    FAST_REBOOT_DUMP_RC=$?
```
yxieca: I don't think you will have a chance to get to here in case failure because we are running this script with '-e' option, when fast-reboot-dump.py fails, it will trigger the error exit.

There is a complicated way to achieve what you want:

https://github.com/Azure/sonic-utilities/blob/master/scripts/fast-reboot#L153

yxieca: I didn't test this, but conceptually, there is a simpler format for your purpose:

FAST_REBOOT_DUMP_RC=0
/usr/bin/fast-reboot-dump.py -t /host/fast-reboot || FAST_REBOOT_DUMP_RC=$?
if [[ FAST_REBOOT_DUMP_RC -ne 0 ]]; then
...
fi

yxieca:
Retest this please

PR #518, opened 2019-04-20
[show] Remove ability to clear counters from 'show' commands

jleveque:
- Remove ability to clear counters from the following `show` commands, because all of these counters can now be cleared using the `sonic-clear` command, which is the appropriate command to use
    - `show interface counters`
    - `show pfc counters`
    - `show queue counters`

- Also fix some comments and help messages


jleveque:
Retest this please.

prsunny:
I thot some tools are still using the `-c` option. This could cause backward compatibility issue right?

jleveque:
@prsunny: In that case, this change should only get merged to the master branch, giving users time to update their tools to use the appropriate `sonic-clear` commands.

PR #529, opened 2019-05-10
Layer 2 Forwarding Enhancements

anilkpandey:
**- What I did**
1. Added code changes related to Layer 2 Forwarding Enhancements:
Added the following commands:

> VLAN range add/del
> VLAN member range add/del
> Static FDB add/del
> FDB aging time configuration
> show fdb aging time
> show mac count
> show vlan count
> added port and vlan option in 'sonic-clear fdb'


prsunny:
I'm not sure if these validations must be here. There is an design discussion for overall Config validator and I don't think we need to scatter these checks in multiple places.

anilkpandey:
Hi @stcheng and @prsunny, 

I just reverted the previous changes for validations and added a new file fdbshow for 'show mac' performance enhancements. Please review the changes.

In case you want to consider the validations changes, let me know, I will readd the changes in that case.

Thanks.

config/main.py
```diff
@@ -739,11 +739,33 @@ def add_vlan_member(ctx, vid, interface_name, untagged):
             interface_name = interface_name_to_alias(interface_name)
             if interface_name is None:
                 ctx.fail("'interface_name' is None!")
-            ctx.fail("{} is already a member of {}".format(interface_name,
-                                                        vlan_name))
+            ctx.fail("{} is already a member of {}".format(interface_name,vlan_name))
         else:
+<<<<<<< HEAD
```
stcheng: please resolve the conflicts

scripts/fdbshow
```diff
@@ -63,25 +63,51 @@ class FdbShow(object):
...
             fdb_entry = s.decode()
             fdb = json.loads(fdb_entry .split(":", 2)[-1])
             if not fdb:
                 continue
-
-            ent = self.db.get_all('ASIC_DB', s, blocking=True)
+            fdb_list.append(fdb)
+            pipe.hgetall(s)
+        values = pipe.execute()
+        client = self.db.redis_clients["ASIC_DB"]
```
prsunny: Why is this repeated?

scripts/fdbshow
```diff
@@ -63,25 +63,51 @@ class FdbShow(object):
...
+        vlans = []
+        vlan_keys = self.db.keys('ASIC_DB', "ASIC_STATE:SAI_OBJECT_TYPE_VLAN:*")
+        for v in vlan_keys:
+            pipe.hgetall(v)
+        vlans = pipe.execute()
+
+        posi = 0
+        for ent in values:
+            if 'SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID' not in ent:
+                posi = posi + 1
```
prsunny: Is this case true, where SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID is not present?

scripts/fdbshow
```diff
@@ -63,25 +63,51 @@ class FdbShow(object):
...
+                if_name = self.if_oid_map[port_id]
+            except :
+                if_name = port_id
+
             if 'vlan' in fdb:
                 vlan_id = fdb["vlan"]
             elif 'bvid' in fdb:
-                vlan_id = port_util.get_vlan_id_from_bvid(self.db, fdb["bvid"])
-            self.bridge_mac_list.append((int(vlan_id),) + (fdb["mac"],) + (if_name,) + (fdb_type,))
+                vlan_id  = vlans[vlan_keys.index("ASIC_STATE:SAI_OBJECT_TYPE_VLAN:{}".format(fdb_list[posi][u'bvid']))] [b"SAI_VLAN_ATTR_VLAN_ID"]
```
prsunny: Do you see considerable improvement over using this vs `get_vlan_id_from_bvid`? Suggest to have this logic in one place.

scripts/fdbshow
anilkpandey: yes its not needed. will remove

scripts/fdbshow
anilkpandey: If an entry is deleted either due to ageout or flush just after we fetched the keys, it will hit this case.

scripts/fdbshow
anilkpandey: get_vlan_id_from_bvid reads from DB. But, we already have all the data retrieved from DB, so just need to parse that.

prsunny: I agree to that. My point is if `get_vlan_id_from_bvid` doesn't make much of a different in performance, suggest to use the API. Any change related to BRIDGE/VLAN/BVID in future (which happened before and we had to scan all the places where `vlan `id was used) will be contained in one place rather than in multiple places.

anilkpandey: understood. get_vlan_id_from_bvid is being used in 2 more places. I will see if I can change those caller also to retrieve all VLANs first and use it. Then I can modify get_vlan_id_from_bvid to not read from DB. Thanks.

prsunny: How much performance advantage are we getting without the call? Do you have any numbers?

anilkpandey: For 32K entries, it takes 7 seconds if I don't call get_vlan_id_from_bvid and it takes 5 min if I call get_vlan_id_from_bvid.

prsunny: oh ok, that's quite an improvement. Lets keep the change then!

prsunny:
retest this please

jleveque:
Retest this please

jleveque:
Retest this please

anilkpandey:
Added code changes related to Layer 2 Forwarding Enhancements

jleveque:
Retest this please

anilkpandey:
@prsunny @stcheng , can you please complete this review? Regarding the test failures, the vnet, vrf and subport tests are failing for all PRs in this repo. Any ideas whats the reason?

anilkpandey:
Retest this please

anilkpandey:
retest this please

anilkpandey:
retest this please

ben-gale:
@prsunny - can this one be merged no pls?

scripts/fast-reboot-dump.py
```diff
@@ -147,6 +147,108 @@ def get_fdb(db, vlan_name, vlan_id, bridge_id_2_iface):
 
     return fdb_entries, available_macs, map_mac_ip
 
+def generate_fdb_entries_2(filename):
```
prsunny, line 150: Why do we have this change in fast-reboot?. @tahmed-dev for viz. fast-reboot enhancement is not mentioned in description. May require additional testing with some recent changes in this area

config/main.py
```diff
@@ -1828,6 +1841,113 @@ def delete_snmptrap_server(ctx, ver):
...
+@mac.command('add')
+@click.argument('mac', metavar='<mac-address as xx:xx:xx:xx:xx:xx>', required=True)
+@click.argument('vlan', metavar='<vlan>', required=True, type=int)
+@click.argument('interface_name', metavar='<interface name>', required=True)
+@click.pass_context
+def add_mac(ctx, mac, vlan, interface_name):
+    db = ctx.obj['db']
+
+    mac_valid = bool(re.match('^' + '[\:\-]'.join(['([0-9a-f]{2})']*6) + '$', mac.lower()))
+    if mac_valid == False:
```
prsunny, line 1900: can we have mac validation in a separate function which can help future use?

config/main.py
```diff
@@ -1828,6 +1841,113 @@ def delete_snmptrap_server(ctx, ver):
...
+
+    if get_interface_naming_mode() == "alias":
+        interface_name = interface_alias_to_name(interface_name)
+        if interface_name is None:
+            ctx.fail("'interface_name' is None!")
+
+    if interface_name_is_valid(interface_name) is False:
+        ctx.fail("Interface name is invalid!!")
+
+    db.set_entry('FDB', (vlan_name, mac), {'port': interface_name })
```
prsunny, line 1929: Currently, there is no config db support for FDB and aging time. For this, we need to to have this [PR](https://github.com/Azure/sonic-swss/pull/885), correct?

show/main.py
```diff
@@ -716,22 +716,59 @@ def pwm_buffer_pool():
...
+@mac.command()
+def aging_time():
+    """Show MAC Aging-Time"""
+    config_db = ConfigDBConnector()
+    config_db.connect()
 
-    cmd = "fdbshow"
+    # Fetching data from config_db for SWITCH
+    switch_table = config_db.get_table('SWITCH')
+    switch_keys = switch_table.keys()
```
prsunny, line 742: can we have this displayed from APP db? So even if there is config, it will display the default.

config/vlan.py
```diff
@@ -201,3 +201,333 @@ def del_vlan_dhcp_relay_destination(db, vid, dhcp_relay_destination_ip):
...
+    vlan_range_validate(ctx, vid1, vid2)
+
+    vid2 = vid2+1
+
+    warning_vlans_list = []
+    warning_membership_list = []
+    warning_ip_list = []
+    clients = db.redis_clients["CONFIG_DB"]
+    pipe = clients.pipeline()
+    vlan_member_keys = db.keys('CONFIG_DB', "*VLAN_MEMBER*")
```
tahmed-dev, line 263: suggest to use `clients.scan(...)` API

scripts/fast-reboot-dump.py
tahmed-dev: is this a replacement of the generate_dfb_entries? if yes, what is the time saving here when using this version?

scripts/fast-reboot-dump.py
```diff
@@ -147,6 +147,108 @@ def get_fdb(db, vlan_name, vlan_id, bridge_id_2_iface):
...
+        vlan_id = int(vlan_ent['SAI_VLAN_ATTR_VLAN_ID'])
+        vlan_oid = vlan_oid_list[posi]
+        map_vlan_id_to_vlan_oid[vlan_id] = vlan_oid
+        map_vlan_oid_to_vlan_id[vlan_oid] = vlan_id
+        posi = posi + 1
+    #print("generate_fdb_entries_2 after vlan-map" + datetime.datetime.now().strftime("%H:%M:%S.%f"))
+
+    #print("generate_fdb_entries_2 before fdb-key-getall" + datetime.datetime.now().strftime("%H:%M:%S.%f"))
+    pipe = client.pipeline()
+    fdb_list = db.keys(db.ASIC_DB, "ASIC_STATE:SAI_OBJECT_TYPE_FDB_ENTRY:*")
```
tahmed-dev, line 197: suggest to use `client.scan(...)` API

scripts/fast-reboot-dump.py
jainp1979: Here, written and used optimized version (generate_fdb_entries_2) of generate_fdb_entries 
1) Modifed the functionality to work based on available mac addresses rather than available vlans 
2) used bulk (pipe) asic-db apis to get vlan and fdb from ASIC db instead of getting the data for individual vlan/fdb entry 
3) stored vlan_oid to vlan_id and vice-versa mapping in local memory instead of getting the same everytime from ASIC-DB 
4) done away with get_fdb function to be able to modify the overall logic and reduce copying of objects and function calls.

measured specifically fast-reboot-dump.py script from FDB perspective and now it takes around 10 sec instead of 12 minutes with 4K empty vlans;
Also tested the modified script with 4k mac addresses in 4k vlans (1 mac per vlan) and it gets completed with in 10 sec.

anilkpandey:
retest this please

config/vlan.py
tahmed-dev: @anilkpandey [client scan](https://redis.io/commands/scan) it use cursor iterator so it does not lock the redis db for long time. [Here](https://github.com/Azure/sonic-py-swsssdk/blob/master/src/swsssdk/configdb.py#L429) is a code that uses client scan in swsssdk.

ben-gale:
> overall, lgtm, seems unit test is missing, requesting few more reviewers

@prsunny - the unit tests are covered in https://github.com/Azure/sonic-swss/pull/885 - however these can't pass without this PR (#529) merging. It also seems like Tamer has reviewed, so this covers your request. Can this one be merged now please?

lguohan:
@ben-gale, sonic utilities recently added unit test infrastructure. now all new features needs unit test. you can checkout the coverage report here. https://sonic-jenkins.westus2.cloudapp.azure.com/job/common/job/sonic-utilities-build/RCov_20Report/

it is important that we are adding unit test for new features to avoid regression.

anilkpan:
retest this please

anilkpan:
retest this please

scripts/fdbshow
```diff
@@ -31,8 +31,8 @@ import json
 import sys
 
 from natsort import natsorted
-from swsssdk import port_util
-from swsscommon.swsscommon import SonicV2Connector
+from swsssdk import SonicV2Connector, port_util
```
lguohan, line 34: we should not use swsssdk, it is deprecating.

anilkpandey:
retest this please

PR #533, opened 2019-05-15
Add SONIC CLI Manual

kannankvs:
**- What I did**

Command Line Interface manual is moved from Azure/SONiC repository to sonic-utilities repository so that all developers can update this manual whenever they do any changes in the code related to CLI.

**- How I did it**

**- How to verify it**

-->

### DIFF
--- a/doc/Command-Reference.md
+++ b/doc/Command-Reference.md
@@ -1,0 +1,3680 @@
+# SONiC COMMAND LINE INTERFACE GUIDE
+
+Table of Contents
+=================
+
+   * [Document History](#document-history)
+   * [Introduction](#introduction)
+   * [Basic Configuration And Show](#basic-configuration-and-show)
+      * [SSH Login](#ssh-login)
+      * [Configuring Management Interface](#configuring-management-interface)
+      * [Config Help](#config-help)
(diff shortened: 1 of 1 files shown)


doc/Command-Reference.md
```diff
@@ -0,0 +1,3680 @@
...
+   * [Watermark Configuration And Show](#watermark-configuration-and-show)
+      * [Watermark Show command](#watermark-show-command)
+      * [Watermark Config command](#watermark-config-command)
+   * [Software Installation Commands](#software-installation-commands)
+      * [SONiC Installer](#sonic-installer)
+   * [Troubleshooting Commands](#troubleshooting-commands)
+   * [Routing Stack Configuration And Show](#routing-stack-configuration-and-show)
+   
+
+# Document History
```
jleveque: Curious if we really need a "Document History" section, since this file will now be committed in the repo, and one will be able to check the history right here in GitHub. Also, the section could grow to be quite long. However, I do see how it could be useful for seeing what has changed at a quick glance.

Please share your thoughts here.

kannankvs: It's just an easy way to track what is going on inside the document instead of going through the github history.

PR #534, opened 2019-05-15
[sonic_installer] If asked to install an image which is already installed, simply set as default

jleveque:
- `sonic_installer install` command now retrieves the version of the binary it is told to install and checks to see if it is in the list of currently installed images. If it is, the command outputs a message stating so and simply attempts to set that image as the default image.

- Also added a check in `sonic_installer install` to compare the image type (ONIE, Aboot, etc.) of the binary image it is told to install against the type of the currently running image. If the types do not match, the command will output a message informing the user of the mismatch and exit in error. However, I also added a `-f|--force` option in order to allow the user to forcefully install an image of a different type, just in case the need were to ever arise.

Related enhancements and refactoring:
- Rename `get_image_type()` to `get_running_image_type()`
- Add `get_binary_image_type(binary_image_path)` helper function
- Add `get_binary_image_version(binary_image_path)` helper function
- Add set_default_image(image) helper function
- Update commands for retrieving SONiC version from Aboot image files as the format has changed

### DIFF
--- a/sonic_installer/main.py
+++ b/sonic_installer/main.py
@@ -79,8 +79,8 @@ def get_binary_image_version(binary_image_path):
         return None
     elif binary_type == IMAGE_TYPE_ABOOT:
         p1 = subprocess.Popen(["unzip", "-p", binary_image_path, "boot0"], stdout=subprocess.PIPE, preexec_fn=default_sigpipe)
-        p2 = subprocess.Popen(["grep", "-m 1", "^image_path"], stdin=p1.stdout, stdout=subprocess.PIPE, preexec_fn=default_sigpipe)
-        p3 = subprocess.Popen(["sed", "-n", r"s/^image_path=\"\$target_path\/image-\(.*\)\"$/\1/p"], stdin=p2.stdout, stdout=subprocess.PIPE, preexec_fn=default_sigpipe)
+        p2 = subprocess.Popen(["grep", "-m 1", "^image_name"], stdin=p1.stdout, stdout=subprocess.PIPE, preexec_fn=default_sigpipe)
+        p3 = subprocess.Popen(["sed", "-n", r"s/^image_name=\"\image-\(.*\)\"$/\1/p"], stdin=p2.stdout, stdout=subprocess.PIPE, preexec_fn=default_sigpipe)
     else:
         p1 = subprocess.Popen(["cat", "-v", binary_image_path], stdout=subprocess.PIPE, preexec_fn=default_sigpipe)
         p2 = subprocess.Popen(["grep", "-m 1", "^image_version"], stdin=p1.stdout, stdout=subprocess.PIPE, preexec_fn=default_sigpipe)
--- a/sonic_installer/main.py
+++ b/sonic_installer/main.py
@@ -53,11 +53,63 @@ def reporthook(count, block_size, total_size):
                                   (percent, progress_size / (1024 * 1024), speed, time_left))
     sys.stdout.flush()
 
-def get_image_type():
+def get_running_image_type():
+    """ Attempt to determine whether we are running an ONIE or Aboot image """
     cmdline = open('/proc/cmdline', 'r')
     if "Aboot=" in cmdline.read():
         return IMAGE_TYPE_ABOOT
     return IMAGE_TYPE_ONIE
+
+# Returns None if image doesn't exist or isn't a regular file
+def get_binary_image_type(binary_image_path):
+    """ Attempt to determine whether this is an ONIE or Aboot image file """
+    if not os.path.isfile(binary_image_path):
+        return None
+
+    with open(binary_image_path) as f:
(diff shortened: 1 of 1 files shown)


sonic_installer/main.py
```diff
@@ -53,12 +53,63 @@ def reporthook(count, block_size, total_size):
...
+    return IMAGE_TYPE_ONIE
+
+# Returns None if image doesn't exist or doesn't appear to be a valid SONiC image file
+def get_binary_image_version(binary_image_path):
+    binary_type = get_binary_image_type(binary_image_path)
+    if not binary_type:
+        return None
+    elif binary_type == IMAGE_TYPE_ABOOT:
+        p1 = subprocess.Popen(["unzip", "-p", binary_image_path, "boot0"], stdout=subprocess.PIPE, preexec_fn=default_sigpipe)
+        p2 = subprocess.Popen(["grep", "-m 1", "^image_path"], stdin=p1.stdout, stdout=subprocess.PIPE, preexec_fn=default_sigpipe)
```
yxieca: s/image_path/image_name/

sonic_installer/main.py
```diff
@@ -53,12 +53,63 @@ def reporthook(count, block_size, total_size):
...
+        p3 = subprocess.Popen(["sed", "-n", r"s/^image_path=\"\$target_path\/image-\(.*\)\"$/\1/p"], stdin=p2.stdout, stdout=subprocess.PIPE, preexec_fn=default_sigpipe)
+    else:
+        p1 = subprocess.Popen(["cat", "-v", binary_image_path], stdout=subprocess.PIPE, preexec_fn=default_sigpipe)
```
yxieca: cat didn't work for me when I tried. you might have to use strings?

sonic_installer/main.py
```diff
@@ -248,24 +301,40 @@ def install(url):
...
         image_path = os.path.join("./", url)
 
-    # Verify that the local file exists and is a regular file
-    # TODO: Verify the file is a *proper SONiC image file*
-    if not os.path.isfile(image_path):
-        click.echo("Image file '{}' does not exist or is not a regular file. Aborting...".format(image_path))
+    running_image_type = get_running_image_type()
+    binary_image_type = get_binary_image_type(image_path)
+    binary_image_version = get_binary_image_version(image_path)
+    if not binary_image_type or not binary_image_version:
```
yxieca: Maybe you could also add a check here to make sure the image type matches?

sonic_installer/main.py
```diff
@@ -248,24 +301,40 @@ def install(url):
...
+        click.echo("Image file does not exist or is not a valid SONiC image file")
         raise click.Abort()
 
-    if get_image_type() == IMAGE_TYPE_ABOOT:
-        run_command("/usr/bin/unzip -od /tmp %s boot0" % image_path)
-        run_command("swipath=%s target_path=/host sonic_upgrade=1 . /tmp/boot0" % image_path)
+    # Is this version already installed?
+    if binary_image_version in get_installed_images():
```
yxieca: If forced, should we remove the existing image and install new one?

sonic_installer/main.py
```diff
@@ -248,24 +301,40 @@ def install(url):
...
+            raise click.Abort()
     else:
-        os.chmod(image_path, stat.S_IXUSR)
-        run_command(image_path)
-        run_command('grub-set-default --boot-directory=' + HOST_PATH + ' 0')
-    run_command("rm -rf /host/old_config")
-    # copy directories and preserve original file structure, attributes and associated metadata
-    run_command("cp -ar /etc/sonic /host/old_config")
+        # Verify that the binary image is of the same type as the running image
+        if (binary_image_type != running_image_type) and not force:
```
yxieca: What is the use case of installing a different type of image?

sonic_installer/main.py
jleveque: I think we can leave that feature for a future PR after discussion.

sonic_installer/main.py
jleveque: The only scenario I can think of is if someone accidentally installed an incorrect image type (which, before this check, was quite easy to do if someone accidentally installed an Aboot image on an ONIE device). The `force` option would allow an escape route, just in case. It should practically never be used.

sonic_installer/main.py
jleveque: Fixed in https://github.com/Azure/sonic-utilities/pull/534/commits/87e733861ed884f420506db258098dd791020d5c

sonic_installer/main.py
jleveque: You forgot to pass the `-v` flag to `cat`. The full command should read as `cat -v sonic-broadcom.bin | grep -m 1 "^image_version" | sed -n "s/^image_version=\"\(.*\)\"$/\1/p"`. This will work for you.

PR #536, opened 2019-05-17
[201811] enable DB migrator code

yxieca:
**- What I did**
Enable the existing DB migration code

**- How to verify it**
warm boot from before PFC_WD table rename to after.

### DIFF
--- a/scripts/db_migrator.py
+++ b/scripts/db_migrator.py
@@ -68,9 +68,9 @@ class DBMigrator():
         #       here we only intended to migrade to DB version 1.0.1.
         #       If new DB version is added in the future, the incremental
         #       upgrade will take care of the subsequent migrations.
-        # self.migrate_pfc_wd_table()
-        # self.set_version('version_1_0_1')
-        # return 'version_1_0_1'
+        self.migrate_pfc_wd_table()
+        self.set_version('version_1_0_1')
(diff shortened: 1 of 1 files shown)


scripts/db_migrator.py
```diff
@@ -68,9 +68,9 @@ def version_unknown(self):
         #       here we only intended to migrade to DB version 1.0.1.
         #       If new DB version is added in the future, the incremental
         #       upgrade will take care of the subsequent migrations.
-        # self.migrate_pfc_wd_table()
-        # self.set_version('version_1_0_1')
-        # return 'version_1_0_1'
+        self.migrate_pfc_wd_table()
+        self.set_version('version_1_0_1')
```
prsunny: is the version `1.0.1` or `1_0_1`? As per comments its `1.0.1`

yxieca: 1_0_1 because it is a string used for function name. '.' doesn't work too well there.

yxieca: 1_0_1 is reformatted version number for 1.0.1. Do you want me to change the comment?

yxieca:
retest this please

scripts/db_migrator.py
prsunny: That's fine.. just want to confirm

PR #542, opened 2019-05-23
neighbor_advertiser: use vlan members to set up acl

trzhang-msft:
- update neighbor_advertiser
- use vlan members to set up acl


trzhang-msft:
retest this please

PR #543, opened 2019-05-27
Fixed issue "CRM module throws error".

iris00522:
- What I did
If we input crm thresholds parameters incompletely, it cause crm module throws error while accessing crm config info.
- How I did it
The parameters which aren't set display with default value.(reference sonic-swss/orchagent/crmorch.h) 
- How to verify it
Config partial thresholds parameters.
Execute CRM show commands and check it not cause crm module throws error.

### DIFF
--- a/crm/main.py
+++ b/crm/main.py
@@ -39,6 +39,10 @@ class Crm:
         """
         CRM Handler to display thresholds information.
         """
+        default_threshold_type = 'percentage'
+        default_threshold_low = 70
+        default_threshold_high = 85
+
         configdb = swsssdk.ConfigDBConnector()
         configdb.connect()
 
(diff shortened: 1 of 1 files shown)


prsunny:
I think this check and the values (out-of-range) should be done by the config validator. I would prefer CLI to provide the expected arguments.

iris00522:
If i perform the cmd
"crm config thresholds acl group high 300
"crm config thresholds acl group type percentage"
It lose config the thresholds acl group low
The table just show the value, witch it can get. 
For example.

root@sonic:~# crm show thresholds acl group
Resource Name Threshold Type Low Threshold High Threshold
--------------- ---------------- --------------- ----------------
acl_group percentage x 300
x means could not get CRM configuration.
or don't  show anything.
root@sonic:~# crm show thresholds acl group
Resource Name Threshold Type Low Threshold High Threshold
--------------- ---------------- --------------- ----------------


which one would prefer?

gord1306:
@prsunny , the issue seems to be the crm_info only get value from configDB, and it cannot get the default value from crmorch. If no value in config DB, the content crm_info will be incomplete for the crm show command.

PR #547, opened 2019-06-04
sonic_sku_create.py util for generating a new sku 

sschlafman:
**- What I did**
     Created a utility for generating a new SKU based on a base sku (-b) and a sku definition file (-f)
**- How I did it**
     Copied the base sku folder and modified port_config.ini based on the sku definition 
**- How to verify it**
     Tested on Mellanox platform.

-->

### DIFF
--- a/scripts/psushow
+++ b/scripts/psushow
@@ -11,25 +11,23 @@ from tabulate import tabulate
 VERSION = '1.0'
 
 def get_psu_status_list():
+    """
+    Build list of PSU/PDB status from STATE_DB PSU_INFO table.
+    Keys may be "PSU 1", "PSU 2" (PSU platform) or "PDB 1", "PDB 2" (PDB platform).
+    Returns empty list when no entries exist (no error; HLD: output is simply no rows).
+    """
     psu_status_list = []
-
     db = SonicV2Connector(host="127.0.0.1")
     db.connect(db.STATE_DB)
 
-    # Currently set chassis_num to 1, need to improve it once new platform API is implemented
-    chassis_num = 1
-    chassis_name = "chassis {}".format(chassis_num)
-    num_psus = db.get(db.STATE_DB, 'CHASSIS_INFO|{}'.format(chassis_name), 'psu_num')
-    if not num_psus:
-        return None
+    keys = db.keys(db.STATE_DB, "PSU_INFO|*")
+    if not keys:
+        return psu_status_list
 
-    for psu_idx, psu_key in enumerate(natsorted(db.keys(db.STATE_DB, "PSU_INFO|*")), start = 1):
-        psu_key = psu_key.replace("PSU_INFO|", "")
+    for psu_idx, psu_key in enumerate(natsorted(keys), start=1):
+        psu_name = psu_key.replace("PSU_INFO|", "")
         psu_status = {}
-
         psu_status['index'] = str(psu_idx)
-
-        psu_name = psu_key
         psu_status['name'] = psu_name
 
         presence = db.get(db.STATE_DB, 'PSU_INFO|{}'.format(psu_name), 'presence')
@@ -61,20 +59,21 @@ def get_psu_status_list():
 
 
 def psu_status_show_table(index):
+    """
+    Show PSU/PDB status table. Columns stay the same for consistent experience (HLD).
+    When no PSU/PDB present, output is simply header and no rows (no error message).
+    """
     psu_status_list = get_psu_status_list()
-
-    if not psu_status_list:
-        return None
-
-    header = ['PSU',  'Model', 'Serial', 'HW Rev', 'Voltage (V)', 'Current (A)', 'Power (W)', 'Status', 'LED']
+    header = ['PSU', 'Model', 'Serial', 'HW Rev', 'Voltage (V)', 'Current (A)', 'Power (W)', 'Status', 'LED']
     status_table = []
 
     if index > 0:
+        if not psu_status_list:
+            print("Error: PSU {} is not available. Number of supported PSUs: 0".format(index))
+            return -1
         if index > len(psu_status_list):
             print("Error: PSU {} is not available. Number of supported PSUs: {}".format(index, len(psu_status_list)))
             return -1
-
-        # Trim the list down to contain only the requested PSU
         psu_status_list = [psu_status_list[index-1]]
 
     for psu_status in psu_status_list:
@@ -88,24 +87,23 @@ def psu_status_show_table(index):
                              psu_status['status'],
                              psu_status['led_status']])
 
-    if status_table:
-        print(tabulate(status_table, header, tablefmt="simple", floatfmt='.2f'))
-
+    print(tabulate(status_table, header, tablefmt="simple", floatfmt='.2f'))
     return 0
 
 
 def psu_status_show_json(index):
+    """
+    Show PSU/PDB status in JSON. When no PSU/PDB present, output [] (no error message).
+    """
     psu_status_list = get_psu_status_list()
 
-    if not psu_status_list:
-        return None
-
     if index > 0:
+        if not psu_status_list:
+            print("Error: PSU {} is not available. Number of supported PSUs: 0".format(index))
+            return -1
         if index > len(psu_status_list):
             print("Error: PSU {} is not available. Number of supported PSUs: {}".format(index, len(psu_status_list)))
             return -1
-
-        # Trim the list down to contain only the requested PSU
         psu_status_list = [psu_status_list[index-1]]
 
     print(json.dumps(psu_status_list, indent=4))
@@ -138,10 +136,10 @@ Examples:
         else:
             ret = psu_status_show_table(psu_index)
 
-        if ret != 0:
-            print("PSU not detected")
+        if ret == -1:
             return 1
-
+        # ret == 0: success (including empty list: header + no rows, or JSON [])
+        # No "PSU not detected" when list is empty (HLD: output is simply no rows)
     return 0
 
 
--- a/tests/psushow_test.py
+++ b/tests/psushow_test.py
@@ -57,6 +57,14 @@ class TestPsushow(object):
         psu_status_list = psushow.get_psu_status_list()
         assert psu_status_list == expected_psu_status_list
 
+    def test_get_psu_status_list_no_psu_keys(self):
+        """STATE_DB has no PSU_INFO keys: early return with empty list."""
+        mock_db = mock.MagicMock()
+        mock_db.STATE_DB = 6
+        mock_db.keys.return_value = []
+        with mock.patch.object(psushow, 'SonicV2Connector', return_value=mock_db):
+            assert psushow.get_psu_status_list() == []
+
     def test_status_table(self, capsys):
         expected_output = '''\
 PSU    Model    Serial                        HW Rev      Voltage (V)    Current (A)    Power (W)  Status    LED
@@ -95,10 +103,9 @@ PSU 2  0J6J4K   CN-0J6J4K-17972-5AF-008M-A00  A                 12.18          1
             captured = capsys.readouterr()
             assert captured.out == expected_output
 
-        # Test trying to display a non-existent PSU
+        # Test trying to display a non-existent PSU (no "PSU not detected" per HLD)
         expected_output = '''\
 Error: PSU 3 is not available. Number of supported PSUs: 2
-PSU not detected
 '''
         for arg in ['-s', '--status']:
             with mock.patch('sys.argv', ['psushow', arg, '-i', '3']):
@@ -193,10 +200,9 @@ PSU not detected
             captured = capsys.readouterr()
             assert captured.out == expected_output
 
-        # Test trying to display a non-existent PSU
+        # Test trying to display a non-existent PSU (no "PSU not detected" per HLD)
         expected_output = '''\
 Error: PSU 3 is not available. Number of supported PSUs: 2
-PSU not detected
 '''
         for arg in ['-j', '--json']:
             with mock.patch('sys.argv', ['psushow', '-s', '-i', '3', arg]):
@@ -204,6 +210,49 @@ PSU not detected
             assert ret == 1
             captured = capsys.readouterr()
             assert captured.out == expected_output
+
+    def test_status_table_empty_no_error(self, capsys):
+        """HLD: when no PSU/PDB, output is simply header and no rows (no 'PSU not detected' error)."""
+        with mock.patch.object(psushow, 'get_psu_status_list', return_value=[]):
+            with mock.patch('sys.argv', ['psushow', '-s']):
+                ret = psushow.main()
+        assert ret == 0
+        captured = capsys.readouterr()
+        # Header only, no data rows, no error message
+        assert 'PSU ' in captured.out and 'Model' in captured.out
+        assert 'PSU not detected' not in captured.out
+
+    def test_status_table_empty_with_index_errors(self, capsys):
+        """Explicit index with no PSUs: error path in psu_status_show_table."""
+        expected = (
+            'Error: PSU 1 is not available. Number of supported PSUs: 0\n'
+        )
+        with mock.patch.object(psushow, 'get_psu_status_list', return_value=[]):
+            with mock.patch('sys.argv', ['psushow', '-s', '-i', '1']):
+                ret = psushow.main()
+        assert ret == 1
+        assert capsys.readouterr().out == expected
+
+    def test_status_json_empty_no_error(self, capsys):
+        """When no PSU/PDB, JSON output is [] and no error."""
+        with mock.patch.object(psushow, 'get_psu_status_list', return_value=[]):
+            with mock.patch('sys.argv', ['psushow', '-s', '-j']):
+                ret = psushow.main()
+        assert ret == 0
+        captured = capsys.readouterr()
+        assert captured.out.strip() == '[]'
+        assert 'PSU not detected' not in captured.out
+
+    def test_status_json_empty_with_index_errors(self, capsys):
+        """Explicit index with no PSUs: error path in psu_status_show_json."""
+        expected = (
+            'Error: PSU 1 is not available. Number of supported PSUs: 0\n'
+        )
+        with mock.patch.object(psushow, 'get_psu_status_list', return_value=[]):
+            with mock.patch('sys.argv', ['psushow', '-s', '-j', '-i', '1']):
+                ret = psushow.main()
+        assert ret == 1
+        assert capsys.readouterr().out == expected
 
     def test_version(self, capsys):
         for arg in ['-v', '--version']:


stcheng:
retest this please

qiluo-msft:
@lguohan Could you help port the uart patches to the new kernel?

sonic_sku_create/sonic_sku_create.py
```diff
@@ -0,0 +1,329 @@
+#! /usr/bin/python
+"""
+using soni_sku_create.py  to create a new SONiC SKU based on a <SKU-DEF>.xml file
```
qiluo-msft: >soni_sku_create [](start = 6, length = 15)

typo

sonic_sku_create/sonic_sku_create.py
```diff
@@ -0,0 +1,329 @@
+#! /usr/bin/python
```
qiluo-msft: >python [](start = 12, length = 6)

Explicitly specify python2 or python3.

qiluo-msft: And nowadays it is recommended this ways: [https://stackoverflow.com/a/19305076/2514803](https://stackoverflow.com/a/19305076/2514803)

---
In reply to: [290530856](https://github.com/Azure/sonic-utilities/pull/547#discussion_r290530856) [](ancestors = 290530856)

sonic_sku_create/sonic_sku_create.py
```diff
@@ -0,0 +1,329 @@
...
+			l2_mode = True
+			sku_mode = True
+		else :
+			l2_mode = False
+			sku_mode = True		
+		
+		if args.file:
+			SKU_DEF_FILE = args.file
+		else:
+			SKU_DEF_FILE = os.getcwd() + '/sku_def.xml'
```
qiluo-msft: >os.getcwd() + '/sku_def.xml [](start = 18, length = 27)

The default path is based on current directory, which may be some random path user is reside in. Let's keep teh argument as required.

sonic_sku_create/sonic_sku_create.py
```diff
@@ -0,0 +1,329 @@
...
+		"""
+		Redis database platform info
+		"""
+		self.metadata = self.db.get_all(self.db.CONFIG_DB, "DEVICE_METADATA|localhost" )
+
+	def sku_def_parser(self,sku_def,sku_base) :
+		try:
+			f = open(sku_def,"r")
+		except IOError:
+			print ("Couln't open file: " + sku_def)
```
qiluo-msft: >print ("Couln't open file: " + sku_def) [](start = 3, length = 39)

print and then go on, then you will hit another exception when you access f.

sonic_sku_create/sonic_sku_create.py
```diff
@@ -0,0 +1,329 @@
...
+		self.platform_specific_dict = {"x86_64-mlnx_msn2700-r0":self.msn2700_specific}
+		self.default_lanes_per_port = []
+		self.platform = 'NA'
+		self.fpp = []
+		self.fpp_split  =  {}
+		self.num_of_fpp = 0
+		self.num_of_lanes_dict = {"1" : 4, "2" : 2, "4" : 1}
+		
+		# Set up db connections
+		self.db = swsssdk.SonicV2Connector(host="127.0.0.1")
```
qiluo-msft: >host="127.0.0.1" [](start = 37, length = 16)

Always prefer unix socket in new code.

sonic_sku_create/sonic_sku_create.py
```diff
@@ -0,0 +1,329 @@
...
+				sku.print_port_config()
+			else:
+				sku.create_sku_dir()
+				sku.create_port_config()
+				print ("Created a new sku ",sku.sku_name)
+			
+		if l2_mode : ##If No SKU name provided by -l2_sku_name then if -file exists, use the sku from the file otherwise use current SKU 
+			if args.l2_sku_name is None :
+				if args.file is None :
+					sku_name = sku.metadata['hwsku']
```
qiluo-msft: >metadata [](start = 20, length = 8)

The script will be useful when we have an old image but we need a new SKU. Any old SKU in the image is not used if we generated a new SKU.

However, L2 mode in this script is not required.

sonic_sku_create/sonic_sku_create.py
```diff
@@ -0,0 +1,329 @@
...
+			port_config.append(line)
+		 
+		port_config.sort(key=lambda x: (int(re.search(('\d+'),x[0]).group(0)))) # sort the list with interface name
+		print(tabulate(port_config, header,tablefmt="plain"))
+		
+		
+	def create_sku_dir(self) :
+		# create a new SKU directory based on the base SKU
+		try:
+			shutil.copytree(self.base_sku_dir, self.new_sku_dir)
```
qiluo-msft: >copytree [](start = 10, length = 8)

If we are overwrite an existing sku dir in the image without `--remove`, we should report the error and exit.

sonic_sku_create/sonic_sku_create.py
```diff
@@ -0,0 +1,329 @@
...
+	"""
+	Process aclstat
+	"""
+
+	def __init__(self):
+		self.metadata = {}
+		self.portconfig_dict = {}
+		self.platform_specific_dict = {"x86_64-mlnx_msn2700-r0":self.msn2700_specific}
+		self.default_lanes_per_port = []
+		self.platform = 'NA'
```
qiluo-msft: >NA [](start = 19, length = 2)

Use None or empty string as initial value, and always check it before using.

sonic_sku_create/sonic_sku_create.py
```diff
@@ -0,0 +1,329 @@
...
+			line_arr = line.split()
+			m = re.match(pattern,line_arr[2])
+			self.default_lanes_per_port.insert(int(m.group(1))-1,line_arr[1])
+		#print(self.default_lanes_per_port)
+		f.close()
+		
+	def set_lanes(self) :
+		#set lanes and index per interfaces based on split
+		lanes_index = PORTCONFIG_HEADER.index('lanes')
+		index_index = PORTCONFIG_HEADER.index('index')
```
qiluo-msft: >index [](start = 41, length = 5)

This lane is optional, such as in /usr/share/sonic/device/x86_64-mlnx_msn2700-r0/Mellanox-SN2700/port_config.ini

sonic_sku_create/sonic_sku_create.py
```diff
@@ -0,0 +1,329 @@
...
+		self.base_sku_name = sku_base
+		self.base_sku_dir = DEFAULT_DEV_PATH + self.platform + '/' + self.base_sku_name + '/'
+		self.base_file_path = self.base_sku_dir + "port_config.ini"
+		self.new_sku_dir = DEFAULT_DEV_PATH+self.platform+"/"+self.sku_name+ '/'
+		for child in root:
+			if child.tag == "Ethernet":
+				for interface in child:
+					for eth_iter in interface.iter():
+						if eth_iter is not None:
+							self.portconfig_dict[eth_iter.get("Index")] = ["Ethernet"+eth_iter.get("Index"),[1,2,3,4], eth_iter.get("Speed"), eth_iter.get("InterfaceName"),  eth_iter.get("Index")]
```
qiluo-msft: >"Index" [](start = 41, length = 7)

I think we could ignore 'Index' in xml, and use InterfaceName as the clue.

sonic_sku_create/sonic_sku_create.py
```diff
@@ -0,0 +1,329 @@
...
+		
+def main():
+	parser = argparse.ArgumentParser(description='Create a new SKU',
+									version='1.0.0',
+									formatter_class=argparse.RawTextHelpFormatter)
+	parser.add_argument('-f', '--file', action='store', help='SKU definition file (default <cwd>/sku_def.xml)', default=None)
+	parser.add_argument('-b', '--base', action='store', help='SKU base definition  ', default=None)
+	parser.add_argument('-r', '--remove', action='store_true', help='Remove SKU folder')
+	parser.add_argument('-c', '--cmd', action='store', choices=['new_sku_only', 'l2_mode_only', 'new_sku_l2'], help='Choose action to preform (Generate a new SKU, Configure L2 mode, Both', default="new_sku_only")
+	parser.add_argument('-l2_sku_name', '--l2_sku_name', action='store', help='SKU name to be used in the L2 configuration mode', default=None)
```
qiluo-msft: >l2_sku_name [](start = 23, length = 11)

In concept, SKU is independent with L2 mode. So there is no l2_sku_name.

We already have feature to config any SKU into L2 mode [https://github.com/Azure/SONiC/wiki/L2-Switch-mode#3-generate-a-configuration-for-l2-switch-mode](https://github.com/Azure/SONiC/wiki/L2-Switch-mode#3-generate-a-configuration-for-l2-switch-mode). If your feature is the same, let's keep original method, and remove feature duplication here.

qiluo-msft (changes requested):
As comments

sonic_sku_create/sonic_sku_create.py
qiluo-msft: As a further design discussion, we could use InterfaceName in minigraph for new SKU to generate the SKU directory. If that is possible, could you help implement a new option in this script?

---
In reply to: [290541078](https://github.com/Azure/sonic-utilities/pull/547#discussion_r290541078) [](ancestors = 290541078)

sonic_sku_create/sonic_sku_create.py
```diff
@@ -0,0 +1,329 @@
+#! /usr/bin/python
+"""
```
qiluo-msft: Move this script into /script directory?
Add it into setup.py?
Add a unit test?

sonic_sku_create/sonic_sku_create.py
```diff
@@ -0,0 +1,329 @@
...
+		subprocess.call("sudo systemctl restart swss ", shell=True)
+
+		print( "Finished L2 mode configuration based on "+l2_sku+" SKU")
+		
+def main():
+	parser = argparse.ArgumentParser(description='Create a new SKU',
+									version='1.0.0',
+									formatter_class=argparse.RawTextHelpFormatter)
+	parser.add_argument('-f', '--file', action='store', help='SKU definition file (default <cwd>/sku_def.xml)', default=None)
+	parser.add_argument('-b', '--base', action='store', help='SKU base definition  ', default=None)
```
qiluo-msft: >None [](start = 91, length = 4)

Actually it's easy to find the base SKU inside a running SONiC, so we could safely have a default value.

```
PLATFORM=`sonic-cfggen -H -v DEVICE_METADATA.localhost.platform`
PRESET=(`head -n 1 /usr/share/sonic/device/$PLATFORM/default_sku`)
echo ${PRESET[0]}
```

sonic_sku_create/sonic_sku_create.py
sschlafman: fixed

sonic_sku_create/sonic_sku_create.py
sschlafman: changed to #! /usr/bin/env python2

sonic_sku_create/sonic_sku_create.py
sschlafman: This is relevant when you have an old SKU and you want to configure it to L2 mode.
If the user will not indicate a specific SKU to configure as L2 mode nor a SKU def file will be provided, the script will use the current SKU to be used in the L2 mode.

sonic_sku_create/sonic_sku_create.py
sschlafman: Fixed

sonic_sku_create/sonic_sku_create.py
sschlafman: Fixed. Verifying SKU definition file is provided when creating a new SKU.
When setting l2_mode_only, it is allowed to omit the file and the current SKU will be used

sonic_sku_create/sonic_sku_create.py
sschlafman: Added a check if the SKU exists, if yes USER should use the -r flag first

sonic_sku_create/sonic_sku_create.py
sschlafman: I think that if you don't have the index column it might affect some of the show commands (transceivers, etc).
Anyway when creating a new SKU it is better to add it.

sonic_sku_create/sonic_sku_create.py
sschlafman: Why does it matter ?
Are you planning to remove the index from the minigraph ?

sonic_sku_create/sonic_sku_create.py
sschlafman: The l2_mode option in the script is just an implementation of the process described (  [https://github.com/Azure/SONiC/wiki/L2-Switch-mode#3-generate-a-configuration-for-l2-switch-mode.]).
Instead of manually doing it step by step, the script is doing everything.
Why do you want to remove it?

mlnx-toolbox/mlnx-toolbox.py
```diff
@@ -0,0 +1,362 @@
+#! /usr/bin/python
+"""
+using soni_sku_create.py  to create a new SONiC SKU based on a <SKU-DEF>.xml file
```
stcheng: typo: sonic

sschlafman: Removed file from Commit not relevant to the SKU create script

sonic_sku_create/sonic_sku_create.py
sschlafman: exiting after printing the error

sonic_sku_create/sonic_sku_create.py
sschlafman: removed the usage of sockets to access DB

sonic_sku_create/sonic_sku_create.py
sschlafman: Changed to using "sonic-cfggen -H -v DEVICE_METADATA.localhost.platform"

sonic_sku_create/sonic_sku_create.py
sschlafman: Moved into scripts directory and added it to setup.py

scripts/sonic_sku_create.py
```diff
@@ -0,0 +1,354 @@
...
+		self.num_of_lanes_dict = {"1" : 4, "2" : 2, "4" : 1}
+		
+		# Set up db connections
+		###self.db = swsssdk.SonicV2Connector(host="127.0.0.1")
+		###self.db.connect(self.db.CONFIG_DB)
+		
+		###self.configdb = swsssdk.ConfigDBConnector()
+		###self.configdb.connect()
+
+		###self.read_metadata()
```
qiluo-msft: Remove unused code.

scripts/sonic_sku_create.py
```diff
@@ -0,0 +1,354 @@
...
+		try:
+			platform = subprocess.check_output("sonic-cfggen -H -v DEVICE_METADATA.localhost.platform",shell=True) #self.metadata['platform']
+			self.platform = platform.rstrip()
+		except KeyError:
```
qiluo-msft: >except KeyError [](start = 2, length = 15)

If this is critical, we should exit.

sonic_sku_create/sonic_sku_create.py
qiluo-msft: You are assume the 'index' field is unique. If yes, can you remove the assumption. An internal 'index' based on row order is good enough, or remove it totally.

---
In reply to: [305300266](https://github.com/Azure/sonic-utilities/pull/547#discussion_r305300266) [](ancestors = 305300266)

sonic_sku_create/sonic_sku_create.py
qiluo-msft: I mean the default value of 'base' is known from the image itself, and we don't need user feed this argument on command line.

---
In reply to: [293090385](https://github.com/Azure/sonic-utilities/pull/547#discussion_r293090385) [](ancestors = 293090385)

scripts/sonic_sku_create.py
```diff
@@ -0,0 +1,358 @@
...
+			print ("Couldn't find platform info in CONFIG_DB DEVICE_METADATA")
```
qiluo-msft: >print [](start = 3, length = 5)

print warning or error messages to stderr, instead of stdout.

scripts/sonic_sku_create.py
sschlafman: FIxed

scripts/sonic_sku_create.py
sschlafman: Fixed

scripts/sonic_sku_create.py
```diff
@@ -0,0 +1,441 @@
+#! /usr/bin/env python2
+"""
+usage: sonic_sku_create.py [-h] [-v] [-f FILE | -m MINIGRAPH] [-b BASE] [-r]
+                           [-c {new_sku_only,l2_mode_only,new_sku_l2}]
+                           [-s SKU] [-p] [-vv]
```
qiluo-msft: >s [](start = 29, length = 1)

Not your fault. Let's make it -k to be consistent with [https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-config-engine/sonic-cfggen](https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-config-engine/sonic-cfggen)

scripts/sonic_sku_create.py
```diff
@@ -0,0 +1,441 @@
...
+		self.num_of_fpp = 0
+		self.num_of_lanes_dict = {"1" : 4, "2" : 2, "4" : 1}
+		
+
+		try:
+			platform = subprocess.check_output("sonic-cfggen -H -v DEVICE_METADATA.localhost.platform",shell=True) #self.metadata['platform']
+			self.platform = platform.rstrip()
+		except KeyError:
+			print ("Couldn't find platform info in CONFIG_DB DEVICE_METADATA", file=sys.stderr)
+			exit()
```
qiluo-msft: >exit [](start = 3, length = 4)

In exception cases, exit with non-zero error codes. Multiple times in this script.

scripts/sonic_sku_create.py
```diff
@@ -0,0 +1,441 @@
...
+
+		print( "Finished L2 mode configuration based on "+l2_sku+" SKU")
+		
+def main():
+	parser = argparse.ArgumentParser(description='Create a new SKU',
+									version='1.0.0',
+									formatter_class=argparse.RawTextHelpFormatter)
+	group = parser.add_mutually_exclusive_group()
+	group.add_argument('-f', '--file', action='store', help='SKU definition from xml file. -f OR -m must be provided when creating a new SKU', default=None)
+	group.add_argument('-m', '--minigraph_file', action='store', help='SKU definition from minigraph file. -f OR -m must be provided when creating a new SKU', default=None)
```
qiluo-msft: >default=None [](start = 156, length = 12)

`default='/etc/sonic/minigraph.xml'`
So we can directly use `-m`

scripts/sonic_sku_create.py
```diff
@@ -0,0 +1,441 @@
...
+	parser.add_argument('-b', '--base', action='store', help='SKU base definition  ', default=None)
+	parser.add_argument('-r', '--remove', action='store_true', help='Remove SKU folder')
+	parser.add_argument('-c', '--cmd', action='store', choices=['new_sku_only', 'l2_mode_only', 'new_sku_l2'], help='Choose action to preform (Generate a new SKU, Configure L2 mode, Both', default="new_sku_only")
+	parser.add_argument('-s', '--sku', action='store', help='SKU name to be used when creating a new SKU or for  L2 configuration mode', default=None)
```
qiluo-msft: >  [](start = 109, length = 1)

extra blank

scripts/sonic_sku_create.py
```diff
@@ -0,0 +1,438 @@
...
+		match = None
+		for device_info in meta.findall(str(QName(minigraph_ns, "DeviceInfo"))):
+			dev_sku = device_info.find(str(QName(minigraph_ns, "HwSku"))).text
+			if dev_sku == hwsku :
+				match = True
+				interfaces = device_info.find(str(QName(minigraph_ns, "EthernetInterfaces"))).findall(str(QName(minigraph_ns1, "EthernetInterface")))
+				for interface in interfaces:
+					alias = interface.find(str(QName(minigraph_ns, "InterfaceName"))).text
+					speed = interface.find(str(QName(minigraph_ns, "Speed"))).text
+					index  = interface.find(str(QName(minigraph_ns, "Index"))).text
```
qiluo-msft: >"Index" [](start = 53, length = 7)

The minigraph parser is reading "Index" column, which is not needed by your algorithm. Please remove.

sschlafman: Removed the usage of parsing the "Index" column form xml and minigraph files

scripts/sonic_sku_create.py
sschlafman: fixed

scripts/sonic_sku_create.py
sschlafman: added a default

liat-grozovik:
retest this please

qiluo-msft:
Please fix remaining LGTM error, and add unit test

PR #560, opened 2019-06-24
[fast reboot] stop removing opennsl module before reboot

yxieca:
**- What I did**

We have been running warm reboot test without removing opennsl for
10k+ iterations and not having issue.

The kexec reboot part is the same for warm reboot and fast reboot. It
should be safe for us to skip this during fast reboot to save some
time too.


yxieca:
retest this please

yxieca:
retest this please

PR #571, opened 2019-07-04
[show] Don't abort in InterfaceAliasConverter ctor if PORT table doesn't exist

jleveque:
When the show command is run, don't abort if InterfaceAliasConverter fails to create port_dict. Instead, print a warning message and continue execution.

Without this change, no show commands will work if there is no PORT table in ConfigDB (e.g., if the device is freshly imaged, but unconfigured. In this situation, one cannot get any system info, as all show commands abort (e.g., `show version`, `show platform summary`, etc.). These commands should work whether or not the device is configured.

### DIFF
--- a/show/main.py
+++ b/show/main.py
@@ -58,8 +58,8 @@ class InterfaceAliasConverter(object):
         self.port_dict = config_db.get_table('PORT')
 
         if not self.port_dict:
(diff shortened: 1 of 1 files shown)


lguohan:
@jleveque , can you identify the branches that needed to be backported?

PR #576, opened 2019-07-06
Fix reboot command for booting to other OS

byu343:
**- What I did**
The reboot command may not be able boot to some OS other than SONiC for the exception in reboot_pre_check(), as a result of "sonic_installer list" throwing exception when the image path of the next OS does not satisfy the pattern of SONiC. 

Add an option '-f' to the reboot command to override the exception in reboot_pre_check(), which can  be used for rebooting regardless of the availability of the next image. Besides, fix the line of 'Next: ' in the output of "sonic_install list", for the case that SONiC is not the next image under the Aboot condition.

**- How I did it**

**- How to verify it**

-->


lguohan:
retest this please

PR #577, opened 2019-07-09
Fix Unable to find the commands supported under "config platform"

geans-pin:
**- What I did**
Config platform is used for mlnx which should not parsed for other platform vendor. So, add the check to hide this for mlnx only.

**- How I did it**
Move the def platform(): define in the checker

**- How to verify it**
sudo config platform

admin@sonic:~$ sudo config platform ?
Usage: config platform [OPTIONS] COMMAND [ARGS]...

Platform-related configuration tasks

Options:
--help Show this message and exit.

admin@sonic:~$ sudo config platform
Usage: config [OPTIONS] COMMAND [ARGS]...

Error: No such command "platform".
admin@sonic:~$


keboliu:
@geans-pin is "mlnx" commands still exposed on other non-mellanox platforms? could you please share the screenshot so I can investigate?

actually "config platform" is not only for Mellanox, the original intention is that other vendors can also add their platform specific commands under "config platform" as Mellanox have done.

geans-pin:
> @geans-pin actually "config platform" is not only for Mellanox, the original intention is that other vendors can also add their platform specific commands under "config platform" as Mellanox have done.

Yes, the problem is the @config.group should not defined in the original place and should move to the platform specific place as the current fix to prevent from other vendor without defining platform specific commands parsing the command. Any other idea ?

@config.group()
    def platform():

jleveque:
@keboliu: Do you see any issues with this approach? I agree with @geans-pin that we should only define a 'platform' group if it will be populated on the running platform.

keboliu:
> @keboliu: Do you see any issues with this approach? I agree with @geans-pin that we should only define a 'platform' group if it will be populated on the running platform.

I tend to agree with the change, but please add some comments to describe this is not Mellanox specific and other vendors can do the same if they need similar CLIs.

geans-pin:
Can someone help to guide how to resolve the following merge issue ? 
default — Build finished. No test results found.

Geans

geans-pin:
PR closed for re-submit

PR #579, opened 2019-07-15
[ecnconfig] Validate input WRED parameters

baiwei0427:
**- What I did**
Validate input WRED parameters, including

1) if the threshold is a non-negative integer
2) if the probability is a non-negative integer between 0 and 100
3) if min threshold is no larger than max threshold

**- How I did it**
Before setting new parameters in config DB, I validate all configuration parameters (including the old ones) using function validate_profile_data

**- How to verify it**
Platform: SONiC on MLNX SN-2700
Input command: "ecnconfig -p AZURE_LOSSLESS -gmax 122 -gmin 123"

No output. But orch agent will crash due to wrong WRED parameters (gmin > gmax). You can use "ps aux|grep orch" to check the state of orch agent.

Invalid gmin (123) and gmax (122). gmin should be no smaller than gmax
Exception caught:  Input arguments error. Invalid WRED profile parameters

Signed-off-by: Wei Bai baiwei0427@gmail.com

### DIFF
--- a/scripts/ecnconfig
+++ b/scripts/ecnconfig
@@ -21,9 +21,9 @@ optional arguments:
   -ymax  --yellow-max          set max threshold for packets marked yellow
   -rmin  --red-min             set min threshold for packets marked red
   -rmax  --red-max             set max threshold for packets marked red
-  -gdrop --green-drop-prob     set max probability for packets marked green
-  -ydrop --yellow-drop-prob    set max probability for packets marked yellow
-  -rdrop --red-drop-prob       set max probability for packets marked red
+  -gdrop --green-drop-prob     set max drop/mark probability for packets marked green
+  -ydrop --yellow-drop-prob    set max drop/mark probability for packets marked yellow
+  -rdrop --red-drop-prob       set max drop/mark probability for packets marked red
 
 2) show and change ECN on/off status on queues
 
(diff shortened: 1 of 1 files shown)


baiwei0427:
retest this please

wendani:
@andriymoroz-mlnx

scripts/ecnconfig
```diff
@@ -21,9 +21,9 @@ optional arguments:
   -ymax  --yellow-max          set max threshold for packets marked yellow
   -rmin  --red-min             set min threshold for packets marked red
   -rmax  --red-max             set max threshold for packets marked red
-  -gdrop --green-drop-prob     set drop probability for packets marked green
-  -ydrop --yellow-drop-prob    set drop probability for packets marked yellow
-  -rdrop --red-drop-prob       set drop probability for packets marked red
+  -gdrop --green-drop-prob     set max probability for packets marked green
```
andriymoroz-mlnx: confusing, if parameter is "drop probability", why changing comment to "max probability"?

andriymoroz-mlnx: maybe "max drop probability"?

baiwei0427: @andriymoroz-mlnx  According to RED algorithm (e.g., Figure 5 in https://conferences.sigcomm.org/sigcomm/2015/pdf/papers/p523.pdf), this parameter determines the maximum probability when the queue size is between Kmin (gmin) and Kmax (gmax). RED can be used to mark packets or drop packets.

baiwei0427:
retest this please

baiwei0427:
retest this please

scripts/ecnconfig
baiwei0427: @andriymoroz-mlnx  maybe "max drop/marking probability ..."

andriymoroz-mlnx: right it can also mark...
ok, max drop/mark probability sounds good for me

baiwei0427: @andriymoroz-mlnx  I think gdrop is a little bit misleading. At the first glance, I think it indicates if dropping green packets (instead of ECN marking). g-max-prob is a much better name. But if we change gdrop in ecnconfig, we may need to change many other places. So let's keep it here.

andriymoroz-mlnx: not sure it is used anywhere, but agree

wendani:
🚢

PR #582, opened 2019-07-16
[config] Do not stop or restart dependent services when reloading config

jleveque:
**- What I did**

Remove explicit calls to stop/restart services which are now dependent on SwSS.

As of https://github.com/Azure/sonic-buildimage/pull/2546, teamd, snmp and dhcp_relay services will be stopped/started/restarted along with swss when it is stopped/started/restarted, respectively.

This prevents issues due to double restarts (e.g., this script would call `systemctl restart swss`, which would now cause teamd to restart, then shortly thereafter it would call `systemctl restart teamd`, which could lead to instability).

Now, stopping/starting/restarting swss will also stop/start/restart teamd, snmp, dhcp_relay and radv services, so we should no longer be stopping/starting/restarting them along with swss.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -288,13 +288,10 @@ def _abort_if_false(ctx, param, value):
 
 def _stop_services():
     services = [
-        'dhcp_relay',
         'swss',
-        'snmp',
         'lldp',
         'pmon',
         'bgp',
-        'teamd',
         'hostcfgd',
     ]
     for service in services:
@@ -312,11 +309,8 @@ def _restart_services():
         'rsyslog-config',
         'swss',
         'bgp',
-        'teamd',
         'pmon',
         'lldp',
-        'snmp',
-        'dhcp_relay',
         'hostcfgd',
     ]
     for service in services:


config/main.py
```diff
@@ -288,13 +288,10 @@ def _abort_if_false(ctx, param, value):
 
 def _stop_services():
     services = [
-        'dhcp_relay',
         'swss',
-        'snmp',
         'lldp',
         'pmon',
         'bgp',
```
nazariig: @jleveque can we also add swss.service dependency for bgp and lldp and eventually have similar behaviour as for teamd? The idea here is to have network and platform parts separated and fully managed by systemd. What do you think?

jleveque: @nazariig: Good suggestions. I'm all in favor of simplifying the service management. I opened a PR to add the same dependency for BGP here (https://github.com/Azure/sonic-buildimage/pull/3078), but as you can see from the review comments, it appears there are corner cases, so we will need to think about a more intricate approach.

Regarding LLDP, it's not as imperative that we restart it along with SwSS. Now that we have lldpmgrd, anytime SwSS restarts it should get notified of port state changes and configure LLDP appropriately. It's not as critical as dhcp_relay or radv which have no way of being updated to port changes, but the dependency wouldn't hurt and would help simplify the service management, so I'd like to get @lguohan's opinion on this.

jleveque:
Retest this please

config/main.py
stepanblyschak: @jleveque If swss service restarts in warm mode does it restart teamd?

jleveque: As long as there is a call to `systemctl [re]start swss`, teamd and the other services which are configured with the same dependency on swss will also be [re]started.

stepanblyschak: @jleveque I believe this breaks swss service level warm restart because teamd will recreate portchannels. Is it going to be fixed?

jleveque: @yxieca / @pavel-shirshov to comment on teamd warm restart.

yxieca: @jleveque teamd is not a dependent of swss from what I see in current code. I think you need to put teamd back.

jleveque: Discussed offline. teamd is still dependent on swss. We will need a more elegant dependency solution in sonic-buildimage to handle the warm-restart case. But this PR should still be valid.

jleveque:
Retest this please

nazariig:
@lguohan please have a look

jleveque:
Retest this please

jleveque:
Retest this please

jleveque:
@yxieca: Now that https://github.com/Azure/sonic-buildimage/pull/2852 has been merged into the 201811 branch, this PR should also be cherry-picked into the 201811 branch.

PR #593, opened 2019-07-31
[config]Fix the command  "config interface pfc" 

tengfei-astfs:
[config]Fix the command  "config interface pfc" 

Signed-off-by: tengfei <tengfei@asterfusion.com>

**- What I did**
Fix the command "config interface pfc asymmetric [OPTIONS] STATUS <interface_name>" returning error
**- How I did it**
Modify the way about obtain interface from  pfc asymmetric function 
**- How to verify it**
config interface pfc asymmetric on Ethernet0
then
show interfaces status


tengfei-astfs:
@jleveque Hi, it's a long time since you dealed with my poll request. Is there any progress ?

PR #599, opened 2019-08-07
ZTP CLI commands

rajendra-dendukuri:
Implemented following commands to use Zero Touch Provisioning

show ztp status
config ztp enable
config ztp disable
config ztp run

Signed-off-by: Rajendra Dendukuri <rajendra.dendukuri@broadcom.com>

**- What I did**
Added following CLI commands for Zero Touch Provisioning feature. 

show ztp status
config ztp enable
config ztp disable
config ztp run

If sonic-ztp package is not included in the image, ztp cli commands display following message.
"ZTP feature unavailable in this image version"

**- How I did it**
Modified config/main.py and show/main.py

**- How to verify it**
Execute individual commands and check ZTP flow

-->


rajendra-dendukuri:
please retest this

rajendra-dendukuri:
please retest this

rajendra-dendukuri:
retest

rajendra-dendukuri:
retest please

lguohan:
retest this please

lguohan:
retest this please

lguohan:
retest this please

show/main.py
```diff
@@ -2390,9 +2390,28 @@ def tablelize(keys, data, enable_table_keys, prefix):
...
+@cli.command()
+@click.argument('status', required=False, type=click.Choice(["status"]))
+@click.option('--verbose', is_flag=True, help="Enable verbose output")
+def ztp(status, verbose):
+    """Show Zero Touch Provisioning status"""
+    if os.path.isfile('/usr/bin/ztp') is False:
+        exit("ZTP feature unavailable in this image version")
+
+    if os.geteuid() != 0:
+        exit("Root privileges are required for this operation")
```
jleveque: We prefer all `show ...` commands to be executable by non-root users, since the commands should be read-only, and not modify any configuration, unless there is a good reason to prevent it. Do you have a reason for this?

rajendra-dendukuri: While it may be preferred to have non-root users access to show available for a command, there is a security angle with ZTP. If we allow it, the URL information where provisioning scripts are present will be accessible to non-root users. These scripts may have information about username/passwords for sudo capable users, and some other similar sensitive information. It may lead to privilege escalation. For the same reason, all the files downloaded and generated by ZTP are created in such a way that they cannot be read by non-root user.  So I do not recommend this change.

jleveque: Thanks for the explanation.

@lguohan: Are you OK with this?

lguohan:
retest this please

show/main.py
lguohan: in general, ztp status should contain generic status that does not have sensitive information. like what is the ztp phase. those should not need root previledge.

jleveque:
@rajendra-dendukuri: Please also fix conflicts.

show/main.py
rajendra-dendukuri: I agree we can show some minimal state machine information. I checked the code and realized that it will require some design changes in the ztp app to store sensitive data and state information separately. Can we take up this non-root access activity as an enhancement as part of a different PR.

ben-gale: Guohan and I discussed 1/29, and I also discussed with Rajendra. The ask is to limit the scope of privilege to *only* the sensitive material (e.g. URLs pointing to files containing user info), such that non-sensitive information (e.g. ZTP admin state) is visible to the non-root user. We understand the ask, and can look into this, but we would like you to consider allowing this to be a later enhancement. That is, let's approve/merge this PR, merge it to the 201911 release branch, and tackle this feedback in parallel. Is that OK?

jleveque:
Retest this please

PR #605, opened 2019-08-14
[neighbor_advertise]: Use the existing MIRROR ACL table

stcheng:
Currently only one IPv4 mirror type ACL table is supported.
Thus, in order to insert an ACL rule, we use the existing table
in the configuration.


yxieca:
retest this please

PR #606, opened 2019-08-14
[config]: Flush the neighbor table when removing the router interface

stcheng:
Manually trigger the flush command to remove the deprecated
neighbors that are associated with the removed router interface.


stcheng:
retest this please

lguohan:
might be related if we remove the ip interface.

PR #607, opened 2019-08-14
[config] Call 'systemctl reset-failed' before 'systemctl restart' when restarting services

jleveque:
This ensures that the process will get restarted, even if systemd has previously placed the service in the "failed" state, as `systemctl reset-failed <service>` will take the service out of the "failed" state.

Also, no longer print the command names as they are run; instead print a message stating that we are restarting each service.

The commands which currently call this function are `config load_minigraph` and `config reload`

This should address https://github.com/Azure/sonic-buildimage/issues/3244

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -296,7 +296,8 @@ def _stop_services():
     ]
     for service in services:
         try:
-            run_command("systemctl stop %s" % service, display_cmd=True)
+            click.echo("Stopping {} ...".format(service))
+            run_command("systemctl stop {}".format(service))
         except SystemExit as e:
             log_error("Stopping {} failed with error {}".format(service, e))
             raise
@@ -315,7 +316,11 @@ def _restart_services():
     ]
     for service in services:
         try:
-            run_command("systemctl restart %s" % service, display_cmd=True)
+            # We first run "systemctl reset-failed" to ensure that we
+            # can restart the service, even if it has entered a failed state
+            click.echo("Restarting {} ...".format(service))
+            run_command("systemctl reset-failed {}".format(service))
+            run_command("systemctl restart {}".format(service))
         except SystemExit as e:
             log_error("Restart {} failed with error {}".format(service, e))
             raise
--- a/config/main.py
+++ b/config/main.py
@@ -287,23 +287,57 @@ def _abort_if_false(ctx, param, value):
         ctx.abort()
 
 def _stop_services():
-    services = [
+    services_to_stop = [
         'swss',
         'lldp',
         'pmon',
         'bgp',
         'hostcfgd',
     ]
-    for service in services:
+
+    for service in services_to_stop:
         try:
-            click.echo("Stopping {} ...".format(service))
+            click.echo("Stopping service {} ...".format(service))
             run_command("systemctl stop {}".format(service))
+
         except SystemExit as e:
             log_error("Stopping {} failed with error {}".format(service, e))
             raise
 
+def _reset_failed_services():
+    services_to_reset = [
+        'bgp',
+        'dhcp_relay',
+        'hostcfgd',
+        'hostname-config',
+        'interfaces-config',
+        'lldp',
+        'ntp-config',
+        'pmon',
+        'radv',
+        'rsyslog-config',
+        'snmp',
+        'swss',
+        'syncd',
+        'teamd'
+    ]
+
+    command = "systemctl --failed | grep failed | awk '{ print $2 }' | awk -F'.' '{ print $1 }'"
+    proc = subprocess.Popen(command, shell=True, stdout=subprocess.PIPE)
+    (out, err) = proc.communicate()
+    failed_services = out.rstrip('\n').split('\n')
+
+    for service in failed_services:
+        if service in services_to_reset:
+            try:
+                click.echo("Resetting failed service {} ...".format(service))
+                run_command("systemctl reset-failed {}".format(service))
+            except SystemExit as e:
+                log_error("Failed to reset service {}".format(service))
+                raise
+
 def _restart_services():
-    services = [
+    services_to_restart = [
         'hostname-config',
         'interfaces-config',
         'ntp-config',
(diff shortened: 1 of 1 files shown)


config/main.py
```diff
@@ -315,7 +316,11 @@ def _restart_services():
     ]
     for service in services:
         try:
-            run_command("systemctl restart %s" % service, display_cmd=True)
+            # We first run "systemctl reset-failed" to ensure that we
+            # can restart the service, even if it has entered a failed state
+            click.echo("Restarting {} ...".format(service))
+            run_command("systemctl reset-failed {}".format(service))
```
nikos-github: I assume the error code returned by reset-failed if the services is not in failed state, doesn't trigger the exception. Looks good to me otherwise.

jleveque: Error code shouldn't trigger the exception. Plus, calling `systemctl reset-failed` on a service not in a failed state will still return 0.

jleveque:
Retest this please

config/main.py
nikos-github: Great. Thank you for fixing this.

jleveque:
I'm now contemplating changing this as follows:

- Instead of calling `systemctl reset-failed <service>` on each service in the list, just call `systemctl reset-failed` once to clear the failed state of ALL failed services (this will also make sure all dependent services are removed from the failed state.

Thoughts, anyone?

avi-milner:
@jleveque  i have tested this and it seems to fix the problem: peformed multiple config load_minigraph without any isse seen

config/main.py
```diff
@@ -313,9 +314,16 @@ def _restart_services():
         'lldp',
         'hostcfgd',
     ]
+
+    # We first run "systemctl reset-failed" to remove the "failed"
+    # status from all services before we attempt to restart them
+    click.echo("Resetting all failed services ...")
+    run_command("systemctl reset-failed")
```
nikos-github: The concern I have is that this could possibly reset more services than the ones on the list we are meant to handle. Not sure if this is correct or desirable.

avi-milner: so maybe change approach to reset counter per service ?

nikos-github: Yes I believe that approach would be better.

jleveque:
Retest this please

avi-milner:
> Retest this please
hi reviwer  requests you to use your first commit that resets counter per service and not globally,
i didnt see that you reverted 2nd commit

jleveque:
@avi-milner: Are you asking that I revert back to the original solution? What about dependent services? What if they (synd, teamd, etc.) are in the failed state? I don't believe they will get restarted.

nikos-github:
> @avi-milner: Are you asking that I revert back to the original solution? What about dependent services? What if they (synd, teamd, etc.) are in the failed state? I don't believe they will get restarted.

Are syncd and teamd systemd controlled?

jleveque:
@nikos-github: Yes. All Docker containers are controlled by their own service. Multiple services are dependent upon SwSS (syncd, teamd, dhcp_relay, radv, snmp), so when SwSS restarts it will also restart its dependent services.

@avi-milner: What do you see wrong with removing all services from the failed list upon reloading config? Reloading config is pretty much like wiping the slate clean. Why not also reset the failed status of all services at that time?

avi-milner:
i was just referring you to nikos change request

נשלח מ-Workspace ONE Boxer

בתאריך 15 באוג׳ 2019 23:59, Nikos <notifications@github.com> כתב:
External Email
________________________________

@avi-milner<https://github.com/avi-milner>: Are you asking that I revert back to the original solution? What about dependent services? What if they (synd, teamd, etc.) are in the failed state? I don't believe they will get restarted.

Are syncd and teamd systemd controlled?

jleveque:
I see. So you're good with this change?

jleveque:
Retest this please

avi-milner:
yes

נשלח מ-Workspace ONE Boxer

בתאריך 16 באוג׳ 2019 01:08, Joe LeVeque <notifications@github.com> כתב:
External Email
________________________________

I see. So you're good with this change?

nikos-github:
@jleveque As I mentioned earlier, this is not the right approach. Resetting all failed services in the sonic device through the use of 'systemctl reset-failed' is not correct. The services need to be explicitly specified. Not all services are under sonic's control in the switch and therefore no such assumption should be made.

jleveque:
Sorry for the confusion earlier. I seem to have gotten your responses mixed up.

@nikos-github: So you believe resetting all failed services upon reloading the configuration is not an assumption we can make? Like I said earlier, reloading configuration pretty much "wipes the slate clean"; at which point we would want to try to restart *all* services, even those which may have failed (which may have been due to bad configuration). Can you provide an example of a service which you would *not* want to remove from the failed list when reloading configuration?

nikos-github:
@jleveque This really depends on how the user has configured their Linux system. We should try to restart all failed sonic services or at least the ones controlled or are dependent by config reload. Not `all` failed services in the system in general. When sonic configuration is reloaded, is should contain itself to sonic services. `systemctl reset-failled` will reset ALL failed services in the system whether they are sonic or not.

jleveque:
@nikos-github: Please review my latest iteration. I now obtain a list of failed services and only restart each one if it is a service we will be attempting to restart (or a dependency thereof) by explicitly checking against a list.

jleveque:
Retest this please

jleveque:
@avi-milner: Can you test the current iteration of this PR to ensure it still fixes your problem?

jleveque:
Retest this please

jleveque:
Retest this please

jleveque:
Retest this please

avi-milner:
> @avi-milner: Can you test the current iteration of this PR to ensure it still fixes your problem?

it works fine for me

config/main.py
```diff
@@ -454,6 +494,9 @@ def load_minigraph():
     if os.path.isfile(db_migrator) and os.access(db_migrator, os.X_OK):
         run_command(db_migrator + ' -o set_version')
 
+    # We first run "systemctl reset-failed" to remove the "failed"
+    # status from all services before we attempt to restart them
+    _reset_failed_services()
```
avi-milner: this would only remove services that are already in failed status

config/main.py
```diff
@@ -398,6 +435,9 @@ def reload(filename, yes, load_sysinfo):
     if os.path.isfile(db_migrator) and os.access(db_migrator, os.X_OK):
         run_command(db_migrator + ' -o migrate')
 
+    # We first run "systemctl reset-failed" to remove the "failed"
+    # status from all services before we attempt to restart them
+    _reset_failed_services()
```
avi-milner: this would only remove services that are already in failed status

PR #608, opened 2019-08-15
[fast/warm reboot] kill radv docker before stopping BGP

yxieca:
**- What I did**

Kill radv docker before stopping BGP. Otherwise it will announce our
departure and cause hosts to lose default gateway.

**- How to verify it**
warm reboot with IPv6 IO going through.

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -374,6 +374,7 @@ fi
 # Kill radv before stopping BGP service to prevent annoucing our departure.
 debug "Stopping radv ..."
 docker kill radv &>/dev/null || [ $? == 1 ]
+systemctl stop radv
 
 # Kill bgpd to start the bgp graceful restart procedure
 debug "Stopping bgp ..."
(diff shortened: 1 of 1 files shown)


scripts/fast-reboot
```diff
@@ -371,6 +371,10 @@ if [[ "$REBOOT_TYPE" = "warm-reboot" || "$REBOOT_TYPE" = "fastfast-reboot" ]]; t
     fi
 fi
 
+# Kill radv before stopping BGP service to prevent annoucing our departure.
+debug "Stopping radv ..."
+docker kill radv &>/dev/null || [ $? == 1 ]
```
lguohan: should also do systemctl stop radv

yxieca: I don't think we auto restart radv yet?

yxieca: Done.

yxieca:
retest this please

PR #612, opened 2019-08-21
[neighbor_advertiser]: Add V6 support to mirror ICMPV6 packets

stcheng:
Add two entries to mirror ICMPV6 type 128 and 129 packets.


scripts/neighbor_advertiser
```diff
@@ -314,7 +317,26 @@ def add_mirror_acl_rule():
...
+        'ICMPV6_TYPE': '128',
+        'mirror_action': MIRROR_SESSION_NAME
+    }
+
+    config_db.set_entry('ACL_RULE',
+            (MIRROR_ACL_TABLEV6_NAME, MIRROR_ACL_RULEV6_1_NAME), acl_rule_info)
+
+    acl_rule_info = {
+        'PRIORITY': '8886',
+        'ICMPV6_TYPE': '129',
```
prsunny: I think only `request `is required.  `reply `could be removed.

PR #621, opened 2019-08-26
[fast-reboot] Check if ASIC config has changed before warm reboot

daall:
**- What I did**
I modified the fast-reboot script to check that certain ASIC configurations haven't changed between the source and destination SONiC image before performing the warm reboot.

**- How I did it**
Azure/sonic-buildimage#3384 adds the relevant checksums at build time. I wrote a script to fetch these checksums from the filesystem and check that they match. This script is called as a pre-check in the warm reboot script.

**- How to verify it**
I built two SONiC images with the same checksum and verified that warm reboot proceeded normally. I also built a third image with a different checksum and verified that warm reboot was aborted when attempting to boot from one of the other two images.

N/A

N/A

-->
DEPENDS Azure/sonic-buildimage#3384

### DIFF
--- a/scripts/asic_config_check
+++ b/scripts/asic_config_check
@@ -19,7 +19,8 @@ function debug()
 
 # Retrieve the source ASIC config checksum from the source image
 # Exits with error SRC_ASIC_CONFIG_NOT_FOUND if the checksum is not found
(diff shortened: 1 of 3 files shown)


yxieca:
retest this please

PR #633, opened 2019-09-06
[FastReboot]: Send SIGINT to all teamd before stop

pavel-shirshov:
**- What I did**
Added sending of SIGINT signal to all active teamd processes before Fast-Reboot

**- How I did it**
using pkill utility

**- How to verify it**
Put the changes from this PR to your DUT.
Then use `fast-reboot` on your dut while capturing packets on one of the portchannel member interfaces of DUT

-->

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -407,12 +407,13 @@ docker kill lldp > /dev/null
 systemctl stop lldp
 
 if [[ "$REBOOT_TYPE" = "fast-reboot" ]]; then
+    # Kill teamd processes inside of teamd container with SIGINT to allow them to send last LACP frames
     # Kill teamd, otherwise it gets down all LAGs
     # We call `docker kill teamd` to ensure the container stops as quickly as possible,
     # then immediately call `systemctl stop teamd` to prevent the service from
     # restarting the container automatically.
     # Note: teamd must be killed before syncd, because it will send the last packet through CPU port
-    # TODO: stop teamd gracefully to allow teamd to send last valid update to be sure we'll have 90 seconds reboot time
+    docker exec -i teamd pkill -INT teamd
     docker kill teamd > /dev/null
     systemctl stop teamd
 fi
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -413,7 +413,7 @@ if [[ "$REBOOT_TYPE" = "fast-reboot" ]]; then
     # then immediately call `systemctl stop teamd` to prevent the service from
     # restarting the container automatically.
     # Note: teamd must be killed before syncd, because it will send the last packet through CPU port
-    docker exec -i teamd pkill -INT teamd
+    docker exec -i teamd pkill -INT teamd || [ $? == 1 ]
     docker kill teamd > /dev/null
     systemctl stop teamd
 fi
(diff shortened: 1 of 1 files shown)


scripts/fast-reboot
```diff
@@ -407,12 +407,13 @@ docker kill lldp > /dev/null
...
 
 if [[ "$REBOOT_TYPE" = "fast-reboot" ]]; then
+    # Kill teamd processes inside of teamd container with SIGINT to allow them to send last LACP frames
     # Kill teamd, otherwise it gets down all LAGs
     # We call `docker kill teamd` to ensure the container stops as quickly as possible,
     # then immediately call `systemctl stop teamd` to prevent the service from
     # restarting the container automatically.
     # Note: teamd must be killed before syncd, because it will send the last packet through CPU port
-    # TODO: stop teamd gracefully to allow teamd to send last valid update to be sure we'll have 90 seconds reboot time
+    docker exec -i teamd pkill -INT teamd
```
yxieca: Might be safer to do something like https://github.com/Azure/sonic-utilities/blob/15e24b0e9b7e9ef30b667219cd40dfb8c0dd60c8/scripts/fast-reboot#L399 in case teamd has quit already.

scripts/fast-reboot
```diff
@@ -407,12 +407,13 @@ docker kill lldp > /dev/null
 systemctl stop lldp
 
 if [[ "$REBOOT_TYPE" = "fast-reboot" ]]; then
+    # Kill teamd processes inside of teamd container with SIGINT to allow them to send last LACP frames
     # Kill teamd, otherwise it gets down all LAGs
```
jleveque: It seems like this line can either be removed or worked into the previous line.

scripts/fast-reboot
```diff
@@ -407,12 +407,13 @@ docker kill lldp > /dev/null
...
     # We call `docker kill teamd` to ensure the container stops as quickly as possible,
     # then immediately call `systemctl stop teamd` to prevent the service from
     # restarting the container automatically.
     # Note: teamd must be killed before syncd, because it will send the last packet through CPU port
-    # TODO: stop teamd gracefully to allow teamd to send last valid update to be sure we'll have 90 seconds reboot time
+    docker exec -i teamd pkill -INT teamd
     docker kill teamd > /dev/null
```
jleveque: Is there no need for a brief sleep (or wait for all teamd processes to exit) after sending the SIGINT and before killing the container? Can we guarantee the teamd processes were all able to receive the signal and send the last LACP frames?

pavel-shirshov: I have the same thought. We can't guarantee that. You're right. But every sleep increase pause of control plane. Also I'm not sure which sleep time is good enough. I feel that 1 sec will work good for 99% cases, but I can't prove it.

yxieca:
retest this please

scripts/fast-reboot
yxieca: I think the process should have been killed when kill command returns.

pavel-shirshov: I'll write a loop to check that.

jleveque: @yxieca: `kill` simply sends a signal to a process. It has no way of knowing whether the signal will cause the process to terminate. Thus, kill returns after it sends the signal; it doesn't wait for anything. A loop to check for the existence of the process(es) like Pavel just added is the best method of ensuring we don't kill the container prematurely.

PR #637, opened 2019-09-09
[warm reboot] Skip ASIC config pre-check if current image does not support it

daall:
**What I did**
I edited the ASIC config check script to abort the check if the current image does not have a checksum.

**How to verify it**
1. Run warm-reboot with the updated check script on a SONiC device that does not contain an asic_config_checksum file.
2. Confirm that warm-reboot proceeds past the pre-check stage.
3. Confirm that the syslog contains messages indicating that there was no checksum found on the current image and that the pre-check is not supported.

### DIFF
--- a/scripts/asic_config_check
+++ b/scripts/asic_config_check
@@ -3,7 +3,7 @@
 # Exit codes
 ASIC_CONFIG_UNCHANGED=0
 ASIC_CONFIG_CHANGED=1
-SRC_ASIC_CONFIG_NOT_FOUND=2
+CURR_ASIC_CONFIG_NOT_FOUND=2
 DST_ASIC_CONFIG_NOT_FOUND=3
 
 # Logging utilities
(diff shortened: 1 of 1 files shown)


scripts/asic_config_check
```diff
@@ -17,14 +17,16 @@ function debug()
...
 {
     if [[ ! -f "/etc/sonic/asic_config_checksum" ]]; then
-            error "ASIC config not found in src image, can't verify changes"
-            exit "${SRC_ASIC_CONFIG_NOT_FOUND}"
+            debug "ASIC config not found in curr image, pre-check not supported"
+            return "${CURR_ASIC_CONFIG_NOT_FOUND}"
     fi
+
+    return "${ASIC_CONFIG_UNCHANGED}"
 }
```
yxieca: This will unconditionally pass check if there is a local checksum.

scripts/asic_config_check
```diff
@@ -17,14 +17,16 @@ function debug()
...
-function GetSourceASICConfigChecksum()
+# Retrieve the current ASIC config checksum from the current image
+# Returns CURR_ASIC_CONFIG_NOT_FOUND if the checksum is not found, ASIC_CONFIG_UNCHANGED otherwise
+function GetCurrentASICConfigChecksum()
 {
     if [[ ! -f "/etc/sonic/asic_config_checksum" ]]; then
-            error "ASIC config not found in src image, can't verify changes"
-            exit "${SRC_ASIC_CONFIG_NOT_FOUND}"
+            debug "ASIC config not found in curr image, pre-check not supported"
+            return "${CURR_ASIC_CONFIG_NOT_FOUND}"
```
yxieca: This return code need to change to unchanged or add code in warm-reboot to allow this code.

PR #648, opened 2019-09-17
PVSTP feature implementation

ph408077:
- Added config/show/debug/clear handling for STP
- Modified other utilities script to include STP

**- What I did**

**- How I did it**

**- How to verify it**

-->


ph408077:
Not sure why 164 tests failed. Not able to access the Details link as well. Please let me know if anything needs to be done from my side.

show/stp.py
```diff
@@ -0,0 +1,421 @@
...
+            else:
+                click.echo("{:17}{:13}{}".format(ifname,"No","NA"))
+
+    pass
+
+
+@spanning_tree.command('root_guard')
+@click.pass_context
+def show_stp_root_guard(ctx):
+    """Show spanning_tree root_guard"""
```
mhampasagar: Does this function need to show all STP enabled pots and VLAN with both "consistent" and "inconsistent" port ? just showing inconsistent ports alone is fine right? otherwise it will end up showing all VLAN, port combinations.

Also does this command take VLAN as argument? its good to have VLAN id as argument to display for specific VLAN

sandeep-kulambi: This is the current design, for this release there are no plans to change this.

We will be addressing this in next release as part of IS-CLI changes
1) Display only inconsistent ports
2) Support the option of taking VLAN as argument

mhampasagar: ok

mhampasagar (dismissed):

PR #650, opened 2019-09-17
Revert "[FastReboot]: Send SIGINT to all teamd before stop"

yxieca:
Reverts Azure/sonic-utilities#633

This change has unexpected side effect causing fast-reboot to have longer data plane down time. Reverting it for now.


yxieca:
retest this please

PR #657, opened 2019-09-19
[fdb]cli: fdb entries are cleared according to vlan or port or vlan&&port

yangshp1987:
Fdb entries can be deleted with user specifying vlan or port or vlan&&port.
fdborch is in charge of deleting fdb entries in another issues#1064,and it had pull request.

**- What I did**

**- How I did it**

**- How to verify it**

-->

### DIFF
--- a/scripts/fdbclear
+++ b/scripts/fdbclear
@@ -42,10 +42,15 @@ def main():
 
     try:
         fdb = FdbClear()
-        if args.vlan is not None:
-            print("command not supported yet.")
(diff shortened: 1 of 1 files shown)


prsunny:
Can you provide the output of the command before/after the fix in description?

yangshp1987:
> Can you provide the output of the command before/after the fix in description?

OK，I will provide it later.

yangshp1987:
@prsunny Could you please review，thanks.

prsunny (approved):

PR #659, opened 2019-09-19
SONiC Management Framework Release 1.0

PrabhuSreenivasan:
SONiC Management Framework Release 1.0

-Combined development from Broadcom and Dell
-Adds support for CLI, REST and GNMI interfaces
-YANG based data modeling
-Schema and syntax validation for Redis DB data using CVL
-written in GO programming language

Added helper scripts syspoll  and syseeprom-to-json to poll system data and convert syseeprom data to JSON format.

### DIFF
--- a/scripts/syseeprom-to-json
+++ b/scripts/syseeprom-to-json
@@ -1,0 +1,31 @@
+#!/usr/bin/awk -f
+
+BEGIN { print "{"; n = 0 }
+
+function sep()
+{
+	if (n > 0)  print ", ";
+	++n;
+}
+
+/Product Name/       { sep(); print "\"" $1 " " $2 "\": \"" $5 "\""; } 
+/Part Number/        { sep(); print "\"" $1 " " $2 "\": \"" $5 "\""; } 
+/Serial Number/      { sep(); print "\"" $1 " " $2 "\": \"" $5 "\""; } 
+/Base MAC Address/   { sep(); print "\"" $1 " " $2 " " $3 "\": \"" $6 "\""; }
+/Manufacture Date/   { sep(); print "\"" $1 " " $2 "\": \"" $5 "\""; } 
+/Device Version/     { sep(); print "\"" $1 " " $2 "\": \"" $5 "\""; } 
+/Label Revision/     { sep(); print "\"" $1 " " $2 "\": \"" $5 "\""; } 
+/Platform Name/      { sep(); print "\"" $1 " " $2 "\": \"" $5 "\""; } 
+/ONIE Version/       { sep(); print "\"" $1 " " $2 "\": \"" $5 "\""; } 
+/MAC Addresses/      { sep(); print "\"" $1 " " $2 "\": " $5; }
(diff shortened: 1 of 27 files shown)


scripts/syspoll
```diff
@@ -0,0 +1,48 @@
...
+            w2 = subprocess.check_output(['bash', '-c', 'cat /proc/{}/stat 2>/dev/null'.format(pid)]).split()
+        except subprocess.CalledProcessError:
+            continue
+        procs[pid] = {'cmd': cmd, 'start': btime + int(w2[21]) / ticks_per_sec, 'user': int(w2[13]), 'system': int(w2[14]), 'mem': int(w[4]), 'cpuitil': float(w[2]), 'memutil': float(w[3])}
+    d['procs'] = procs
+
+    f = open(tmpfile, 'w')
+    json.dump(d, f)
+    f.close()
+    os.rename(tmpfile, outfile)
```
renukamanavalan: This would fail if outfile already exist.
Would you like to unlink to ensure.

scripts/syspoll
```diff
@@ -0,0 +1,48 @@
...
+
+    time.sleep(1)
```
renukamanavalan: What is this sleep for?

scripts/syseeprom-to-json
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/awk -f
```
renukamanavalan: Can you please explain, how is this used?

scripts/syspoll
hpersh: No.  On Linux, this is an atomic operation, with the destination file being silently replaced.

scripts/syspoll
hpersh: To update the output file once per second.

scripts/syseeprom-to-json
hpersh: This is a utility to convert the text output of the syseeprom dump utility into JSON.

scripts/syspoll
```diff
@@ -0,0 +1,48 @@
...
+import time
+
+progname = sys.argv[0]
+
+outfile = '/var/platform/system'
+tmpfile = '/var/platform/system.tmp'
+
+ticks_per_sec = os.sysconf(os.sysconf_names['SC_CLK_TCK'])
+
+while True:
```
renukamanavalan: I see it as a pretty handy tool to collect all info in human readable format. 

Can you help us understand the use case of keeping a file in sync with system state, by running every second ?

[Sorry for the delay. The comments from other PR drowned the rest :-) ]

hpersh: This was the initial implementation of a back-end for a CLI command to show system status -- the front-end would simply read a known file, which is constantly updated by this back-end daemon.
I would say that, one day soon, the redis DB should be used, and be updated by the daemon, rather than a file.

renukamanavalan:
Hi Howard, 
Based on Joe's comments, let me revert this. Can you please raise a PR for sonic-buildimage repo?
Thank you,
Renuka

Issue #658, opened 2019-09-19
In alias naming_mode, the interface alias in some show commands is incorrect.


leoli-nps:
**Description**

1. Interface name and alias
```
admin@tau1t8-01:~$ show interfaces status
      Interface            Lanes    Speed    MTU         Alias             Vlan    Oper    Admin            Type    Asym PFC
---------------  ---------------  -------  -----  ------------  ---------------  ------  -------  --------------  ----------
      Ethernet0                8      10G   9100   Ethernet1/1           routed    down     down  SFP/SFP+/SFP28         off
      Ethernet1                9      10G   9100   Ethernet2/1            trunk      up       up  SFP/SFP+/SFP28         off
      Ethernet2               10      10G   9100   Ethernet3/1            trunk      up       up  SFP/SFP+/SFP28         off
      Ethernet3               11      10G   9100   Ethernet4/1            trunk      up       up  SFP/SFP+/SFP28         off
      Ethernet4               12      10G   9100   Ethernet5/1            trunk      up       up  SFP/SFP+/SFP28         off
      Ethernet5               13      10G   9100   Ethernet6/1            trunk      up       up  SFP/SFP+/SFP28         off
      Ethernet6               14      10G   9100   Ethernet7/1            trunk      up       up  SFP/SFP+/SFP28         off
      Ethernet7               15      10G   9100   Ethernet8/1            trunk      up       up  SFP/SFP+/SFP28         off
      Ethernet8               16      10G   9100   Ethernet9/1            trunk      up       up  SFP/SFP+/SFP28         off
      Ethernet9               17      10G   9100  Ethernet10/1            trunk      up       up  SFP/SFP+/SFP28         off
     Ethernet10               18      10G   9100  Ethernet11/1            trunk      up       up  SFP/SFP+/SFP28         off
     Ethernet11               19      10G   9100  Ethernet12/1            trunk      up       up  SFP/SFP+/SFP28         off
     Ethernet12               20      10G   9100  Ethernet13/1            trunk      up       up  SFP/SFP+/SFP28         off
     Ethernet13               21      10G   9100  Ethernet14/1            trunk      up       up  SFP/SFP+/SFP28         off
     Ethernet14               22      10G   9100  Ethernet15/1            trunk      up       up  SFP/SFP+/SFP28         off
     Ethernet15               23      10G   9100  Ethernet16/1            trunk      up       up  SFP/SFP+/SFP28         off
[...]

PR #664, opened 2019-09-21
[show] Properly replace port name with alias in command output

jleveque:
**- What I did**

Properly replace port name with alias in command output. If the device is configured such that any port names happen to be substrings of port aliases (e.g., name  = "Ethernet1", alias = "Ethernet1/1"), alias naming mode would output incorrect aliases (e.g., "Ethernet1/1/1") because it would consider the substring of the alias as a port name.

Previous solution checked for word boundaries (`\b`) on either side of the port name. This did not work properly if there existed a port alias which contained a port name followed by a forwardslash, as the regex considers the forwardslash as a word boundary. Thus, it would get detected as a port name when, in fact, it was actually just a part of an alias name.

This fix will only replace a port name if it is at the start of a line or is preceded by whitespace (`(^|\s)`) **and** it is at the end of a line or is followed by whitespace or a comma followed by whitespace (`($|,{{0,1}}\s)`).

Also added a separate clause to convert the output of `sudo teamshow`, where port names can be followed by '(D)', '(S)', '(D*)' or '(S*)'.

Fixes: https://github.com/Azure/sonic-utilities/issues/658

**- How to verify it**
1. Configure the device such that some or all port names are substrings of some/all port aliases
2. Set interface naming mode to "alias" mode
3. Run all show commands and verify the aliases displayed are correct

### DIFF
--- a/show/main.py
+++ b/show/main.py
@@ -333,20 +333,41 @@ def run_command_in_alias_mode(command):
                 elif output[0].isdigit():
                     output = "    " + output
                 print_output_in_alias_mode(output, index)
+
             elif command.startswith("nbrshow"):
                 """show arp"""
                 index = 2
                 if "Vlan" in output:
                     output = output.replace('Vlan', '  Vlan')
                 print_output_in_alias_mode(output, index)
+
+            elif command.startswith("sudo teamshow"):
+                """
+                sudo teamshow
+                Search for port names either at the start of a line or preceded immediately by
+                whitespace and followed immediately by either the end of a line or whitespace
+                OR followed immediately by '(D)', '(S)', '(D*)' or '(S*)'
+                """
+                converted_output = raw_output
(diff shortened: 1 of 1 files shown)


jleveque:
Retest this please

jleveque:
@leoli-nps: Can you please test out this change and confirm whether it fixes #658?

leoli-nps:
> @leoli-nps: Can you please test out this change and confirm whether it fixes #658?
@jleveque , I have tested it with your changes and found that the alias is not displayed correctly for the interface name followed by a character. See the comments on [#658](https://github.com/Azure/sonic-utilities/issues/658) for details.

jleveque:
Retest this please

jleveque:
Retest this please

jleveque:
Retest this please

jleveque:
Retest this please

PR #668, opened 2019-09-24
[warm/fast reboot] some service docker might have been stopped already

yxieca:
**- What I did**

Some services are depending on swss and might have been stopped at the
time the loop of docker killing was executed.

**- How to verify it**
Without the change, continuous fast-reboot seems to hit the issue fairly frequently where snmp docker has been stopped when reaching the docker kill loop. And that would cause test to fail.

With the fix, the snmp docker could still be killed already, but test would continue.

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -476,10 +476,16 @@ debug "Stopped  syncd ..."
 # We call `docker kill ...` to ensure the container stops as quickly as possible,
 # then immediately call `systemctl stop ...` to prevent the service from
 # restarting the container automatically.
+debug "Stopping all remaining containers ..."
 for CONTAINER_NAME in $(docker ps --format '{{.Names}}'); do
-    docker kill $CONTAINER_NAME > /dev/null || [ $? == 1 ]
+    CONTAINER_STOP_RC=0
+    docker kill $CONTAINER_NAME &> /dev/null || CONTAINER_STOP_RC=$?
     systemctl stop $CONTAINER_NAME
+    if [[ CONTAINER_STOP_RC -ne 0 ]]; then
+        debug "Failed killing container $CONTAINER_NAME RC $CONTAINER_STOP_RC ."
+    fi
 done
+debug "Stopped all remaining containers ..."
 
 # Stop the docker container engine. Otherwise we will have a broken docker storage
 systemctl stop docker.service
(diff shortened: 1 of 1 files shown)


yxieca:
retest this please

scripts/fast-reboot
```diff
@@ -477,7 +477,7 @@ debug "Stopped  syncd ..."
 # then immediately call `systemctl stop ...` to prevent the service from
 # restarting the container automatically.
 for CONTAINER_NAME in $(docker ps --format '{{.Names}}'); do
-    docker kill $CONTAINER_NAME > /dev/null
+    docker kill $CONTAINER_NAME > /dev/null || [ $? == 1 ]
```
jleveque: Per our discussion, it might be best to continue on all non-zero exit codes here, preferably logging them to syslog for posterity.

qiluo-msft: >docker kill $CONTAINER_NAME > /dev/null || [ $? == 1 ] [](start = 4, length = 54)

Why this line's redirect and pipe are different with L467

jleveque: Good observation. I think the ordering in L467 is incorrect.

yxieca: Yeah, we discussed this mistake and never fixed it :-) I'll add the change to review later when tested.

yxieca: Updated the change.

yxieca: Change is done. Thanks for noticing!

scripts/fast-reboot
```diff
@@ -476,10 +478,16 @@ debug "Stopped  syncd ..."
 # We call `docker kill ...` to ensure the container stops as quickly as possible,
 # then immediately call `systemctl stop ...` to prevent the service from
 # restarting the container automatically.
+debug "Stopping all remaining containers ..."
 for CONTAINER_NAME in $(docker ps --format '{{.Names}}'); do
-    docker kill $CONTAINER_NAME > /dev/null
+    CONTAINER_STOP_RC=0
+    docker kill $CONTAINER_NAME &> /dev/null || CONTAINER_STOP_RC=$?
```
jleveque: You could combine L483-484 into one:
```
CONTAINER_STOP_RC=$(docker kill $CONTAINER_NAME &> /dev/null)
```

yxieca: I though this would cause failure quit when docker kill fails?

yxieca: We are running with -e. Unfortunately what you proposed would cause failure abort.

jleveque: Right. I forgot about `-e` :-(

yxieca:
retest this please

PR #669, opened 2019-09-25
[config]: Add hostname command to config CLI

samaity:
Signed-off-by: Sangita Maity <sangitamaity0211@gmail.com>

**- What I did**
added new command **change-hostname** to config CLI so that the user can change the hostname on a SONiC device without traffic being impacted.

**- How I did it**
update **"hostname"** key to the **"DEVICE_METADATA|localhost"** field in config DB using sonic-cfggen and restart the **hostname-config** service.

**- How to verify it**

> 
changed the hostname from **csw06** to **csw06-sonic**

```
admin@csw06:~$ sudo config change-hostname csw06-sonic
Running command: service hostname-config restart
Please note loaded setting will be lost after system reboot. To preserve setting, run config save.


admin@csw06:~$ sudo config save -y


admin@csw06:~$ exit
logout
Connection to 172.x.x.x closed.


➜  ~ ssh admin@172.x.x.x
admin@172.x.x.x's password:
Linux csw06-sonic 4.9.0-9-amd64 #1 SMP Debian 4.9.168-1+deb9u3 (2015-12-19) x86_64
Unauthorized access and/or use are prohibited.
All access and/or use are subject to monitoring.


Help:    http://azure.github.io/SONiC/


Last login: Tue Sep 24 18:11:57 2019 from 172.x.x.x
admin@csw06-sonic:~$    -> **hostname got updated successfully**
```

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -525,16 +525,16 @@ def load_minigraph():
 
 
 #
-# 'change-hostname' command
+# 'hostname' command
 #
-@config.command('change-hostname')
-@click.argument('hostname', metavar='<hostname>', required=True)
-def change_hostname(hostname):
+@config.command('hostname')
+@click.argument('new_hostname', metavar='<new_hostname>', required=True)
+def hostname(new_hostname):
     """Change Hostname on a SONiC device without impacting the traffic."""
 
     config_db = ConfigDBConnector()
     config_db.connect()
-    config_db.mod_entry('DEVICE_METADATA' , 'localhost', {"hostname" : hostname})
+    config_db.mod_entry('DEVICE_METADATA' , 'localhost', {"hostname" : new_hostname})
     try:
         command = "service hostname-config restart"
         run_command(command, display_cmd=True)
--- a/config/main.py
+++ b/config/main.py
@@ -530,7 +530,7 @@ def load_minigraph():
 @config.command('hostname')
 @click.argument('new_hostname', metavar='<new_hostname>', required=True)
 def hostname(new_hostname):
-    """Change Hostname on a SONiC device without impacting the traffic."""
+    """Change device hostname without impacting the traffic."""
 
     config_db = ConfigDBConnector()
     config_db.connect()
--- a/config/main.py
+++ b/config/main.py
@@ -525,6 +525,25 @@ def load_minigraph():
 
 
 #
+# 'change-hostname' command
+#
+@config.command('change-hostname')
+@click.argument('hostname', metavar='<hostname>', required=True)
+def change_hostname(hostname):
+    """Change Hostname on a SONiC device without impacting the traffic."""
+
+    config_db = ConfigDBConnector()
+    config_db.connect()
+    config_db.mod_entry('DEVICE_METADATA' , 'localhost', {"hostname" : hostname})
+    try:
+        command = "service hostname-config restart"
+        run_command(command, display_cmd=True)
+    except SystemExit as e:
+        click.echo("Restarting hostname-config  service failed with error {}".format(e))
+        raise
+    click.echo("Please note loaded setting will be lost after system reboot. To preserve setting, run `config save`.")
+
+#
 # 'portchannel' group ('config portchannel ...')
 #
 @config.group()
--- a/doc/Command-Reference.md
+++ b/doc/Command-Reference.md
@@ -220,8 +220,8 @@ This command lists all the possible configuration commands at the top level.
     aaa                    AAA command line
     acl                    ACL-related configuration tasks
     bgp                    BGP-related configuration tasks
-    change-hostname        Change Hostname on a SONiC device without...
     ecn                    ECN-related configuration tasks
+    hostname               Change Hostname on a SONiC device without...
     interface              Interface-related configuration tasks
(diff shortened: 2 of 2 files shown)


config/main.py
```diff
@@ -524,6 +524,25 @@ def load_minigraph():
...
+@config.command('change-hostname')
+@click.argument('hostname', metavar='<hostname>', required=True)
+def change_hostname(hostname):
+    """Change Hostname on a SONiC device without impacting the traffic."""
+
+    config_db = ConfigDBConnector()
+    config_db.connect()
+    config_db.mod_entry('DEVICE_METADATA' , 'localhost', {"hostname" : hostname})
+    try:
+        command = "service hostname-config restart"
```
prsunny: Do we need to restart `hostname-config` service? Isn't [this ](https://github.com/Azure/sonic-buildimage/blob/master/files/image_config/hostcfgd/hostcfgd#L312)handles the config change?

samaity: Not sure. 🤔 I had to restart to "hostname-config" to make it effective.

jleveque: Hmm... I believe Prince is correct -- hostcfgd should now pick up the change and apply it without a restart. This change went in ~4 months ago. @samaity: is you hostcfgd up-to-date? If so then maybe there's a bug in hostcfgd?

config/main.py
```diff
@@ -524,6 +524,25 @@ def load_minigraph():
     click.echo("Please note setting loaded from minigraph will be lost after system reboot. To preserve setting, run `config save`.")
 
 
+#
+# 'change-hostname' command
+#
+@config.command('change-hostname')
```
jleveque: Should we simply call it "hostname" (`config hostname <new_hostname>`)?

config/main.py
samaity: @jleveque I just checked again. It seems **hostcfgd** updates all the docker containers' hostname without a restart but not the SONiC switch's hostname itself. As per [#2924](https://github.com/Azure/sonic-buildimage/pull/2924) PR, **hostcfgd** keeps containers' hostname up-to-date. 

Am I missing something?

jleveque: @samaity: You're correct. It looks like the author of https://github.com/Azure/sonic-buildimage/pull/2924 was only concerned with updating the hostname inside the containers. Ultimately, this functionality should also be present for the host OS. This is the eventual goal for all SONiC components. However, this would involve eliminating hostname-config.service and moving that functionality into hostcfgd, which would require a good deal of testing. That is outside the scope of this PR. This solution is acceptable for now.

doc/Command-Reference.md
```diff
@@ -220,6 +220,7 @@ This command lists all the possible configuration commands at the top level.
     aaa                    AAA command line
     acl                    ACL-related configuration tasks
     bgp                    BGP-related configuration tasks
+    change-hostname        Change Hostname on a SONiC device without...
```
jleveque: Please also change this to reflect the new subcommand name.

Also please update the PR title to reflect this change.

samaity: _modified as requested._

jleveque:
Retest this please

jleveque:
Retest this please

config/main.py
```diff
@@ -524,6 +524,25 @@ def load_minigraph():
     click.echo("Please note setting loaded from minigraph will be lost after system reboot. To preserve setting, run `config save`.")
 
 
+#
+# 'hostname' command
+#
+@config.command('hostname')
+@click.argument('new_hostname', metavar='<new_hostname>', required=True)
+def hostname(new_hostname):
+    """Change Hostname on a SONiC device without impacting the traffic."""
```
jleveque: @samaity: Since the docstrings get truncated when displayed in the help message, I think you should shorten this message so that more of it is displayed. The phrase "on a SONiC device" is unnecessary, as the CLI is only used on SONiC devices.

`"""Change device hostname without impacting traffic."""`

jleveque:
Retest this please

qiluo-msft:
Retest this please

config/main.py
```diff
@@ -524,6 +524,25 @@ def load_minigraph():
...
+# 'hostname' command
+#
+@config.command('hostname')
+@click.argument('new_hostname', metavar='<new_hostname>', required=True)
+def hostname(new_hostname):
+    """Change device hostname without impacting the traffic."""
+
+    config_db = ConfigDBConnector()
+    config_db.connect()
+    config_db.mod_entry('DEVICE_METADATA' , 'localhost', {"hostname" : new_hostname})
```
prsunny: I've one question... What if 'localhost' is not present in 'DEVICE_METADATA'? Does this create a new one?

qiluo-msft:
Retest this please

config/main.py
samaity: I verified this thing, @prsunny. It will add the "hostname" key to the "DEVICE_METADATA|localhost" field in config DB.

jleveque:
Retest this please

qiluo-msft:
Retest this please

jleveque:
Retest this please

jleveque:
Retest this please

jleveque:
Retest this please

PR #670, opened 2019-09-25
[neighbor_advertiser] Verify that DIPs returned from ferret are not in device VLAN

daall:
**- What I did**
I added a check to neighbor_advertiser to make sure that the DIPs returned from ferret are not in any of the device's VLANs.

**- How I did it**
- I added a helper method to query CONFIG_DB for VLAN interfaces and check that a given IP does not fall in that range.
- I used the helper method to check that DIPs returned from ferret did not fall in any of the device's VLANs.
- I added a retry to attempt to get other DIPs from ferret before moving on to the next ferret VIP.

**- How to verify it**
1. Run the script against a response object that contains an IPv4 address that falls inside one of the DUTs VLANs. The script should retry 3 times and then fail.
2. Run the script against a response object that contains an IPv4 address that does not fall inside of the DUTs VLANs. The script should succeed.

Signed-off-by: Danny Allen (daall@microsoft.com)

### DIFF
--- a/scripts/neighbor_advertiser
+++ b/scripts/neighbor_advertiser
@@ -328,7 +328,6 @@ def post_neighbor_advertiser_slice(ferret_service_vip):
     url = 'http://{}:85{}{}'.format(ferret_service_vip, FERRET_NEIGHBOR_ADVERTISER_API_PREFIX, get_switch_name())
     response = None
 
-    ferret_server_ipv4_addr = None
     for i in range(DEFAULT_FERRET_QUERY_RETRIES):
         try:
             response = requests.post(url, json = request_slice, timeout = DEFAULT_REQUEST_TIMEOUT)
@@ -336,6 +335,7 @@ def post_neighbor_advertiser_slice(ferret_service_vip):
             log_error('The request failed, vip: {}, error: {}'.format(ferret_service_vip, e))
             return None
 
+        # Handle response errors
         if not response:
             log_error('Failed to set up neighbor advertiser slice, vip: {}, no response obtained'.format(ferret_service_vip))
             return None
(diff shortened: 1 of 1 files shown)


daall:
retest this please

scripts/neighbor_advertiser
```diff
@@ -308,24 +331,30 @@ def post_neighbor_advertiser_slice(ferret_service_vip):
...
-    else:
-        log_error('Failed to set up neighbor advertiser slice, vip: {}, no response obtained'.format(ferret_service_vip))
 
-    return ferret_server_ipv4_addr
+        if not is_dip_in_device_vlan(ferret_server_ipv4_addr):
+            log_info('Successfully set up neighbor advertiser slice, vip: {}, dip: {}'.format(ferret_service_vip, ferret_server_ipv4_addr))
+            return ferret_server_ipv4_addr
+
+    log_error('Failed to set up neighbor advertiser slice, vip: {}, returned dips were in device VLAN'.format(ferret_service_vip))
+    return None
```
yxieca: Not a major issue.

But I suggest you to follow the convention above, put failure handling code in if block and the main return becomes successful return.

daall:
retest this please

scripts/neighbor_advertiser
```diff
@@ -308,23 +328,34 @@ def post_neighbor_advertiser_slice(ferret_service_vip):
...
-        ferret_server_ipv4_addr = neighbor_advertiser_configuration['ipv4Addr']
-        log_info('Successfully set up neighbor advertiser slice, vip: {}, dip: {}'.format(ferret_service_vip, ferret_server_ipv4_addr))
-    elif response:
-        log_error('Failed to set up neighbor advertiser slice, vip: {}, error_code: {}, error_content: {}'.format(ferret_service_vip, response.status_code, response.content))
-    else:
-        log_error('Failed to set up neighbor advertiser slice, vip: {}, no response obtained'.format(ferret_service_vip))
+        ferret_dip = neighbor_advertiser_configuration['ipv4Addr']
+
+        if not is_dip_in_device_vlan(ferret_dip):
+            save_as_json(neighbor_advertiser_configuration, NEIGHBOR_ADVERTISER_RESPONSE_CONFIG_PATH)
```
yxieca: I think you want to generate a log here when "is_dip_in_device_vlan(ferret_dip)" is true and the main branch can log success and return directly?

scripts/neighbor_advertiser
```diff
@@ -82,16 +84,34 @@ def log_error(msg):
...
+is_dip_in_device_vlan = None
+
+
+def get_device_vlan_checker():
+    global is_dip_in_device_vlan
+
+    vlan_interfaces = config_db.get_table('VLAN_INTERFACE')
+
+    def check_device(ferret_server_ip_addr):
+        return any(IPAddress(ferret_server_ip_addr) in IPNetwork(vlan_int[1]) for vlan_int in vlan_interfaces.iterkeys())
```
lguohan: >vlan_int[1]) for vlan_int in vlan_interfaces.iterkeys [](start = 65, length = 53)

what if vlan_interface has ipv6 address? we should only filter ipv4 address here.

lguohan: we also have a case where vlan interface only contains interface name, but not IP address. we should check vlan_int[1] is available or not. you can sync with prince on this case.

---
In reply to: [328399686](https://github.com/Azure/sonic-utilities/pull/670#discussion_r328399686) [](ancestors = 328399686)

yxieca: Good catch!

scripts/neighbor_advertiser
```diff
@@ -308,24 +328,36 @@ def post_neighbor_advertiser_slice(ferret_service_vip):
...
+            log_error('Failed to set up neighbor advertiser slice, vip: {}, error_code: {}, error_content: {}'.format(ferret_service_vip, response.status_code, response.content))
+            return None
 
-    if response and response.ok:
         neighbor_advertiser_configuration = json.loads(response.content)
-        save_as_json(neighbor_advertiser_configuration, NEIGHBOR_ADVERTISER_RESPONSE_CONFIG_PATH)
         ferret_server_ipv4_addr = neighbor_advertiser_configuration['ipv4Addr']
+
+        # Retry the request if the provided DIP is in the device VLAN
+        if is_dip_in_device_vlan(ferret_server_ipv4_addr):
```
prsunny: So this is the global variable assigned to a function name. Why don't we define this as a function itself for better clarity?

prsunny: In that case, a class would be ideal. But another (simple) option is to make the Vlan interface global and fetch it in main. What do you think?

scripts/neighbor_advertiser
```diff
@@ -82,16 +84,50 @@ def log_error(msg):
...
+    global vlan_interface_query
+
+    # Lazy load the vlan interfaces the first time we run this check.
+    if not vlan_interface_query:
+        vlan_interface_query = config_db.get_table('VLAN_INTERFACE')
+
+    ferret_dip = IPAddress(ferret_dip)
+
+    for vlan_interface in vlan_interface_query.iterkeys():
+        vlan_subnet = vlan_interface[1]
```
prsunny: As discussed, you can reuse this function for this check - `is_ip_prefix_in_key(self, key)`

yxieca:
retest this please

PR #671, opened 2019-09-25
[Mellanox] Don't start & stop pmon when config reload

stephenxs:
**- What I did**
Don't start/stop pmon during config reload since pmon is triggered by syncd.
This PR depends on [[Mellanox] Stop pmon ahead of syncd #3505](https://github.com/Azure/sonic-buildimage/pull/3505)

**- How I did it**
For Mellanox platform, remove pmon from the services_to_stop and service_to_start list.

**- How to verify it**
execute "config reload" and "systemctl restart swss.service" and check whether all dockers start.

-->

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1387,7 +1387,7 @@ def add_syslog_server(ctx, syslog_ip_address):
     if syslog_ip_address in syslog_servers:
         click.echo("Syslog server {} is already configured".format(syslog_ip_address))
         return
-    else: 
+    else:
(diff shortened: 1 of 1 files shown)


config/main.py
```diff
@@ -287,13 +297,22 @@ def _abort_if_false(ctx, param, value):
         ctx.abort()
 
 def _stop_services():
-    services_to_stop = [
-        'swss',
-        'lldp',
-        'pmon',
```
nazariig: @stephenxs no need to have a separate service list. Just check the pmon presence and remove it only for Mellanox platforms.

PR #681, opened 2019-09-27
Need to store hostname in configDB

fk410167:
**- What I did**
After configure hostname,hostname changed but it is not reflecting in configDB(running config) so config save/reboot still shows old hostname.
So store hostname in ConfigDB.

**- How I did it**
As part of change_hostname API, store the new hostname in ConfigDB.

**- How to verify it**
\# config hostname \<new-name\>
\# show host
\# show run all | grep hostname
\<reboot\>
\# show run all | grep hostname (to verify that the new-name is present as hostname)

-->

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -245,6 +245,11 @@ def _change_hostname(hostname):
         run_command('hostname -F /etc/hostname', display_cmd=True)
         run_command('sed -i "/\s{}$/d" /etc/hosts'.format(current_hostname), display_cmd=True)
         run_command('echo "127.0.0.1 {}" >> /etc/hosts'.format(hostname), display_cmd=True)
+        config_db = ConfigDBConnector()
+        config_db.connect()
+        curr_host_name = config_db.get_entry('DEVICE_METADATA', "localhost").get('hostname')
+        if (curr_host_name != hostname):
+            config_db.mod_entry('DEVICE_METADATA', "localhost", {'hostname': hostname})
 
 def _clear_qos():
     QOS_TABLE_NAMES = [


fk410167:
Retest this please

config/main.py
```diff
@@ -245,6 +245,11 @@ def _change_hostname(hostname):
         run_command('hostname -F /etc/hostname', display_cmd=True)
         run_command('sed -i "/\s{}$/d" /etc/hosts'.format(current_hostname), display_cmd=True)
         run_command('echo "127.0.0.1 {}" >> /etc/hosts'.format(hostname), display_cmd=True)
+        config_db = ConfigDBConnector()
+        config_db.connect()
+        curr_host_name = config_db.get_entry('DEVICE_METADATA', "localhost").get('hostname')
```
prsunny: This hostname is invoked from L492, where it is already saved to config_db. Where is the command to change hostname?

jleveque: Agree with @prsunny. This function is only called from `load_mgmt_config()`, after it retrieves the value from config_db. What is the need for this?

Also, have you seen https://github.com/Azure/sonic-utilities/pull/669, which adds a command for updating the hostname?

jleveque: In fact, the functionality of this `_change_hostname()` function is now redundant with the hostname-config service. This function should be removed, and load_mgmt_config() should instead call `sudo systemctl restart hostname-config.service`

fk410167: @prsunny, @jleveque,
It seems the 'config hostname \<new-name\>' command has been added to our organisation's internal branch but **not** ported to master branch yet. My changes are on top of that. Below is the excerpt,

```
#
# 'hostname' command ('config hostname name')
#
@config.command()
@click.argument('name', required=True, metavar="<name>")
def hostname(name):
  _change_hostname(name, display=False)
```

https://github.com/Azure/sonic-utilities/pull/669 is approved and might get merged, this '_change_hostname()' functionality would become redundant. 

I will close my PR.

fk410167:
Similar to https://github.com/Azure/sonic-utilities/pull/669 ... Closing my PR

PR #688, opened 2019-10-01
Add CLI support for configurable drop counters

daall:
Add CLI support for configurable drop counters
- Adds dropconfig script to configure drop counters
- Adds dropstat script to show and clear drop counters
- Adds handlers for drop counters to show, config, and sonic-clear

**New command output (if the output of a command-line utility has changed)**
- A command reference update has been provided

**Details if related**
Depends on:
- Azure/sonic-sairedis#520
- Azure/sonic-swss#1075


daall:
I got some feedback on the behavior of the scripts that I wanted to note here:
- It would be helpful to print the descriptions for the counters in the header of show drops counts so that users know what they're looking at
- Because drop reasons are very platform specific it would probably be good to include the check for supported counters and drop reasons at this stage and work with the platform vendors to support the query APIs along with the drop counters

scripts/dropconfig
```diff
@@ -0,0 +1,306 @@
...
+        print(tabulate(table, drop_counter_capability_header, tablefmt='simple', stralign='left'))
+
+        for counter, capabilities in device_caps.iteritems():
+            supported_reasons = deserialize_reason_list(capabilities.get('reasons', ''))
+            if supported_reasons:
+                print('\n{}'.format(counter))
+                for reason in supported_reasons:
+                    print('\t{}'.format(reason))
+    
+    def create_counter(self, counter_name, alias, group, counter_type, description, reasons):
```
dev-aviz: How about user wanted to update any reasons to the existing countes? nice to have some method to do that?

scripts/dropconfig
```diff
@@ -0,0 +1,306 @@
...
+                          counter.get('reason', ''),
+                          counter.get('description', '')))
+        print(tabulate(table, drop_counter_config_header, tablefmt='simple', stralign='left'))
+
+    def print_device_capabilities(self):
+        """
+            Prints out the capabilities that this device has
+        """
+
+        device_caps = self.get_device_capabilities()
```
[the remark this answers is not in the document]
daall: I agree! It wasn't clear if we would have the support to do this at this stage but it looks like we will. I need to do a little bit of work in sai redis to expose the capabilties but once that is done and I can test it out I'll update this script to do that check.

scripts/dropconfig
```diff
@@ -0,0 +1,306 @@
...
+        if not counter_name:
+            raise Exception('No counter name provided')
+
+        if not self.counter_name_in_use(counter_name):
+            raise Exception('Counter \'{}\' not found'.format(counter_name))
+
+        self.config_db.set_entry(DEBUG_COUNTER_CONFIG_TABLE, counter_name, None)
+        self.config_db.delete_table(self.config_db.serialize_key((DROP_REASON_TABLE_PREFIX, counter_name)))
+
+    def add_reasons(self, counter_name, reasons):
```
dev-aviz: Do user need to reset the drop counter explicitly when drop_reasons are added/removed?

daall: Good catch! I don't believe the counter gets cleared when drop reasons are added and removed from the counter but I need to verify.

lguohan:
can you follow example here to add unit test for your show command. 

https://github.com/Azure/sonic-utilities/tree/master/sonic-utilities-tests

show/main.py
```diff
@@ -1865,6 +1865,53 @@ def table(table_name, verbose):
...
+@cli.group(cls=AliasedGroup, default_if_no_args=False)
+def drops():
+    """Show drop counter related information"""
+    pass
+
+# 'config' subcommand ("show drops config")
+@drops.command()
+@click.option('-g', '--group', required=False)
+@click.option('--verbose', is_flag=True, help="Enable verbose output")
+def config(group, verbose):
```
jleveque: Suggest to rename command using full word, "configuration", as the Click library will automatically handle tab-completion and abbreviation.

config/main.py
```diff
@@ -1286,6 +1286,60 @@ def incremental(file_name):
...
+    if desc:  command += " -d '{}'".format(desc)
+
+    run_command(command, display_cmd=verbose)
+
+@drops.command()
+@click.argument("counter_name", type=str, required=True)
+@click.option('-v', '--verbose', is_flag=True, help="Enable verbose output")
+def delete(counter_name, verbose):
+    """Delete an existing drop counter"""
+    command = "dropconfig -c uninstall -n {}".format(counter_name)
```
qiluo-msft: >counter_name [](start = 53, length = 12)

Use `[https://docs.python.org/3/library/shlex.html#shlex.quote`](https://docs.python.org/3/library/shlex.html#shlex.quote`) to prevent input injection risk.

config/main.py
```diff
@@ -1286,6 +1286,60 @@ def incremental(file_name):
...
+#
+# 'drops' group ('config drops ...')
+#
+@config.group()
+def drops():
+    """Drop counter related configuration tasks"""
+    pass
+
+#
+# 'init' subcommand ('config drops init')
```
qiluo-msft: >drops [](start = 29, length = 5)

how about `dropcounter`?

scripts/dropconfig
```diff
@@ -0,0 +1,362 @@
...
+                        reason = reason[len(out_drop_reason_prefix):]
+                    print('\t{}'.format(reason))
+
+    def create_counter(self, counter_name, alias, group, counter_type, description, reasons):
+        """
+            Creates a new counter configuration
+        """
+
+        if not counter_name:
+            raise Exception('Counter name not provided')
```
qiluo-msft: >Exception [](start = 18, length = 9)

Use specific exception type?

scripts/dropconfig
```diff
@@ -0,0 +1,362 @@
...
+        for reason in reasons:
+            self.config_db.set_entry(self.config_db.serialize_key((DROP_REASON_TABLE_PREFIX, counter_name)), reason, {})
+
+        drop_counter_entry = {'type' : counter_type}
+
+        if alias:
+            drop_counter_entry['alias'] = alias
+        if group:
+            drop_counter_entry['group'] = group
+        if description:
```
qiluo-msft: >description [](start = 11, length = 11)

Do you support empty description? If yes, if-block will be skipped.

scripts/dropstat
```diff
@@ -0,0 +1,397 @@
...
+
+    def get_port_state(self, port_name):
+        """
+            Get the state of the given port.
+        """
+        full_table_id = PORT_STATUS_TABLE_PREFIX + port_name
+        admin_state = self.db.get(self.db.APPL_DB, full_table_id, PORT_ADMIN_STATUS_FIELD)
+        oper_state = self.db.get(self.db.APPL_DB, full_table_id, PORT_OPER_STATUS_FIELD)
+        if admin_state is None or oper_state is None:
+             return PORT_STATE_NA
```
qiluo-msft: >  [](start = 12, length = 1)

Try find formatting issues by [http://pep8online.com](http://pep8online.com)

config/main.py
daall: That is a super cool tool, thank you for sharing! I'll go ahead and update this.

daall: It looks like `shlex.quote` was added in python 3, which we don't support yet.

qiluo-msft: https://docs.python.org/2/library/shlex.html#shlex.shlex.quotes

qiluo-msft: Is it possible to change this script to python3 syntax?

daall:
> need unit test for show commands

added!

config/main.py
daall: I don't think so but maybe @jleveque can confirm. In either case, python2 is being deprecated soon so we do need some plan to move over to python3 in SONiC.

scripts/dropconfig
daall: They don't appear to be automatically cleared when drop reasons are added/removed. Do you have any concerns with this approach?

config/main.py
qiluo-msft: Agree. Please created an issue or add TODO comment in code.

daall:
retest default please

daall:
retest this please

config/main.py
daall: Created #721

daall:
> Looks good to me. Please check other review comments.
> In future, please minimize force-push, because we are using internal tool to compare between iterations, and force push will lose comment context.

Will do, thanks for the heads up! And thanks for the review! :)

scripts/dropconfig
daall: The counters *are* cleared when drop reasons are added and removed.

clear/main.py
```diff
@@ -194,6 +194,12 @@ def pfccounters():
     command = "pfcstat -c"
     run_command(command)
 
+@cli.command()
+def dropcounter():
```
jleveque: Suggest pluralizing the subcommand name, "dropcounters". The Click library will handle abbreviation and tab-completion, so "dropcounter" will still work. This suggestion would apply to all commands in this PR.

config/main.py
```diff
@@ -1591,6 +1591,79 @@ def incremental(file_name):
...
+    run_command(command, display_cmd=verbose)
+
+
+#
+# 'delete_counter' subcommand ('config dropcounter delete_counter')
+#
+@dropcounter.command()
+@click.argument("counter_name", type=str, required=True)
+@click.option('-v', '--verbose', is_flag=True, help="Enable verbose output")
+def delete_counter(counter_name, verbose):
```
jleveque: Is there a reason to include the suffix "_counter" here? It seems a bit redundant. Is it to distinguish it from the "_reasons" subcommands commands below?

config/main.py
```diff
@@ -1591,6 +1591,79 @@ def incremental(file_name):
...
+#
+@dropcounter.command()
+@click.argument("counter_name", type=str, required=True)
+@click.argument("counter_type", type=str, required=True)
+@click.argument("reasons",      type=str, required=True)
+@click.option("-a", "--alias", type=str, help="Alias for this counter")
+@click.option("-g", "--group", type=str, help="Group for this counter")
+@click.option("-d", "--desc",  type=str, help="Description for this counter")
+@click.option('-v', '--verbose', is_flag=True, help="Enable verbose output")
+def initialize_counter(counter_name, alias, group, counter_type, desc, reasons, verbose):
```
jleveque: Is there a reason to include the suffix "_counter" here? It seems a bit redundant. Is it to distinguish it from the "_reasons" subcommands commands below?

config/main.py
daall: That's right, although with the _reasons suffix I think it might be clear enough without the _counter suffix.

jleveque: That's kinda what I was thinking, as well.

PR #703, opened 2019-10-16
[neighbor_advertiser]: Adapt to different mirror ACL table names

stcheng:
ACL table names varies on different platforms.
The current assumption is that the name would contain either
EVERFLOW or EVERFLOWV6.

Thus a function find_mirror_table_name is added to find the exact
names for both the v4 and v6 mirror table names.

### DIFF
--- a/scripts/neighbor_advertiser
+++ b/scripts/neighbor_advertiser
@@ -36,6 +36,7 @@ SYSLOG_IDENTIFIER = 'neighbor_advertiser'
 #
 
 MIRROR_SESSION_NAME = 'neighbor_advertiser'
+MIRROR_ACL_TABLE_PREFIX = 'SONIC_'
 MIRROR_ACL_TABLE_NAME = 'EVERFLOW'
 MIRROR_ACL_TABLEV6_NAME = 'EVERFLOWV6'
 MIRROR_ACL_RULE_NAME = 'rule_arp'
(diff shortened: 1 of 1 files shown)


scripts/neighbor_advertiser
```diff
@@ -397,25 +413,29 @@ def add_mirror_session(dst_ipv4_addr):
...
     acl_rule_info = {
         'PRIORITY': '8887',
         'ICMPV6_TYPE': '135',
         'mirror_action': MIRROR_SESSION_NAME
     }
 
-    config_db.set_entry('ACL_RULE',
-            (MIRROR_ACL_TABLEV6_NAME, MIRROR_ACL_RULEV6_NAME), acl_rule_info)
+    if v6_table:
+        print v6_table
```
yxieca: Is the debug code you forgot to remove?

scripts/neighbor_advertiser
```diff
@@ -384,10 +384,26 @@ def save_as_json(obj, file_path):
...
 
 
+def find_mirror_table_name():
+    acl_tables = config_db.get_keys("ACL_TABLE")
+    v4_table, v6_table = "", ""
+    for table in acl_tables:
+        if MIRROR_ACL_TABLE_NAME in table and MIRROR_ACL_TABLEV6_NAME not in table:
+            v4_table = table
+        elif MIRROR_ACL_TABLEV6_NAME in table:
+            v6_table = table
```
yxieca: If you test MIRROR_ACL_TABLEV6_NAME first, then you don't need 2 conditions for v4_table name, right?

scripts/neighbor_advertiser
```diff
@@ -384,10 +384,26 @@ def save_as_json(obj, file_path):
         json.dump(obj, outfile, sort_keys = True)
 
 
+def find_mirror_table_name():
+    acl_tables = config_db.get_keys("ACL_TABLE")
+    v4_table, v6_table = "", ""
+    for table in acl_tables:
+        if MIRROR_ACL_TABLE_NAME in table and MIRROR_ACL_TABLEV6_NAME not in table:
+            v4_table = table
+        elif MIRROR_ACL_TABLEV6_NAME in table:
```
prsunny: just curious, isn't check for equality better?

scripts/neighbor_advertiser
```diff
@@ -384,10 +384,26 @@ def save_as_json(obj, file_path):
         json.dump(obj, outfile, sort_keys = True)
 
 
+def find_mirror_table_name():
```
yxieca: Can you add a comment for this function regarding why you need to do this look up? I think the reason is not clear and could be forgotten over time.

scripts/neighbor_advertiser
```diff
@@ -397,25 +413,29 @@ def add_mirror_session(dst_ipv4_addr):
...
         'mirror_action': MIRROR_SESSION_NAME
     }
 
-    config_db.set_entry('ACL_RULE',
-            (MIRROR_ACL_TABLE_NAME, MIRROR_ACL_RULE_NAME), acl_rule_info)
+    if v4_table:
+        config_db.set_entry('ACL_RULE',
+                (v4_table, MIRROR_ACL_RULE_NAME), acl_rule_info)
 
     acl_rule_info = {
```
daall: I don't think it makes a huge difference but is there any reason these are defined outside of the if-check if they're only used inside of the if-checks?

stcheng:
updated the PR. thanks for reviewing.

PR #714, opened 2019-10-23
[neighbor_advertiser]: Add sleep in setting mirror session and ACL rules

stcheng:
The sleep(3) is to ensure that orchagent could execute tasks by the
correct sequence so that mirror session and ACL rules could be
successfully created/removed.

### DIFF
--- a/scripts/neighbor_advertiser
+++ b/scripts/neighbor_advertiser
@@ -15,6 +15,7 @@ import argparse
 import syslog
 import traceback
 import subprocess
+import time
 import sonic_device_util
 from swsssdk import ConfigDBConnector
 from swsssdk import SonicV2Connector
@@ -28,6 +29,7 @@ from netaddr import IPAddress, IPNetwork
 DEFAULT_DURATION = 300
 DEFAULT_REQUEST_TIMEOUT = 2
 DEFAULT_FERRET_QUERY_RETRIES = 3
+DEFAULT_CONFIG_DB_WAIT_TIME = 3 # seconds
 SYSLOG_IDENTIFIER = 'neighbor_advertiser'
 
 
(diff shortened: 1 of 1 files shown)


scripts/neighbor_advertiser
```diff
@@ -30,7 +31,6 @@ DEFAULT_REQUEST_TIMEOUT = 2
 DEFAULT_FERRET_QUERY_RETRIES = 3
 SYSLOG_IDENTIFIER = 'neighbor_advertiser'
 
-
```
trzhang-msft: keep two newlines between sections

scripts/neighbor_advertiser
```diff
@@ -45,6 +45,7 @@ VXLAN_TUNNEL_NAME = 'neighbor_advertiser'
 VXLAN_TUNNEL_MAP_PREFIX = 'map_'
 
 
+
```
trzhang-msft: remove extra newline

scripts/neighbor_advertiser
```diff
@@ -442,6 +443,7 @@ def add_mirror_acl_rule():
 
 def set_mirror_tunnel(ferret_server_ip):
     add_mirror_session(ferret_server_ip)
+    time.sleep(3) # Ensure the mirror session is created before creating the rules
```
trzhang-msft: replace 3 with a const to be defined above in 'config' ?

scripts/neighbor_advertiser
```diff
@@ -466,6 +468,7 @@ def remove_mirror_acl_rule():
 
 def reset_mirror_tunnel():
     remove_mirror_acl_rule()
+    time.sleep(3) # Ensure the rules are removed before removing the mirror session
```
trzhang-msft: same as above comment

stcheng:
@trzhang-msft thanks for reviewing! I've resolved your comments.

stcheng:
retest this please

scripts/neighbor_advertiser
```diff
@@ -442,6 +444,8 @@ def add_mirror_acl_rule():
 
 def set_mirror_tunnel(ferret_server_ip):
     add_mirror_session(ferret_server_ip)
+    # Ensure the mirror session is created before creating the rules
+    time.sleep(DEFAULT_CONFIG_DB_WAIT_TIME)
```
lguohan: do we have retry logic in the orchagent?

yxieca: https://github.com/Azure/sonic-swss/pull/1104

PR #724, opened 2019-11-06
<issues#723>add queuestat -p Ethernetx -c to clear one port queue

yangshp1987:
**- What I did**
fix command "queuestat -p Ethernetx -c" to clear one port
**- How I did it**
queuestat script add port_clear parametre in function save_fresh_stats.
**- How to verify it**

image

image

### DIFF
--- a/scripts/queuestat
+++ b/scripts/queuestat
@@ -231,7 +231,7 @@ class Queuestat(object):
         else:
             self.cnstat_print(port, cnstat_dict)
 
-    def save_fresh_stats(self):
+    def save_fresh_stats(self, port_clear):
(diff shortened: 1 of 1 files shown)


lguohan:
-p and -c should be exclusive. It would be better to keep that way. in this case, can you add argument to -c so that it can take argument. if it has argument, then clear one port. if there is no argument, then it clears all ports.

But it is better to keep -p and -c option exclusive.

yangshp1987:
OK，I get it. I regarded -p and -c as independent parametres. I think -p means for just one port and -c means clearing.So I used -p and -c as clearing stats for one port.

PR #726, opened 2019-11-06
Do not start pfcwd for M0 devices

neethajohn:

**- What I did**
Disable pfcwd for M0 devices

**- How I did it**
Get the device type from the minigraph and use that for comparison

**- How to verify it**
 - Modified device type in the minigraph to 'MgmtToRRouter' and verified that pfcwd is not started when load_minigraph is executed.
  - For other types, pfcwd was started when load_minigraph was executed

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -532,8 +532,8 @@ def load_minigraph():
     proc = subprocess.Popen(command, shell=True, stdout=subprocess.PIPE)
     device_type, err = proc.communicate()
     if err:
-        click.echo("Could not get the device type from minigraph, exiting")
-        sys.exit(1)
+        click.echo("Could not get the device type from minigraph, setting device type to Unknown")
+        device_type = 'Unknown'
     else:
         device_type = device_type.strip()
 
--- a/config/main.py
+++ b/config/main.py
@@ -527,6 +527,16 @@ def load_minigraph():
     """Reconfigure based on minigraph."""
     log_info("'load_minigraph' executing...")
 
+    # get the device type
+    command = "{} -m -v DEVICE_METADATA.localhost.type".format(SONIC_CFGGEN_PATH)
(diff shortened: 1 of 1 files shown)


config/main.py
```diff
@@ -527,6 +527,16 @@ def load_minigraph():
     """Reconfigure based on minigraph."""
     log_info("'load_minigraph' executing...")
 
+    # get the device type
+    command = "{} -m -v DEVICE_METADATA.localhost.type".format(SONIC_CFGGEN_PATH)
+    proc = subprocess.Popen(command, shell=True, stdout=subprocess.PIPE)
+    device_type, err = proc.communicate()
+    if err:
+        click.echo("Could not get the device type from minigraph, exiting")
+        sys.exit(1)
```
yxieca: Do we have legitimate case where device type is not available in minigraph?

yxieca: From a cautious perspective. I am leaning towards letting this error go and enable pfcwd. There might be a default or test mingiraph missing device type?

config/main.py
```diff
@@ -540,7 +550,8 @@ def load_minigraph():
         command = "{} -H -m --write-to-db".format(SONIC_CFGGEN_PATH)
     run_command(command, display_cmd=True)
     client.set(config_db.INIT_INDICATOR, 1)
-    run_command('pfcwd start_default', display_cmd=True)
+    if device_type != 'MgmtToRRouter':
```
prsunny: Isn't it better to check if device_type contains 'Mgmt' ?

config/main.py
neethajohn: will make that change

config/main.py
neethajohn: as per our discussion, leaving this as is

yxieca:
retest this please

PR #727, opened 2019-11-06
Make configlet application script idempotent for updates.

renukamanavalan:
Delete of non-existing table/keys is a no-op, hence can be ignored.

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/scripts/configlet
+++ b/scripts/configlet
@@ -100,14 +100,27 @@ def init():
 
 def db_update(t, k, lst):
     init()
-    db.mod_entry(t, k, lst)
+    to_upd = False
+    data = db.get_entry(t, k)
+    for i in lst.keys():
+        if not data.has_key(i) or data[i] != lst[i]:
+            to_upd = True
+            break
+
+    if to_upd:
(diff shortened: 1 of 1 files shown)


yxieca:
Why is this change not needed for master?

renukamanavalan:
Should be for master. Let me create PR for master.

scripts/configlet
```diff
@@ -100,14 +100,27 @@ def init():
...
+    data = db.get_entry(t, k)
+    for i in lst.keys():
+        if not data.has_key(i) or data[i] != lst[i]:
+            to_upd = True
+            break
+
+    if to_upd:
+        db.mod_entry(t, k, lst)
 
 def db_filtered_upd(t, k, lst):
```
lguohan: maybe db_replace is a more accurate name here?

renukamanavalan:
Opened PR for master.
https://github.com/Azure/sonic-utilities/pull/728

scripts/configlet
lguohan: db_delete_fields

PR #729, opened 2019-11-07
kdump support

olivier-singla:
In the event of a kernel crash, we need to gather as much information as possible to understand and identify the root cause of the crash. Currently, the kernel does not provide much information, which make kernel crash investigation difficult and time consuming.

Fortunately, there is a way in the kernel to provide more information in the case of a kernel crash. kdump is a feature of the Linux kernel that creates crash dumps in the event of a kernel crash. This PR will add kernel kdump support. Please note that there is another PR in sonic-utilities which is also needed:
https://github.com/Azure/sonic-buildimage/pull/3722

An extension to the CLI utilities config and show is provided to configure and manage kdump:
 - view kdump status (enabled/disabled, active, configuration, stored crash files)
 - enable / disable kdump functionality
 - configure kdump (how many kernel crash logs can be saved, memory
   allocated for capture kernel)
 - view kernel crash logs

There is a design document which describes this kdump implementation:
https://github.com/Azure/SONiC/pull/510

**- What I did**
Added kdump support

**- How I did it**
Please read HLD:
https://github.com/Azure/SONiC/pull/510

**- How to verify it**
config kdump enable
config save 
reboot
echo 1 > /proc/sys/kernel/sysrq ; echo c > /proc/sysrq-trigger
show kdump
show kdump log 1 20

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1242,7 +1242,7 @@ def kdump():
 
 @kdump.command()
 def disable():
-    """Disable kdump operation!"""
+    """Disable kdump operation"""
     config_db = ConfigDBConnector()
     if config_db is not None:
         config_db.connect()
--- a/config/main.py
+++ b/config/main.py
@@ -1250,17 +1250,13 @@ def disable():
         run_command("sonic-kdump-config --disable")
 
 @kdump.command()
-@click.option('-y', '--yes', is_flag=True, help="Force rebooting after enabling kdump")
-def enable(yes):
+def enable():
     """Enable kdump operation"""
     config_db = ConfigDBConnector()
     if config_db is not None:
         config_db.connect()
         config_db.mod_entry("KDUMP", "config", {"enabled": "true"})
-        if yes:
-            run_command("sonic-kdump-config --enable --yes")
-        else:
-            run_command("sonic-kdump-config --enable")
+        run_command("sonic-kdump-config --enable")
 
 @kdump.command()
 @click.argument('kdump_memory', metavar='<kdump_memory>', required=True)
--- a/config/main.py
+++ b/config/main.py
@@ -1233,6 +1233,55 @@ def shutdown():
     """Shut down BGP session(s)"""
     pass
 
+@config.group()
+def kdump():
+    """ Configure kdump """
+    if os.geteuid() != 0:
+        exit("Root privileges are required for this operation")
+    pass
+
+@kdump.command()
+def disable():
+    """Disable kdump operation!"""
+    config_db = ConfigDBConnector()
+    if config_db is not None:
+        config_db.connect()
+        config_db.mod_entry("KDUMP", "config", {"enabled": "false"})
+        run_command("sonic-kdump-config --disable")
+
+@kdump.command()
+@click.option('-y', '--yes', is_flag=True, help="Force rebooting after enabling kdump")
+def enable(yes):
+    """Enable kdump operation"""
+    config_db = ConfigDBConnector()
+    if config_db is not None:
+        config_db.connect()
+        config_db.mod_entry("KDUMP", "config", {"enabled": "true"})
+        if yes:
+            run_command("sonic-kdump-config --enable --yes")
+        else:
+            run_command("sonic-kdump-config --enable")
+
+@kdump.command()
+@click.argument('kdump_memory', metavar='<kdump_memory>', required=True)
+def memory(kdump_memory):
+    """Set memory allocated for kdump capture kernel"""
+    config_db = ConfigDBConnector()
+    if config_db is not None:
+        config_db.connect()
+        config_db.mod_entry("KDUMP", "config", {"memory": kdump_memory})
+        run_command("sonic-kdump-config --memory %s" % kdump_memory)
+
+@kdump.command()
+@click.argument('kdump_num_dumps', metavar='<kdump_num_dumps>', required=True, type=int)
+def num_dumps(kdump_num_dumps):
+    """Set max number of dump files for kdump"""
+    config_db = ConfigDBConnector()
+    if config_db is not None:
+        config_db.connect()
+        config_db.mod_entry("KDUMP", "config", {"num_dumps": kdump_num_dumps})
+        run_command("sonic-kdump-config --num_dumps %d" % kdump_num_dumps)
+
 # 'all' subcommand
 @shutdown.command()
 @click.option('-v', '--verbose', is_flag=True, help="Enable verbose output")
--- a/scripts/reboot
+++ b/scripts/reboot
@@ -1,5 +1,12 @@
 #!/bin/bash
 
+# Reboot immediately if we run the kdump capture kernel
+VMCORE_FILE=/proc/vmcore
+if [ -e $VMCORE_FILE -a -s $VMCORE_FILE ]; then
+        debug "We have a /proc/vmcore, then we just kdump'ed"
+        /sbin/reboot
+fi
+
 REBOOT_USER=$(logname)
 REBOOT_TIME=$(date)
 PLATFORM=$(sonic-cfggen -H -v DEVICE_METADATA.localhost.platform)
--- a/scripts/sonic-kdump-config
+++ b/scripts/sonic-kdump-config
@@ -1,0 +1,633 @@
+#!/usr/bin/python
+'''
+Copyright 2019 Broadcom. The term "Broadcom" refers to Broadcom Inc.
+and/or its subsidiaries.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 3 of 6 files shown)


lguohan:
retest this please

show/main.py
```diff
@@ -1792,6 +1792,87 @@ def vlan():
...
+@kdump.command('status', default=True)
+def status():
+    """Show kdump status"""
+    run_command("sonic-kdump-config --status")
+    run_command("sonic-kdump-config --memory")
+    run_command("sonic-kdump-config --num_dumps")
+    run_command("sonic-kdump-config --files")
+
+@kdump.command('memory')
+def memory():
```
lguohan: there is duplicated code in get_kdump_memory, please remove such duplication.

show/main.py
```diff
@@ -1792,6 +1792,87 @@ def vlan():
...
+        if table_data is not None:
+            config_data = table_data.get('config')
+            if config_data is not None:
+                kdump_memory_from_db = config_data.get('memory')
+                if kdump_memory_from_db is not None:
+                    kdump_memory = kdump_memory_from_db
+    click.echo("Memory Reserved: %s" % kdump_memory)
+
+@kdump.command('num_dumps')
+def num_dumps():
```
lguohan: duplicated code in sonic-kdump-config

show/main.py
```diff
@@ -1792,6 +1792,87 @@ def vlan():
...
+#
+# 'kdump command ("show kdump ...")
+#
+@cli.group(cls=AliasedGroup, default_if_no_args=True, )
+def kdump():
+    """Show kdump configuration, status and information """
+    pass
+
+@kdump.command('enabled')
+def enabled():
```
lguohan: duplicated code in sonic-kdump-config, please remove

scripts/sonic-kdump-config
```diff
@@ -0,0 +1,633 @@
...
+        else:
+            parser.print_help()
+            sys.exit(1)
+    except Exception as e:
+        print_err('Error! Exception[%s] occured while processing the command sonic-kdump-config %s.' %(str(e), sys.argv[1]))
+        sys.exit(1)
+
+    if changed:
+        if options.yes:
+            perform_reboot()
```
lguohan: yes option is unclear to user. also user can choose to do warm reboot or fast reboot, recommend to do not integrate reboot into this config tool.

scripts/sonic-kdump-config
```diff
@@ -0,0 +1,633 @@
...
+            print_err("Unable to write KDUMP_NUM_DUMPS into %s" % kdump_cfg)
+            sys.exit(1)
+    else:
+        print_err("Error while writing KDUMP_NUM_DUMPS into %s" % kdump_cfg)
+        sys.exit(1)
+
+## Perform reboot command
+#
+#  We trigger a system reboot, WITHOUT asking for a confirmation
+def perform_reboot():
```
lguohan: user can choose warm reboot or fast reboot, better not implement reboot in this config tool.

scripts/sonic-kdump-config
```diff
@@ -0,0 +1,633 @@
...
+    memory = get_kdump_memory()
+    num_dumps = get_kdump_num_dumps()
+    if verbose:
+        print("configDB: kdump_enabled=%d memory=[%s] num_nums=%d" % (kdump_enabled, memory, num_dumps))
+
+    current_img = get_current_image();
+    if verbose:
+        print("Current image=[%s]" % current_img)
+    try:
+        lines = [line.rstrip('\n') for line in open(grub_cfg)]
```
lguohan: you cannot assume grub here. sonic also works on aboot. please detect the image type. if it is aboot, then you need to quit.

config/main.py
```diff
@@ -1156,6 +1156,55 @@ def shutdown():
...
+@config.group()
+def kdump():
+    """ Configure kdump """
+    if os.geteuid() != 0:
+        exit("Root privileges are required for this operation")
+    pass
+
+@kdump.command()
+def disable():
+    """Disable kdump operation!"""
```
jleveque: Remove exclamation point from docstring.

config/main.py
```diff
@@ -1156,6 +1156,55 @@ def shutdown():
...
+def disable():
+    """Disable kdump operation"""
+    config_db = ConfigDBConnector()
+    if config_db is not None:
+        config_db.connect()
+        config_db.mod_entry("KDUMP", "config", {"enabled": "false"})
+        run_command("sonic-kdump-config --disable")
+
+@kdump.command()
+@click.option('-y', '--yes', is_flag=True, help="Force rebooting after enabling kdump")
```
lguohan: this comments need to be modified.

scripts/sonic-kdump-config
```diff
@@ -0,0 +1,616 @@
...
+        rewrite_grub_cfg(lines, grub_cfg)
+
+    write_use_kdump(1)
+
+    return changed
+
+## Command: Enable kdump on Next image only
+#
+#  @param verbose If True, the function will display a few additional information
+#  @param no      If True, will not display reboot needed message
```
lguohan: do we still need this param now?

scripts/sonic-kdump-config
```diff
@@ -0,0 +1,616 @@
...
+            print_err("Unable to write KDUMP_NUM_DUMPS into %s" % kdump_cfg)
+            sys.exit(1)
+    else:
+        print_err("Error while writing KDUMP_NUM_DUMPS into %s" % kdump_cfg)
+        sys.exit(1)
+
+## Command: Enable kdump
+#
+#  @param verbose If True, the function will display a few additinal information
+#  @param no      If True, will not display reboot needed message
```
lguohan: remove this parameter since it is not needed.?

scripts/sonic-kdump-config
```diff
@@ -0,0 +1,616 @@
...
+#  @return        True is the grub configuration has changed, and False if it has not
+def cmd_kdump_enable(verbose, no):
```
Staphylo: Could the parsing of grub.cfg and the logic be separated into different functions?
The idea would be facilitating the integration of other bootloaders than grub by adding an abstraction layer.

I would propose something like the following in the case of cmd_kdump_enable. Some other places in this file would need some changes too. If a cmd_function refers to grub rather than a function of bootloader then it's lacking genericity.
```python

bootloader = get_bootloader() # this would return Grub() or Aboot() based on autodetection
if bootloader is None:
   print("Feature not supported on this platform")
   return False
image = get_current_image()
old_crashkernel = bootloader.get_crash_kernel( image )
new_crashkernel = "crashkernel: ..."
if old_crashkernel != new_crashkernel:
   bootloader.set_crash_kernel( image, new_crashkernel )
```

lguohan: @olivier-singla , since the sonic support multiple boot loader, not just grub. We have aboot, now we are working on u-boot. The concern here is that current code does not provide good abstraction for the community to add support for other types of boot load. Can you simplify the code logic and define the api. So that others can provide their own implementation for their own bootloader?

scripts/sonic-kdump-config
```diff
@@ -0,0 +1,632 @@
...
+def cmd_kdump_disable(verbose):
+
+    kdump_enabled = get_kdump_administrative_mode()
+    memory = get_kdump_memory()
+    num_dumps = get_kdump_num_dumps()
+    if verbose:
+        print("configDB: kdump_enabled=%d memory=[%s] num_nums=%d" % (kdump_enabled, memory, num_dumps))
+
+    if os.path.exists(grub_cfg):
+        return kdump_disable_grub(verbose, kdump_enabled, memory, num_dumps)
```
Staphylo: It's better than the previous version but I still feel some abstraction is lacking.
Basically with this design for every operation each bootloader has to be added inside a if/elif/elif. This is redundant and unnecessary.
Both the `kdump_disable_grub` and `kdump_config_next_grub` functions have a mix of common logic tied with grub specific code. Adding a new bootloader to this file would likely end up with a lot of copy pasting.

Could we instead just rely on a `get_bootloader()` function? This function would hold the if/elif/elif that returns the Bootloader instance or raise UnknownBootloaderError.
Each bootloader would then override the method of the base class Bootloader to add support.

Adding a new Bootloader would now just be about adding a new class definition and changing 2 lines in ` `get_bootloader` which is a much better design.

lguohan: @olivier-singla , can you address them?

olivier-singla: I am not sure more abstraction will make things easier or simpler when adding support for another boot loader. There is little code duplicated between `kdump_enable_grub`, `kdump_config_next_grub` and `cmd_kdump_disable_grub`. grub and u-boot (for instance) use a different mechanism: grub use a flat text file, where u-boot use environment variables.

olivier-singla:
retest please

lguohan:
retest this please

scripts/reboot
```diff
@@ -1,5 +1,12 @@
 #!/bin/bash
 
+# Reboot immediately if we run the kdump capture kernel
+VMCORE_FILE=/proc/vmcore
+if [ -e $VMCORE_FILE -a -s $VMCORE_FILE ]; then
+        debug "We have a /proc/vmcore, then we just kdump'ed"
+        /sbin/reboot
```
stephenxs: Is it possible that to call `/sbin/reboot` ahead of echoing `user request reboot` to /host/reboot-cause/reboot-cause.txt fails the reboot cause?

PR #737, opened 2019-11-14
[bugfix]: allow loopback vrf migration from version 1_0_1

lguohan:
commit fe4446a4808bc98ca2b24b0a9afe822aaa50a7ce fix loopback vrf migration.

but the commit does not enable db migration from version 1_0_1.

Now, bump version 1_0_2 and allow db fix from 1_0_1

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/scripts/db_migrator.py
+++ b/scripts/db_migrator.py
@@ -35,7 +35,7 @@ class DBMigrator():
                      none-zero values.
               build: sequentially increase within a minor version domain.
         """
-        self.CURRENT_VERSION = 'version_1_0_1'
+        self.CURRENT_VERSION = 'version_1_0_2'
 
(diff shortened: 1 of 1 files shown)


scripts/db_migrator.py
```diff
@@ -116,18 +116,26 @@ def version_unknown(self):
...
-        self.set_version('version_1_0_1')
-        return 'version_1_0_1'
-
+        self.set_version('version_1_0_2')
+        return 'version_1_0_2'
 
     def version_1_0_1(self):
         """
         Current latest version. Nothing to do here.
         """
```
yxieca: Can you remove the claim of "Current latest version"?

PR #740, opened 2019-11-15
Platform plugin for Fast-reboot/warm-reboot 

santhosh-kt:
**- What I did**
Added support for calling a platform specific plugin in fastboot, warmboot
**- How I did it**
Changes applied to 'scripts/fast-reboot'
**- How to verify it**
Write a platform plugin script and execute fast-reboot command
Plugin script will be executed and then reboot will be invoked

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -13,6 +13,9 @@ STRICT=no
 REBOOT_METHOD="/sbin/kexec -e"
 ASSISTANT_IP_LIST=""
 ASSISTANT_SCRIPT="/usr/bin/neighbor_advertiser"
+DEVPATH="/usr/share/sonic/device"
+PLATFORM=$(sonic-cfggen -H -v DEVICE_METADATA.localhost.platform)
+PLATFORM_PLUGIN="${REBOOT_TYPE}_plugin"
 
 # Require 100M available on the hard drive for warm reboot temp files,
 # Size is in 1K blocks:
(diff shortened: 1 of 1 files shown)


yxieca:
Can you give an example of how this platform plugin will be used, and why?

Arul-Jeniston-Mc:
- Dell hardware stores all reset/reboot reasons (including cold and warm cpu resets) in a register. This hardware register value persists till next reset/reboot occurs. 
- As the hardware doesn't have a way to detect fast-boot, following the fast-boot, hardware reset-cause register would end up having the previous reset reason value which is a stale value.
- This plugin is added to find this situation. Otherwise, following fast-boot, the reset reason API (that reads the data from hardware) returns incorrect reset reason.

For example:-
1. Run 1:  User power-cycled the switch. .Hardware reset-cause register has value 0x11 which means power-cycle.
2. Run 2: reboot command executed from SONiC. Hardware reset-cause register value changed to 0x55 which means warm-cpu reboot.
3. Run 3: hardware watchdog resets the switch. Hardware reset-cause register value changed to 0x33 which means watchdog reset.
4. Run 4: fast-boot command executed from SONiC. Hardware reset-cause register value remains un-changed as 0x33 which means watchdog reset.
5. Run 5: show reboot reason call reset reason API which refers the hardware and returns watchdog reset. This is incorrect. The above change addresses this problem.

yxieca:
retest this please

santhosh-kt:
Please include this change for 201811 branch also.

PR #746, opened 2019-11-20
[config] Add 'feature' subcommand

pra-moh:
**- What I did**
added command to configure feature status 
**- How I did it**
By adding in main.py 
**- How to verify it**
config feature_status name value

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -2171,7 +2171,7 @@ def delete(ctx):
 # 
 @config.command('feature')
 @click.argument('name', metavar='<feature-name>', required=True)
-@click.argument('state', metavar='<enabled/disabled>', required=True, type=click.Choice(["enabled", "disabled"]))
+@click.argument('state', metavar='<feature-state>', required=True, type=click.Choice(["enabled", "disabled"]))
 def feature_status(name,state):
     """ Configure status of feature"""
     config_db = ConfigDBConnector()
--- a/config/main.py
+++ b/config/main.py
@@ -2167,13 +2167,12 @@ def delete(ctx):
     config_db.set_entry('SFLOW', 'global', sflow_tbl['global'])
 
 #
-# 'feature_status' command ('config feature name state')
+# 'feature' command ('config feature name state')
 # 
 @config.command('feature')
-@click.pass_context
 @click.argument('name', metavar='<feature-name>', required=True)
 @click.argument('state', metavar='<enabled/disabled>', default='disabled', required=True, type=click.Choice(["enabled", "disabled"]))
-def feature_status(ctx,name,state):
+def feature_status(name,state):
     """ Configure status of feature"""
     config_db = ConfigDBConnector()
     config_db.connect()
@@ -2186,7 +2185,7 @@ def feature_status(ctx,name,state):
         click.echo("{} feature doesnt exist".format(name))
         return
     if not state:
-        click.echo("Empty state value")
+        click.echo("Invalid feature state")
         return
 
     config_db.mod_entry('FEATURE', name,{'status': state})
--- a/config/main.py
+++ b/config/main.py
@@ -2172,7 +2172,7 @@ def delete(ctx):
 @config.command('feature')
 @click.argument('name', metavar='<feature-name>', required=True)
 @click.argument('state', metavar='<feature-state>', required=True, type=click.Choice(["enabled", "disabled"]))
-def feature_status(name,state):
+def feature_status(name, state):
     """ Configure status of feature"""
     config_db = ConfigDBConnector()
     config_db.connect()
--- a/config/main.py
+++ b/config/main.py
@@ -2167,18 +2167,16 @@ def delete(ctx):
     config_db.set_entry('SFLOW', 'global', sflow_tbl['global'])
 
 #
-# 'feature_status' command ('config feature_status name value')
+# 'feature_status' command ('config feature name state')
 # 
-@config.command('feature_status')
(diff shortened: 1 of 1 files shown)


config/main.py
```diff
@@ -2166,6 +2166,35 @@ def delete(ctx):
     sflow_tbl['global'].pop('agent_id')
     config_db.set_entry('SFLOW', 'global', sflow_tbl['global'])
 
+#
+# 'feature_status' command ('config feature_status name value')
```
jleveque: I think we should simply name the subcommand "feature" (`config feature ...`)

config/main.py
```diff
@@ -2166,6 +2166,35 @@ def delete(ctx):
...
+# 
+@config.command('feature_status')
+@click.pass_context
+@click.argument('name', metavar='<feature-name>', required=True)
+@click.argument('value', metavar='<enabled/disabled>', required=True)
```
jleveque: I think the argument name should be `state`, and instead of setting the metavar to `<enabled/disabled>`, you should set it to "<state>" and set `type=click.Choice(["enabled", "disabled"])` to allow the user to only enter those two options. Then you don't need the `accepted_values` variable below, either.

config/main.py
```diff
@@ -2166,6 +2166,35 @@ def delete(ctx):
...
+def feature_status(ctx,name,value):
+    """ Configure status of feature"""
+    config_db = ConfigDBConnector()
+    config_db.connect()
+    accepted_values = {'enabled', 'disabled'}
+    keyname = 'FEATURE' + name
```
jleveque: keyname is not used anywhere. Remove?

config/main.py
```diff
@@ -2166,6 +2166,30 @@ def delete(ctx):
     sflow_tbl['global'].pop('agent_id')
     config_db.set_entry('SFLOW', 'global', sflow_tbl['global'])
 
+#
+# 'feature_status' command ('config feature name state')
+# 
+@config.command('feature')
+@click.pass_context
```
jleveque: Click context is not used in function. Remove?

config/main.py
```diff
@@ -2166,6 +2166,30 @@ def delete(ctx):
...
+    """ Configure status of feature"""
+    config_db = ConfigDBConnector()
+    config_db.connect()
+    status_data = config_db.get_entry('FEATURE', name)
+
+    if not name:
+        click.echo("Invalid feature name")
+        return
+    if not status_data:
+        click.echo("{} feature doesnt exist".format(name))
```
jleveque:
```suggestion
        click.echo(" Feature '{}' doesn't exist".format(name))
```

config/main.py
```diff
@@ -2166,6 +2166,30 @@ def delete(ctx):
...
+        return
+    if not state:
+        click.echo("Empty state value")
```
jleveque: Suggest changing message to "Invalid feature state"

config/main.py
```diff
@@ -2166,6 +2166,29 @@ def delete(ctx):
...
+def feature_status(name,state):
+    """ Configure status of feature"""
+    config_db = ConfigDBConnector()
+    config_db.connect()
+    status_data = config_db.get_entry('FEATURE', name)
+
+    if not name:
+        click.echo("Invalid feature name")
+        return
+    if not status_data:
```
hui-ma: why does it matter and is considered as a failure case?

jleveque: The thought here, I believe, is that if status_data is None, there was no key named "<name>" in the FEATURES table, therefore <name> is not a valid feature to modify.

pra-moh: That's correct

config/main.py
```diff
@@ -2166,6 +2166,23 @@ def delete(ctx):
     sflow_tbl['global'].pop('agent_id')
     config_db.set_entry('SFLOW', 'global', sflow_tbl['global'])
 
+#
+# 'feature' command ('config feature name state')
+# 
+@config.command('feature')
+@click.argument('name', metavar='<feature-name>', required=True)
+@click.argument('state', metavar='<enabled/disabled>', required=True, type=click.Choice(["enabled", "disabled"]))
```
jleveque: Please change metavar value to reflect usage instead of possible values ( `<feature-state>`?), now that possible values are handled by click.Choice.

config/main.py
```diff
@@ -2166,6 +2166,23 @@ def delete(ctx):
     sflow_tbl['global'].pop('agent_id')
     config_db.set_entry('SFLOW', 'global', sflow_tbl['global'])
 
+#
+# 'feature' command ('config feature name state')
+# 
+@config.command('feature')
+@click.argument('name', metavar='<feature-name>', required=True)
+@click.argument('state', metavar='<feature-state>', required=True, type=click.Choice(["enabled", "disabled"]))
+def feature_status(name,state):
```
jleveque: Please add a space after the comma.

config/main.py
```diff
@@ -2166,6 +2166,23 @@ def delete(ctx):
...
+    """ Configure status of feature"""
+    config_db = ConfigDBConnector()
+    config_db.connect()
+    status_data = config_db.get_entry('FEATURE', name)
+
+    if not status_data:
+        click.echo(" Feature '{}' doesn't exist".format(name))
+        return
+
+    config_db.mod_entry('FEATURE', name,{'status': state})
```
jleveque: Please add a space after the second comma. Sorry I missed this one earlier.

jleveque:
Retest this please

PR #750, opened 2019-11-22
[neighbor_advertiser] Adds initial support for HTTPS to neighbor advertiser

daall:

**- What I did**
Added a call to the HTTPS endpoint for Ferret instances.

**- How I did it**
Since some Ferret instances may still be using HTTP, I added a new function that attempts to call the new HTTPS endpoint before failing over to HTTP.

**- How to verify it**
1. Attempt to run neighbor_advertiser against a ferret VIP that supports HTTPS. The script should run successfully.
2. Attempt to run neighbor_advertiser against a ferret VIP that does not currently support HTTP. The script should print info messages to the syslog indicating that the HTTPS endpoint was not found, and then succeed using HTTP.

### DIFF
--- a/scripts/neighbor_advertiser
+++ b/scripts/neighbor_advertiser
@@ -357,7 +357,7 @@ def wrapped_ferret_request(request_slice, https_endpoint, http_endpoint):
                                      timeout=DEFAULT_REQUEST_TIMEOUT,
                                      verify=False)
     except ConnectTimeout:
-        log_info("Failed to connect to Ferret endpoint, trying the backup...")
+        log_info("HTTPS Ferret endpoint not found, trying HTTP...")
         response = requests.post(http_endpoint,
                                  json=request_slice,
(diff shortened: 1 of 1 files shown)


scripts/neighbor_advertiser
```diff
@@ -338,17 +340,40 @@ def construct_neighbor_advertiser_slice():
...
     return slice_obj
 
+def wrapped_ferret_request(request_slice, https_endpoint, http_endpoint):
+    """
+        Attempts to reach ferret by first trying HTTPS, failing over to HTTP in
+        case of failure (e.g. timeout, endpoint not found, etc.).
+    """
+    response = None
+
+    # FIXME: While we transition to HTTPS we're disabling the verify field. We
```
yxieca: This is probably more a NOTE than FIXME

yxieca:
retest this please

PR #757, opened 2019-12-01
[neighbor advertiser] catch all exceptions while trying https endpoint

yxieca:
**- What I did**

When connecting https endpoint failed, we need to try http endpoint.
Therefore we need to catch all exceptions.

**- How to verify it**
warm-reboot-control-plane-assistant test. Also tested with https endpoint success code and timeout.

### DIFF
--- a/scripts/neighbor_advertiser
+++ b/scripts/neighbor_advertiser
@@ -355,6 +355,7 @@ def wrapped_ferret_request(request_slice, https_endpoint, http_endpoint):
                                      json=request_slice,
                                      timeout=DEFAULT_REQUEST_TIMEOUT,
                                      verify=False)
+            response.raise_for_status()
     except Exception as e:
         log_info("Connect HTTPS Ferret endpoint failed (error: {}), trying HTTP...".format(e))
         response = requests.post(http_endpoint,
--- a/scripts/neighbor_advertiser
+++ b/scripts/neighbor_advertiser
@@ -21,7 +21,6 @@ import sonic_device_util
 from swsssdk import ConfigDBConnector
 from swsssdk import SonicV2Connector
 from netaddr import IPAddress, IPNetwork
-from requests.exceptions import ConnectTimeout
 
 
 #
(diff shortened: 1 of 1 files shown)


yxieca:
retest this please

scripts/neighbor_advertiser
```diff
@@ -356,8 +355,8 @@ def wrapped_ferret_request(request_slice, https_endpoint, http_endpoint):
                                      json=request_slice,
                                      timeout=DEFAULT_REQUEST_TIMEOUT,
                                      verify=False)
-    except ConnectTimeout:
-        log_info("HTTPS Ferret endpoint not found, trying HTTP...")
+    except Exception as e:
```
daall: @trzhang-msft @yxieca is it ever possible for the HTTPS endpoint to return a response with an error code (like 401, 403, etc.)? If it is possible, do we want to fail immediately if such an error code is returned or do we want to try the HTTP endpoint?

yxieca: It might be returning these codes. But I think we should still retry until http endpoint also failed. Bypassing http endpoint would be tossing away a chance to work. IMHO.

daall: Right, that's my concern. I think if one of those error codes is returned then there won't be an exception, so this method will just return the response to the post_neighbor_advertiser_slice method. That method returns failure if the response code is not 200. So, if we think one of these error codes might be returned by the HTTPS endpoint we might want to move that error handling logic into this method so that we can retry with the HTTP endpoint if need be.

yxieca: I see what you mean. Good catch!

trzhang-msft: @daall is correct. HTTP/HTTPs status codes include non-200, like 400 for bad request. But whether non-200 is treated as an exception is fully client side logic, and I'm not sure about the python built-in behavior here.

trzhang-msft:
@yxieca @daall @lguohan Starting next deployment, Ferret will only support HTTPS.

PR #758, opened 2019-12-03
[neighbor advertiser] raise exception when http endpoint return failure

yxieca:
**- What I did**
raise exception when http endpoint return failure to match the HTTPS handling flow.

### DIFF
--- a/scripts/neighbor_advertiser
+++ b/scripts/neighbor_advertiser
@@ -380,14 +380,6 @@ def post_neighbor_advertiser_slice(ferret_service_vip):
             log_error('The request failed, vip: {}, error: {}'.format(ferret_service_vip, e))
             return None
 
-        # Handle response errors
-        if not response:
-            log_error('Failed to set up neighbor advertiser slice, vip: {}, no response obtained'.format(ferret_service_vip))
-            return None
-        if response and not response.ok:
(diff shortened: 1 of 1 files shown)


jleveque:
Retest this please

yxieca:
retest this please

scripts/neighbor_advertiser
```diff
@@ -361,6 +361,7 @@ def wrapped_ferret_request(request_slice, https_endpoint, http_endpoint):
         response = requests.post(http_endpoint,
                                  json=request_slice,
                                  timeout=DEFAULT_REQUEST_TIMEOUT)
+        response.raise_for_status()
```
daall: I think this change makes sense, but it makes this part of `post_neighbor_advertiser_slice` obsolete since an exception will always be thrown if the response does not have one of the 2xx codes:
```
# Handle response errors
        if not response:
            log_error('Failed to set up neighbor advertiser slice, vip: {}, no response obtained'.format(ferret_service_vip))
            return None
        if response and not response.ok:
            log_error('Failed to set up neighbor advertiser slice, vip: {}, error_code: {}, error_content: {}'.format(ferret_service_vip, response.status_code, response.content))
            return None
```

I'd say you should probably delete this from `post_neighbor_advertiser_slice` since it's no longer going to be used. We could also incorporate some of the logging in the wrapper method if we want more specific debugging information. Let me know what you think.

yxieca:
retest this please

PR #763, opened 2019-12-12
Enable HW Watchdog for fast/warm reboot

sujinmkang:
To prevent from device hanging during fast/warm reboot

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -513,6 +513,12 @@ if [ -x /sbin/hwclock ]; then
     /sbin/hwclock -w || /bin/true
 fi
 
+# Enable Watchdog Timer
+if [[ -e /usr/local/bin/platform_watchdog_enable.sh ]]; then
+    debug "Enabling Watchdog before fast-reboot"
+    /usr/local/bin/platform_watchdog_enable.sh
+fi
+
 # Reboot: explicity call Linux native reboot under sbin
 debug "Rebooting with ${REBOOT_METHOD} to ${NEXT_SONIC_IMAGE} ..."
 exec ${REBOOT_METHOD}
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -514,9 +514,9 @@ if [ -x /sbin/hwclock ]; then
 fi
 
 # Enable Watchdog Timer
-if [[ -e /usr/local/bin/platform_watchdog_enable.sh ]]; then
+if [[ -x /usr/bin/watchdog ]]; then
     debug "Enabling Watchdog before fast-reboot"
-    /usr/local/bin/platform_watchdog_enable.sh
+    /usr/bin/watchdog -e
 fi
 
 # Reboot: explicity call Linux native reboot under sbin
--- a/scripts/watchdog
+++ b/scripts/watchdog
@@ -1,0 +1,72 @@
+#!/bin/bash
+
+VERBOSE=no
+
+function debug()
+{
+    if [[ x"${VERBOSE}" == x"yes" ]]; then
+        echo "$date $@"
+    fi
+    logger "$@"
+}
+
+function disable_watchdog()
+{
+    if [[ -x /usr/local/bin/platform_watchdog_disable.sh ]]; then
+        debug "Disabling Watchdog..."
+	/usr/local/bin/platform_watchdog_disable.sh
+    fi
+}
+
+function enable_watchdog()
+{
+    if [[ -x /usr/local/bin/platform_watchdog_enable.sh ]]; then
+        debug "Enabling Watchdog..."
+	/usr/local/bin/platform_watchdog_enable.sh
+    fi
+}
+
+SCRIPT=$0
+
+function show_help_and_exit()
+{
+    echo "Usage ${SCRIPT} [options]"
+    echo "    Request Enabling/Disabling Watchdog. Invoke platform-specific tool when available."
+    echo " "
+    echo "    Available options:"
+    echo "        -h, -? : getting this help"
+    echo "        -e     : enabling platform hw watchdog"
+    echo "        -d     : disabling platform hw watchdog"
+
+    exit 0
+}
+
+function parse_options()
+{
+    while getopts "h?ved" opt; do
+        case ${opt} in
+            h|\? )
+                show_help_and_exit
+                ;;
+            v )
+                VERBOSE=yes
+                ;;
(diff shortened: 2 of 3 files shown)


scripts/fast-reboot
```diff
@@ -513,6 +513,12 @@ if [ -x /sbin/hwclock ]; then
     /sbin/hwclock -w || /bin/true
 fi
 
+# Enable Watchdog Timer
+if [[ -e /usr/local/bin/platform_watchdog_enable.sh ]]; then
```
yxieca: Please test with '-x' instead

scripts/fast-reboot
```diff
@@ -513,6 +513,12 @@ if [ -x /sbin/hwclock ]; then
...
+    debug "Enabling Watchdog before fast-reboot"
+    /usr/local/bin/platform_watchdog_enable.sh
```
yxieca: Please don't call Dell watchdog script from here. Call a secondary script to call different script according to the platform.

scripts/watchdog
```diff
@@ -0,0 +1,72 @@
...
+function debug()
+{
+    if [[ x"${VERBOSE}" == x"yes" ]]; then
+        echo "$date $@"
+    fi
+    logger "$@"
+}
+
+function disable_watchdog()
+{
```
yxieca: Please detect platform first. This change only works for Dell. You also need to do things differently for Arista/Mellanox. And preferably, generate a log message state platform watchdog support is missing otherwise.

yxieca:
Please change this file: https://github.com/Azure/sonic-utilities/blob/master/setup.py#L30 to include the new script, so that it will be installed to /usr/bin. Please add the new script name according to the alphabetic order.

setup.py
```diff
@@ -68,6 +68,7 @@ def get_test_suite():
         'scripts/mmuconfig',
         'scripts/fast-reboot',
         'scripts/fast-reboot-dump.py',
+        'scripts/watchdog',
```
yxieca: Can you add it according to Alphabetic order?

scripts/watchdog
```diff
@@ -0,0 +1,149 @@
...
+    fi
+
+    return 1
+}
+
+
+function isArista() {
+    local -r _ARISTA_PLATFORM="x86_64-arista*"
+
+    if [[ "${ONIE_SWITCH_ASIC}" =~ "${_ARISTA_PLATFORM}" ]]; then
```
yxieca: ABOOT_PLATFORM

scripts/watchdog
```diff
@@ -0,0 +1,149 @@
...
+
+if [[ x"${ACTION}" == x"${ENABLE}" ]]; then
+    enable_watchdog
+fi
+
+if [[ x"${ACTION}" == x"${DISABLE}" ]]; then
+    disable_watchdog
+fi
+
+exit 1
```
yxieca: Please don't return 1. This will be treated as failure by caller.

scripts/watchdog
```diff
@@ -0,0 +1,149 @@
...
+    fi
+
+    return 1
+}
+
+
+function isArista() {
+    local -r _ARISTA_PLATFORM="x86_64-arista"
+
+    if [[ "${ABOOT_PLATFORM}" =~ "${_ARISTA_PLATFORM}" ]]; then
```
jleveque: Why quote the operands here, but not in the regex compare in isDellS6100() above? I suggest making them match. Either add quotes in isDellS6100() or remove the quotes here.

scripts/watchdog
```diff
@@ -0,0 +1,149 @@
...
+
+function disable_watchdog()
+{
+    if isDellS6100; then
+        if [[ -x /usr/local/bin/platform_watchdog_disable.sh ]]; then
+            debug "Disabling Watchdog..."
+            /usr/local/bin/platform_watchdog_disable.sh
+        fi
+    fi
+    if isMLNX; then
```
jleveque: Since these `if` statements are mutually exclusive, I suggest changing this to an `elif`.

scripts/watchdog
```diff
@@ -0,0 +1,149 @@
...
+        debug "Disabling Watchdog..."
+        # call watchdog api for mlnx
+    fi
+    if isArista; then
```
jleveque: Since these `if` statements are mutually exclusive, I suggest changing this to an `elif`.

scripts/watchdog
```diff
@@ -0,0 +1,149 @@
...
+
+function enable_watchdog()
+{
+    if isDellS6100; then
+        if [[ -x /usr/local/bin/platform_watchdog_enable.sh ]]; then
+            debug "Enabling Watchdog..."
+	    /usr/local/bin/platform_watchdog_enable.sh
+        fi
+    fi
+    if isMLNX; then
```
jleveque: Since these `if` statements are mutually exclusive, I suggest changing this to an `elif`.

scripts/watchdog
```diff
@@ -0,0 +1,149 @@
...
+        debug "Enabling Watchdog..."
+        # call watchdog api for mlnx
+    fi
+    if isArista; then
```
jleveque: Since these `if` statements are mutually exclusive, I suggest changing this to an `elif`.

scripts/watchdog
```diff
@@ -0,0 +1,149 @@
...
+        fi
+    fi
+}
+
+function enable_watchdog()
+{
+    if isDellS6100; then
+        if [[ -x /usr/local/bin/platform_watchdog_enable.sh ]]; then
+            debug "Enabling Watchdog..."
+	    /usr/local/bin/platform_watchdog_enable.sh
```
jleveque: Improper indent depth

scripts/watchdog
```diff
@@ -0,0 +1,149 @@
...
+if [[ "$EUID" -ne 0 ]]; then
+    echo "This command must be run as root" >&2
+    exit 1
+fi
+
+if [[ x"${ACTION}" == x"${ENABLE}" ]]; then
+    enable_watchdog
+fi
+
+if [[ x"${ACTION}" == x"${DISABLE}" ]]; then
```
jleveque: Since these `if` statements are mutually exclusive, I suggest changing this to an `elif`. You could then add an `else` statement which prints and/or logs an error/usage statement.

yxieca:
retest this please

PR #770, opened 2019-12-17
A generic JSON file updater, which can add/update-existing attributes.

renukamanavalan:
This tool would be used to update /etc/sonic/core_analyzer.rc.json file to add
credentials by HW proxy.

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/scripts/update_json.py
+++ b/scripts/update_json.py
@@ -1,0 +1,55 @@
+#! /usr/bin/env python
+
+import os
+import sys
+import json
+import argparse
+
+TMP_SUFFIX = ".tmp"
+BAK_SUFFIX = ".bak"
+
+def dict_update(dst, patch):
+    for k in patch.keys():
+        if type(patch[k]) == dict:
+            dst[k] = dict_update(dst[k], patch[k])
+        else:
+            dst[k] = patch[k]
+    return dst
+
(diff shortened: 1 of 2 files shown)


scripts/update_json.py
```diff
@@ -0,0 +1,55 @@
...
+    with open(tmpf, "w") as f:
+        json.dump(dst, f, indent = 4)
+
+    os.rename(rcf, bakf)
+    os.rename(tmpf, rcf)
+
+
+def main():
+    parser=argparse.ArgumentParser(description="Update JSON based file")
+    parser.add_argument("-r", "--rc", help="JSON file to be updated")
```
jleveque: Why "r"? What does it stand for? And long arg name should be more descriptive than "rc".

scripts/update_json.py
```diff
@@ -0,0 +1,55 @@
...
+
+    with open(tmpf, "w") as f:
+        json.dump(dst, f, indent = 4)
+
+    os.rename(rcf, bakf)
+    os.rename(tmpf, rcf)
+
+
+def main():
+    parser=argparse.ArgumentParser(description="Update JSON based file")
```
jleveque: Add spaces around assignment equals sign: `parser = argparse....`

PR #789, opened 2020-01-22
[fdbshow][nbrshow] Print interface OID in lieu of name if there is no OID->interface name mapping

jleveque:
In release branches prior to 201911, FDB or ARP learnt on VLAN interface LAG port will cause nothing to appear in the "show arp" or "show mac" output because there existed no OID->interface name never had the mappings for LAG interface.

This issue has since been fixed in master (and 201911). However, making this change in master is not harmful and will prevent a regression from causing no output to be displayed for the interface name in the future.


lguohan:
retest this please

PR #798, opened 2020-02-03
[show] Add subcommand to display the status of auto-restart feature for each container

yozhao101:
**- What I did**
Since we introduced the auto-restart feature for each container, we need add a show subcommand to display the current status of auto-restart feature for all containers or a specific container.

**- How I did it**
We define a function named autorestart to show the status of this features. This function will accept an option parameter which is the container name. If this parameter is not specified, this function will by default show the status of all containers. Otherwise it will show the status of specific container.

**- How to verify it**
We can run the command **show container feature autorestart** or **show container feature autorestart container_name** to verify.

### DIFF
--- a/show/main.py
+++ b/show/main.py
@@ -2813,6 +2813,7 @@ def features():
 def container():
     """Show container"""
     pass
+
 #
 # 'feature' group (show container feature ...)
 #
@@ -2824,7 +2825,6 @@ def feature():
 #
 # 'autorestart' subcommand (show container feature autorestart)
 #
-
 @feature.command('autorestart', short_help="Show whether the auto-restart feature for container(s) is enabled or disabled")
 @click.argument('container_name', required=False)
 def autorestart(container_name):
--- a/show/main.py
+++ b/show/main.py
@@ -2845,6 +2845,5 @@ def autorestart(container_name):
             body.append([name, container_feature_table[name]['auto_restart']])
     click.echo(tabulate(body, header))
 
-
 if __name__ == '__main__':
     cli()
--- a/show/main.py
+++ b/show/main.py
@@ -2814,7 +2814,7 @@ def container()
     """Show container feature"""
     pass
 
-@container.command('feature')
+@container.group('feature')
 def feature()
     pass
 
--- a/show/main.py
+++ b/show/main.py
@@ -2807,27 +2807,31 @@ def features():
     click.echo(tabulate(body, header))
 
 #
-# 'container' group
+# 'container' group (show container ...)
 #
-@cli.group(cls=AliasedGroup, default_if_no_args=False)
+@cli.group(name='container', invoke_without_command=False)
 def container()
     """Show container"""
     pass
 #
-# 'feature' group
+# 'feature' group (show container feature ...)
 #
-@container.group(cls=AliasedGroup, default_if_no_args=False)
+@container.group(name='feature', invoke_without_command=False)
 def feature()
     """Show container feature"""
     pass
 
-# show whether the auto-restart feature for containers or a specific container
-# is enabled or disabled.
+#
+# 'autorestart' subcommand (show container feature autorestart)
+#
 
 @feature.command('autorestart')
 @click.argument('container_name', required=False)
 def autorestart(container_name):
-    """Show container feature autorestart"""
+    """show whether the auto-restart feature for containers or a specific container
+       is enabled or disabled.
+    """
+
     config_db = ConfigDBConnector()
     config_db.connect()
     header = ['ContainerName', 'Status']
--- a/show/main.py
+++ b/show/main.py
@@ -2816,7 +2816,7 @@ def container()
 #
 # 'feature' group
 #
-@container.group('feature')
+@container.group(cls=AliasedGroup, default_if_no_args=False)
 def feature()
     """Show container feature"""
     pass
--- a/show/main.py
+++ b/show/main.py
@@ -2810,14 +2810,14 @@ def features():
 # 'container' group (show container ...)
 #
 @cli.group(name='container', invoke_without_command=False)
-def container()
+def container():
     """Show container"""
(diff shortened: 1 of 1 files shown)


show/main.py
```diff
@@ -2806,5 +2806,26 @@ def features():
         body.append([key, status_data[key]['status']])
     click.echo(tabulate(body, header))
 
+# show whether the auto-restart feature for containers or a specific container
+# is enabled or disabled.
+
+@cli.command('autorestart')
```
jleveque: Since we will have multiple container features including autorestart, this command should be nested in a group (or a couple groups), maybe `show container feature autorestart`. This way we can add other things under both the "container" and "feature" groups in the future.

show/main.py
```diff
@@ -2806,5 +2806,26 @@ def features():
...
+@click.option('-c', '--container-name', required=False)
```
jleveque: No need for a flag. You can just add an optional parameter, which is more in line with our other SONiC commands. If the optional parameter is present (e.g., `show container feature autorestart swss`), limit your output to only that container. If it is not present (e.g., `show container feature autorestart`), display the status of all containers. Take a look through this file, there are many examples of this style.

show/main.py
yozhao101: Please help me review it.

show/main.py
```diff
@@ -2806,5 +2806,41 @@ def features():
         body.append([key, status_data[key]['status']])
     click.echo(tabulate(body, header))
 
+#
+# 'container' group
+#
+@cli.group(cls=AliasedGroup, default_if_no_args=False)
```
jleveque: There is no need for this group to be aliased. Try using this definition instead:

```
@cli.group(name='container', invoke_without_command=False)
```

show/main.py
```diff
@@ -2806,5 +2806,41 @@ def features():
         body.append([key, status_data[key]['status']])
     click.echo(tabulate(body, header))
 
+#
+# 'container' group
```
jleveque: Provide example after. E.g.,

`# 'container' group (show container ...)`

This helps the reader understand the nesting better.

show/main.py
```diff
@@ -2806,5 +2806,41 @@ def features():
...
+#
+@cli.group(cls=AliasedGroup, default_if_no_args=False)
+def container()
+    """Show container"""
+    pass
+#
+# 'feature' group
+#
+@container.group(cls=AliasedGroup, default_if_no_args=False)
```
jleveque: There is no need for this group to be aliased. Try using this definition instead:

```
@container.group(name='feature', invoke_without_command=False)
```

show/main.py
```diff
@@ -2806,5 +2806,41 @@ def features():
...
+    """Show container feature"""
+    pass
+
+# show whether the auto-restart feature for containers or a specific container
+# is enabled or disabled.
+
+@feature.command('autorestart')
+@click.argument('container_name', required=False)
+def autorestart(container_name):
+    """Show container feature autorestart"""
```
jleveque: This docstring gets used as the help statement for the command. Please make it a brief explanation of what the command does.

show/main.py
yozhao101: Fixed.

show/main.py
yozhao101: Added this example.

show/main.py
yozhao101: Fixed.

show/main.py
yozhao101: Added an explanation.

show/main.py
```diff
@@ -2806,5 +2806,44 @@ def features():
...
+
+#
+# 'autorestart' subcommand (show container feature autorestart)
+#
+
+@feature.command('autorestart')
+@click.argument('container_name', required=False)
+def autorestart(container_name):
+    """show whether the auto-restart feature for containers or a specific container
+       is enabled or disabled.
```
jleveque: Good, but a bit too wordy. Try, `Show whether the auto-restart feature for container(s) is enabled or disabled`

Also, if the docstring is longer than 45 characters, it will get truncated. However, I discovered recently that one can work around this by defining the parameter `short_help` instead of using a docstring (see https://github.com/pallets/click/issues/486). I think we should probably just move over to this method everywhere. Can you please try that method instead? Example:

```
@feature.command('autorestart', short_help="Show whether the auto-restart feature for container(s) is enabled or disabled")
```

yozhao101: Will try the second method.

show/main.py
```diff
@@ -2810,27 +2810,24 @@ def features():
...
+def feature():
     """Show container feature"""
     pass
 
 #
 # 'autorestart' subcommand (show container feature autorestart)
 #
 
-@feature.command('autorestart')
+@feature.command('autorestart', short_help="Show whether the auto-restart feature for contain(s) is enabled or disabled")
```
jleveque: Typo: "contain**er**(s):

show/main.py
yozhao101: I tested the second method on DUT and It did work!

show/main.py
yozhao101: Fixed.

show/main.py
```diff
@@ -2806,5 +2806,40 @@ def features():
...
+
+#
+# 'autorestart' subcommand (show container feature autorestart)
+#
+@feature.command('autorestart', short_help="Show whether the auto-restart feature for container(s) is enabled or disabled")
+@click.argument('container_name', required=False)
+def autorestart(container_name):
+    config_db = ConfigDBConnector()
+    config_db.connect()
+    header = ['ContainerName', 'Status']
```
jleveque: It's OK to add a space in the column header, i.e., `Container Name`

yozhao101: Will update.

yozhao101: Fixed.

PR #799, opened 2020-02-04
[MultiDB] refactor docker warm reboot codes

dzhangalibaba:
* use swsssdk lib API instead of redis-cli


dzhangalibaba:
@qiluo-msft

PR #801, opened 2020-02-05
[config] Add subcommand to configure the status of auto-restart feature for each container

yozhao101:
- What I did
Since we introduced the auto-restart feature for each container, we need add a config subcommand to change the current status (enabled/disabled) of auto-restart feature for a specific container.

- How I did it
We define a function named autorestart to change the status of a container. This function will accept two parameters which are the container name and new status. If one of parameters is not specified, this function will show the information and then exit.

- How to verify it
We can run the command **config container feature autorestart container_name autorestart_status**.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -2468,5 +2468,41 @@ def feature_status(name, state):
 
     config_db.mod_entry('FEATURE', name, {'status': state})
 
+#
+# 'container' group ('config container ...')
+#
+@config.group(name='container', invoke_with_command=False)
+def container():
+    """Config container"""
+    pass
+
+#
+# 'feature' group ('config container feature ...')
+#
+@container.group(name='feature', invoke_with_command=False)
+def feature():
+    """Config container feature"""
+    pass
+
+#
+# 'autorestart' subcommand ('config container feature autorestart ...')
+#
+@feature.command(name='autorestart', short_help="Configure the status of autorestart feature for specific container")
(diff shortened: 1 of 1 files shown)


yozhao101:
@jleveque Sure.

config/main.py
```diff
@@ -2468,5 +2468,41 @@ def feature_status(name, state):
 
     config_db.mod_entry('FEATURE', name, {'status': state})
 
+#
+# 'container' group ('config container ...')
+#
+@config.group(name='container', invoke_with_command=False)
+def container():
+    """Config container"""
```
jleveque: Change help string to `Modify configuration of containers`

Also update this in Command reference PR.

config/main.py
```diff
@@ -2468,5 +2468,41 @@ def feature_status(name, state):
...
+    pass
+
+#
+# 'feature' group ('config container feature ...')
+#
+@container.group(name='feature', invoke_with_command=False)
+def feature():
+    """Config container feature"""
```
jleveque: Change help string to `Modify configuration of container features`

Also update this in Command reference PR.

config/main.py
yozhao101: Fixed.

config/main.py
yozhao101: Fixed.

jleveque:
Retest this please

PR #812, opened 2020-02-13
[config/main.py] Modify reload() function to load configuration from init_cfg.json implicitly

yozhao101:
**- What I did**
I modified the function reload() such that it will load the configuration files config_db.json and init_cfg.json into Config_DB.

**- How I did it**

**- How to verify it**

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -546,7 +546,7 @@ def reload(filename, yes, load_sysinfo):
         command = "{} -H -k {} --write-to-db".format(SONIC_CFGGEN_PATH, cfg_hwsku)
         run_command(command, display_cmd=True)
 
-    if os.path.isfile('/etc/sonic/init_cfg.json'):
+    if os.path.isfile(INIT_CFG_FILE):
         command = "{} -j {} -j {} --write-to-db".format(SONIC_CFGGEN_PATH, INIT_CFG_FILE, filename)
     else:
         command = "{} -j {} --write-to-db".format(SONIC_CFGGEN_PATH, filename)
--- a/config/main.py
+++ b/config/main.py
@@ -27,6 +27,8 @@ SONIC_CFGGEN_PATH = '/usr/local/bin/sonic-cfggen'
 SYSLOG_IDENTIFIER = "config"
 VLAN_SUB_INTERFACE_SEPARATOR = '.'
 
+INIT_CFG_FILE = '/etc/sonic/init_cfg.json'
+
 # ========================== Syslog wrappers ==========================
 
 def log_debug(msg):
@@ -506,20 +508,18 @@ def save(filename):
 @config.command()
 @click.option('-y', '--yes', is_flag=True)
 @click.argument('filename', default='/etc/sonic/config_db.json', type=click.Path(exists=True))
-@click.argument('init_cfg_file', default='/etc/sonic/init_cfg.json', type=click.Path(exists=True))
-def load(filename, init_cfg_file, yes):
-    """Import previous saved config DB dump file and init_cfg.json."""
+def load(filename, yes):
+    """Import a previous saved config DB dump file."""
     if not yes:
-        click.confirm('Load config from the files %s and %s?' % (filename, init_cfg_file), abort=True)
-    command = "{} -j {} -j {} --write-to-db".format(SONIC_CFGGEN_PATH, filename, init_cfg_file)
+        click.confirm('Load config from the files %s' % filename, abort=True)
+    command = "{} -j {} --write-to-db".format(SONIC_CFGGEN_PATH, filename)
     run_command(command, display_cmd=True)
 
 @config.command()
 @click.option('-y', '--yes', is_flag=True)
 @click.option('-l', '--load-sysinfo', is_flag=True, help='load system default information (mac, portmap etc) first.')
 @click.argument('filename', default='/etc/sonic/config_db.json', type=click.Path(exists=True))
-@click.argument('init_cfg_file', default='/etc/sonic/init_cfg.json', type=click.Path(exists=True))
-def reload(filename, init_cfg_file, yes, load_sysinfo):
+def reload(filename, yes, load_sysinfo):
     """Clear current configuration and import previous saved config DB dump file and init_cfg.json."""
     if not yes:
         click.confirm('Clear current config and reload config from the files %s and %s?' % (filename, init_cfg_file), abort=True)
@@ -546,7 +546,10 @@ def reload(filename, init_cfg_file, yes, load_sysinfo):
         command = "{} -H -k {} --write-to-db".format(SONIC_CFGGEN_PATH, cfg_hwsku)
         run_command(command, display_cmd=True)
 
-    command = "{} -j {} -j {} --write-to-db".format(SONIC_CFGGEN_PATH, filename, init_cfg_file)
+    if os.path.isfile('/etc/sonic/init_cfg.json'):
+        command = "{} -j {} -j {} --write-to-db".format(SONIC_CFGGEN_PATH, filename, INIT_CFG_FILE)
+
+    command = "{} -j {} --write-to-db".format(SONIC_CFGGEN_PATH, filename)
     run_command(command, display_cmd=True)
     client.set(config_db.INIT_INDICATOR, 1)
 
--- a/config/main.py
+++ b/config/main.py
@@ -522,7 +522,7 @@ def load(filename, yes):
 def reload(filename, yes, load_sysinfo):
     """Clear current configuration and import a previous saved config DB dump file."""
     if not yes:
-        click.confirm('Clear current config and reload config from the file?' % filename, abort=True)
+        click.confirm('Clear current config and reload config from the file %s?' % filename, abort=True)
 
     log_info("'reload' executing...")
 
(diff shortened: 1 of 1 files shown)


config/main.py
```diff
@@ -506,21 +506,23 @@ def save(filename):
...
-    """Import a previous saved config DB dump file."""
+@click.argument('init_cfg_file', default='/etc/sonic/init_cfg.json', type=click.Path(exists=True))
+def load(filename, init_cfg_file, yes):
+    """Import previous saved config DB dump file and init_cfg.json."""
     if not yes:
-        click.confirm('Load config from the file %s?' % filename, abort=True)
-    command = "{} -j {} --write-to-db".format(SONIC_CFGGEN_PATH, filename)
+        click.confirm('Load config from the files %s and %s?' % (filename, init_cfg_file), abort=True)
+    command = "{} -j {} -j {} --write-to-db".format(SONIC_CFGGEN_PATH, filename, init_cfg_file)
     run_command(command, display_cmd=True)
```
jleveque: On second thought, I think we should not load init_cfg.json during `config load`, as this command is meant to load the contents of config_db.json over the existing running config. Since the init_cfg.json should not change until the device is upgraded, there should be no need to load init_cfg.json here. Please undo these changes.

config/main.py
```diff
@@ -506,21 +506,23 @@ def save(filename):
...
 
 @config.command()
 @click.option('-y', '--yes', is_flag=True)
 @click.option('-l', '--load-sysinfo', is_flag=True, help='load system default information (mac, portmap etc) first.')
 @click.argument('filename', default='/etc/sonic/config_db.json', type=click.Path(exists=True))
-def reload(filename, yes, load_sysinfo):
-    """Clear current configuration and import a previous saved config DB dump file."""
+@click.argument('init_cfg_file', default='/etc/sonic/init_cfg.json', type=click.Path(exists=True))
```
jleveque: I don't think we need to make this file a user-configurable option. Initial configuration is built into the image and should be fixed. Just make it a constant.

config/main.py
```diff
@@ -544,7 +546,7 @@ def reload(filename, yes, load_sysinfo):
         command = "{} -H -k {} --write-to-db".format(SONIC_CFGGEN_PATH, cfg_hwsku)
         run_command(command, display_cmd=True)
 
-    command = "{} -j {} --write-to-db".format(SONIC_CFGGEN_PATH, filename)
+    command = "{} -j {} -j {} --write-to-db".format(SONIC_CFGGEN_PATH, filename, init_cfg_file)
```
jleveque: Here, you should check for the presence of the init_cfg.json file and run the appropriate command based on whether or not it exists. See how it's done in the load_minigraph command on lines 609-612.

config/main.py
yozhao101: Undo the changes.

config/main.py
yozhao101: Define a constant string to represent the path of file init_cfg.json and please review it.

config/main.py
```diff
@@ -518,9 +520,9 @@ def load(filename, yes):
 @click.option('-l', '--load-sysinfo', is_flag=True, help='load system default information (mac, portmap etc) first.')
 @click.argument('filename', default='/etc/sonic/config_db.json', type=click.Path(exists=True))
 def reload(filename, yes, load_sysinfo):
-    """Clear current configuration and import a previous saved config DB dump file."""
+    """Clear current configuration and import previous saved config DB dump file and init_cfg.json."""
```
jleveque: I don't think it's necessary to change this line to mention init_cfg.json. It may confuse users. It should happen silently. Please change back.

config/main.py
```diff
@@ -518,9 +520,9 @@ def load(filename, yes):
...
     if not yes:
-        click.confirm('Clear current config and reload config from the file %s?' % filename, abort=True)
+        click.confirm('Clear current config and reload config from the files %s and %s?' % (filename, init_cfg_file), abort=True)
```
jleveque: I don't think it's necessary to change this line to mention init_cfg.json. It may confuse users. It should happen silently. Please change back.

config/main.py
yozhao101: Added a if statement to decide whether the file init_cfg.json exists or not.

config/main.py
```diff
@@ -544,6 +546,9 @@ def reload(filename, yes, load_sysinfo):
         command = "{} -H -k {} --write-to-db".format(SONIC_CFGGEN_PATH, cfg_hwsku)
         run_command(command, display_cmd=True)
 
+    if os.path.isfile('/etc/sonic/init_cfg.json'):
+        command = "{} -j {} -j {} --write-to-db".format(SONIC_CFGGEN_PATH, filename, INIT_CFG_FILE)
+
     command = "{} -j {} --write-to-db".format(SONIC_CFGGEN_PATH, filename)
```
jleveque: This line should be part of an 'else' clause, Otherwise you're loading 'filename' twice if init_cfg.json exists.

yozhao101: My bad and add an else statement.

config/main.py
yozhao101: Fixed.

config/main.py
yozhao101: Fixed.

config/main.py
```diff
@@ -544,7 +546,11 @@ def reload(filename, yes, load_sysinfo):
         command = "{} -H -k {} --write-to-db".format(SONIC_CFGGEN_PATH, cfg_hwsku)
         run_command(command, display_cmd=True)
 
-    command = "{} -j {} --write-to-db".format(SONIC_CFGGEN_PATH, filename)
+    if os.path.isfile('/etc/sonic/init_cfg.json'):
+        command = "{} -j {} -j {} --write-to-db".format(SONIC_CFGGEN_PATH, filename, INIT_CFG_FILE)
```
jleveque: Have you confirmed this is the appropriate order of the "-j" arguments to first load the init_cfg.json file followed by the config_db.json file? It looks like sonic-cfggen loads the files in the order they are supplied: https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-config-engine/sonic-cfggen#L258

yozhao101: For the statement of line 550, sonic-cfggen will first load config_db.json and then init_cfg.json, right?

jleveque: That's what it looks like to me. Have you confirmed this? We want to first load init_cf.json, then load config_db.json, overwriting any values which exist in both. Please confirm.

yozhao101: I think sonic-cfggen should first load init_cfg.json and then config_db.json.

jleveque: If it loads in the order provided, it would first load config_db.json and then init_cfg.json. Please test and verify the behavior and report back.

jleveque: You just made a commit to change the order, but you haven't reported back your results. Can you please confirm the order by testing?

yozhao101: Joe, I tested on DUT using command **_sonic-cfggen -j /etc/sonic/init_cfg.json -j /etc/sonic/config_db.json --write-to-db_**. In init_cfg.json file, the initial values for "auto_restart" and "high_mem_alert" are "disabled". In config_db.json, the initial values for these two keys are "enabled". After running this command and then using command **_show container feature autorestart_**, the key for "auto_restart" will be overrode to "enabled".

jleveque: OK, thanks. Can you also confirm that if you swap the order of the files back to `sonic-cfggen -j /etc/sonic/config_db.json -j /etc/sonic/init_cfg.json --write-to-db` that the results will be wrong, as we assume?

yozhao101: Yes, I tested on DUT using the command **_sonic-cfggen -j /etc/sonic/config_db.json -j /etc/sonic/init_cfg.json `--write-to-db_**. Initially the key "auto_restart“ for database in init_cfg.json is set to "disabled" and this key for database is set to "enabled" in config_db.json. After running this command and the command` show container feature autorestart`, the value for auto_restart will be overrode to "disabled".

jleveque: Great. Thanks! I believe we saw at least one place in sonic-buildimage where the order is reversed. When you begin work in there, please check for and fix any instances of this.

yozhao101: Yes. for sonic-cfggen command, I found the following locations which we can review and may add init_cfg.json. Some locations already load init_cfg.json.  files/image_config/updategraph/updategraph ===> line 10; files/scripts/configdb-load.sh ===> line 10;  files/build_templates/docker_image_ctl.j2 ===> line 111; files/image_config/config-setup/config-setup ===> line 110 and 190;  I will check and change the order.

config/main.py
```diff
@@ -544,7 +546,11 @@ def reload(filename, yes, load_sysinfo):
         command = "{} -H -k {} --write-to-db".format(SONIC_CFGGEN_PATH, cfg_hwsku)
         run_command(command, display_cmd=True)
 
-    command = "{} -j {} --write-to-db".format(SONIC_CFGGEN_PATH, filename)
+    if os.path.isfile('/etc/sonic/init_cfg.json'):
```
jleveque: Since you created a constant, please use it here, also:

```
    if os.path.isfile(INIT_CFG_FILE):
```

yozhao101: Fixed.

PR #819, opened 2020-02-23
[reboot] make sure the reboot happens even if platform reboot failed

yxieca:
**- What I did**

Platform reboot tool could fail if the platform driver didn't load properly. In this case, a reboot is required to recover the device. We need to make sure the reboot tool is robust.

### DIFF
--- a/scripts/reboot
+++ b/scripts/reboot
@@ -145,8 +145,7 @@ if [ -x ${DEVPATH}/${PLATFORM}/${PLAT_REBOOT} ]; then
     # Sleep 1 second before calling /sbin/reboot to accommodate situation #1 above.
     sleep 1
 
-    VERBOSE=yes debug "Reboot failed!" >&2
+    VERBOSE=yes debug "Platform specific reboot failed! Issuing OS-level reboot ..." >&2
 fi
 
 exec /sbin/reboot $@
-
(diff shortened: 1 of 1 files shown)


scripts/reboot
```diff
@@ -134,11 +134,19 @@ fi
...
+    #
+    # 1. The vendor platform reboot returned after scheduled the platform specific reboot.
+    #    This is a vendor platform reboot code bug but it happens.
+    # 2. The vendor platform reboot failed. e.g. due to platform driver didn't load properly.
+    #
+    # As result if the reboot script reaches here. We should make the reboot happen.
+    # Sleep 1 second before calling /sbin/reboot to accommodate situation #1 above.
+    sleep 1
+
+    VERBOSE=yes debug "Reboot failed!" >&2
```
jleveque: Suggest to change message to "Platform-specific reboot failed!"

scripts/reboot
```diff
@@ -134,11 +134,19 @@ fi
...
 fi
 
-# Should never reach here
-VERBOSE=yes debug "Reboot failed!" >&2
-exit 1
+exec /sbin/reboot $@
```
jleveque: Suggest to also echo another message to the user before issuing /sbin/reboot (and possibly also log it), like "Issuing OS-level reboot..."

scripts/reboot
yxieca: Good idea! Done.

scripts/reboot
yxieca: This is a common path when lack of platform reboot tool. Added suggested wording above.

jleveque: Understood, but if the user sees the message "Platform-specific reboot failed!" from above, I think it would also be nice to inform them that the system is still going to reboot. Plus, if this is logged, we can determine if for some reason a platform-specific reboot is not attempted when we believe it should have benn.

yxieca: I initially thought the common successful path should be quiet. But since we logged issuing platform specific reboot anyways. I guess it doesn't hurt to make it symmetric.

PR #823, opened 2020-02-27
[fast-reboot] Fix kernel panic related to irq after fast-reboot

byu343:
**- What I did**
After the change in master branch updating SAI from 3.5.3.1m-25 to 3.7.3.2, we always found kernel panic after running fast-reboot command in testing SONiC with traffic. In the up path of fast-reboot, we can find warning messages like "unhandled irq 16 error" before kernel panic, which implies that some components are not properly closed in the down path. 

This fix will unload certain kernel modules by stopping opennsl before fast-reboot, which is suggested by BRCM. 

Note that another part of the fix is to add 'ExecStop=-/etc/init.d/opennsl-modules stop' to sonic-buildimage:platform/broadcom/saibcm-modules/systemd/opennsl-modules.service, which will be included in another pull request.

**- How I did it**

**- How to verify it**

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -529,6 +529,13 @@ then
   systemctl stop nps-modules-`uname -r`.service || debug "Ignore stopping nps service error $?"
 fi
 
+# Stop opennsl modules for Broadcom platform
+if [[ "$sonic_asic_type" = 'broadcom' ]];
+then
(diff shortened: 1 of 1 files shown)


lguohan:
retest this please

lguohan:
retest this please

PR #829, opened 2020-03-03
ssd_mitigation_changes

rkdevi27:
**- What I did**
I have added the ssd mitigation changes which collects ssd logs for S6100 with 3IE3 ssd.
**- How I did it**
I have added the following changes:
       Collect ssd logs before fast-reboot.
       Added after removing a sonic image and before installing an image.
       Block the fast-reboot for 2 minutes from the time of sonic image installation.
**- How to verify it**
         I have tested the functionality and attaching the logs here.

[console_logs.txt](https://github.com/Azure/sonic-utilities/files/4282056/console_logs.txt)


yxieca:
I don't feel safe with the approach in the combination of changes.

If sonic image installation needs to do fstrim, then it should not return until the fstrim is done. Otherwise, if every installation calls fstrim and prevent fast-reboot for 2 minutes, then the upgrade pipeline will never complete, because fast-reboot will always fail.

yxieca:
We need to discuss this change before moving forward.

rkdevi27:
I have reverted the fstrim changes and 2 minutes delay changes and updated the PR.

rkdevi27:
Retest default please

rkdevi27:
Retest default please

scripts/fast-reboot
```diff
@@ -16,6 +16,7 @@ ASSISTANT_SCRIPT="/usr/bin/neighbor_advertiser"
 DEVPATH="/usr/share/sonic/device"
 PLATFORM=$(sonic-cfggen -H -v DEVICE_METADATA.localhost.platform)
 PLATFORM_PLUGIN="${REBOOT_TYPE}_plugin"
+PLATFORM_LOG_SSD_HEALTH="platform_log_ssd_health"
```
sujinmkang, line 19: as I suggested in https://github.com/Azure/sonic-buildimage/pull/4214, can we make the platform_log_ssd_health as a common utility to check the ssd status (maybe we can do specifically for innodisk's "3IE3" type or generically for all ssd devices)?

rkdevi27: Sujin,
As suggested, I have made log_ssd_health as generic and it has been called with ExecStartPre in fstrim.service. Removed the check for 3IE3 and now it will collect smartctl output before fast-reboot and fstrim service start in all platforms. I have verified the fix in few Dell Platforms(S6100 and Z9264F)

rkdevi27:
Retest default please

sujinmkang:
retest default please

sujinmkang:
Retest this please

PR #833, opened 2020-03-10
[db_migrator]Do DB migration for buffer pool size change on Mellanox platform

keboliu:
**- What I did**
Do DB migration for buffer pool size change introduced by new SDK, this is specifically for the Mellanox platform.

**- How I did it**
Advance the DB version to a new version 1_0_3, and add Mellanox specific DB migration functions.
Inside the function to migrate the buffer pool size to the latest values.

**- How to verify it**
test sonic to sonic upgrade from between various version and different topology.


scripts/db_migrator.py
```diff
@@ -96,6 +100,89 @@ def migrate_interface_table(self):
...
+            or pool_size_in_db_list == spc2_t0_default_value or pool_size_in_db_list == spc2_t1_default_value:
+            # Generate buffer configuration from latest template
+            tmp_json_file = "/tmp/mlnx_buffers_all.json"
+            device_buffer_template_path = os.path.join("/usr/share/sonic/device/", platform, hwsku, 'buffers.json.j2')
+            command = "sonic-cfggen -d -t {} >{}".format(device_buffer_template_path, tmp_json_file)
+            try:
+                subprocess.check_call(command, shell=True)
+            except:
+                log_error("failed to generate new mlnx buffer configuration, skip migration")
+                return False
```
yxieca: Can you make code robust enough so that failure is not happening?

Having a failure in migration path would leave config_db not migrated, and likely will cause orchagent to crash in the future. right?

scripts/db_migrator.py
```diff
@@ -96,6 +100,89 @@ def migrate_interface_table(self):
...
+        pools_in_db = buffer_pool_conf.keys()
+
+        # Buffer pool numbers is different with default, don't need migrate
+        if len(pools_in_db) != len(buffer_pools):
+            return True
+
+        # If some buffer pool is not default ones, don't need migrate
+        for buffer_pool in buffer_pools:
+            if buffer_pool not in pools_in_db:
+                return True
```
yxieca: Should this be continue instead of return? Do you need to migrate other buffer pool(s) that might still have default values?

scripts/db_migrator.py
```diff
@@ -96,6 +100,89 @@ def migrate_interface_table(self):
...
+            try:
+                with open(tmp_json_file) as tmp_buffer_file:
+                    buffer_json = json.load(tmp_buffer_file)
+                    for pool in buffer_pools:
+                        self.configDB.set_entry('BUFFER_POOL', pool, buffer_json['BUFFER_POOL'][pool])
+                    log_info("Successfully migrate mlnx buffer pool size to the latest.")
+                    return True
+            except IOError:
+                log_error("failed to open tmp mlnx buffer json configuration file, skip migration")
+                return False
```
yxieca: Same comment here. Can you make it not fail?

yxieca:
Can this change survive following scenarios:

1. Newly installed sonic image boots up, no minigraph, no config_db.json.
2. 0 size config_db.json (might be json parsing error in this case)
3. empty config_db.json file (contents is '{}')

db_migration towards config_db.json could be too late? migration is invoked after database has loaded the config_db.json into memory.

scripts/db_migrator.py
keboliu: deprecate the original way to generate a tmp buffer config json file with new configurations, instead, use predefined new buffer pool configuration dict,  this will reduce the possible fail case during handling the tmp files.

scripts/db_migrator.py
keboliu: the idea is that if the user changed the configuration,  and here we migrate it arbitrarily, then we are changing the user's configuration w/o notice, IMO the proper way here should elaborate the issue in release notes(we are going to add this issue to 201911 version release notes when we have a GA version) about this issue, and give a guideline to the user for the handle.

scripts/db_migrator.py
keboliu: as above comments

keboliu:
> Can this change survive following scenarios:
> 
> 1. Newly installed sonic image boots up, no minigraph, no config_db.json.
> 2. 0 size config_db.json (might be json parsing error in this case)
> 3. empty config_db.json file (contents is '{}')
> 
> db_migration towards config_db.json could be too late? migration is invoked after database has loaded the config_db.json into memory.

enhanced the logic to handle more corner cases. the cases you mentioned have been covered.
per my test and code reading, it's true that db_migrator only update the Redis DB, and will keep config_db.json untouched, shall we add a step to dump configuration to config_db.json as the last step of db migration?

liat-grozovik:
retest this please

keboliu:
retest this please

yxieca:
> per my test and code reading, it's true that db_migrator only update the Redis DB, and will keep config_db.json untouched, shall we add a step to dump configuration to config_db.json as the last step of db migration?

I don't think we need to save in-memory db back to config_db.json. Because we can always migrate and migrate is one step of load config_db.json. If the migration failed, we definitely don't want to write back. right?

As such, I think you should leave /etc/sonic/config_db.sjon alone. So that it is consistent by itself. And an updated migration code could handle it in the future.

prsunny:
> > per my test and code reading, it's true that db_migrator only update the Redis DB, and will keep config_db.json untouched, shall we add a step to dump configuration to config_db.json as the last step of db migration?
> 
> I don't think we need to save in-memory db back to config_db.json. Because we can always migrate and migrate is one step of load config_db.json. If the migration failed, we definitely don't want to write back. right?
> 
> As such, I think you should leave /etc/sonic/config_db.sjon alone. So that it is consistent by itself. And an updated migration code could handle it in the future.

Agree to Ying.

keboliu:
> > per my test and code reading, it's true that db_migrator only update the Redis DB, and will keep config_db.json untouched, shall we add a step to dump configuration to config_db.json as the last step of db migration?
> 
> I don't think we need to save in-memory db back to config_db.json. Because we can always migrate and migrate is one step of load config_db.json. If the migration failed, we definitely don't want to write back. right?
> 
> As such, I think you should leave /etc/sonic/config_db.sjon alone. So that it is consistent by itself. And an updated migration code could handle it in the future.

@yxieca @prsunny  then we are ok,  I will not do further change to save migrated DB to config_db.json.   Thanks.

scripts/db_migrator.py
```diff
@@ -96,6 +100,108 @@ def migrate_interface_table(self):
...
+                        return False
+                    if 'BUFFER_POOL' in config_db.keys():
+                        buffer_pool_conf = config_db['BUFFER_POOL']
+                    else:
+                        # No buffer pool size configuration exist, no need to migration
+                        log_info("No buffer pool configuration, no need to migrate")
+                        return True
+            except IOError:
+                log_error("failed to open config_db.json file, skip migration")
+                return False
```
yxieca: I think you should remove the if block and only keep the else block. In other words, only migrate the in memory database.

keboliu: Removed. 
In the very beginning, the idea was to get info from config_db.json if it exists, since eventually it will be loaded to the in-memory database, so we can just get data from in-memory DB.

And also added special handling for 3800 platforms, since it has unique buffer pool configuration compare to other Spectrum 2 platform: https://github.com/Azure/sonic-buildimage/pull/4239

PR #838, opened 2020-03-17
Fixes bug for PFCWD feature parameters

shlomibitton:
- What I did
The feature allows setting 'detection_time', 'restoration_time' and 'pollling_interval' PFCWD to an interface. The 'pollling_interval' must be lower than 'detection_time' and 'restoration_time'.
The fix is checking if there is a lower value of 'detection_time' or 'restoration_time' than the 'pollling_interval' value entered by the user in config DB, if yes exit with error code 1.

- How I did it
Checking the config DB for interfaces PFCWD values.

- How to verify it
Try adding 'pollling_interval' greater than one of the values of PFCWD interfaces.

- Previous command output (if the output of a command-line utility has changed)
No Output.

- New command output (if the output of a command-line utility has changed)
unable to use polling_interval = #ms, value is bigger or equal to the minimum in PFCWD table

### DIFF
--- a/pfcwd/main.py
+++ b/pfcwd/main.py
@@ -204,12 +204,12 @@ def interval(poll_interval):
             restoration_time_entry_value = int(configdb.get_entry(CONFIG_DB_PFC_WD_TABLE_NAME, entry).get('restoration_time'))
             if ((detection_time_entry_value != None) and (detection_time_entry_value < entry_min)):
                 entry_min = detection_time_entry_value
-                entry_min_str = "detection_time"
+                entry_min_str = "detection time"
             if ((restoration_time_entry_value != None) and (restoration_time_entry_value < entry_min)):
                 entry_min = restoration_time_entry_value
-                entry_min_str = "restoration_time"
+                entry_min_str = "restoration time"
         if entry_min < poll_interval:
-            print >> sys.stderr, "unable to use polling_interval = {}ms, value is bigger than one of the configured {} values, please choose a smaller polling_interval".format(poll_interval,entry_min_str)
+            print >> sys.stderr, "unable to use polling interval = {}ms, value is bigger than one of the configured {} values, please choose a smaller polling_interval".format(poll_interval,entry_min_str)
             exit(1)
         
         pfcwd_info['POLL_INTERVAL'] = poll_interval
--- a/pfcwd/main.py
+++ b/pfcwd/main.py
@@ -3,6 +3,7 @@
 import click
 import swsssdk
 import os
+import sys
 from tabulate import tabulate
 from natsort import natsorted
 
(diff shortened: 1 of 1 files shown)


shlomibitton:
@volodymyrsamotiy @liat-grozovik 
please review

pfcwd/main.py
```diff
@@ -194,9 +194,28 @@ def interval(poll_interval):
...
+                res_str = "restoration_time"
+        if entry_min < poll_interval:
+            poll_interval = entry_min - 1
+            print "polling_interval is greater than {}, using polling_interval = {}ms".format(res_str,poll_interval)
+        
         pfcwd_info['POLL_INTERVAL'] = poll_interval
-
-    configdb.mod_entry(CONFIG_DB_PFC_WD_TABLE_NAME, "GLOBAL", pfcwd_info)
+        if(poll_interval < 100):
+            print "unable to use polling_interval = {}ms, value is less than 100".format(poll_interval)
```
lguohan: should go to stderr, and also return error code 1

pfcwd/main.py
```diff
@@ -194,9 +194,28 @@ def interval(poll_interval):
...
+            restoration_time_entry_value = int(configdb.get_entry(CONFIG_DB_PFC_WD_TABLE_NAME, entry).get('restoration_time'))
+            if ((detection_time_entry_value != None) and (detection_time_entry_value < entry_min)):
+                entry_min = detection_time_entry_value
+                res_str = "detection_time"
+            if ((restoration_time_entry_value != None) and (restoration_time_entry_value < entry_min)):
+                entry_min = restoration_time_entry_value
+                res_str = "restoration_time"
+        if entry_min < poll_interval:
+            poll_interval = entry_min - 1
+            print "polling_interval is greater than {}, using polling_interval = {}ms".format(res_str,poll_interval)
```
lguohan: output should go to stderr, and also return error code 1

shlomibitton: @lguohan 
at this point we are using a different **valid** value for the polling interval.
you think we should not allow it? if the input is not valid just exit without using an alternative value?

neethajohn: It's better to exit at this point and allow the user to decide the value

pfcwd/main.py
neethajohn: This code can be removed if you are exiting above.

neethajohn: This code can be removed if you are exiting when polling interval ends up being less

pfcwd/main.py
```diff
@@ -194,9 +195,23 @@ def interval(poll_interval):
...
+        for entry in pfcwd_table:
+            if("Ethernet" not in entry):
+                continue
+            detection_time_entry_value = int(configdb.get_entry(CONFIG_DB_PFC_WD_TABLE_NAME, entry).get('detection_time'))
+            restoration_time_entry_value = int(configdb.get_entry(CONFIG_DB_PFC_WD_TABLE_NAME, entry).get('restoration_time'))
+            if ((detection_time_entry_value != None) and (detection_time_entry_value < entry_min)):
+                entry_min = detection_time_entry_value
+            if ((restoration_time_entry_value != None) and (restoration_time_entry_value < entry_min)):
+                entry_min = restoration_time_entry_value
+        if entry_min <= poll_interval:
```
neethajohn: The condition should be just '<'. A user will never be able to configure the upper bound value which is 3000 because of "=" check. Also most of the current device configs have the same value for detection, restoration and poll interval. We will be restricting that config as well

pfcwd/main.py
```diff
@@ -194,9 +195,23 @@ def interval(poll_interval):
...
+            print >> sys.stderr, "unable to use polling_interval = {}ms, value is bigger or equal to the minimum in PFCWD table".format(poll_interval)
```
neethajohn: Can you make this msg more meaningful? just saying 'minimum' is hard for the user to figure out the cause of the failure

shlomibitton: @neethajohn 
please review and comment if this is clear enough

pfcwd/main.py
```diff
@@ -204,10 +204,12 @@ def interval(poll_interval):
             restoration_time_entry_value = int(configdb.get_entry(CONFIG_DB_PFC_WD_TABLE_NAME, entry).get('restoration_time'))
             if ((detection_time_entry_value != None) and (detection_time_entry_value < entry_min)):
                 entry_min = detection_time_entry_value
+                entry_min_str = "detection_time"
```
neethajohn: 'detection time' is preferred in the user msg as opposed to 'detection_time'

pfcwd/main.py
```diff
@@ -204,10 +204,12 @@ def interval(poll_interval):
...
             if ((restoration_time_entry_value != None) and (restoration_time_entry_value < entry_min)):
                 entry_min = restoration_time_entry_value
-        if entry_min <= poll_interval:
-            print >> sys.stderr, "unable to use polling_interval = {}ms, value is bigger or equal to the minimum in PFCWD table".format(poll_interval)
+                entry_min_str = "restoration_time"
```
neethajohn: same as above

shlomibitton: @neethajohn
please review and approve the changes.

pfcwd/main.py
shlomibitton: @neethajohn
please review and approve the changes.

pfcwd/main.py
shlomibitton: @lguohan 
please review

PR #846, opened 2020-03-21
explicitly specify command with underscores

lguohan:
Starting click 7.0. The default behavior of a command with under
scores will be replace with dashes.

this is to address the above default behavior change, so that
the command remains the same.

more details can be found:

https://github.com/pallets/click/issues/1123

**- What I did**

**- How I did it**

**- How to verify it**
tested for under current click and click 7.0

under buster
```
admin@vlab-01:/$ sudo config load
load              load_mgmt_config  load_minigraph    
admin@vlab-01:/$ sudo config load
```

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -641,7 +641,7 @@ def reload(filename, yes, load_sysinfo):
     log_info("'reload' restarting services...")
     _restart_services()
 
-@config.command()
+@config.command("load_mgmt_config")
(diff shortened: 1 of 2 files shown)


jleveque:
@lguohan: I'm not sure if you want to handle other CLI utilities in this PR also, but I know for a fact `sonic_installer` has many subcommands which contain underscores.

lguohan:
@jleveque , i agree with you.

PR #856, opened 2020-03-24
[multi-asic]: Update reload of systemd services to support multi-asic platforms

SuvarnaMeenakshi:
**- What I did**
For multi-asic platform, there can be multiple instances of some systemd services. currently config reload CLI stops and restarts only single instance of systemd service. 
Modified stop/reset failed-status and restart system services function to perform action on multiple instances of systemd services.

**- How I did it**
1. Read plaform/asic.conf file to get the number of ASICs for a platform.
2. Get the services which are multiple instances from generated_services.conf
3. For single ASIC platform, continue to stop and restart services as it is done today.
4. For multiple ASIC platform, stop and restart each instance for multi-instance service. stop and restart of single instance will be done the same way for single and multi-asic platforms.

**- How to verify it**
Bring up a single ASIC VS image.
Load config_db.json/minigraph.xml
Update config_db.json or load a new config_db.json.
Do a "config reload".
Verify that all required services are stopped and restarted.
Verify all services are up.
Verify if the configuration/interfaces of previous configuration is completely removed and new configuration is applied.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -97,7 +97,7 @@ def execute_systemctl(list_of_services, action):
                     click.echo("Executing {} of service {}@{}...".format(action, service, inst))
                     run_command("systemctl {} {}@{}.service".format(action, service, inst))
                 except SystemExit as e:
-                    log_error("Failed to execute {} of service {}@{} with error {}".format(action, service, e))
+                    log_error("Failed to execute {} of service {}@{} with error {}".format(action, service, inst, e))
                     raise
 
 def run_command(command, display_cmd=False, ignore_error=False):
--- a/config/main.py
+++ b/config/main.py
@@ -439,13 +439,13 @@ def _get_num_asic():
     platform = _get_platform()
     asic_conf_file = os.path.join('/usr/share/sonic/device/', platform, ASIC_CONF_FILENAME)
     if not os.path.isfile(asic_conf_file):
-        num_asic=1
+        num_asic = 1
     else:
         with open(asic_conf_file) as conf_file:
             for line in conf_file:
-                line_info=line.split('=')
+                line_info = line.split('=')
                 if line_info[0].lower() == "num_asic":
-                     num_asic=int(line_info[1])
+                     num_asic = int(line_info[1])
     return num_asic
 
 def _get_sonic_generated_services():
--- a/config/main.py
+++ b/config/main.py
@@ -445,7 +445,7 @@ def _get_num_asic():
             for line in conf_file:
                 line_info = line.split('=')
                 if line_info[0].lower() == "num_asic":
-                     num_asic = int(line_info[1])
+                    num_asic = int(line_info[1])
     return num_asic
 
 def _get_sonic_generated_services():
--- a/config/main.py
+++ b/config/main.py
@@ -77,38 +77,28 @@ except KeyError, TypeError:
 # Execute action on list of systemd services
 def execute_systemctl(list_of_services, action):
     num_asic = _get_num_asic()
-    generated_services_list, generated_multi_instance_services = _get_sonic_generated_services()
+    generated_services_list, generated_multi_instance_services = _get_sonic_generated_services(num_asic)
     if ((generated_services_list == []) and
         (generated_multi_instance_services == [])):
         log_error("Failed to get generated services")
         return
 
     for service in list_of_services:
-        if (service + '.service' in generated_services_list): 
+        if (service + '.service' in generated_services_list):
             try:
-                click.echo("Executing {} action on service {}...".format(action, service))
+                click.echo("Executing {} of service {}...".format(action, service))
                 run_command("systemctl {} {}".format(action, service))
             except SystemExit as e:
-                log_error("Failed to execute {} for service {} with error {}".format(action, service, e))
+                log_error("Failed to execute {} of service {} with error {}".format(action, service, e))
                 raise
-        elif (service + '.service' in generated_multi_instance_services):
-            if num_asic == 1:
+        if (service + '.service' in generated_multi_instance_services):
+            for inst in range(num_asic):
                 try:
-                    click.echo("Executing {} action on service {}...".format(action, service))
-                    run_command("systemctl {} {}".format(action, service))
+                    click.echo("Executing {} of service {}@{}...".format(action, service, inst))
+                    run_command("systemctl {} {}@{}.service".format(action, service, inst))
                 except SystemExit as e:
-                    log_error("Failed to execute {} for service {} with error {}".format(action, service, e))
+                    log_error("Failed to execute {} of service {}@{} with error {}".format(action, service, e))
                     raise
-            else:
-                for inst in range(num_asic):
-                    try:
-                        click.echo("Executing {} action on service {}@{}...".format(action, service, inst))
-                        run_command("systemctl {} {}@{}.service".format(action, service, inst))
-                    except SystemExit as e:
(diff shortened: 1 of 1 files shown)


config/main.py
```diff
@@ -490,24 +534,37 @@ def _restart_services():
...
     for service in services_to_restart:
-        if service + '.service' not in generated_services_list:
-            continue
-        try:
-            click.echo("Restarting service {} ...".format(service))
-            run_command("systemctl restart {}".format(service))
-        except SystemExit as e:
-            log_error("Restart {} failed with error {}".format(service, e))
-            raise
+        if ((service + '.service' in generated_services_list) or
```
lguohan: duplicated code as above. please use function to reduce the duplicated code.

lguohan: looks like the code is duplicated three times

SuvarnaMeenakshi: updated code as per comment.

config/main.py
```diff
@@ -419,25 +472,7 @@ def _stop_services():
...
         'nat'
     ]
-    generated_services_list = _get_sonic_generated_services()
-
-    if generated_services_list is None:
-        log_error("Failed to get generated services")
-        return
-
-    if asic_type == 'mellanox' and 'pmon' in services_to_stop:
-        services_to_stop.remove('pmon')
```
lguohan: i do not see this code in your new code? is it missing?

config/main.py
```diff
@@ -395,14 +429,33 @@ def _get_platform():
...
+    generated_multi_instance_services = []
     with open(SONIC_GENERATED_SERVICE_PATH) as generated_service_file:
         for line in generated_service_file:
-            generated_services_list.append(line.rstrip('\n'))
-    return None if not generated_services_list else generated_services_list
+            if '@' in line:
+                 line=line.replace('@', '')
+                 generated_multi_instance_services.append(line.rstrip('\n'))
+            else:
+                generated_services_list.append(line.rstrip('\n'))
```
lguohan: the indent does not match line 455.

config/main.py
```diff
@@ -419,25 +472,7 @@ def _stop_services():
...
-        if service + '.service' not in generated_services_list:
-            continue
-        try:
-            click.echo("Stopping service {} ...".format(service))
-            run_command("systemctl stop {}".format(service))
-
-        except SystemExit as e:
-            log_error("Stopping {} failed with error {}".format(service, e))
-            raise
+    execute_systemctl(services_to_stop, SYSTEMCTL_STOP)
```
lguohan: please check the return results and act accordingly.

config/main.py
```diff
@@ -27,9 +27,14 @@
 SONIC_CFGGEN_PATH = '/usr/local/bin/sonic-cfggen'
 SYSLOG_IDENTIFIER = "config"
 VLAN_SUB_INTERFACE_SEPARATOR = '.'
+ASIC_CONF_FILENAME = 'asic.conf'
 
 INIT_CFG_FILE = '/etc/sonic/init_cfg.json'
 
+SYSTEMCTL_STOP="stop"
```
lguohan: -> SYSTEMCTL_ACTION_STOP

config/main.py
```diff
@@ -69,6 +74,35 @@ def log_error(msg):
...
+def execute_systemctl(list_of_services, action):
+    num_asic = _get_num_asic()
+    generated_services_list, generated_multi_instance_services = _get_sonic_generated_services()
+    if ((generated_services_list == []) and
+        (generated_multi_instance_services == [])):
+        log_error("Failed to get generated services")
+        return
+
+    for service in list_of_services:
+        if ((service + '.service' in generated_services_list) or
```
lguohan: it is more clear to have such logic

if service in generate_serivces_list:
   ...
else if service in multi_instance_service:
   if num_asic == 1:
   else:
else:
    log_warning("service xxx no in the list")

config/main.py
```diff
@@ -395,14 +429,33 @@ def _get_platform():
...
+    platform = _get_platform()
+    asic_conf_file = os.path.join('/usr/share/sonic/device/', platform, ASIC_CONF_FILENAME)
+    if not os.path.isfile(asic_conf_file):
+        num_asic=1
+    else:
+        with open(asic_conf_file) as conf_file:
+            for line in conf_file:
+                line_info=line.split('=')
+                if line_info[0].lower() == "num_asic":
+                     print line_info[1]
```
lguohan: why do you print here.

config/main.py
SuvarnaMeenakshi: Update as per comment.

config/main.py
SuvarnaMeenakshi: execute_systemctl() - this function prints error message or handles exception itself, should we further get return value and act?

config/main.py
SuvarnaMeenakshi: Updated as per comment.

config/main.py
SuvarnaMeenakshi: Updated as per comment.

config/main.py
SuvarnaMeenakshi: Added while testing, removed this.

config/main.py
SuvarnaMeenakshi: added back, accidentally removed in previous commit.

config/main.py
```diff
@@ -395,14 +435,32 @@ def _get_platform():
                 return tokens[1].strip()
     return ''
 
+def _get_num_asic():
+    platform = _get_platform()
+    asic_conf_file = os.path.join('/usr/share/sonic/device/', platform, ASIC_CONF_FILENAME)
+    if not os.path.isfile(asic_conf_file):
+        num_asic=1
```
lguohan: the indent is not consistent, should have space around the '='

config/main.py
```diff
@@ -395,14 +435,32 @@ def _get_platform():
...
+    else:
+        with open(asic_conf_file) as conf_file:
+            for line in conf_file:
+                line_info=line.split('=')
```
lguohan: the indent is not consistent, should have space around the '='

config/main.py
```diff
@@ -395,14 +435,32 @@ def _get_platform():
...
+                if line_info[0].lower() == "num_asic":
+                     num_asic=int(line_info[1])
```
lguohan: the indent is not consistent, should have space around the '='

config/main.py
```diff
@@ -69,6 +74,41 @@ def log_error(msg):
...
+
+    for service in list_of_services:
+        if (service + '.service' in generated_services_list): 
+            try:
+                click.echo("Executing {} action on service {}...".format(action, service))
+                run_command("systemctl {} {}".format(action, service))
+            except SystemExit as e:
+                log_error("Failed to execute {} for service {} with error {}".format(action, service, e))
+                raise
+        elif (service + '.service' in generated_multi_instance_services):
```
judyjoseph: Do we have a case where a particular service is there in both in generated_services_list and generated_multi_instance_services list ? like may be database where we have a global database service and unit based database services for namespaces.

config/main.py
SuvarnaMeenakshi: Fixed

config/main.py
SuvarnaMeenakshi: Fixed

config/main.py
SuvarnaMeenakshi: Fixed

config/main.py
SuvarnaMeenakshi: currently there is no such service that is both global and multi-instance apart from database. But I have updated the logic to take care of this type of service, in case something like that gets added in future.

PR #857, opened 2020-03-24
[config] Implement a process level lock

praveen-li:
**- What I did**
Implement a config process level lock.

**- How I did it**
Changes:
1.) Implement a class, which uses hsetnx for lock.
2.) lock is expired within timeout period or will be released by owner.
3.) After -y prompt, lock is reacquired, because timer could have expired,
    before user enters yes.
Signed-off-by: Praveen Chaudhary pchaudhary@linkedin.com

#### Why lock is acquired at process level, not just when we interact with DB.
command such as port breakout interect with DB 3 times:
1.) Read Config DB,
---Processing---
2.) Delete Ports
---Processing---
3.) Add Ports.
Here we need to take the lock before step one and then have to release the lock after step three. If we take the lock only while interacting with DB then we have to take lock three times in above situation and that also doesn't solve the problem, because some process can update the DB in between Step 1 and step 2.

Similar command can be written in future with config validation in picture.

**- How to verify it**

Testing:

Test 1: acquire a lock with command config. Run another config command in between to make sure lock is not acquired. Then say yes on the prompt of first command before timer expires.

Expectation: Second command should not acquire lock, first command should re-extend the timer.

Terminal 1:
```
admin@lnos-x1-a-fab01:~$ sudo config save
:::Lock Acquired:::
Existing file will be overwritten, continue? [y/N]: y
:::Lock Timer Extended:::
:::Lock Released:::
```
Terminal 2:
```
admin@lnos-x1-a-fab01:~$ sudo config save
:::Can not acuire lock, Abort:::
Lock PID: 10877 and self.pid:10890
```

Test 2: run first command, let the timer expires, run second command, then immediately say yes on the prompt of first command.

Expectation: Second command should acquire lock, first command should not be acquire lock after prompt. 

Terminal 1:
```
admin@lnos-x1-a-fab01:~$ sudo config save
:::Lock Acquired:::
Existing file will be overwritten, continue? [y/N]: y
:::Can not acquire lock LOCK PID: 10903 and self.pid:10877:::
:::Lock PID: 10903 and self.pid:10877:::
```

Terminal 2:
```
admin@lnos-x1-a-fab01:~$ sudo config load
:::Lock Acquired:::
Load config from the file /etc/sonic/config_db.json? [y/N]: N
Aborted!
:::Lock PID: None and self.pid: 10903:::
```

Test 3: run first command, let the timer expires, then say yes on the prompt of first command to reacquire the lock.

```
admin@lnos-x1-a-fab01:~$ sudo config save
:::Lock Acquired:::
Existing file will be overwritten, continue? [y/N]: y
:::Lock Reacquired:::
:::Lock Released:::
admin@lnos-x1-a-fab01:~$
```

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -553,18 +553,17 @@ class ConfigDbLock():
             # expire call.
             if self.client.hsetnx(self.lockName, "PID", self.pid):
                 self.client.expire(self.lockName, self.timeout)
-                print(":::Lock Acquired:::")
             # if lock exists but expire timer not running, run expire time and
             # abort.
             elif not self.client.ttl(self.lockName):
                 self.client.expire(self.lockName, self.timeout)
-                print(":::Can not acquire lock, Reset Timer & Abort:::");
+                click.echo(":::Can not acquire lock, Reset Timer & Abort:::");
                 sys.exit(1)
             else:
-                print(":::Can not acquire lock, Abort:::");
+                click.echo(":::Can not acquire lock, Abort:::");
                 sys.exit(1)
         except Exception as e:
-            print(":::Exception: {}:::".format(e))
+            click.echo(":::Exception: {}:::".format(e))
             sys.exit(1)
         return
 
@@ -573,7 +572,6 @@ class ConfigDbLock():
             # Try to set lock first
             if self.client.hsetnx(self.lockName, "PID", self.pid):
                 self.client.expire(self.lockName, self.timeout)
-                print(":::Lock Reacquired:::")
             # if lock exists, check who owns it
             else:
                 p = self.client.pipeline(True)
@@ -582,18 +580,17 @@ class ConfigDbLock():
                 # if current process holding then extend the timer
                 if p.hget(self.lockName, "PID") == str(self.pid):
                     self.client.expire(self.lockName, self.timeout)
-                    print(":::Lock Timer Extended:::");
                     p.unwatch()
                     return
                 else:
                     # some other process is holding the lock.
-                    print(":::Can not acquire lock LOCK PID: {} and self.pid:{}:::".\
+                    click.echo(":::Can not acquire lock LOCK PID: {} and self.pid:{}:::".\
                         format(p.hget(self.lockName, "PID"), self.pid))
                     p.unwatch()
                     sys.exit(1)
 
         except Exception as e:
-            print(":::Exception: {}:::".format(e))
+            click.echo(":::Exception: {}:::".format(e))
             sys.exit(1)
         return
 
@@ -607,15 +604,14 @@ class ConfigDbLock():
                 p.multi()
                 p.delete(self.lockName)
                 p.execute()
-                print(":::Lock Released:::");
                 return
             else:
                 # some other process s holding the lock.
-                print(":::Lock PID: {} and self.pid:{}:::".\
+                click.echo(":::Lock PID: {} and self.pid:{}:::".\
                     format(p.hget(self.lockName, "PID"), self.pid))
             p.unwatch()
         except Exception as e:
-            print("Exception: {}".format(e))
+            click.echo("Exception: {}".format(e))
         return
 
     def __del__(self):
--- a/config/main.py
+++ b/config/main.py
@@ -622,7 +622,7 @@ class ConfigDbLock():
         return
 # end of class configdblock
 
-cdblock = ConfigDbLock()
+configdb_lock = ConfigDbLock()
 # This is our main entrypoint - the main 'config' command
 @click.group(context_settings=CONTEXT_SETTINGS)
 def config():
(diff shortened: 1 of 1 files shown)


config/main.py
```diff
@@ -519,7 +519,100 @@ def is_ipaddress(val):
         return False
     return True
 
+# class for locking entire config process
+class configdblock():
```
jleveque, line 523: Suggest rename class using Pascal case, which will also help readability: `ConfigDbLock`

praveen-li: Sure, I will change the name, it shd be pascal case only right not camelcase. Also plz let me know if any other changes are needed before I raise 2nd commit. Thx.

praveen-li:
@jleveque : kindly let me know if any more review comments. This feature will be pretty useful to avoid multiple writes to config DB through config command. Thanks a lot.

config/main.py
```diff
@@ -519,7 +519,100 @@ def is_ipaddress(val):
...
+            db_kwargs = dict()
+            configdb = ConfigDBConnector(**db_kwargs)
+            configdb.connect()
+
+            self.client = configdb.get_redis_client('CONFIG_DB')
+            # Set lock and expire time. Process may get killed b/w set lock and
+            # expire call.
+            if self.client.hsetnx(self.lockName, "PID", self.pid):
+                self.client.expire(self.lockName, self.timeout)
+                print(":::Lock Acquired:::")
```
jleveque, line 545: Replace `print` statements with `click.echo` statements for consistency.

Also, I feel like we don't need to output success messages, only failure messages. Do you agree?

praveen-li: @jleveque : sure I will replace print statements with click.echo.

Network operation folks may have few scripts which uses "config <>" commands. I expect few failures there, once this change is merged (in case, there are few scripts running simultaneously). So yeah, failure case information is useful.

But I can convert successful cases to debug messages, they were more or less for testing. I agree with you.

praveen-li: resolved.

config/main.py
```diff
@@ -519,7 +519,99 @@ def is_ipaddress(val):
...
+        except Exception as e:
+            click.echo(":::Exception: {}:::".format(e))
+        return
+
+    def __del__(self):
+        self._releaseLock()
+        return
+# end of class configdblock
+
+cdblock = ConfigDbLock()
```
jleveque, line 614: Can you also rename this variable to make it more readable? "cdblock" can be mistaken for "CD Block". Maybe `configdb_lock`?

praveen-li: I think, let`s use\force camel case for variables in this file. I did that already inside the class, so If you do not prefer otherwise I will use below var name.
 
```
configDbLock = ConfigDbLock()
```

jleveque: For our style in SONiC, we tend to use Pascal case for class names and lower_with_underscores for variable names.

praveen-li: @jleveque , cool thx for sharing, I will stick to it then.

praveen-li:
retest this please

praveen-li:
@jleveque 

Hi Joe, I see a successful test run here:
https://sonic-jenkins.westus2.cloudapp.azure.com/job/common/job/sonic-utilities-build-pr/1236/

Latest test run has few failures, I feel they are random, because I have not changed logic of code. Also I see these kind of failures for other PRs as well. Let me know your view.

jleveque:
Retest this please

praveen-li:
@jleveque 

Hi Joe, 

The tests which are falling in the last test run were actually passing in previous test. So I think it's somewhat random failures right now. Let me know if anything pending from my side on this PR. Thanks a lot for the review.

jleveque:
Retest this please

praveen-li:
@jleveque , seems all test passed. Let me know if you can review and merge it. Thx a lot.

config/main.py
```diff
@@ -530,7 +530,99 @@ def is_ipaddress(val):
...
+                else:
+                    # some other process is holding the lock.
+                    click.echo(":::Unable to reacquire lock (lock PID: {}, self.pid: {}):::".\
+                        format(p.hget(self.lockName, "PID"), self.pid))
+                    p.unwatch()
+                    sys.exit(1)
+
+        except Exception as e:
+            click.echo(":::Exception: {}:::".format(e))
+            sys.exit(1)
```
jleveque, line 594: This block isn't indented correctly. Can you please fix the indentation and remove the extra blank line above it?

praveen-li: I can remove the extra line[591], but I think rest is indented correctly, i.e. 593, 594 are 4 space + from 592. 
Note line 595 is in function block.

jleveque: Thanks. I believe the extra blank line caused an optical illusion; it looked like the `except` statement was aligned with the function definition.

praveen-li:
@jleveque  I think, we looks good approving it :). Thx.

PR #877, opened 2020-04-10
[config] Add support for multi-ASIC devices

judyjoseph:
**- What I did**
Changes to the following config commands and scripts for Multi-ASIC devices.

> **config load** : supports config load across all namespaces 
> **config save** : supports config save across all namespaces 
> **config reload** : supports reload of all services across namespaces.

**- How I did it**

The following are the changes introduced in this PR

**The changes done here are  based on the namespace support added in https://github.com/Azure/sonic-py-swsssdk/pull/63**

(a) Added API's to 
      > is_multi_asic()
      > get_all_namespaces()
      > validate_namespace(namespace)

(b) config load/save/reload commands :  the action is taken on the linux host + if we are in multi-asic platform repeat the action in the namespaces. The filename argument is optional as before. If not given these are the default files used ( below eg: for a multi-ASIC device with 3 ASIC's )
/etc/sonic/config_db.json     --> for global linux host database
/etc/sonic/config_db0.json   --> for "asic0" namespace
/etc/sonic/config_db1.json   --> for "asic1" namespace
/etc/sonic/config_db2.json   --> for "asic2" namespace

If the filename argument needs to be provided it needs to be given as a string of config file names separated by comma

(c) Updates to the db_migrator.py script. 

**- How to verify it**
Verified on multi-ASIC platform

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -12,9 +12,7 @@ import netifaces
 
 import sonic_device_util
 import ipaddress
-from swsssdk import ConfigDBConnector
-from swsssdk import SonicV2Connector
-from swsssdk import SonicDBConfig
+from swsssdk import ConfigDBConnector, SonicV2Connector, SonicDBConfig
 from minigraph import parse_device_desc_xml
 
 import aaa
@@ -147,7 +145,7 @@ def get_all_namespaces():
             elif metadata['localhost']['sub_role'] == 'BackEnd':
                 back_ns.append(namespace)
 
-    return {'front_ns':front_ns, 'back_ns':back_ns}
+    return {'front_ns': front_ns, 'back_ns': back_ns}
 
 # Validate whether a given namespace name is valid in the device.
 def validate_namespace(namespace):
@@ -595,7 +593,9 @@ config.add_command(nat.nat)
                 expose_value=False, prompt='Existing files will be overwritten, continue?')
 @click.argument('filename', required=False)
 def save(filename):
-    """Export current config DB to a file on disk."""
+    """Export current config DB to a file on disk.\n
+       [FILENAME] : Filenames separated by comma with no spaces in between.
+    """
     num_asic = _get_num_asic()
     cfg_files = []
 
@@ -608,7 +608,7 @@ def save(filename):
         cfg_files = filename.split(',')
 
         if len(cfg_files) != num_cfg_file:
-            click.echo("Input {} config file[s] separated by comma if needed".format(num_cfg_file))
+            click.echo("Input {} config file[s] separated by comma for multiple files ".format(num_cfg_file))
             return
 
     """In case of multi-asic mode we have additional config_db{NS}.json files for
@@ -633,7 +633,6 @@ def save(filename):
         if namespace is None:
             command = "{} -d --print-data > {}".format(SONIC_CFGGEN_PATH, file)
         else:
-            namespace = "{}{}".format(NAMESPACE_PREFIX, inst)
             command = "{} -n {} -d --print-data > {}".format(SONIC_CFGGEN_PATH, namespace, file)
 
         run_command(command, display_cmd=True)
@@ -642,7 +641,9 @@ def save(filename):
 @click.option('-y', '--yes', is_flag=True)
 @click.argument('filename', required=False)
 def load(filename, yes):
-    """Import a previous saved config DB dump file."""
+    """Import a previous saved config DB dump file.
+       [FILENAME] : Filenames separated by comma with no spaces in between.
+    """
     if filename is None:
         message = 'Load config from the default config file[s] ?'
     else:
@@ -663,7 +664,7 @@ def load(filename, yes):
         cfg_files = filename.split(',')
 
         if len(cfg_files) != num_cfg_file:
-            click.echo("Input {} config file[s] separated by comma if needed".format(num_cfg_file))
+            click.echo("Input {} config file[s] separated by comma for multiple files ".format(num_cfg_file))
             return
 
     """In case of multi-asic mode we have additional config_db{NS}.json files for
@@ -693,7 +694,6 @@ def load(filename, yes):
         if namespace is None:
             command = "{} -j {} --write-to-db".format(SONIC_CFGGEN_PATH, file)
         else:
-            namespace = "{}{}".format(NAMESPACE_PREFIX, inst)
             command = "{} -n {} -j {} --write-to-db".format(SONIC_CFGGEN_PATH, namespace, file)
 
         run_command(command, display_cmd=True)
@@ -704,7 +704,9 @@ def load(filename, yes):
 @click.option('-l', '--load-sysinfo', is_flag=True, help='load system default information (mac, portmap etc) first.')
 @click.argument('filename', required=False)
 def reload(filename, yes, load_sysinfo):
-    """Clear current configuration and import a previous saved config DB dump file."""
+    """Clear current configuration and import a previous saved config DB dump file.
+       [FILENAME] : Filenames separated by comma with no spaces in between.
+    """
     if filename is None:
         message = 'Clear current config and reload config from the default config file[s] ?'
     else:
@@ -727,7 +729,7 @@ def reload(filename, yes, load_sysinfo):
         cfg_files = filename.split(',')
 
         if len(cfg_files) != num_cfg_file:
-            click.echo("Input {} config file[s] separated by comma if needed".format(num_cfg_file))
+            click.echo("Input {} config file[s] separated by comma for multiple files ".format(num_cfg_file))
             return
 
     if load_sysinfo:
--- a/config/main.py
+++ b/config/main.py
@@ -705,7 +705,7 @@ def load(filename, yes):
 @click.argument('filename', required=False)
 def reload(filename, yes, load_sysinfo):
     """Clear current configuration and import a previous saved config DB dump file.
-       <filename> : Names of configuration file(s) to reload, separated by comma with no spaces in between
+       <filename> : Names of configuration file(s) to load, separated by comma with no spaces in between
     """
     if filename is None:
         message = 'Clear current config and reload config from the default config file(s) ?'
--- a/config/main.py
+++ b/config/main.py
@@ -27,9 +27,8 @@ SONIC_CFGGEN_PATH = '/usr/local/bin/sonic-cfggen'
 SYSLOG_IDENTIFIER = "config"
 VLAN_SUB_INTERFACE_SEPARATOR = '.'
 ASIC_CONF_FILENAME = 'asic.conf'
-PLATFORM_ROOT_PATH = '/usr/share/sonic/device'
 DEFAULT_CONFIG_DB_FILE = '/etc/sonic/config_db.json'
-NS_PREFIX = 'asic'
+NAMESPACE_PREFIX = 'asic'
 
 INIT_CFG_FILE = '/etc/sonic/init_cfg.json'
 
@@ -137,7 +136,7 @@ def get_all_namespaces():
 
     if is_multi_asic():
         for asic in range(num_asics):
-            namespace = "{}{}".format(NS_PREFIX, asic)
+            namespace = "{}{}".format(NAMESPACE_PREFIX, asic)
             config_db = ConfigDBConnector(use_unix_socket_path=True, namespace=namespace)
             config_db.connect()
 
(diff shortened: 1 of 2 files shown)


config/main.py
```diff
@@ -27,6 +27,9 @@
 SYSLOG_IDENTIFIER = "config"
 VLAN_SUB_INTERFACE_SEPARATOR = '.'
 ASIC_CONF_FILENAME = 'asic.conf'
+PLATFORM_ROOT_PATH = '/usr/share/sonic/device'
```
jleveque, line 30: This constant appears to be unused. Remove?

config/main.py
```diff
@@ -27,6 +27,9 @@
...
+DEFAULT_CONFIG_DB_FILE = '/etc/sonic/config_db.json'
+NS_PREFIX = 'asic'
```
jleveque, line 32: Suggest to rename to `NAMESPACE_PREFIX` for readability

config/main.py
```diff
@@ -542,32 +587,96 @@ def config():
...
 @config.command()
 @click.option('-y', '--yes', is_flag=True)
 @click.option('-l', '--load-sysinfo', is_flag=True, help='load system default information (mac, portmap etc) first.')
-@click.argument('filename', default='/etc/sonic/config_db.json', type=click.Path(exists=True))
+@click.argument('filename', default=DEFAULT_CONFIG_DB_FILE, type=click.Path(exists=True))
 def reload(filename, yes, load_sysinfo):
     """Clear current configuration and import a previous saved config DB dump file."""
     if not yes:
-        click.confirm('Clear current config and reload config from the file %s?' % filename, abort=True)
+        click.confirm('Clear current config and reload config ?', abort=True)
```
jleveque, line 679: So, for multi-ASIC platforms, there is no way to specify a file to load from because there are multiple files involved, is this correct? If so, maybe we should consider fixing the file for single-ASIC platforms as well, so that the behavior is consistent for both.

judyjoseph: Yes there are multiple files that need to be specified, might be difficult for user also to give as input. Are you referring to removing the filename option itself from the command ?  if we remove won't it affect backward compatibility with some scripts using this command may be ?

jleveque: Yes, we would need to check. Most of the time we just load from the default conifg_db.json file.

config/main.py
judyjoseph: Removed.

config/main.py
judyjoseph: Renamed to NAMESPACE_PREFIX

config/main.py
judyjoseph: I am finding couple of testcases which uses filename option to specify the config_db.json file, there could be other use cases as well. 
Hence I have updated the implementation so that the user can input specific config_db.json files even for namespace. This will make sure the behavior is consistent for all platforms. Please take a look.

scripts/db_migrator.py
```diff
@@ -291,14 +293,22 @@ def main():
                         required = False,
                         help = 'the unix socket that the desired database listens on',
                         default = None )
+        parser.add_argument('-n',
+                        dest='namespace',
+                        metavar='namespace details',
```
lguohan, line 298: >namespace details [](start = 33, length = 17)

asic namespace

scripts/db_migrator.py
```diff
@@ -291,14 +293,22 @@ def main():
...
+                        type = str,
+                        required = False,
+                        help = 'The namespace whose DB instance we need to connect',
```
lguohan, line 301: >namespace [](start = 36, length = 9)

asic namespace

config/main.py
```diff
@@ -542,32 +586,132 @@ def config():
 config.add_command(nat.nat)
 
 @config.command()
+@click.option('-n', '--namespace', help='Namespace name')
 @click.option('-y', '--yes', is_flag=True, callback=_abort_if_false,
-                expose_value=False, prompt='Existing file will be overwritten, continue?')
-@click.argument('filename', default='/etc/sonic/config_db.json', type=click.Path())
-def save(filename):
+                expose_value=False, prompt='Existing files will be overwritten, continue?')
+@click.argument('filename',nargs=-1)
```
lguohan, line 592: keep a space after comma

config/main.py
```diff
@@ -542,32 +586,132 @@ def config():
 config.add_command(nat.nat)
 
 @config.command()
+@click.option('-n', '--namespace', help='Namespace name')
```
lguohan, line 589: Asic namespace

config/main.py
```diff
@@ -542,32 +586,132 @@ def config():
...
-def save(filename):
+                expose_value=False, prompt='Existing files will be overwritten, continue?')
+@click.argument('filename',nargs=-1)
+def save(namespace, filename):
     """Export current config DB to a file on disk."""
-    command = "{} -d --print-data > {}".format(SONIC_CFGGEN_PATH, filename)
-    run_command(command, display_cmd=True)
+    if namespace:
+        if not is_multi_asic():
+            click.echo("Namespace is not significant in a Single ASIC platform")
```
lguohan, line 597: use this is more clear. "Asic namespace is not enabled on this platform!"

config/main.py
```diff
@@ -542,32 +586,132 @@ def config():
...
+            return None
+        if not validate_namespace(namespace):
+            click.echo("Invalid Namespace entered {}".format(namespace))
```
lguohan, line 600: Invalid Asic namespace {}.

config/main.py
```diff
@@ -542,32 +586,132 @@ def config():
...
+            return None
+        if not len(filename):
+            inst = namespace[len(NAMESPACE_PREFIX)]
+            cfg_file = "/etc/sonic/config_db{}.json".format(inst)
+        else:
+            cfg_file = filename[0]
```
lguohan, line 606: I do not understand the logic here. what is filename[0]?

config/main.py
```diff
@@ -542,32 +586,132 @@ def config():
...
-                expose_value=False, prompt='Existing file will be overwritten, continue?')
-@click.argument('filename', default='/etc/sonic/config_db.json', type=click.Path())
-def save(filename):
+                expose_value=False, prompt='Existing files will be overwritten, continue?')
+@click.argument('filename',nargs=-1)
+def save(namespace, filename):
     """Export current config DB to a file on disk."""
-    command = "{} -d --print-data > {}".format(SONIC_CFGGEN_PATH, filename)
-    run_command(command, display_cmd=True)
+    if namespace:
```
lguohan, line 595: is there a use case to save config db for each individual namespace? if not, then we do not need to provide this option. I feel like everytime we want to save config db, we should save all config dbs.

config/main.py
judyjoseph: Ok filename is returned as a tuple, as this is of type nargs=-1. But yes this is wrong in this case as here the tuple has only one element and I can just use cfg_file = filename.

Will fix it and update this and other similar places.

config/main.py
```diff
@@ -542,32 +586,132 @@ def config():
...
+        run_command(command, display_cmd=True)
+
+        """In case of multi-asic mode we have additional config_db{NS}.json files for
+           various namespaces created per ASIC. {NS} is the namespace index.
+        """
+        if is_multi_asic():
+            num_asic = _get_num_asic()
+            for inst in range(num_asic):
+                namespace = "{}{}".format(NAMESPACE_PREFIX, inst)
+                # Get the file from user input, else take the default file /etc/sonic/config_db{NS_id}.json
```
lguohan, line 626: if user provide input, we should make sure user provide all 7 filenames. we should not sometimes take user input, sometime use the default values.

config/main.py
lguohan: I also suggest for multi-npu case, the filename should be a a list files separated by comma as a single string.

config/main.py
```diff
@@ -542,32 +586,132 @@ def config():
...
+                command = "{} -n {} -d --print-data > {}".format(SONIC_CFGGEN_PATH, namespace, cfg_file)
+                run_command(command, display_cmd=True)
 
 @config.command()
+@click.option('-n', '--namespace', help='Namespace name')
 @click.option('-y', '--yes', is_flag=True)
-@click.argument('filename', default='/etc/sonic/config_db.json', type=click.Path(exists=True))
-def load(filename, yes):
+@click.argument('filename',nargs=-1)
+def load(filename, yes, namespace):
```
lguohan, line 639: check my comments on the save comments. i think they apply to this command as well.

config/main.py
```diff
@@ -584,26 +728,59 @@ def reload(filename, yes, load_sysinfo):
...
-    if load_sysinfo:
-        command = "{} -H -k {} --write-to-db".format(SONIC_CFGGEN_PATH, cfg_hwsku)
-        run_command(command, display_cmd=True)
 
-    if os.path.isfile(INIT_CFG_FILE):
-        command = "{} -j {} -j {} --write-to-db".format(SONIC_CFGGEN_PATH, INIT_CFG_FILE, filename)
-    else:
-        command = "{} -j {} --write-to-db".format(SONIC_CFGGEN_PATH, filename)
+    """ In Single AISC platforms we have single DB service. In multi-ASIC platforms we have a global DB
+        service running in the host + DB services running in the namespace created per ASIC.
```
lguohan, line 733: DB services running in each ASIC namespace

config/main.py
```diff
@@ -584,26 +728,59 @@ def reload(filename, yes, load_sysinfo):
...
+        if inst is -1:
+            namespace = ''
+        else:
+            namespace = "{}{}".format(NAMESPACE_PREFIX, inst)
 
-    run_command(command, display_cmd=True)
-    client.set(config_db.INIT_INDICATOR, 1)
+        # Get the config file, based on user input 
+        if len(filename) > inst+1:
+            cfg_file = filename[inst+1]
```
lguohan, line 747: check the user input first.

config/main.py
```diff
@@ -584,26 +728,59 @@ def reload(filename, yes, load_sysinfo):
...
+        if load_sysinfo:
+            command = "{} -H -k {} -n \"{}\" --write-to-db".format(SONIC_CFGGEN_PATH, cfg_hwsku, namespace)
+            run_command(command, display_cmd=True)
+
+        # For the database service running in linux host we use the file user gives as input
+        # or by default DEFAULT_CONFIG_DB_FILE. In the case of database service running in namespace,
+        # the default config_db<namespaceID>.json format is used.
+
+        if os.path.isfile(INIT_CFG_FILE):
+            command = "{} -j {} -j {} -n \"{}\" --write-to-db".format(SONIC_CFGGEN_PATH, INIT_CFG_FILE, cfg_file, namespace)
```
lguohan, line 772: why quote is needed for namespace variable, but not others?

config/main.py
```diff
@@ -584,26 +728,59 @@ def reload(filename, yes, load_sysinfo):
...
+    """ In Single AISC platforms we have single DB service. In multi-ASIC platforms we have a global DB
+        service running in the host + DB services running in the namespace created per ASIC.
+        In the below logic, we get all namespaces in this platform and add an empty namespace ''
+        denoting the current namespace which we are in ( the linux host )
+    """
+    num_asic = _get_num_asic()
+    for inst in range(-1, num_asic-1):
+        # Get the namespace name, for linux host it is '' empty namespace string.
+        if inst is -1:
+            namespace = ''
```
lguohan, line 741: for global db, you do not need to supply -n option.

config/main.py
judyjoseph: Ok, I will remove the namespace option. So the only option available to the user is "filename" which he could give as input.

config/main.py
judyjoseph: This code change was done earlier as per the initial approach for SonicV2Connector/ConfigDBConnector class .. where we had the default namespace as ''. Now that we have changed that to default None, I have updated the code likewise.

config/main.py
judyjoseph: This again  was done earlier as per the initial approach for SonicV2Connector/ConfigDBConnector class. Now I have added explicit check for namespace is None and added the -n flag.

judyjoseph:
There are some testfailures here as the code change needs latest sonic-py-swsssdk , which is still not submodule updated in sonic-buildimage. Please ignore them while reviewing.

qiluo-msft:
@lguohan Why LGTM checker is not enabled in this PR? Could you help setup?

lguohan:
image

it says enabled.

scripts/db_migrator.py
judyjoseph: updated.

scripts/db_migrator.py
judyjoseph: updated.

config/main.py
judyjoseph: Made the change to tahe the argument from user as a single string with files separated by comma.

config/main.py
judyjoseph: updated

config/main.py
judyjoseph: updated.

config/main.py
judyjoseph: Ok, now I have removed the namespace option to the commands - the user can either not provide any filenames , if he provides he needs to provide all namespacecount+1 of them

config/main.py
judyjoseph: Take care of.

config/main.py
judyjoseph: updated.

judyjoseph:
retest this please

config/main.py
```diff
@@ -14,6 +14,7 @@
 import ipaddress
 from swsssdk import ConfigDBConnector
 from swsssdk import SonicV2Connector
+from swsssdk import SonicDBConfig
```
jleveque, line 17: Suggest combining lines 15-17 into one as follows:

```
from swsssdk import ConfigDBConnector, SonicV2Connector, SonicDBConfig
```

config/main.py
```diff
@@ -114,6 +117,48 @@ def run_command(command, display_cmd=False, ignore_error=False):
...
+            config_db = ConfigDBConnector(use_unix_socket_path=True, namespace=namespace)
+            config_db.connect()
+
+            metadata = config_db.get_table('DEVICE_METADATA')
+            if metadata['localhost']['sub_role'] == 'FrontEnd':
+                front_ns.append(namespace)
+            elif metadata['localhost']['sub_role'] == 'BackEnd':
+                back_ns.append(namespace)
+
+    return {'front_ns':front_ns, 'back_ns':back_ns}
```
jleveque, line 150: Add space after colons. E.g.,

```
{'front_ns': front_ns, 'back_ns': back_ns}
```

config/main.py
judyjoseph: Updated.

config/main.py
judyjoseph: Updated.

config/main.py
```diff
@@ -543,34 +590,148 @@ def config():
...
-                expose_value=False, prompt='Existing file will be overwritten, continue?')
-@click.argument('filename', default='/etc/sonic/config_db.json', type=click.Path())
+                expose_value=False, prompt='Existing files will be overwritten, continue?')
+@click.argument('filename', required=False)
 def save(filename):
-    """Export current config DB to a file on disk."""
-    command = "{} -d --print-data > {}".format(SONIC_CFGGEN_PATH, filename)
-    run_command(command, display_cmd=True)
+    """Export current config DB to a file on disk.\n
+       [FILENAME] : Filenames separated by comma with no spaces in between.
```
jleveque, line 597: Suggest changing this line to the following:

```
<filename> : Names of configuration file(s) to load, separated by comma with no spaces in between
```

config/main.py
```diff
@@ -594,7 +594,7 @@ def config():
 @click.argument('filename', required=False)
 def save(filename):
     """Export current config DB to a file on disk.\n
-       [FILENAME] : Filenames separated by comma with no spaces in between.
+       <filename> : Names of configuration file(s) to use, separated by comma with no spaces in between
```
jleveque, line 597: s/use/save/

config/main.py
```diff
@@ -642,12 +642,12 @@ def save(filename):
 @click.argument('filename', required=False)
 def load(filename, yes):
     """Import a previous saved config DB dump file.
-       [FILENAME] : Filenames separated by comma with no spaces in between.
+       <filename> : Names of configuration file(s) to use, separated by comma with no spaces in between
```
jleveque, line 645: s/use/load/

config/main.py
```diff
@@ -705,12 +705,12 @@ def load(filename, yes):
 @click.argument('filename', required=False)
 def reload(filename, yes, load_sysinfo):
     """Clear current configuration and import a previous saved config DB dump file.
-       [FILENAME] : Filenames separated by comma with no spaces in between.
+       <filename> : Names of configuration file(s) to use, separated by comma with no spaces in between
```
jleveque, line 708: s/use/reload/

config/main.py
```diff
@@ -705,7 +705,7 @@ def load(filename, yes):
 @click.argument('filename', required=False)
 def reload(filename, yes, load_sysinfo):
     """Clear current configuration and import a previous saved config DB dump file.
-       <filename> : Names of configuration file(s) to use, separated by comma with no spaces in between
+       <filename> : Names of configuration file(s) to reload, separated by comma with no spaces in between
```
jleveque, line 708: s/reload/load/. "Reload" simply means clear the DB first; the file(s) just get loaded. "Reload" is a bit confusing terminology as it is. Sorry to nitpick.

judyjoseph: Thanks Joe, updated. yes the wording to use "reload" was confusing.

jleveque:
Looks good to me. Will wait for @lguohan to review again.

rlhui:
retest vs please

rlhui:
retest vsimage please

config/main.py
```diff
@@ -543,34 +590,148 @@ def config():
...
+        if cfg_files:
+            file = cfg_files[inst+1]
+        else:
+            if namespace is None:
+                file = DEFAULT_CONFIG_DB_FILE
+            else:
+                file = "/etc/sonic/config_db{}.json".format(inst)
+
+        # if any of the config files in linux host OR namespace is not present, return
+        if not os.path.isfile(file):
```
prsunny, line 690: This check fails the TSG step - https://github.com/Azure/SONiC/wiki/L2-Switch-mode
`The config_db file /dev/stdin doesn't exist`

judyjoseph: Raised PR to resolve this with https://github.com/Azure/sonic-utilities/pull/1115

PR #881, opened 2020-04-17
[fanshow] Add column drawer name and led status to output of show platform fan

Junchao-Mellanox:
**- What I did**

1. Add a column to display fan drawer name
2. Add a column to display fan led status

**- How I did it**

**- How to verify it**

Manual test

```
        FAN     Speed    Direction     Presence    Status          Timestamp
-----------  --------  -----------  -----------  --------  -----------------
       fan1       91%       intake      Present        OK  20200414 09:41:52
       fan2       81%       intake      Present        OK  20200414 09:41:52

```

```
  Drawer    LED          FAN     Speed    Direction     Presence    Status          Timestamp
--------  -----  -----------  --------  -----------  -----------  --------  -----------------
 drawer1  green         fan1       92%       intake      Present        OK  20200414 09:39:52
 drawer1  green         fan2       81%       intake      Present        OK  20200414 09:39:52

```

### DIFF
--- a/scripts/fanshow
+++ b/scripts/fanshow
@@ -11,7 +11,7 @@ from swsssdk import SonicV2Connector
 from natsort import natsorted
 
 
-header = ['Drawer', 'FAN', 'Speed', 'Direction', 'Presence', 'Status', 'LED', 'Timestamp']
+header = ['Drawer', 'LED', 'FAN', 'Speed', 'Direction', 'Presence', 'Status', 'Timestamp']
 
 FAN_TABLE_NAME = 'FAN_INFO'
 DRAWER_FIELD_NAME = 'drawer_name'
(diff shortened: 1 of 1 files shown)


jleveque:
I'm OK with this change, but as discussed offline, we now assign LEDs on a per-fan-drawer basis, so having the LED color in each fan row here could be a bit confusing to the user. One option would be to separate the output by fan drawer similar to the following:

```
Drawer Name: drawer1   
        LED: green
 Fan status:

   Fan Name     Speed    Direction     Presence    Status          Timestamp
-----------  --------  -----------  -----------  --------  -----------------
       fan1       92%       intake      Present        OK  20200414 09:39:52
       fan2       81%       intake      Present        OK  20200414 09:39:52
```

Thoughts?

Junchao-Mellanox:
> I'm OK with this change, but as discussed offline, we now assign LEDs on a per-fan-drawer basis, so having the LED color in each fan row here could be a bit confusing to the user. One option would be to separate the output by fan drawer similar to the following:
> 
> ```
> Drawer Name: drawer1   
>         LED: green
>  Fan status:
> 
>    Fan Name     Speed    Direction     Presence    Status          Timestamp
> -----------  --------  -----------  -----------  --------  -----------------
>        fan1       92%       intake      Present        OK  20200414 09:39:52
>        fan2       81%       intake      Present        OK  20200414 09:39:52
> ```
> 
> Thoughts?

This would be clearer for Mellanox platform. But I am not sure if it is OK for those platform which has no fan drawer. I prefer to keep current implementation.

PR #891, opened 2020-04-24
new commands to add a range of vlans and add a member to a range of vlans

anilkpan:
**- What I did**
Added new commands to be able to add a range of vlans and add a member of a range of vlans
**- How I did it**
Added the following new commands:
create a range of vlans
delete a range o vlans
add a member of range of vlans
remove a member from a range of vlans

**- How to verify it**

config vlan range add 2 50
config vlan member range add 2 50 Ethernet0
config vlan member range del 2 50 Ethernet0
config vlan range del 2 50

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1135,6 +1135,285 @@ def del_vlan_member(ctx, vid, interface_name):
     db.set_entry('VLAN', vlan_name, vlan)
     db.set_entry('VLAN_MEMBER', (vlan_name, interface_name), None)
 
+ Validate VLAN range.
+#
+def vlan_range_validate(ctx, vid1, vid2):
+    vlan1 = 'Vlan{}'.format(vid1)
+    vlan2 = 'Vlan{}'.format(vid2)
+
+    if vlan_id_is_valid(vid1) is False:
+        ctx.fail("{} is not within allowed range of 1 through 4094".format(vlan1))
+    if vlan_id_is_valid(vid2) is False:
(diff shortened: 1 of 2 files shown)


config/main.py
```diff
@@ -1333,6 +1344,369 @@ def del_vlan_dhcp_relay_destination(ctx, vid, dhcp_relay_destination_ip):
...
+    pipe = clients.pipeline()
+    for vid in range (vid1, vid2):
+        vlan = 'Vlan{}'.format(vid)
+
+        if len(db.get_entry('VLAN', vlan)) != 0:
+            if warning is True:
+                warning_vlans_list.append(vid)
+            continue
+
+        pipe.hmset('VLAN|{}'.format(vlan),  {'vlanid': vid})
```
lguohan, line 1405: need use swsssdk as abstract layer of API. using redis client api make hard to support multi-db implementation in the future.

anilkpan: @lguohan if I understood correctly, should I add a new API in swss-common/swsssdk and call the API from here instead of using redis client API directly?

ben-gale: Anil - he's asking you to look in https://github.com/Azure/sonic-swss-common for a suitable DB access function and use this. If you don't find one that meets your needs then please add it - thanks.

anilkpan:
Added new APIs for bulk update:
https://github.com/Azure/sonic-py-swsssdk/pull/82

PR #898, opened 2020-04-29
Improved route_check tool and adopt to 20191130 image.

renukamanavalan:
1) Add syslog support.
2) Enable forever periodic scan
3) Skip link local addresses
4) Skip eth0 routes
3) Adopt to 20191130 changes
    3.1) APPl-DB INTF_TABLE may not have IP address
    3.2) nexthop is never empty

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/scripts/route_check.py
+++ b/scripts/route_check.py
@@ -147,13 +147,19 @@ def get_interfaces():
     print_message(syslog.LOG_DEBUG, json.dumps({"APPL_DB_INTF": sorted(intf)}, indent=4))
     return sorted(intf)
 
-def filter_out_rt(ifname, keys):
+def filter_out_local_interfaces(keys):
     rt = []
+    local_if = set(['eth0', 'lo', 'docker0'])
+
     db = ConfigDBConnector()
     db.db_connect('APPL_DB')
-
+    
     for k in keys:
-        if (db.get_entry('ROUTE_TABLE', k)['ifname'] != ifname):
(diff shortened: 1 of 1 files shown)


scripts/route_check.py
```diff
@@ -93,103 +100,138 @@ def do_diff(t1, t2):
 def get_routes():
     db = ConfigDBConnector()
     db.db_connect('APPL_DB')
-    print_message(MODE_DEBUG, "APPL DB connected for routes")
+    print_message(syslog.LOG_DEBUG, "APPL DB connected for routes")
     keys = db.get_keys('ROUTE_TABLE')
```
prsunny, line 104: This covers the regular route case. But since we also have VNET routes (VNET_ROUTE_TABLE, VNET_ROUTE_TUNNEL_TABLE), it will show mismatch when run on switches with VNETs. Just thinking how we should handle that, or extend this script to cover it as a later commit? @lguohan

renukamanavalan: As this does not affect 20181130, we will commit this PR and take VNET routes handling in a separate PR.

renukamanavalan:
lgtm says, I fixed an issue that was there in previous code.

lguohan:
is this needed for 201811 image? does not seem to match the title of this pr

yxieca:
201811 branch is only taking bug fixes at this point.

PR #901, opened 2020-05-03
[show] Fix abbreviations for 'show ip bgp ...' commands

jleveque:
Currently, the `AliasedGroup` class in show/main.py handles finding a subcommand to run based on an abbreviation. Thus, we need to import in the bgp files and use it as the base class for the 'bgp' groups in order to allow abbreviations to work for the subcommands in these groups.


rlhui:
@jleveque, there's conflict.  please raise a separate PR for 201911 branch.

jleveque:
Opened separate PR against 201911 branch here: https://github.com/Azure/sonic-utilities/pull/909

PR #910, opened 2020-05-12
Multi-asic changes for config bgp commands and utilities.

judyjoseph:
**- What I did**

Changes to config bgp commands for Multi-ASIC devices.

**- How I did it**

The changes done here are based on the namespace support added in SonicDBConfig/ConfigDBConnector/SonicV2Connector classes in the PR  **Azure/sonic-py-swsssdk#63**
> To connect to DB in a particular namespace pass the parameter namespace=namespaceName
> To connect to the DB in the local context ( where ever we are ) don't pass any namespace input, it would be defaulted to '' (empty string ). This is the default option, so that the changes made here are backward compatible with single ASIC platforms.

**The following are the changes introduced in this PR**

(a) added a new parameter of **config_db** to the bgp neighbor helper API's. This approach is better, as now we have to do this once **config_db = ConfigDBConnector()** in the CLI handler and the config_db instance is passed down as argument to the API. 

(b) Modified the API _get_all_neighbor_ipaddresses() to accept an additional parameter **ignore_hosts** if there are any hosts to be ignored when forming the ip_addr neighbor list.

(c) config bgp commands : In multi-ASIC devices we have both internal BGP sessions between the bgp dockers running in different namespaces and EBGP sessions with external neighbors.
   > config bgp startup/shutdown all : performs "action" only on the external BGP neighbors.
   > config bgp startup/shutdown/remove neighbor : performs "action" on any internal/external BGP neighbor based on user input. User has to be careful of not giving the neighbor address of IBGP sessions.

**- How to verify it**
Verified on a multi-ASIC devices, by running the Config commands.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -167,6 +167,23 @@ def validate_namespace(namespace):
     else:
         return False
 
+"""In case of Multi-Asic platform, Each ASIC will have a host name and we call it internal hosts.
+   So we loop through the databases in different namespaces and get the hostname
+"""
+def get_all_internal_hosts():
+    internal_hosts = []
+    num_asics = _get_num_asic()
+
+    if is_multi_asic():
+        for asic in range(num_asics):
+            namespace = "{}{}".format(NAMESPACE_PREFIX, asic)
+            config_db = ConfigDBConnector(use_unix_socket_path=True, namespace=namespace)
+            config_db.connect()
+            metadata = config_db.get_table('DEVICE_METADATA')
+            internal_hosts.append(metadata['localhost']['hostname'])
+
+    return internal_hosts
+
 def interface_alias_to_name(interface_alias):
     """Return default interface name if alias name is given as argument
     """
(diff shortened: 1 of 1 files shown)


config/main.py
```diff
@@ -199,6 +199,23 @@ def validate_namespace(namespace):
     else:
         return False
 
+"""In case of Multi-Asic platform, Each ASIC will have a host name and we call it internal hosts.
+   So we loop through the databases in different namespaces and get the hostname
+"""
```
jleveque, line 204: Prefer to reserve`"""` for docstrings and use `#` for comments

config/main.py
```diff
@@ -1687,20 +1704,53 @@ def num_dumps(kdump_num_dumps):
...
-    _change_bgp_session_status(ipaddr_or_hostname, 'down', verbose)
+    namespaces = [DEFAULT_NAMESPACE]
+    found_neighbor = False
+    if is_multi_asic:
+        ns_list = get_all_namespaces()
+        namespaces = ns_list['front_ns'] + ns_list['back_ns']
+
+    """Connect to CONFIG_DB in linux host (in case of single ASIC) or CONFIG_DB in all the
+       namespaces (in case of multi ASIC) and do the sepcified "action" on the BGP neighbor(s)
+    """
```
jleveque, line 1745: Prefer to reserve`"""` for docstrings and use `#` for comments

config/main.py
```diff
@@ -1711,20 +1761,55 @@ def startup():
...
+    int_hosts = []
+
+    if is_multi_asic:
+        ns_list = get_all_namespaces()
+        namespaces = ns_list['front_ns']
+        int_hosts = get_all_internal_hosts()
+
+    """Connect to CONFIG_DB in linux host (in case of single ASIC) or CONFIG_DB in all the
+       namespaces (in case of multi ASIC) and do the sepcified "action" on the BGP neighbor(s)
+    """
```
jleveque, line 1778: Prefer to reserve`"""` for docstrings and use `#` for comments

config/main.py
```diff
@@ -1711,20 +1761,55 @@ def startup():
...
+    namespaces = [DEFAULT_NAMESPACE]
+    found_neighbor = False
+
+    if is_multi_asic:
+        ns_list = get_all_namespaces()
+        namespaces = ns_list['front_ns'] + ns_list['back_ns']
+
+    """Connect to CONFIG_DB in linux host (in case of single ASIC) or CONFIG_DB in all the
+       namespaces (in case of multi ASIC) and do the sepcified "action" on the BGP neighbor(s)
+    """
```
jleveque, line 1804: Prefer to reserve`"""` for docstrings and use `#` for comments

config/main.py
```diff
@@ -1738,8 +1823,27 @@ def remove():
...
+    namespaces = [DEFAULT_NAMESPACE]
+    removed_neighbor = False
+
+    if is_multi_asic:
+        ns_list = get_all_namespaces()
+        namespaces = ns_list['front_ns'] + ns_list['back_ns']
+
+    """Connect to CONFIG_DB in linux host (in case of single ASIC) or CONFIG_DB in all the
+       namespaces (in case of multi ASIC) and do the sepcified "action" on the BGP neighbor(s)
+    """
```
jleveque, line 1838: Prefer to reserve`"""` for docstrings and use `#` for comments

config/main.py
judyjoseph: Done.

config/main.py
judyjoseph: Done.

config/main.py
judyjoseph: Done.

config/main.py
judyjoseph: Done.

config/main.py
judyjoseph: Done.

jleveque:
LGTM. Wait for Guohan and Pavel to review.

rlhui:
retest this please.

jleveque:
Retest this please

config/main.py
```diff
@@ -351,88 +351,95 @@ def get_interface_naming_mode():
...
-    config_db = ConfigDBConnector()
-    config_db.connect()
     entry = config_db.get_entry('BGP_NEIGHBOR', ipaddress)
     return True if entry else False
 
-def _get_all_neighbor_ipaddresses():
+def _get_all_neighbor_ipaddresses(config_db, ignore_local_hosts=False):
     """Returns list of strings containing IP addresses of all BGP neighbors
+       if the flag ignore_local_hosts is set to True, additional check to see if 
+       if the BGP neighbor AS number is same as local BGP AS number, if so ignore that neigbor.
```
pavel-shirshov, line 368: so the name of the function is missleading, because you're filtering out peer with local bgp AS?
Also, we could have dynamic bgp neighbors and bgp_monitor entries.
Should we include them here?

config/main.py
```diff
@@ -351,88 +351,95 @@ def get_interface_naming_mode():
...
+    if _is_neighbor_ipaddress(config_db, ipaddr_or_hostname.lower()):
         ip_addrs.append(ipaddr_or_hostname.lower())
     else:
         # If <ipaddr_or_hostname> is not the IP address of a neighbor, check to see if it's a hostname
-        ip_addrs = _get_neighbor_ipaddress_list_by_hostname(ipaddr_or_hostname)
+        ip_addrs = _get_neighbor_ipaddress_list_by_hostname(config_db, ipaddr_or_hostname)
 
     if not ip_addrs:
-        click.get_current_context().fail("Could not locate neighbor '{}'".format(ipaddr_or_hostname))
+        return False
```
pavel-shirshov, line 411: why the error message was removed?

config/main.py
judyjoseph: This was an existing function, to which I added a new parameter ignore_local_hosts to do bgp AS based filtering. I will change the name of the function.
 
Currently all the functions look in the BGP_NEIGHBOR table alone, we need change in all API's to check the other BGP tables as well. BGP DYNAMIC PEERS and BGP MONITORS. We could do it separately in a different PR if that is ok

config/main.py
judyjoseph: These error message are not removed, the "same" message is now moved into the bgp command handlers ( eg: https://github.com/Azure/sonic-utilities/blob/33899cfe36018b8c6b03e9dbd125dfe117c8b56a/config/main.py#L1709 ). This is because for a specific neighbor shut/start case in multi-asic platform, we check in all bgp instances if the neighbor exists before declaring "not exist" with the ERROR message.

PR #914, opened 2020-05-13
Changes to make lldp show command for multi-npu platforms.

abdosi:
**- What I did**
Enhanced below command to work for multi-npu platforms:- 
show lldp table
show lldp neighbours [INTERFACE]

In the O/P we will filter out the LLDP information of Back plane Port/asic

Both this command are now wrapper of script/lldpshow which is now enhanced as below

usage: lldpshow [-h] [-v] [-d] [-p PORT]

Display the LLDP neighbors

optional arguments:
  -h, --help            show this help message and exit
  -v, --version         show program's version number and exit
  -d, --detail          LLDP neighbors detail information
  -p PORT, --port PORT  LLDP neighbors detail information for given port

                                      Examples:
                                      lldpshow
                                      lldpshow -d
                                      lldpshow -d -p Ethernet0

**- How I did it**
Updating script/lldpshow and using it
**- How to verify it**
Verified on both Single and Multi-NPU platforms
No change in O/P


abdosi:
retest this please

abdosi:
@jleveque @lguohan Can you please review this.

abdosi:
@jleveque can you please review this.

scripts/lldpshow
```diff
@@ -25,30 +25,65 @@ import re
...
+        SonicDBConfig.load_sonic_global_db_config()
 
-    def get_info(self):
+        # For multi-npu platforms we will get only front-panel interface to display
+        namespaces = sonic_device_util.get_all_namespaces()
+        per_npu_configdb = {}
+        for instance_num, front_asic_namespaces in enumerate(namespaces['front_ns']):
+            per_npu_configdb[front_asic_namespaces] = ConfigDBConnector(use_unix_socket_path=True, namespace=front_asic_namespaces)
+            per_npu_configdb[front_asic_namespaces].connect()
+            self.lldp_interface.append('')
```
jleveque, line 51: You are appending the empty string here to signify the host (i.e., no) namespace, correct? I think you should create a constant for this so that it is more clear, like in the other PRs.

scripts/lldpshow
```diff
@@ -25,30 +25,65 @@ import re
...
+            self.lldp_instance.append(instance_num)
+            keys = per_npu_configdb[front_asic_namespaces].get_keys("PORT")
+            for key in keys:
+                if key.startswith(BACKEND_ASIC_INTERFACE_NAME_PREFIX):
+                    continue
+                self.lldp_interface[instance_num] += key + ' '
+
+        # LLDP running in host namespace
+        self.lldp_instance.append('')
+        self.lldp_interface.append('')
```
jleveque, line 61: As above, should use a constant.

scripts/lldpshow
abdosi: @jleveque Done

scripts/lldpshow
abdosi: @jleveque Done

abdosi:
> As comments. Also I notice this file uses "NPU" and "ASIC" interchangeably. As I have mentioned before, I think we need to unify this terminology. I have created an issue to track this: #922

@jleveque In this file made everything as ASIC

jleveque:
> > As comments. Also I notice this file uses "NPU" and "ASIC" interchangeably. As I have mentioned before, I think we need to unify this terminology. I have created an issue to track this: #922
> 
> @jleveque In this file made everything as ASIC

@abdosi: I believe you mean you changed all references to "NPU", not "ASIC".

scripts/lldpshow
```diff
@@ -25,30 +25,72 @@ import re
...
 import xml.etree.ElementTree as ET
 from tabulate import tabulate
+import argparse
+import sonic_device_util
+from swsssdk import ConfigDBConnector, SonicDBConfig
+BACKEND_NPU_INTERFACE_NAME_PREFIX = 'Ethernet-BP'
+
+LLDP_INTERFACE_LIST_IN_HOST_NAMESPACE ''
+LLDP_INSTANCE_IN_HOST_NAMESPACE ''
+LLDP_DEFAULT_INTERFACE_LIST_IN_NPU_NAMESPACE ''
```
jleveque, line 35: Why is NPU namespace also an empty string?

abdosi: @jleveque This is just initial/default value to start with. The Interface list will get appended in the loop for asic namespace. We start with '' and then append to it.

abdosi:
> > > As comments. Also I notice this file uses "NPU" and "ASIC" interchangeably. As I have mentioned before, I think we need to unify this terminology. I have created an issue to track this: #922
> > 
> > 
> > @jleveque In this file made everything as ASIC
> 
> @abdosi: I believe you mean you changed all references to "NPU", not "ASIC".

@jleveque I think git had some issue with file update . It is "ASIC" since in minigraph we have ASIC already and we can not change that. Going foward as part of Issue #922 we will make NPU as ASIC

scripts/lldpshow
jleveque: I'm confused as to why the host namespace is the "default" in the NPU namespace. Maybe the naming is unclear?

abdosi: @jleveque  for LLDP we have instance in both host (for eth0) and per asic (for physical ports) namespace.

For host namespace using macro LLDP_INSTANCE_IN_HOST_NAMESPACE and LLDP_INTERFACE_LIST_IN_HOST_NAMESPACE.

For per asic namespace using macro LLDP_DEFAULT_INTERFACE_LIST_IN_ASIC_NAMESPACE which is just init value and the interface list gets populated in the loop.

abdosi: @jleveque did I clarify your above concern ? let me know

jleveque: I guess I'm still confused. So the empty string here does not denote the namespace, but rather denotes an interface name? Why do we need to populate an empty interface name here?

abdosi: @jleveque Yes, We have empty string for both namespace and interface-list for host. 

Empty interface-list is to keep logic generic since for host lldp docker as we do not filter of any interface and the command will be just 'lldpctl' + <Empty string>.
In case user has passed specific interface (-p option)  command will be 'lldpctl' + < Interface-Num>.

For asic namespace we filter out Backplane interface (Ethernet-BP) and command will be 'lldpctl' + <Front Interface list> 
or in case user has passed interface (-p option) command will be 'lldpctl' + < Interface-Num>.

jleveque: I see. Thanks for the explanation. I guess it might just be the similarity of the constant names that is throwing me off here. I'd also like to get Pavel's input here, who is more of a SME on LLDP.

scripts/lldpshow
```diff
@@ -115,11 +168,31 @@ class Lldpshow(object):
...
+
+    parser.add_argument('-d', '--detail', action='store_true', help='LLDP neighbors detail information', default=False)
+    parser.add_argument('-p', '--port', type=str, help='LLDP neighbors detail information for given port', default=None)
+    args = parser.parse_args()
+
+    lldp_detail_info = args.detail
+    lldp_port = args.port
+
+    if lldp_port and not lldp_detail_info:
+        parser.error('-d is required when -p is set.')
```
pavel-shirshov, line 189: why do we need this? Can we set -d automatically when -p?

abdosi: @pavel-shirshov yes that is fine too. Updated accordingly.

abdosi: @pavel-shirshov can you please review/approve the PR.  Above comment has been addressed.

scripts/lldpshow
abdosi: @jleveque can you please approve this. pavel has approved it.

PR #915, opened 2020-05-14
[warm-reboot]: added pre-check for ISSU file

vadymhlushko-mlnx:
Signed-off-by: Vadym Hlushko <vadymh@mellanox.com>

**- What I did**
Added pre-check for file /host/warmboot/issu_bank.txt in warm-reboot script.
Added instruction on how to recover issu_bank.txt.

**- How I did it**
Modify the 'fast-reboot' script

**- How to verify it**
Need to somehow corrupt /host/warmboot/issu_bank.txt (list below) and then run the warm-reboot command.

1. Remove issu_bank.txt
2. Clear issu_bank.txt
3. Change characters count in issu_bank.txt
4. Write invalid content into issu_bank.txt

`root@arc-switch1041:~# warm-reboot`
`Warning: Stopping telemetry.service, but it can still be activated by:`
`  telemetry.timer`

`root@arc-switch1041:~# warm-reboot`
`(/host/warmboot/issu_bank.txt) does NOT exist or empty ...`
`To recover (/host/warmboot/issu_bank.txt) file, do the following:`
`$ docker exec -it syncd sx_api_dbg_generate_dump.py`
`$ docker exec -it syncd cat /tmp/sdkdump | grep 'ISSU Bank'`
`Command above will print the VALUE of ISSU BANK - 0 or 1, use this VALUE in the next command`
`$ printf VALUE > /host/warmboot/issu_bank.txt`

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -168,26 +168,39 @@ function recover_issu_bank_file_instruction()
 function check_issu_bank_file()
 {
     ISSU_BANK_FILE=/host/warmboot/issu_bank.txt
-    MLNX_ISSU_BANK_BROKEN=102
+    MLNX_ISSU_BANK_BROKEN_BEFORE_PRE_SHUTDOWN=102
+    MLNX_ISSU_BANK_BROKEN_AFTER_PRE_SHUTDOWN=103
 
     if [[ ! -s "$ISSU_BANK_FILE" ]]; then
         error "(${ISSU_BANK_FILE}) does NOT exist or empty ..."
         recover_issu_bank_file_instruction
-        exit "${MLNX_ISSU_BANK_BROKEN}"
+	if [[ "$1" = true ]]; then
+            exit "${MLNX_ISSU_BANK_BROKEN_BEFORE_PRE_SHUTDOWN}"
+	else
+            exit "${MLNX_ISSU_BANK_BROKEN_AFTER_PRE_SHUTDOWN}"
+	fi
     fi
 
     issu_file_chars_count=`stat -c %s ${ISSU_BANK_FILE}`;
     if [[ $issu_file_chars_count != 1 ]]; then
         error "(${ISSU_BANK_FILE}) characters count is NOT valid ..."
         recover_issu_bank_file_instruction
-        exit "${MLNX_ISSU_BANK_BROKEN}"
+	if [[ "$1" = true ]]; then
+            exit "${MLNX_ISSU_BANK_BROKEN_BEFORE_PRE_SHUTDOWN}"
+	else
+            exit "${MLNX_ISSU_BANK_BROKEN_AFTER_PRE_SHUTDOWN}"
+	fi
     fi
 
     issu_file_content=`awk '{print $0}' ${ISSU_BANK_FILE}`
     if [[ "$issu_file_content" != "0" && "$issu_file_content" != "1" ]]; then
         error "(${ISSU_BANK_FILE}) has invalid content ..."
         recover_issu_bank_file_instruction
-        exit "${MLNX_ISSU_BANK_BROKEN}"
+	if [[ "$1" = true ]]; then
+            exit "${MLNX_ISSU_BANK_BROKEN_BEFORE_PRE_SHUTDOWN}"
+	else
+            exit "${MLNX_ISSU_BANK_BROKEN_AFTER_PRE_SHUTDOWN}"
+	fi
     fi
 }
 
@@ -519,8 +532,10 @@ systemctl stop swss
 if [[ "$REBOOT_TYPE" = "warm-reboot" || "$REBOOT_TYPE" = "fastfast-reboot" ]]; then
     initialize_pre_shutdown
 
+    BEFORE_PRE_SHUTDOWN=true
+
     if [[ "x$sonic_asic_type" == x"mellanox" ]]; then
-        check_issu_bank_file
+        check_issu_bank_file "$BEFORE_PRE_SHUTDOWN"
     fi
 
     request_pre_shutdown
@@ -528,7 +543,7 @@ if [[ "$REBOOT_TYPE" = "warm-reboot" || "$REBOOT_TYPE" = "fastfast-reboot" ]]; t
     wait_for_pre_shutdown_complete_or_fail
 
     if [[ "x$sonic_asic_type" == x"mellanox" ]]; then
-        check_issu_bank_file
+        check_issu_bank_file "!$BEFORE_PRE_SHUTDOWN"
     fi
 
     # Warm reboot: dump state to host disk
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -158,52 +158,32 @@ function request_pre_shutdown()
 
 function recover_issu_bank_file_instruction()
 {
-    error "To recover (${ISSU_BANK_FILE}) file, do the following:"
-    error "$ docker exec -it syncd sx_api_dbg_generate_dump.py"
-    error "$ docker exec -it syncd cat /tmp/sdkdump | grep 'ISSU Bank'"
-    error "Command above will print the VALUE of ISSU BANK - 0 or 1, use this VALUE in the next command"
-    error "$ printf VALUE > /host/warmboot/issu_bank.txt"
+    debug "To recover (${ISSU_BANK_FILE}) file, do the following:"
+    debug "$ docker exec -it syncd sx_api_dbg_generate_dump.py"
+    debug "$ docker exec -it syncd cat /tmp/sdkdump | grep 'ISSU Bank'"
+    debug "Command above will print the VALUE of ISSU BANK - 0 or 1, use this VALUE in the next command"
+    debug "$ printf VALUE > /host/warmboot/issu_bank.txt"
 }
 
 function check_issu_bank_file()
 {
     ISSU_BANK_FILE=/host/warmboot/issu_bank.txt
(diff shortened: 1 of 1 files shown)


scripts/fast-reboot
```diff
@@ -156,6 +156,42 @@ function request_pre_shutdown()
...
+    ISSU_BANK_FILE=/host/warmboot/issu_bank.txt
+    MLNX_ISSU_BANK_BROKEN=102
+
+    if [[ ! -s "$ISSU_BANK_FILE" ]]; then
+        error "(${ISSU_BANK_FILE}) does NOT exist or empty ..."
+        recover_issu_bank_file_instruction
+        exit "${MLNX_ISSU_BANK_BROKEN}"
+    fi
+
+    issu_file_chars_count=`wc -m ${ISSU_BANK_FILE} | awk '{print $1}'`;
```
qiluo-msft, line 179: >wc -m ${ISSU_BANK_FILE} | awk '{print $1}' [](start = 27, length = 42)

A more simple solution
```
stat -c %s FILE
```

scripts/fast-reboot
```diff
@@ -483,8 +519,16 @@ systemctl stop swss
...
 
+    if [[ "x$sonic_asic_type" == x"mellanox" ]]; then
+        check_issu_bank_file
+    fi
+
     request_pre_shutdown
 
+    if [[ "x$sonic_asic_type" == x"mellanox" ]]; then
+        check_issu_bank_file
+    fi
```
qiluo-msft, line 530: Should the 2nd check after `wait_for_pre_shutdown_complete_or_fail`?

yxieca: This is kind of too late.

The check happens past the point of no return and the warm reboot will happen regardless. Can this check happen with rest of the pre-checks?

stepanblyschak: To my knowledge, issu_bank.txt is written on ISSU start == request_pre_shutdown. There is actually no point to check it before - it can even not exist.  The corruption may happen only after ISSU start, but since it is no return point then this check is useless and I don't see a good point to place a check like this. If the file is corrupted before ISSU start it is ok, ISSU start will write the correct value to the file. If the file is corrupted after ISSU start we can't really perform a check since this is too late. Now we can either proceed and rely on the fix in recent SDK or abort warm reboot at the stage when cold reboot is required to restore the system as done in this PR.

qiluo-msft: @stepanblyschak Could you explain "ISSU start == request_pre_shutdown" in details?

Let's assume the file is already corrupted in a running switch, and we run `sudo warm-reboot`. How should precheck to protect?

stepanblyschak: request_pre_shutdown -> syncd sets PRESHUTDOWN attribute, mellanox SAI calls ISSU start -> issu start writes to /host/warmboot/issu_bank.txt.

The precheck could be only done if you want to abort warm reboot. However, it will result in a broken switch that anyway will require cold reboot (not just fixing /host/warmboot/issu_bank.txt).

qiluo-msft: My question is still applicable.

---
In reply to: [427514083](https://github.com/Azure/sonic-utilities/pull/915#discussion_r427514083) [](ancestors = 427514083)

qiluo-msft: Offline discussed with Ying, this one is resolved.

---
In reply to: [427688717](https://github.com/Azure/sonic-utilities/pull/915#discussion_r427688717) [](ancestors = 427688717)

scripts/fast-reboot
qiluo-msft: Did you try? This comment is still applicable.

---
In reply to: [427512540](https://github.com/Azure/sonic-utilities/pull/915#discussion_r427512540) [](ancestors = 427512540)

vadymhlushko-mlnx: > Did you try? This comment is still applicable.
> 
> In reply to: [427512540](https://github.com/Azure/sonic-utilities/pull/915#discussion_r427512540) [](ancestors = 427512540)

Fixed

scripts/fast-reboot
vadymhlushko-mlnx: > My question is still applicable.
> 
> In reply to: [427514083](https://github.com/Azure/sonic-utilities/pull/915#discussion_r427514083) [](ancestors = 427514083)

Fixed

scripts/fast-reboot
```diff
@@ -483,10 +519,18 @@ systemctl stop swss
...
+
     request_pre_shutdown
 
     wait_for_pre_shutdown_complete_or_fail
 
+    if [[ "x$sonic_asic_type" == x"mellanox" ]]; then
+        check_issu_bank_file
```
yxieca, line 531: Sorry for the last minute issue.

How can we distinguish if the error message came out from line 531 or 523? Are you able to passing in a differentiating token to check issu_bank_file and use that to distinguish if the error happens before or after pre-shutdown?

vadymhlushko-mlnx: > Sorry for the last minute issue.
> 
> How can we distinguish if the error message came out from line 531 or 523? Are you able to passing in a differentiating token to check issu_bank_file and use that to distinguish if the error happens before or after pre-shutdown?

Fixed

scripts/fast-reboot
```diff
@@ -156,6 +156,55 @@ function request_pre_shutdown()
...
+    MLNX_ISSU_BANK_BROKEN_BEFORE_PRE_SHUTDOWN=102
+    MLNX_ISSU_BANK_BROKEN_AFTER_PRE_SHUTDOWN=103
+
+    if [[ ! -s "$ISSU_BANK_FILE" ]]; then
+        error "(${ISSU_BANK_FILE}) does NOT exist or empty ..."
+        recover_issu_bank_file_instruction
+	if [[ "$1" = true ]]; then
+            exit "${MLNX_ISSU_BANK_BROKEN_BEFORE_PRE_SHUTDOWN}"
+	else
+            exit "${MLNX_ISSU_BANK_BROKEN_AFTER_PRE_SHUTDOWN}"
```
yxieca, line 180: Sorry I missed this issue in my second round of review.

Exit here will cause the fast/warm reboot to abort. I don't think this is what we want here. Because at the time when this check happens, we've already shutdown some critical services, e.g. BGP, RADV. Depending on if pre-shutdown is called, we could also lose syncd. It is unfortunate that there was an issue. But I think at this point, we would have to proceed with reboot and take the consequence. The logic is that aborting reboot sequence will require human intervention to recover. Continue rebooting will cause data plane disruption in the future. That sucks, but that would self heal. Right?

I think here the best we can do is to generate a log message so that we have log to look into after reboot is done and learn what happened when.

scripts/fast-reboot
```diff
@@ -168,26 +168,39 @@ function recover_issu_bank_file_instruction()
...
     ISSU_BANK_FILE=/host/warmboot/issu_bank.txt
-    MLNX_ISSU_BANK_BROKEN=102
+    MLNX_ISSU_BANK_BROKEN_BEFORE_PRE_SHUTDOWN=102
+    MLNX_ISSU_BANK_BROKEN_AFTER_PRE_SHUTDOWN=103
 
     if [[ ! -s "$ISSU_BANK_FILE" ]]; then
         error "(${ISSU_BANK_FILE}) does NOT exist or empty ..."
         recover_issu_bank_file_instruction
-        exit "${MLNX_ISSU_BANK_BROKEN}"
+	if [[ "$1" = true ]]; then
```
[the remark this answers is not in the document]
vadymhlushko-mlnx: Fixed

scripts/fast-reboot
vadymhlushko-mlnx: > Sorry I missed this issue in my second round of review.
> 
> Exit here will cause the fast/warm reboot to abort. I don't think this is what we want here. Because at the time when this check happens, we've already shutdown some critical services, e.g. BGP, RADV. Depending on if pre-shutdown is called, we could also lose syncd. It is unfortunate that there was an issue. But I think at this point, we would have to proceed with reboot and take the consequence. The logic is that aborting reboot sequence will require human intervention to recover. Continue rebooting will cause data plane disruption in the future. That sucks, but that would self heal. Right?
> 
> I think here the best we can do is to generate a log message so that we have log to look into after reboot is done and learn what happened when.

Fixed

scripts/fast-reboot
```diff
@@ -483,10 +512,18 @@ systemctl stop swss
 if [[ "$REBOOT_TYPE" = "warm-reboot" || "$REBOOT_TYPE" = "fastfast-reboot" ]]; then
     initialize_pre_shutdown
 
+    if [[ "x$sonic_asic_type" == x"mellanox" ]]; then
+        check_issu_bank_file
```
qiluo-msft, line 516: >check_issu_bank_file [](start = 8, length = 20)

This check is before `request_pre_shutdown`. If you find anything wrong, you may still `exit` the script and control/data plane are still running.

vadymhlushko-mlnx: Fixed

qiluo-msft:
Retest this please

scripts/fast-reboot
yxieca: Sorry I didn't see this earlier, even before pre-shutdown, the swss has been stopped. Exiting here will leave the device eventually in failure mode requires human interaction.

I think both check should generate log event stating problem found, but not exiting.

Is there a chance that there was a problem before pre-shutdown, but pre-shutdown fixed it?

yxieca: The only safe place to check and exit is here: https://github.com/Azure/sonic-utilities/blob/master/scripts/fast-reboot#L255

PR #918, opened 2020-05-18
handled the priority of ACL rules to be case insensitive

madhanmellanox:
- What I did
I made the priority of ACL rule props in the ACL commands to be case insensitive.

- How I did it
Instead of checking the ACL ruleprops dictionary for "PRIORITY" based key, I made it the actual key to be searched for.

- How to verify it
I verified the fix by testing on my switch using different case sensitive options of keyword "Priority".

### DIFF
--- a/scripts/aclshow
+++ b/scripts/aclshow
@@ -169,8 +169,13 @@ class AclStat(object):
                     self.get_counter_value(rule_key, 'packets') == 'N/A'):
                 continue
             rule = self.acl_rules[rule_key]
+            rule_priority = -1
+            if "PRIORITY" in rule:
+                rule_priority = rule['PRIORITY']
+            elif "priority" in rule:
+                rule_priority = rule['priority']
             line = [rule_key[1], rule_key[0],
-                    rule['PRIORITY'],
+                    rule_priority,
                     self.get_counter_value(rule_key, 'packets'),
                     self.get_counter_value(rule_key, 'bytes')]
             aclstat.append(line)
--- a/scripts/aclshow
+++ b/scripts/aclshow
@@ -170,10 +170,9 @@ class AclStat(object):
                 continue
             rule = self.acl_rules[rule_key]
             rule_priority = -1
-            if "PRIORITY" in rule:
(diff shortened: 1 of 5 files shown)


sonic-utilities-tests/mock_tables/config_db.json
```diff
@@ -67,7 +67,7 @@
     "ACL_RULE|DATAACL|RULE_4": {
             "L4_SRC_PORT": "4661",
             "PACKET_ACTION": "FORWARD",
-            "PRIORITY": "9996"
+            "priority": "9996"
```
stepanblyschak, line 70: better to create a new rule for the test than modify existing

scripts/aclshow
```diff
@@ -169,8 +169,13 @@ class AclStat(object):
                     self.get_counter_value(rule_key, 'packets') == 'N/A'):
                 continue
             rule = self.acl_rules[rule_key]
+            rule_priority = -1
+            if "PRIORITY" in rule:
+                rule_priority = rule['PRIORITY']
+            elif "priority" in rule:
+                rule_priority = rule['priority']
```
stepanblyschak, line 176: what if priority is 'pRiOrItY' ?

sonic-utilities-tests/mock_tables/config_db.json
madhanmellanox: added a new rule in the next commit.

scripts/aclshow
madhanmellanox: addressed this code review comment in the next commit.

scripts/aclshow
```diff
@@ -169,8 +169,12 @@ class AclStat(object):
                     self.get_counter_value(rule_key, 'packets') == 'N/A'):
                 continue
             rule = self.acl_rules[rule_key]
+            rule_priority = -1
```
stepanblyschak, line 172: should it be N/A as in acl_loader/main.py?

madhanmellanox: rule_priority is expecting int, that is why initialized it to -1.

PR #919, opened 2020-05-19
Changes to commands  config reload/load-minigraph

abdosi:
**- What I did**
Changes to make commands config reload/load-minigraph
to have option of not stop/start docker services. This is not needed
when this commands are called from config-setup shell script. PR: https://github.com/Azure/sonic-buildimage/pull/4609 used this change.
**- How I did it**
Added new option -n (--no-service-restart) with default as False
so that existing command behavior remains same.
**- How to verify it**
Verified Manually and as well when invoke with config-setup

### DIFF
--- a/doc/Command-Reference.md
+++ b/doc/Command-Reference.md
@@ -3269,9 +3269,13 @@ NOTE: Management interface IP address and default route (or specific route) may 
 When user specifies the optional argument "-y" or "--yes", this command forces the loading without prompting the user for confirmation.
 If the argument is not specified, it prompts the user to confirm whether user really wants to load this configuration file.
 
+When user specifies the optional argument "-n" or "--no-service-restart", this command loads the configuration without restarting various services
+running on the device. One of the use case for this option is during boot time when config-setup service loads minigraph configuration and there is no services
+running on the device.
+
 - Usage:
   ```
-  config load_minigraph [-y|--yes]
+  config load_minigraph [-y|--yes] [-n|--no-service-restart]
   ```
 
 - Example:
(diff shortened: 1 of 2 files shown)


doc/Command-Reference.md
```diff
@@ -3269,9 +3269,13 @@ NOTE: Management interface IP address and default route (or specific route) may
 When user specifies the optional argument "-y" or "--yes", this command forces the loading without prompting the user for confirmation.
 If the argument is not specified, it prompts the user to confirm whether user really wants to load this configuration file.
 
+When user specifies the optional argument "-n" or "--no-service-restart", this command loads the configuration without restarting various services
```
jleveque, line 3272: s/various services/dependent services/

doc/Command-Reference.md
```diff
@@ -3304,9 +3308,13 @@ NOTE: Management interface IP address and default route (or specific route) may
 When user specifies the optional argument "-y" or "--yes", this command forces the loading without prompting the user for confirmation.
 If the argument is not specified, it prompts the user to confirm whether user really wants to load this configuration file.
 
+When user specifies the optional argument "-n" or "--no-service-restart", this command clear and loads the configuration without restarting various services
```
jleveque, line 3311: s/various services/dependent services/

abdosi:
> When adding/removing commands or modifying their names/options/arguments, please also modify docs/CommandReference.md to reflect the changes.

@jleveque Thanks for pointing out. Done

doc/Command-Reference.md
```diff
@@ -3304,9 +3308,13 @@ NOTE: Management interface IP address and default route (or specific route) may
...
+running on the device. One of the use case for this option is during boot time when config-setup service loads exisitng old configuration and there is no services
```
jleveque, line 3312: s/One of the use case/One use case/
s/exisitng/existing/
s/there is no services/there are no services/

doc/Command-Reference.md
```diff
@@ -3269,9 +3269,13 @@ NOTE: Management interface IP address and default route (or specific route) may
 When user specifies the optional argument "-y" or "--yes", this command forces the loading without prompting the user for confirmation.
 If the argument is not specified, it prompts the user to confirm whether user really wants to load this configuration file.
 
+When user specifies the optional argument "-n" or "--no-service-restart", this command loads the configuration without restarting various services
+running on the device. One of the use case for this option is during boot time when config-setup service loads minigraph configuration and there is no services
```
jleveque, line 3273: s/One of the use case/One use case/
s/there is no services/there are no services/

PR #938, opened 2020-06-09
201911 fixes #4716 show ipv6 interface neighbor_ip N/A issue

tim-rj:
**- Why I did it**
When exec 'show ipv6 interfaces', the neighbor_ip will always be 'N/A' which will confuse the consumers. After checking we found that this is a show issue. As we know, there will be one local linked IPV6 address for every valid interface. In show/main.py, the interface's ipv6 neighbor ip will always be written by this local linked IPV6's neighbor_ip address which is 'N/A'. And i think this part code of 'show ipv6 interface' was copied from 'show ip interface' before, right? Yea, some change is necessary!
**- What I did**
This is bug fixing.
**- How I did it**
Code change.
**- How to verify it**
upgrate config_db.config attached in Azure/sonic-buildimage#4716 and show ipv6 interfaces

### DIFF
--- a/show/main.py
+++ b/show/main.py
@@ -1504,12 +1504,14 @@ def interfaces():
 
         if netifaces.AF_INET6 in ipaddresses:
             ifaddresses = []
+            neighbor_name = 'N/A'
+            neighbor_ip = 'N/A'
             for ipaddr in ipaddresses[netifaces.AF_INET6]:
-                neighbor_name = 'N/A'
-                neighbor_ip = 'N/A'
                 local_ip = str(ipaddr['addr'])
                 netmask = ipaddr['netmask'].split('/', 1)[-1]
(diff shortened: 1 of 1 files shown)


prsunny:
Is this issue happening in master as well? If not, what is the fix in master? Ideally we want to cherry-pick bug fix from master to 201911.

tim-rj:
> 
> 
> Is this issue happening in master as well? If not, what is the fix in master? Ideally we want to cherry-pick bug fix from master to 201911.

Hi, this issue should exist in master as well, I will try to compile bsp module based on master's kernel as soon as possible. The key point to reproduce is that the port should 'UP' and has a local linked address, then the neighbor_ip will be written by 'N/A'  .

prsunny:
Ok, then you should be raising the PR to master and can mark for cherry-pick to 201911.

PR #765, opened 2019-12-13
[config] Add ConfigMgmt class for config validation, delete ports, add ports

praveen-li:
**- What I did**
 Provided a class in sonic_util config for 
-- Config Validation,
-- BreakOut Ports

### Test cases are added in this PR but described in a dummy PR https://github.com/Azure/sonic-utilities/pull/947

**- How I did it**
--ConfigMgmt Class uploads config data from config_Db.json or from config db using SONiC-py-swsssdk APIs . 
--Port, vlan and acl related config is cropped out of entire config, this is because yang models are available only for Vlan, Acl and Port as of now. 
--Cropped portion of config DB will be given as input to Python Yang LIB while loading config data.
 
--ConfigMgmt Class provides APIs to breakout, delete and add Ports. 
--Using Python yang LIB APIs, this class finds dependencies for each of the port which must be deleted and will delete dependencies. 
-- At last after successful validation, new config is written in config DB.
-- While addition, relevant default config is found for ports which are being added. Then entire new config [default + port] is added to current config in data tree.
-- At last after successful validation, new config is written in config DB.

**- How to verify it**
On Sonic Switch Ran below Tests 
### Note: Below tests are done with PR 766 and PR 857 changes.
```
:::Lock Acquired:::
:::Lock Timer Extended:::

Running Breakout Mode : 4x25G 
Target Breakout Mode : 2x50G

Ports to be deleted : 
 {
    "Ethernet2": "25000", 
    "Ethernet3": "25000", 
    "Ethernet0": "25000", 
    "Ethernet1": "25000"
}
Ports to be added : 
 {
    "Ethernet2": "50000", 
    "Ethernet0": "50000"
}

After running Logic to limit the impact

Final list of ports to be deleted : 
 {
    "Ethernet2": "25000", 
    "Ethernet3": "25000", 
    "Ethernet0": "25000", 
    "Ethernet1": "25000"
} 
Final list of ports to be added :  
 {
    "Ethernet2": "50000", 
    "Ethernet0": "50000"
}
Loaded below Yang Models
['sonic-acl', 'sonic-extension', 'sonic-head', 'sonic-interface', 'sonic-loopback-interface', 'sonic-port', 'sonic-portchannel', 'sonic-vlan']
Reading data from Redis configDb
Note: Below table(s) have no YANG models:
VERSIONS, DEVICE_METADATA, DEVICE_NEIGHBOR, LOCK, FLEX_COUNTER_TABLE, BREAKOUT_CFG, 
[...]

PR #947, opened 2020-06-18
[sonic-utilities-tests/config_mgmt_test.py]: Unit Tests for config_mg…

praveen-li:
…mt.py library.

Signed-off-by: Praveen Chaudhary pchaudhary@linkedin.com

This is a dummy PR just for better description. Committed in [https://github.com/Azure/sonic-utilities/pull/765/commits/77e3734c66339f5b92be560c10f4d8a291ccb7e7]
#### Depend on https://github.com/Azure/sonic-buildimage/pull/4798.

**- What I did**
Write Unit tests for Config_mgmt library introduced in PR #765.

**- How I did it**
Key Decisions:
-- Config_mgmt interaction with Redis DB:
These tests will be executed at build time, the entire interaction with Redis DB is mocked using:
    import mock_tables.dbconnector
This mocks below classes used in config_mgmt
from swsssdk import ConfigDBConnector, SonicV2Connector, port_util

-- Mock sonic_yang or not:
Most of the functions in config_mgmt calls APIs from sonic_yang_mgmt (https://github.com/Azure/sonic-buildimage/tree/master/src/sonic-yang-mgmt) library. It is important to catch any compatibility issues during build time rather than run time. So it is decided not to mock sonic_yang but use real instance. Due to this sonic_utilities will have a build time dependency on sonic_yang_mgmt.

-- Unit Test each function or the functions exposed to caller:
	1.	To unit test each function, we need to mock sonic_yang, which is against the previous decision.  
	2.	Also calling the functions which are exposed to the caller is sufficient to unit test all functions. Because functions like breakOutPort internally call every function in the library, and the expected result is not produced if internal functions are not working properly. 
	3.	In the real scenario, the caller will use this library for config validation and to break out the port in a continuous manner. To emulate a real scenario it is decided to exercise only the functions exposed to the caller. 

Check Result:
-- How to check expected result for breakout Command:
	1.	A successful call to breakOutPort API results in a call to writeConfigDB API twice. 
One time with config to be deleted and second time with config to be added. Both calls are necessary and should be done in order. This is tested by mocking writeConfigDB API as below:
```
cmdpb = config_mgmt.ConfigMgmtDPB(source=config_mgmt.CONFIG_DB_JSON_FILE)
cmdpb.writeConfigDB = MagicMock(return_value=True)
```
Then we can check result as below:
```
calls = [call(dConfig), call(aConfig)]
        assert cmdpb.writeConfigDB.call_count == 2
        cmdpb.writeConfigDB.assert_has_calls(calls, any_order=False)
```
Note: it is important to check that both calls are done, rather than just checking final config.

Tests: 
-- Create the configMgmt Class object and validate config.

-- Have a table without yang model in config, Check if this table is returned by tablesWithoutYang()

-- Test configWithKeys() to search ports in the config.

-- Important Test case, Test below sequence for port breakout:
```
        #Ethernet8: start from 4x25G-->2x50G with -f -l

        #Ethernet8: move from 2x50G-->1x100G without force, list deps

        # Ethernet8: move from 2x50G-->1x100G with force, where deps exists

        # Ethernet8: move from 1x100G-->4x25G without force, no deps

        # Ethernet8: move from 4x25G-->1x100G with force, no deps

        # Ethernet8: move from 1x100G-->1x50G(2)+2x25G(2) with -f -l,

        # Ethernet4: breakout from 4x25G to 2x50G with -f -l
```
Note: this is important to do this sequence in one test case, because it is good to test with continuous  change in Config. Also this way, a new configuration for each test is not needed.
 
Expected Result:
All automated tests should pass.

Each breakout test is checked for expected result  as below:
Have a hard coded value of expected dConfig and aConfig, where
dConfig: Config to be deleted.
aConfig: Config to be deleted.

    And check if writeConfigDB is called using the above configurations.
Example: 
```
dConfig = {u'PORT': {u'Ethernet8': None}}

aConfig = {u'ACL_TABLE': {u'NO-NSW-PACL-V4': {u'ports': ['Ethernet0', \
        'Ethernet4', 'Ethernet8', 'Ethernet10']}, u'NO-NSW-PACL-TEST': {u'ports': \
        ['Ethernet11']}}, u'INTERFACE': {u'Ethernet11|2a04:1111:40:a709::1/126': \
        {u'scope': u'global', u'family': u'IPv6'}, u'Ethernet11': {}}, \
        u'VLAN_MEMBER': {u'Vlan100|Ethernet8': {u'tagging_mode': u'untagged'}, \
        u'Vlan100|Ethernet11': {u'tagging_mode': u'untagged'}}, u'PORT': \
        {'Ethernet8': {'speed': '50000', 'lanes': '73,74'}, 'Ethernet10': \
        {'speed': '25000', 'lanes': '75'}, 'Ethernet11': {'speed': '25000', 'lanes': '76'}}}
        
self.checkResult(cmdpb, dConfig, aConfig)
```
Actual Result:
```
============================================================ test session starts ============================================================
platform linux2 -- Python 2.7.13, pytest-3.0.6, py-1.4.32, pluggy-0.4.0
rootdir: /sonic/src/sonic-utilities/sonic-utilities-tests, inifile: pytest.ini
plugins: cov-2.4.0
collected 4 items

config_mgmt_test.py ....

========================================================= 4 passed in 0.67 seconds ==========================================================
```

**- How to verify it**
Actual Result:
```
============================================================ test session starts ============================================================
platform linux2 -- Python 2.7.13, pytest-3.0.6, py-1.4.32, pluggy-0.4.0
rootdir: /sonic/src/sonic-utilities/sonic-utilities-tests, inifile: pytest.ini
plugins: cov-2.4.0
collected 4 items

config_mgmt_test.py ....

========================================================= 4 passed in 0.67 seconds ==========================================================
```

### DIFF
--- a/sonic-utilities-tests/config_mgmt_test.py
+++ b/sonic-utilities-tests/config_mgmt_test.py
@@ -1,0 +1,475 @@
+import imp
+import os
+import mock_tables.dbconnector
+# import file under test i.e. config_mgmt.py
+imp.load_source('config_mgmt', \
+    os.path.join(os.path.dirname(__file__), '..', 'config', 'config_mgmt.py'))
+import config_mgmt
+
+from unittest import TestCase
+from mock import MagicMock, call
+from json import dump
+
+class TestConfigMgmt(TestCase):
+
+    def setUp(self):
+        config_mgmt.YANG_DIR = "../../sonic-yang-models/yang-models/"
+        config_mgmt.CONFIG_DB_JSON_FILE = "startConfigDb.json"
+        config_mgmt.DEFAULT_CONFIG_DB_JSON_FILE = "portBreakOutConfigDb.json"
+        return
+
+    def test_config_validation(self):
+        iConfig = dict(configDbJson)
+        self.writeJson(iConfig, config_mgmt.CONFIG_DB_JSON_FILE)
+        cm = config_mgmt.ConfigMgmt(source=config_mgmt.CONFIG_DB_JSON_FILE)
+        assert cm.validateConfigData() == True
+        return
+
+    def test_table_without_yang(self):
+        iConfig = dict(configDbJson)
+        unknown = {"unknown_table": {"ukey": "uvalue"}}
+        self.updateConfig(iConfig, unknown)
+        self.writeJson(iConfig, config_mgmt.CONFIG_DB_JSON_FILE)
+        cm = config_mgmt.ConfigMgmt(source=config_mgmt.CONFIG_DB_JSON_FILE)
+        #assert "unknown_table" in cm.tablesWithoutYang()
+        return
+
+    def test_search_keys(self):
+        iConfig = dict(configDbJson)
+        self.writeJson(iConfig, config_mgmt.CONFIG_DB_JSON_FILE)
+        cmdpb = config_mgmt.ConfigMgmtDPB(source=config_mgmt.CONFIG_DB_JSON_FILE)
+        out = cmdpb.configWithKeys(portBreakOutConfigDbJson, \
+            ["Ethernet8","Ethernet9"])
+        assert "VLAN" not in out.keys()
+        assert "INTERFACE" not in out.keys()
+        for k in out['ACL_TABLE'].keys():
+            # only ports must be chosen
+            len(out['ACL_TABLE'][k]) == 1
+        out = cmdpb.configWithKeys(portBreakOutConfigDbJson, \
+            ["Ethernet10","Ethernet11"])
+        assert "INTERFACE" in out.keys()
+        for k in out['ACL_TABLE'].keys():
+            # only ports must be chosen
+            len(out['ACL_TABLE'][k]) == 1
+        return
+
+    def test_break_out(self):
+        # prepare default config
+        self.writeJson(portBreakOutConfigDbJson, \
+            config_mgmt.DEFAULT_CONFIG_DB_JSON_FILE)
+        # prepare config dj json to start with
+        iConfig = dict(configDbJson)
+        #Ethernet8: start from 4x25G-->2x50G with -f -l
+        self.dpb_port8_4x25G_2x50G_f_l(iConfig)
+        #Ethernet8: move from 2x50G-->1x100G without force, list deps
+        self.dpb_port8_2x50G_1x100G(iConfig)
+        # Ethernet8: move from 2x50G-->1x100G with force, where deps exists
+        self.dpb_port8_2x50G_1x100G_f(iConfig)
+        # Ethernet8: move from 1x100G-->4x25G without force, no deps
+        self.dpb_port8_1x100G_4x25G(iConfig)
+        # Ethernet8: move from 4x25G-->1x100G with force, no deps
+        self.dpb_port8_4x25G_1x100G_f(iConfig)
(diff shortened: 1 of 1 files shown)


sonic-utilities-tests/config_mgmt_test.py
```diff
@@ -0,0 +1,475 @@
...
+        aConfig = {u'ACL_TABLE': {u'NO-NSW-PACL-V4': \
+        {u'ports': ['Ethernet0', 'Ethernet8', 'Ethernet10', 'Ethernet4']}}, \
+        u'PORT': {'Ethernet4': {'speed': '50000', 'lanes': '69,70'}, \
+        'Ethernet6': {'speed': '50000', 'lanes': '71,72'}}}
+        self.checkResult(cmdpb, dConfig, aConfig)
+        self.postUpdateConfig(iConfig, dConfig, aConfig)
+        return
+
+###########GLOBAL Configs#####################################
+configDbJson =  {
```
samaity, line 292: I was thinking if you could create a folder containing two mock data json files like configDbJson and portBreakOutConfigDbJson and then import it to this function. 
something like below.
[This Function](https://github.com/Azure/sonic-utilities/blob/master/sonic-utilities-tests/acl_loader_test.py#L17) in acl_loader_test.py file uses [acl_input](https://github.com/Azure/sonic-utilities/tree/master/sonic-utilities-tests/acl_input) folder.
It will help to structurize this test by dividing the code and the input data. What do you think?

sonic-utilities-tests/config_mgmt_test.py
```diff
@@ -0,0 +1,475 @@
...
+        # create object
+        self.writeJson(iConfig, config_mgmt.CONFIG_DB_JSON_FILE)
+        cmdpb = config_mgmt.ConfigMgmtDPB(source=config_mgmt.CONFIG_DB_JSON_FILE)
+        # mock funcs
+        cmdpb.writeConfigDB = MagicMock(return_value=True)
+        cmdpb._verifyAsicDB = MagicMock(return_value=True)
+        return cmdpb
+
+    # Generate port to deleted, added and lanes and speed setting based on
+    # current and new mode.
```
praveen-li, line 103: Write one example

praveen-li: Example: 
For generate_args(8, 73, '4x25G', '2x50G'):
output:
(
['Ethernet8', 'Ethernet9', 'Ethernet10', 'Ethernet11'], 
['Ethernet8', 'Ethernet10'], 
{'Ethernet8': {'lanes': '73,74', 'speed': '50000'}, 'Ethernet10': {'lanes': '75,76', 'speed': '50000'}})

For generate_args(8, 73, '1x100G, ''4x25G'):
output:
(['Ethernet8'], 
['Ethernet8', 'Ethernet9', 'Ethernet10', 'Ethernet11'], 
{'Ethernet8': {'lanes': '73', 'speed': '25000'}, 'Ethernet9': {'lanes': '74', 'speed': '25000'}, 'Ethernet10': {'lanes': '75', 'speed': '25000'}, 'Ethernet11': {'lanes': '76', 'speed': '25000'}})

sonic-utilities-tests/config_mgmt_test.py
praveen-li: Since right now file is relatively small, It seems okie to have json in file itself.

PR #949, opened 2020-06-18
Save the PCIE device status to STATE DB

sujinmkang:
**- What I did**
Add the pcie device status to STATE_DB
**- How I did it**
Update the state_db during PcieUtil.get_pcie_check()
**- How to verify it**
sudo pcieutil pcie-check
and then read the state db to see if it's updated properly.
ex) 
user@server:~$redis-cli -n 6 GET "PCIE_STATE|system"
"1"

### DIFF
--- a/pcieutil/main.py
+++ b/pcieutil/main.py
@@ -181,8 +181,10 @@ def pcie_check():
             log_warning("PCIe Device: " +  item["name"] + " Not Found")
             err+=1
     if err:
+        os.system('redis-cli -n 6 SET "PCIE_STATE|system" "0"')
         print "PCIe Device Checking All Test ----------->>> FAILED"
     else:
+        os.system('redis-cli -n 6 SET "PCIE_STATE|system" "1"')
         print "PCIe Device Checking All Test ----------->>> PASSED"
         
 


sujinmkang:
retest this please

jleveque:
Why do we want to update the State DB when the CLI utility is executed? Shouldn't updating the DB be handled by a service or daemon to ensure it's done consistently?

sujinmkang:
@jleveque both pcied and pcie-check service is using this pcieutil to check the pcie device and this pcieutil is actually check the overall status of all pcie devices.  And later if we want to add a logic to device's status (like only ASIC chipsets or external devices on  the pcie buses) then this place will be better place than each service.

jleveque:
I see. I'm not sure this is the best approach. We have been moving in the opposite direction with our other CLI utilities, where a daemon periodically polls the hardware and updates the State DB, and the CLI utility simply reads the latest data from the DB and prints the info to the screen. I'd like to hear @lguohan's opinion on this approach.

pcieutil/main.py
```diff
@@ -181,8 +181,10 @@ def pcie_check():
             log_warning("PCIe Device: " +  item["name"] + " Not Found")
             err+=1
     if err:
+        os.system('redis-cli -n 6 SET "PCIE_STATE|system" "0"')
         print "PCIe Device Checking All Test ----------->>> FAILED"
     else:
+        os.system('redis-cli -n 6 SET "PCIE_STATE|system" "1"')
```
jleveque, line 187: While I'm not sure if we should update the DB here, if we do decide to take this approach, calling `redis-cli` is not the proper method. You should import the swss-common library and use it to interact with the DB.

sujinmkang:
@jleveque Even this utility is called by both pcie service and pcied, I think your question is still valid, I want to also listen more opinion from others and I will also think about the better/ best solution to update the state db.

sujinmkang:
Close this PR since we will update the STATE_DB from the service and the daemon.

PR #950, opened 2020-06-18
[show] Gracefully handle case where transceiver doesn't support DOM threshold

chiourung:
What I did:
    Don't raise exception if the key does not exist.

**- What I did**

**- How I did it**

**- How to verify it**


jleveque:
Updated PR title.

chiourung:
This has been fixed in commit: b9d962d4

PR #953, opened 2020-06-19
[sonic_installer] Refactor sonic_installer code

Staphylo:
Add a new Bootloader abstraction.
This makes it easier to add bootloader specific behavior while keeping
the main logic identical.
It is also a step that will ease the introduction of secureboot which
relies on bootloader specific behaviors.

Shuffle code around to get rid of the hacky if/else all over the place.
There are now 3 bootloader classes
 - AbootBootloader
 - GrubBootloader
 - UbootBootloader

There was almost no logic change in any of the implementations.
Only the AbootBootloader saw some small improvements.
More will follow in subsequent changes.

<!--
Please make sure you've read and understood our contributing guidelines:
https://github.com/Azure/SONiC/blob/gh-pages/CONTRIBUTING.md

** Make sure all your commits include a signature generated with `git commit -s` **

If this is a bug fix, make sure your description includes "closes #xxxx",
"fixes #xxxx" or "resolves #xxxx" so that GitHub automatically closes the related
issue when the PR is merged.

If you are adding/modifying/removing any command or utility script, please also
make sure to add/modify/remove any unit tests from the sonic-utilities-tests
directory as appropriate.

If you are modifying or removing an existing 'show', 'config' or 'sonic-clear'
subcommand, or you are adding a new subcommand, please make sure you also
update the Command Line Reference Guide (doc/Command-Reference.md) to reflect
your changes.

### DIFF
--- a/sonic_installer/bootloader/aboot.py
+++ b/sonic_installer/bootloader/aboot.py
@@ -1,0 +1,125 @@
+"""
+Bootloader implementation for Aboot used on Arista devices
+"""
+
+import collections
+import os
+import re
+import subprocess
+
+import click
+
+from ..common import (
+   HOST_PATH,
+   IMAGE_DIR_PREFIX,
+   IMAGE_PREFIX,
+   run_command,
+)
+from .bootloader import Bootloader
+
+_secureboot = None
+def isSecureboot():
+    global _secureboot
+    if _secureboot is None:
+        with open('/proc/cmdline') as f:
+           m  = re.search(r"secure_boot_enable=[y1]", f.read())
+        _secureboot = bool(m)
+    return _secureboot
+
+class AbootBootloader(Bootloader):
+
+    NAME = 'aboot'
+    BOOT_CONFIG_PATH = os.path.join(HOST_PATH, 'boot-config')
+    DEFAULT_IMAGE_PATH = '/tmp/sonic_image.swi'
+
+    def _boot_config_read(self, path=BOOT_CONFIG_PATH):
+        config = collections.OrderedDict()
+        with open(path) as f:
+            for line in f.readlines():
+                line = line.strip()
+                if not line or line.startswith('#') or '=' not in line:
+                    continue
+                key, value = line.split('=', 1)
+                config[key] = value
+        return config
+
+    def _boot_config_write(self, config, path=BOOT_CONFIG_PATH):
+        with open(path, 'w') as f:
+            f.write(''.join('%s=%s\n' % (k, v) for k, v in config.items()))
+
+    def _boot_config_set(self, **kwargs):
+        path = kwargs.pop('path', self.BOOT_CONFIG_PATH)
+        config = self._boot_config_read(path=path)
+        for key, value in kwargs.items():
+            config[key] = value
+        self._boot_config_write(config, path=path)
+
+    def _swi_image_path(self, image):
+        image_dir = image.replace(IMAGE_PREFIX, IMAGE_DIR_PREFIX)
+        if isSecureboot():
+           return 'flash:%s/sonic.swi' % image_dir
+        return 'flash:%s/.sonic-boot.swi' % image_dir
+
+    def get_current_image(self):
+        with open('/proc/cmdline') as f:
+            current = re.search(r"loop=/*(\S+)/", f.read()).group(1)
+        return current.replace(IMAGE_DIR_PREFIX, IMAGE_PREFIX)
+
+    def get_installed_images(self):
+        images = []
+        for filename in os.listdir(HOST_PATH):
+            if filename.startswith(IMAGE_DIR_PREFIX):
+                images.append(filename.replace(IMAGE_DIR_PREFIX, IMAGE_PREFIX))
+        return images
+
+    def get_next_image(self):
+        config = self._boot_config_read()
+        match = re.search(r"flash:/*(\S+)/", config['SWI'])
+        return match.group(1).replace(IMAGE_DIR_PREFIX, IMAGE_PREFIX)
+
+    def set_default_image(self, image):
+        image_path = self._swi_image_path(image)
+        self._boot_config_set(SWI=image_path, SWI_DEFAULT=image_path)
+        return True
+
+    def set_next_image(self, image):
+        image_path = self._swi_image_path(image)
+        self._boot_config_set(SWI=image_path)
+        return True
+
+    def install_image(self, image_path):
+        run_command("/usr/bin/unzip -od /tmp %s boot0" % image_path)
+        run_command("swipath=%s target_path=/host sonic_upgrade=1 . /tmp/boot0" % image_path)
+
+    def remove_image(self, image):
+        nextimage = self.get_next_image()
+        current = self.get_current_image()
+        if image == nextimage:
+            image_path = self._swi_image_path(current)
+            self._boot_config_set(SWI=image_path, SWI_DEFAULT=image_path)
+            click.echo("Set next and default boot to current image %s" % current)
+
+        image_dir = image.replace(IMAGE_PREFIX, IMAGE_DIR_PREFIX)
+        click.echo('Removing image root filesystem...')
+        subprocess.call(['rm','-rf', os.path.join(HOST_PATH, image_dir)])
+        click.echo('Image removed')
+
+    def get_binary_image_version(self, image_path):
+        try:
+            version = subprocess.check_output(['/usr/bin/unzip', '-qop', image_path, '.imagehash'])
+        except subprocess.CalledProcessError:
+            return None
+        return IMAGE_PREFIX + version.strip()
+
+    def verify_binary_image(self, image_path):
+        try:
+            subprocess.check_call(['/usr/bin/unzip', '-tq', image_path])
+            # TODO: secureboot check signature
+        except subprocess.CalledProcessError:
+            return False
+        return True
+
+    @classmethod
+    def detect(cls):
+        with open('/proc/cmdline') as f:
+            return 'Aboot=' in f.read()
--- a/sonic_installer/bootloader/bootloader.py
+++ b/sonic_installer/bootloader/bootloader.py
@@ -1,0 +1,50 @@
+"""
+Abstract Bootloader class
+"""
+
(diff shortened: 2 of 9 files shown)


jleveque:
Thanks for this enhancement, Samuel! Would you mind pulling the new classes (Bootloader, AbootBootloader, GrubBootloader, UbootBootloader) out into their own sepearate files? This will help make main.py more readable/managable and keep the different bootloader logic isolated.

Staphylo:
Will do!

Do you have a preference for the location of these files?
Also because there are some common constants that are needed by all these classes I will have to move them to a separate file.

I'm thinking of
```
sonic_installer/__init__.py
sonic_installer/aboot.py
sonic_installer/consts.py
sonic_installer/bootloader.py
sonic_installer/grub.py
sonic_installer/main.py
sonic_installer/uboot.py
```

Alternatively
```
sonic_installer/__init__.py
sonic_installer/bootloader/__init__.py # Bootloader class definition
sonic_installer/bootloader/aboot.py
sonic_installer/bootloader/grub.py
sonic_installer/bootloader/uboot.py
sonic_installer/consts.py
sonic_installer/main.py
```

Let me know what you think.

jleveque:
@Staphylo: I prefer the bootloader/ subdirectory in the latter, but I believe the base Bootloader class definition should not live in the __init__.py file, but rather a separate file, like bootloader/bootloader.py or bootloader/bootloader_base.py because the primary purpose of the __init__.py file is to initialize the package, not to hold content. Thanks!

sonic_installer/common.py
```diff
@@ -0,0 +1,27 @@
+
+import signal
```
qiluo-msft, line 2: Add shebang line?

jleveque: Shebangs only serve a purpose in executable scripts. I prefer to keep them out of library files, as they are not meant to be executed directly. However, there is benefit in otherwise labeling a file with the Python version it was written for. Maybe simply a header docstring for library files?

sonic_installer/bootloader/aboot.py
```diff
@@ -0,0 +1,106 @@
...
+
+import click
+
+from ..common import (
+   HOST_PATH,
+   IMAGE_DIR_PREFIX,
+   IMAGE_PREFIX,
+   run_command,
+)
+from .bootloader import Bootloader
```
qiluo-msft, line 15: Is it possible to move common header lines into `__init__.py` ?

sonic_installer/common.py
qiluo-msft, line 2: Seems this file is better fit into bootloader directory.

sonic_installer/bootloader/bootloader.py
```diff
@@ -0,0 +1,70 @@
...
+class Bootloader(object):
+
+    NAME = None
+    DEFAULT_IMAGE_PATH = '/tmp/sonic_image'
+
+    def get_current_image(self):
+        """returns name of the current image"""
+        with open('/proc/cmdline') as f:
+            current = re.search(r"loop=/*(\S+)/", f.read()).group(1)
+        return current.replace(IMAGE_DIR_PREFIX, IMAGE_PREFIX)
```
jleveque, line 21: I understand you created function bodies for get_current_image(), get_binary_image_version(), and verify_binary_image() in this base class because the functionality is shared between GRUB and Uboot. However, I wonder if we should make this class completely abstract and define those functions in the subclasses. Yes, it would mean a small amount of code is duplicated, but it would mean each bootloader is encapsulated. I don't know if it makes sense to have "default" methods here. Thoughts?

sonic_installer/main.py
```diff
@@ -1,54 +1,52 @@
...
-BOOTLOADER_TYPE_GRUB = 'grub'
-BOOTLOADER_TYPE_UBOOT = 'uboot'
-ARCH = platform.machine()
-BOOTLOADER = BOOTLOADER_TYPE_UBOOT if ("arm" in ARCH) or ("aarch64" in ARCH) else BOOTLOADER_TYPE_GRUB
+
+from .common import run_command
+
+from .bootloader.aboot import AbootBootloader
+from .bootloader.grub import GrubBootloader
+from .bootloader.uboot import UbootBootloader
```
qiluo-msft, line 16: You can import many in `bootloader.__init__.py`, and here `from bootloader import AbootBootloader, GrubBootloader, UbootBootloader`.

The benefit is hiding the file structure inside subdir.

sonic_installer/bootloader/bootloader.py
```diff
@@ -0,0 +1,70 @@
...
+
+        return IMAGE_PREFIX + version_num
+
+    def verify_binary_image(self, image_path):
+        """verify that the image is supported by the bootloader"""
+        return os.path.isfile(image_path)
+
+    @classmethod
+    def detect(cls):
+        return False
```
qiluo-msft, line 69: How about creating a factory method in base class and create subclass after detecting. It is weird to let every subclasses detect one after another.

sonic_installer/common.py
Staphylo: I agree with Joe regarding a the shebang. It's not necessary unless the python file is +x and meant to be executed directly.
I can add a docstring to all the files.

sonic_installer/bootloader/aboot.py
Staphylo: There wouldn't be any advantage of doing so since the import would still be necessary for every python file. In the end you need to import what you need having import indirection doesn't really simplify anything and makes code browsing more cumbersome.

sonic_installer/common.py
Staphylo: Sure I can move it to the file that needs it.
It felt like a useful helper to have around and in `common.py` but if you prefer I can move it to the appropriate file now and once someone needs it somewhere else they'll move it back.

sonic_installer/bootloader/bootloader.py
Staphylo: That was indeed the purpose. Your comment makes sense to me.
What do you feel about creating an `OnieInstallerBootloader` class between `Bootloader` and `(GrubBootloader, UbootBootloader)` ?
In the end the only thing shared by these 2 bootloaders is the format of the installer image.

sonic_installer/main.py
Staphylo: Joe seemed like he was not too keen on having code in `__init__.py`.
If he is fine with such a change I will do it and likely move the `BOOTLOADERS` definition there too.

sonic_installer/bootloader/bootloader.py
Staphylo: Do you want something like the following ?

```
class Bootloader(object):
     @classmethod
     def get_bootloader(cls):
           return <detectedBootloader>
```

With this approch Bootloader can never have access to BOOTLOADER variable since it would be a cyclic dependency.

The only option would be to add a `register_bootloader` and have every python file call it or use a decorator on the class.

```
_bootloaders = []

def register_bootloader(cls):
    _bootloader.append(cls)

class Bootloader(object):
     @classmethod
     def get_bootloader(cls)
          return [b for b in _bootloaders if b.detect()[0]
```

But overall it's just adding complexity for something that will never scale that much.

Also there is no other way around letting each bootloader detect if they are being in used.
Otherwise you would need to move bootloader specific code outside of their own files which I believe defeats the purpose of this approach in a first place.

sonic_installer/bootloader/bootloader.py
Staphylo: Let me know what you think of this new version.

sonic_installer/main.py
jleveque: I am opposed to defining classes in the \_\_init\_\_.py file. As I stated above, the primary purpose of the \_\_init\_\_.py file is to initialize the package, not to hold content. Qi's request here is reasonable. This person's description aligns with my thoughts on the usage of \_\_init\_\_.py: https://mikegrouchy.com/blog/be-pythonic-__init__py

sonic_installer/bootloader/bootloader.py
jleveque: I think this proposal is reasonable, and eliminates the duplicate code via the intermediate `OnieInstallerBootloader` class.

sonic_installer/common.py
qiluo-msft: Please help move it. I see so many reference to this file from bootloader.

---
In reply to: [443153279](https://github.com/Azure/sonic-utilities/pull/953#discussion_r443153279) [](ancestors = 443153279)

sonic_installer/main.py
qiluo-msft: The `get_bootloader` function looks good to me.

There is a good reference if related: [https://mehulkar.com/blog/2017/04/the-factory-pattern-and-circular-dependencies/](https://mehulkar.com/blog/2017/04/the-factory-pattern-and-circular-dependencies/)

---
In reply to: [443153452](https://github.com/Azure/sonic-utilities/pull/953#discussion_r443153452) [](ancestors = 443153452)

sonic_installer/common.py
Staphylo: Please be more argumentative about what problem you want me to solve here and how.
I fail to see the logic of your comment since main.py and all bootloader files need access to the variables/functions exposed by common.py
Moving common.py to bootloader would make no sense to me since main.py should not grab the `run_command` from bootloader since it doesn't have anything to do with what a bootloader is.
It is a utility module and to me lies at the perfect place.

sonic_installer/main.py
Staphylo: Sure I'll move some of the declaration to __init__.py

Regarding your link Qi, I personally disagree with the logic used by this post.
For me a Base class should never have a function that refers and import all its base classes.
IMO it's bad design, I much prefer self registering class over this approach.
Also this example is simplisistic now you have 1M animals, good luck with your 2M line function.
If the logic to detect each animal gets complicated now you have a huge spaghetti mess function that offloads the animal specific detection to the base class which I would never agree to in a review.

I'll do another pass and you guys can comment on my new approach.

jleveque:
Retest this please

qiluo-msft:
Retest this please

PR #960, opened 2020-06-24
Update main.py

zhouglgh:
'state == ok ' was  no longer supported now.

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -2214,7 +2214,7 @@ def bind(ctx, interface_name, vrf_name):
     state_db = SonicV2Connector(host='127.0.0.1')
     state_db.connect(state_db.STATE_DB, False)
     _hash = '{}{}'.format('INTERFACE_TABLE|', interface_name)
-    while state_db.get(state_db.STATE_DB, _hash, "state") == "ok":
+    while state_db.get_all(state_db.STATE_DB, _hash) != None:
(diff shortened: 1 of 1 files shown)


config/main.py
```diff
@@ -2214,7 +2214,7 @@ def bind(ctx, interface_name, vrf_name):
     state_db = SonicV2Connector(host='127.0.0.1')
     state_db.connect(state_db.STATE_DB, False)
     _hash = '{}{}'.format('INTERFACE_TABLE|', interface_name)
-    while state_db.get(state_db.STATE_DB, _hash, "state") == "ok":
+    while state_db.get_all(state_db.STATE_DB, _hash) != None:
```
prsunny, line 2217: Not sure I understand this change. Why do you say it is no longer supported? Is there any PR that you can point which resulted in this?

zhouglgh: For example：
We have Vrf10 and Ethernet10 with 10.10.10.10.
using commands "sudo config interface bind Ethernet10 Vrf10" will lead to an error: There is no "master Vrf10" when use command "ip a show Ethernet10" to check its master.

jleveque (changes requested):
Please update PR title to be more descriptive of the patch.

PR #962, opened 2020-06-24
Enhance SONiC with kubernetes management commands

renukamanavalan:
**Config commands:**
```
sudo config kubernetes server ip <IP address>
sudo config kubernetes server insecure <True/False>
sudo config kubernetes server disable <True/False>
sudo config kubernetes label add <key> <value>
sudo config kubernetes label drop <key>
sudo config kubernetes join [-a/--async] [-f/--force]
sudo config kubernetes reset

```
**Show commands:**
```
show kube server
show kube status
show kube nodes
show kube pods

```
A python wrapper for kube_join for invocation by other tools.

**- What I did**

**- How I did it**

**- How to verify it**


config/kube.py
```diff
@@ -0,0 +1,392 @@
...
+    """Specify a kubernetes ckuster VIP"""
+    if not netaddr.IPAddress(vip):
+        click.echo('Invalid IP address %s' % vip)
+        return
+    _update_kube_server('IP', vip)
+server.add_command(ip)
+
+#cmd kubernetes server insecure
+@click.command()
+@click.argument('option', type=click.Choice(["True", "False"]))
```
jleveque, line 346: We don't use "True"/"False" choices in any other commands that I'm aware of. Suggest changing to "on"/"off" to align with other commands?

config/kube.py
```diff
@@ -0,0 +1,392 @@
...
+def insecure(option):
+    """Specify a kubernetes ckuster VIP access as insecure or not"""
+    _update_kube_server('insecure', option)
+server.add_command(insecure)
+
+#cmd kubernetes server disable
+@click.command()
+@click.argument('option', type=click.Choice(["True", "False"]))
```
jleveque, line 354: We don't use "True"/"False" choices in any other commands that I'm aware of. Suggest changing to "on"/"off" to align with other commands?

config/kube.py
```diff
@@ -0,0 +1,392 @@
...
+    """ Server configuration """
+    pass
+kubernetes.add_command(server)
+
+
+#cmd kubernetes server IP
+@click.command()
+@click.argument('vip')
+def ip(vip):
+    """Specify a kubernetes ckuster VIP"""
```
jleveque, line 337: Typo: s/ckuster/cluster/

config/kube.py
```diff
@@ -0,0 +1,392 @@
...
+        click.echo('Invalid IP address %s' % vip)
+        return
+    _update_kube_server('IP', vip)
+server.add_command(ip)
+
+#cmd kubernetes server insecure
+@click.command()
+@click.argument('option', type=click.Choice(["True", "False"]))
+def insecure(option):
+    """Specify a kubernetes ckuster VIP access as insecure or not"""
```
jleveque, line 348: Typo: s/ckuster/cluster/

config/kube.py
```diff
@@ -0,0 +1,392 @@
...
+    _update_kube_server('insecure', option)
+server.add_command(insecure)
+
+#cmd kubernetes server disable
+@click.command()
+@click.argument('option', type=click.Choice(["True", "False"]))
+def disable(option):
+    """Specify a kubernetes ckuster VIP access is disabled or not"""
```
jleveque, line 356: Typo: s/ckuster/cluster/

config/kube.py
renukamanavalan: Switched to On/Off

config/kube.py
renukamanavalan: Switched to On/Off

config/kube.py
renukamanavalan: fixed.

config/kube.py
renukamanavalan: fixed.

config/kube.py
renukamanavalan: fixed.

renukamanavalan:
The last/outstanding alerts from lgtm can be ignored. It looks like, the imports are unused, but they are used, through class method of click.

jleveque:
@renukamanavalan: It appears the LGTM check is correct -- those imports are unused. You don't access the kube module in either the show/main.py or config/main.py files.

renukamanavalan:
Not really. of course I tested this code to work.

**Here is some high level insight:**
For example, in case of config, the when main.py imports,  kube, the kube.py adds "kubernetes" as command to config group. The config is declared via a class method of Click.group. Being class method, it is globally available. This class method uses Abbreviation group to hold all registered commands.  Being class method, it has to maintain a single instance of Abbreviation, hence kube.by, can join itself with all other commands added to config, via main.py, transparently. Now the Abbreviation group has the command, which is used by click. Also all the entire kube.py is imported, hence when sub-commands from kube are invoked, they are available locally.

**Test:**
When you type "sudo config <tab>",  it shows "kubernetes"
sudo config kubernetes <tab> -- shows all the subcommands and so on


```
admin@str-s6000-acs-13:/usr/lib/python2.7/dist-packages/config$ sudo config 
aaa                     interface               loopback                reload                  vlan
acl                     interface_naming_mode   mirror_session          route                   vrf
bgp                     is_loopback_name_valid  nat                     save                    warm_restart
container               kdump                   ntp                     sflow                   watermark
dropcounters            **kubernetes**              pfcwd                   snmpagentaddress        ztp
ecn                     load                    platform                snmptrap                
feature                 load_mgmt_config        portchannel             syslog                  
hostname                load_minigraph          qos                     tacacs         

admin@str-s6000-acs-13:/usr/lib/python2.7/dist-packages/config$ sudo config **kube** 
join    label   reset   server  
admin@str-s6000-acs-13:/usr/lib/python2.7/dist-packages/config$ sudo config kube server 
disable   insecure  ip        

```
Precisely same  with show too.

The magic is from the class method of click.
The lGTM complains, because it does not see, main.py calling any method from kube (no kube.* calls).

jleveque:
But those subcommands which access the kube module reside in separate files, and those files import kube, so that should be all which is necessary.

show/kube.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/python -u
+# -*- coding: utf-8 -*-
+    
+from show.main import AliasedGroup, cli, run_command
+import os
+from swsssdk import ConfigDBConnector
+
+KUBECTL_CMD = "kubectl --kubeconfig /etc/sonic/kube_admin.conf"
```
qiluo-msft, line 8: >/etc/sonic/kube_admin.conf [](start = 36, length = 26)

What happens if this file is missing?

show/kube.py
```diff
@@ -0,0 +1,65 @@
...
+REDIS_KUBE_TABLE = 'KUBERNETES_MASTER'
+REDIS_KUBE_KEY = 'SERVER'
+
+
+def hostname():
+    return os.uname()[1]
+
+def _get_db_data(table, key):
```
qiluo-msft, line 16: >_get_db_data [](start = 4, length = 12)

_get_configdb_table_key

show/kube.py
```diff
@@ -0,0 +1,65 @@
...
+
+@kubernetes.command()
+def status():
+    """Descibe this node"""
+    run_command("{} describe node {}".format(KUBECTL_CMD, hostname()))
+
+@kubernetes.command()
+def server():
+    """Show kube configuration"""
+    db_data = _get_db_data(REDIS_KUBE_TABLE, REDIS_KUBE_KEY)
```
qiluo-msft, line 58: >db_data [](start = 4, length = 7)

kube_fvs

show/kube.py
renukamanavalan: Good catch. Thanks. Code updated to handle.

show/kube.py
renukamanavalan: Changed to _get_configdb_data. Table & key are implied through arg names.

show/kube.py
renukamanavalan: I don't  get this comment. The current o/p is as below (_similar to 'show aaa'_)

admin@str-s6000-acs-13:~$ show kube server
KUBERNETES_MASTER SERVER IP 40.77.230.238
KUBERNETES_MASTER SERVER insecure True
KUBERNETES_MASTER SERVER disable false
admin@str-s6000-acs-13:~$

config/main.py
```diff
@@ -621,6 +621,9 @@ def config():
 
 config.add_command(aaa.aaa)
 config.add_command(aaa.tacacs)
+
+import kube
```
jleveque, line 625: Change to `from . import kube`. This should fix the LGTM alerts.

Also add a comment like you did in show/main.py ("# Import kube commands")

show/main.py
```diff
@@ -1621,6 +1621,9 @@ def protocol(verbose):
     from .bgp_frr_v6 import bgp
     ipv6.add_command(bgp)
 
+# Import kube commands
+import kube
+
```
jleveque, line 1626: Change to `from . import kube`. This should fix the LGTM alerts.

renukamanavalan: Thanks Joe! Trying...

jleveque: It didn't solve the problem. Can you refactor your code to use the `add_command` syntax above and move this import to the top of the file, rather than trying to import the entire file inline in the middle? It's not "Pythonic".

renukamanavalan: That would cause LGTM to raise cyclic warning, as kube imports main too. 

The right way to do is to re-org/re-factor the main.py such that, all helper commands go to a common.py. This way kube would just include common.py (not main.py) and main.py would do config.add_command(kubernetes) and all good.

This is a bigger job. I would rather take that as a separate PR, dedicated to re-factor.

jleveque: I completely agree. I have opened an issue to track this: https://github.com/Azure/sonic-utilities/issues/967. However, you're going to cause 2 LGTM alerts either way, so why not do the refactor now to prevent extra work in the future?

renukamanavalan: I did start and then realized (especially show/main.py), it is much more invasive. Then I decided not to clutter this work item, with changes that are not related to it.

I do agree, that we need to this. Thanks for filing the issue.

renukamanavalan: BTW, you might have noticed this already, that two pys importing each other is already there. For an an example show/main.py and show/bgp_frr_v4.py import each other. This is because, bgp_v4_frr uses some helper functions in main.py. In some cases, to avoid, they duplicate the code.

A refactoring would touch many .py files. So it would be a standalone task by itself, as cleanup.

renukamanavalan:
retest default please

renukamanavalan:
retest default please.

renukamanavalan:
retest this please

renukamanavalan:
retest this please

renukamanavalan:
retest this please

renukamanavalan:
retest this please

daall:
retest this please

renukamanavalan:
retest this please

config/main.py
renukamanavalan: Alerts fixed.

show/main.py
renukamanavalan: Alerts are fixed. Thanks to all suggestions, which helped.

show/main.py
```diff
@@ -1621,6 +1621,13 @@ def protocol(verbose):
     from .bgp_frr_v6 import bgp
     ipv6.add_command(bgp)
 
+# Import kube commands
+from . import kube
+
+# A dummy call to denote that this kube is not unused
+# Else LGTM is unable to understand the indirect use through cli.group class method
+kube.hostname()
```
jleveque, line 1629: Thanks for fixing the alerts. However, I think I'd rather have the alerts than the dummy function call. Since I opened the issue regarding restructuring the files, at least the alerts serve as a reminder that this needs attention.

Other thoughts here?

renukamanavalan: I too was thinking in the same line ... But later realized that it would not work, for the following reason.

W.r.t kube, we clearly know, that the "alert from LGTM" is incorrect. This is because, it is tough for a tool to figure out this indirect usage link. But if I leave this alert as "not fixed", anytime someone touches the config/main.py or show/main.py, this alert would appear and the author & reviewers of that PR are less likely to have any idea on this indirect usage. 

Hence in cases, where we know clearly, suppressing explicitly with comments help other devs understand.

renukamanavalan: This code is not the first one to introduce this issue, of .py importing each other. It has been there but suppressed, by duplicating code or other means, which is in sync with your point.

I guess, the issue you filed should be good enough. Else we have to watch every PR that touches these files and remember to add a comment, on why-to-ignore.

jleveque: Yeah, I see both sides to this, as well, but I think it's best to remove unnecessary code and like you said, watch every PR that touches these files and remember to add a comment about why to ignore. This might also push us to do the refactor sooner :)

renukamanavalan: I don't really intend to carry a baggage around. Once  we finish the job, finish it clean and move on,  unless I put the refactoring work on myself in the next few weeks. At this point, we have no idea, by when this work could get done. 

BTW, I added explicit comments on that single line of code, which nearly a no-op.

show/kube.py
qiluo-msft: Sorry I mean the var name: db_data -> kube_fvs
`db_data` is too general to me.

---
In reply to: [446318041](https://github.com/Azure/sonic-utilities/pull/962#discussion_r446318041) [](ancestors = 446318041)

renukamanavalan: done.

config/kube.py
```diff
@@ -0,0 +1,359 @@
...
+    _do_join(db_data['IP'], db_data['insecure'])
+
+
+
+@click.group(cls=AbbreviationGroup)
+def kubernetes():
+    """kubernetes command line"""
+    pass
+
+config.add_command(kubernetes)
```
qiluo-msft, line 276: If you move this line into main.py, you don't need to bypass LGTM alert.

renukamanavalan: No. That brings in another alert, because each file is importing each other. BTW, that was my  original code.

Thanks for the suggestion, though.

jleveque: True, but then we get a cyclical import alert. Renuka and I discussed this in the comments above. We have other files causing the cyclical import alerts. I have opened an issue to track the proper fix: https://github.com/Azure/sonic-utilities/issues/967

qiluo-msft: And you don't need to `from config.main import config`

---
In reply to: [451872241](https://github.com/Azure/sonic-utilities/pull/962#discussion_r451872241) [](ancestors = 451872241)

renukamanavalan: I need other functions from main.py, unless I duplicate.

renukamanavalan: Right solution is re-factoring as suggested by Joe. There are already other issues due to this lack of re-factoring, but suppressed. Need a proper cleanup.

show/kube.py
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/python -u
+# -*- coding: utf-8 -*-
+    
+import os
+
+def hostname():
```
qiluo-msft, line 6: Duplicated code should be moved into swsssdk. Or try similar function `socket.gethostname()` ?

config/kube.py
```diff
@@ -0,0 +1,359 @@
...
+
+logger_identity = "kube_config"
+log_level = syslog.LOG_DEBUG
+
+KUBE_ADMIN_CONF = "/etc/sonic/kube_admin.conf"
+KUBELET_YAML = "/var/lib/kubelet/config.yaml"
+KUBELET_SERVICE = "/etc/systemd/system/multi-user.target.wants/kubelet.service"
+
+SERVER_ADMIN_URL = "https://{}/admin.conf"
+KUBEADM_JOIN_CMD = "kubeadm join --discovery-file /etc/sonic/kube_admin.conf --node-name {}"
```
qiluo-msft, line 33: >/etc/sonic/kube_admin.conf [](start = 50, length = 26)

Use KUBE_ADMIN_CONF?

show/kube.py
jleveque: swsssdk is deprecated. swsscommon is the new standard.

Most of these functions don't necessarily belong in swsscommon; they belong in a shared library in sonic-utilities which all utilities can import.

config/kube.py
```diff
@@ -0,0 +1,359 @@
...
+        print("{}: {}:{}: {}".format(logger_identity, fname, ln, m))
+
+def _log_info(m):
+    _log_msg(syslog.LOG_INFO, inspect.stack()[1][1], inspect.stack()[1][2], m)
+
+def _log_err(m):
+    _log_msg(syslog.LOG_ERR, inspect.stack()[1][1], inspect.stack()[1][2], m)
+
+def _log_debug(m):
+    _log_msg(syslog.LOG_DEBUG, inspect.stack()[1][1], inspect.stack()[1][2], m)
```
qiluo-msft, line 57: Can you reuse existing log functions?

config/kube.py
```diff
@@ -0,0 +1,359 @@
...
+
+LOCK_FILE = "/var/lock/kube_join.lock"
+
+def _log_msg(lvl, fname, ln, m):
+    if lvl <= log_level:
+        syslog.openlog(logger_identity)
+        syslog.syslog(lvl, m)
+        syslog.closelog()
+
+    if (
```
qiluo-msft, line 43: >( [](start = 7, length = 1)

Make first condition in the same line of `(` or make `)` below last condition

config/kube.py
```diff
@@ -0,0 +1,359 @@
...
+
+def _log_debug(m):
+    _log_msg(syslog.LOG_DEBUG, inspect.stack()[1][1], inspect.stack()[1][2], m)
+
+def _do_exit(m):
+    m = "FATAL failure: {}. Exiting...".format(m)
+    _log_msg(syslog.LOG_ERR, inspect.stack()[1][1], inspect.stack()[1][2], m)
+    raise SystemExit(m) 
+
+def _get_db_data(table, key):
```
qiluo-msft, line 64: >_get_db_data [](start = 4, length = 12)

Same as `_get_configdb_data()`, reuse?

config/kube.py
```diff
@@ -0,0 +1,359 @@
...
+            server, fname, update_file)
+    run_command(cmd)
+
+    shutil.copyfile(update_file, KUBE_ADMIN_CONF)
+
+    run_command("rm -f {} {}".format(fname, update_file))
+
+
+def _is_connected(server=""):
+    if (
```
qiluo-msft, line 128: >( [](start = 7, length = 1)

the same

config/kube.py
```diff
@@ -0,0 +1,359 @@
...
+            if d:
+                o = urlparse(d)
+                if o.hostname:
+                    return not server or server == o.hostname
+    return False
+                
+
+
+
+def _get_labels():
```
qiluo-msft, line 148: Too many empty lines

jleveque: Agreed. @renukamanavalan: Until we have an automated linter in place, I suggest passing all Python code through http://pep8online.com/ and make changes based on the issues it finds (you can ignore the "line too long" errors, unless the line is exceptionally long).

config/kube.py
renukamanavalan: done

config/kube.py
renukamanavalan: Two reasons:
1) These functions provide additional info, function name & line no.
2) In case of interactive mode, it prints those log lines to the stdout, which is very helpful, to get additional data in stdout, for the user invoking these commands.

show/kube.py
renukamanavalan: This helps override the hostname for kube use, as needed.

config/kube.py
renukamanavalan: done

config/kube.py
renukamanavalan: I will defer this to re-factoring, as it is not right to include the file from show into config.

config/kube.py
renukamanavalan: done

show/kube.py
renukamanavalan: Even if we fix on hostname, again it is part of re-factoring.

config/kube.py
renukamanavalan: You can see, every .py has its own get/set of config DB else duplicated all over. A re-factoring work is required to cleanup all.

config/kube.py
renukamanavalan: Verified against the link provided by Joe. Thanks Joe!

renukamanavalan:
retest this please.

renukamanavalan:
retest this please

renukamanavalan:
retest this please

renukamanavalan:
retest this please

renukamanavalan:
retest this please

show/main.py
renukamanavalan: All LGTM comments resolved w/o any hack.

renukamanavalan:
retest this please

config/main.py
```diff
@@ -19,6 +19,7 @@
 from utilities_common.intf_filter import parse_interface_in_filter
 from utilities_common.util_base import UtilHelper
 from portconfig import get_child_ports, get_port_config_file_name
+from utilities_common.common import *
```
qiluo-msft, line 22: >* [](start = 36, length = 1)

Do not import wildcard in prod code.

qiluo-msft:
> def  interface_is_in_vlan(vlan_member_table, interface_name):

Use one space instead of two, also below.

---
Refers to: config/main.py:741 in 775e1f8. [](commit_id = 775e1f8b2da60dfbd7abbf227ff25fd8cd5b696b, deletion_comment = False)

qiluo-msft:
>     for intf in interface_table.keys():

If this is a dict, you don't need `.keys()`

---
Refers to: config/main.py:759 in 775e1f8. [](commit_id = 775e1f8b2da60dfbd7abbf227ff25fd8cd5b696b, deletion_comment = False)

config/main.py
```diff
@@ -73,45 +74,6 @@ def log_error(msg):
     syslog.closelog()
```
qiluo-msft, line 74: Code duplicated with common.py

config/main.py
renukamanavalan: LGTM complained too. Fixed per LGTM suggestion.
"`NEWImport pollutes the enclosing namespace, as the imported module utilities_common.common does not define '__all__'.`"

```
__all__ = ['log_info', 'log_err', 'log_warning', 'log_debug',
        'do_exit', 'get_configdb_data', 'run_command',
        'AbbreviationGroup']

```

renukamanavalan:
> > def  interface_is_in_vlan(vlan_member_table, interface_name):
> 
> Use one space instead of two, also below.
> 
> Refers to: config/main.py:741 in [775e1f8](https://github.com/Azure/sonic-utilities/commit/775e1f8b2da60dfbd7abbf227ff25fd8cd5b696b). [](commit_id = [775e1f8](https://github.com/Azure/sonic-utilities/commit/775e1f8b2da60dfbd7abbf227ff25fd8cd5b696b), deletion_comment = False)

Done

config/main.py
renukamanavalan: Good catch. Missed to remove. Will  put an update.

qiluo-msft:
> def  interface_is_in_vlan(vlan_member_table, interface_name):

One more in below function.

---
In reply to: [663732872](https://github.com/Azure/sonic-utilities/pull/962#issuecomment-663732872) [](ancestors = 663732872)

---
Refers to: config/main.py:741 in 775e1f8. [](commit_id = 775e1f8b2da60dfbd7abbf227ff25fd8cd5b696b, deletion_comment = False)

lguohan:
need to add unit test for this pr.

qiluo-msft:
Please help resolve conflict.

utilities_common/common.py
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/python -u
```
jleveque, line 1: Are you adding this file by accident?

renukamanavalan: Good catch. Yes accidental. Let me remove

renukamanavalan: No this file has been there, for a while. Part of re-factoring as requested by Qi. I just noticed, that recent merge has run_command in cli.py. So I dropped it from common.py.

Is there any moving around, you would prefer ?

jleveque: You are adding this file, it hasn't been there for a while.

renukamanavalan: Yes in this PR, few commits ago. There is no change since last approval, except dropping run_command, as it is now in cli.py.

Do you have any comments for this file?

utilities_common/common.py
```diff
@@ -0,0 +1,101 @@
...
+def log_error(msg, print_msg=False):
+    _log_msg(syslog.LOG_ERR, print_msg, inspect.stack()[1][1], inspect.stack()[1][2], msg)
+
+
+def log_warning(msg, print_msg=False):
+    _log_msg(syslog.LOG_WARNING, print_msg, inspect.stack()[1][1], inspect.stack()[1][2], msg)
+
+
+def log_debug(msg, print_msg=False):
+    _log_msg(syslog.LOG_DEBUG, print_msg, inspect.stack()[1][1], inspect.stack()[1][2], msg)
```
jleveque, line 48: These syslog wrappers are not needed, as common logging functionality is now handled by sonic-py-common. Please remove.

utilities_common/common.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def get_configdb_data(table, key):
+    config_db = ConfigDBConnector()
+    config_db.connect()
+    data = config_db.get_table(table)
+    return data[key] if key in data else None
+
+
+class AbbreviationGroup(click.Group):
```
jleveque, line 64: This class is now a duplicate, as it has already been moved to cli.py. Please remove.

utilities_common/common.py
```diff
@@ -0,0 +1,101 @@
...
+import os
+import syslog
+import click
+import inspect
+from swsssdk import ConfigDBConnector
+
+SYSLOG_IDENTIFIER = "config"
+
+def get_hostname():
+    return os.uname()[1]
```
jleveque, line 18: This is not CLI-specific. This is better suited for sonic-py-common.

utilities_common/common.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def log_debug(msg, print_msg=False):
+    _log_msg(syslog.LOG_DEBUG, print_msg, inspect.stack()[1][1], inspect.stack()[1][2], msg)
+
+
+def do_exit(msg):
+    m = "FATAL failure: {}. Exiting...".format(msg)
+    _log_msg(syslog.LOG_ERR, True, inspect.stack()[1][1], inspect.stack()[1][2], m)
+    raise SystemExit(m)
```
jleveque, line 54: This belongs in cli.py.

utilities_common/common.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def get_configdb_data(table, key):
+    config_db = ConfigDBConnector()
+    config_db.connect()
+    data = config_db.get_table(table)
+    return data[key] if key in data else None
```
jleveque, line 61: This seems better suited for db.py.

utilities_common/common.py
jleveque: Comments added.

renukamanavalan: Thanks Joe! I will get to each.

utilities_common/common.py
renukamanavalan: done.

utilities_common/common.py
renukamanavalan: Moved to sonic_py_common/device_info.py

utilities_common/common.py
renukamanavalan: done

utilities_common/common.py
renukamanavalan: done.

utilities_common/common.py
renukamanavalan: done.

config/main.py
```diff
@@ -929,6 +929,9 @@ def config(ctx):
 config.add_command(nat.nat)
 config.add_command(vlan.vlan)
 
+from .kube import kubernetes
```
jleveque, line 932: Can you move this import to the top of the file now? Are the cyclical dependencies fixed?

show/kube.py
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/python -u
+# -*- coding: utf-8 -*-
+
+import os
+import click
+
+from sonic_py_common import device_info
+from utilities_common.db import Db
+import utilities_common.cli as clicommon
```
jleveque, line 9: Organize per PEP8 standards:

```python
import os

import click
from utilities_common.db import Db

```

show/main.py
```diff
@@ -2424,5 +2424,8 @@ def tunnel():
 
     click.echo(tabulate(table, header))
 
+from .kube import kubernetes
```
jleveque, line 2427: Can you move this import to the top of the file now? Are the cyclical dependencies fixed?

utilities_common/cli.py
```diff
@@ -499,3 +499,10 @@ def run_command(command, display_cmd=False, ignore_error=False, return_cmd=False
     rc = proc.poll()
     if rc != 0:
         sys.exit(rc)
+
+
+def do_exit(msg):
+    m = "FATAL failure: {}. Exiting...".format(msg)
+    _log_msg(syslog.LOG_ERR, True, inspect.stack()[1][1], inspect.stack()[1][2], m)
+    raise SystemExit(m)
+
```
jleveque, line 508: Remove blank line from EOF. Just ensure file ends with a single newline.

utilities_common/db.py
```diff
@@ -8,3 +8,9 @@ def __init__(self):
         self.db.connect(self.db.APPL_DB)
         self.db.connect(self.db.CONFIG_DB)
         self.db.connect(self.db.STATE_DB)
+
+    def get_data(self, table, key):
+        data = self.cfgdb.get_table(table)
+        return data[key] if key in data else None
+
+
```
jleveque, line 16: Remove blank lines from EOF. Just ensure file ends with a single newline.

config/kube.py
```diff
@@ -0,0 +1,336 @@
...
+import fcntl
+import yaml
+import netaddr
+import shutil
+from urlparse import urlparse
+from swsssdk import ConfigDBConnector
+from utilities_common.db import Db
+from sonic_py_common import device_info
+import utilities_common.cli as clicommon
+from .utils import log
```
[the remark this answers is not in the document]
renukamanavalan: done. Thanks for sample too.

config/kube.py
```diff
@@ -0,0 +1,336 @@
...
+def drop(key):
+    """Drop a label from this node"""
+    if not key:
+        click.echo('Require key to drop')
+        return
+    _label_node("{}-".format(key))
+
+label.add_command(drop)
+
+
```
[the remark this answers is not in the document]
renukamanavalan: done.

config/main.py
renukamanavalan: yep. Moved.

show/kube.py
renukamanavalan: done.

show/main.py
renukamanavalan: Yes & moved.

utilities_common/cli.py
renukamanavalan: done

utilities_common/db.py
renukamanavalan: done.

config/kube.py
```diff
@@ -0,0 +1,337 @@
...
+    pass
+
+
+# cmd kubernetes join [-f/--force]
+@click.command()
+@click.option('-f', '--force', help='Force a join', is_flag=True)
+def join(force):
+    kube_join(force=force)
+
+kubernetes.add_command(join)
```
jleveque, line 251: Rather than calling `kubernetes.add_command(x)`, you should simply be able to change the decorator from:

```python
@click.command()
```

to:

```python
@kubernetes.command()
```

config/kube.py
```diff
@@ -0,0 +1,337 @@
...
+
+
+# cmd kubernetes reset
+@click.command()
+def reset():
+    kube_reset()
+
+kubernetes.add_command(reset)
```
jleveque, line 259: Rather than calling `kubernetes.add_command(x)`, you should simply be able to change the decorator from:

```python
@click.command()
```

to:

```python
@kubernetes.command()
```

config/kube.py
```diff
@@ -0,0 +1,337 @@
...
+
+
+# cmd kubernetes server
+@click.group()
+def server():
+    """ Server configuration """
+    pass
+
+kubernetes.add_command(server)
```
jleveque, line 268: Rather than calling `kubernetes.add_command(x)`, you should simply be able to change the decorator from:

```python
@click.command()
```

to:

```python
@kubernetes.command()
```

config/kube.py
```diff
@@ -0,0 +1,337 @@
...
+@click.command()
+@click.argument('vip')
+def ip(vip):
+    """Specify a kubernetes cluster VIP"""
+    if not netaddr.IPAddress(vip):
+        click.echo('Invalid IP address %s' % vip)
+        return
+    _update_kube_server('IP', vip)
+
+server.add_command(ip)
```
jleveque, line 281: Rather than calling `kubernetes.add_command(x)`, you should simply be able to change the decorator from:

```python
@click.command()
```

to:

```python
@server.command()
```

config/kube.py
```diff
@@ -0,0 +1,337 @@
...
+
+
+# cmd kubernetes server insecure
+@click.command()
+@click.argument('option', type=click.Choice(["on", "off"]))
+def insecure(option):
+    """Specify a kubernetes cluster VIP access as insecure or not"""
+    _update_kube_server('insecure', option == "on")
+
+server.add_command(insecure)
```
jleveque, line 291: Rather than calling `kubernetes.add_command(x)`, you should simply be able to change the decorator from:

```python
@click.command()
```

to:

```python
@server.command()
```

config/kube.py
```diff
@@ -0,0 +1,337 @@
...
+
+
+# cmd kubernetes server disable
+@click.command()
+@click.argument('option', type=click.Choice(["on", "off"]))
+def disable(option):
+    """Specify a kubernetes cluster VIP access is disabled or not"""
+    _update_kube_server('disable', option == "on")
+
+server.add_command(disable)
```
jleveque, line 301: Rather than calling `kubernetes.add_command(x)`, you should simply be able to change the decorator from:

```python
@click.command()
```

to:

```python
@server.command()
```

config/kube.py
```diff
@@ -0,0 +1,337 @@
...
+
+
+# cmd kubernetes label
+@click.group()
+def label():
+    """ label configuration """
+    pass
+
+kubernetes.add_command(label)
```
jleveque, line 310: Rather than calling `kubernetes.add_command(x)`, you should simply be able to change the decorator from:

```python
@click.command()
```

to:

```python
@kubernetes.command()
```

config/kube.py
```diff
@@ -0,0 +1,337 @@
...
+@click.argument('key', required=True)
+@click.argument('val', required=True)
+def add(key, val):
+    """Add a label to this node"""
+    if not key or not val:
+        click.echo('Require key & val')
+        return
+    _label_node("{}={}".format(key, val))
+
+label.add_command(add)
```
jleveque, line 324: Rather than calling `kubernetes.add_command(x)`, you should simply be able to change the decorator from:

```python
@click.command()
```

to:

```python
@label.command()
```

config/kube.py
```diff
@@ -0,0 +1,337 @@
...
+@click.command()
+@click.argument('key', required=True)
+def drop(key):
+    """Drop a label from this node"""
+    if not key:
+        click.echo('Require key to drop')
+        return
+    _label_node("{}-".format(key))
+
+label.add_command(drop)
```
jleveque, line 337: Rather than calling `kubernetes.add_command(x)`, you should simply be able to change the decorator from:

```python
@click.command()
```

to:

```python
@label.command()
```

config/main.py
```diff
@@ -28,6 +28,7 @@
 import nat
 import feature
 import vlan
+import kube
```
jleveque, line 31: Please add this in alphabetical order

show/main.py
```diff
@@ -25,6 +25,7 @@
 import feature
 import interfaces
 import vlan
+import kube
```
jleveque, line 28: Please add this in alphabetical order

config/kube.py
renukamanavalan: done

renukamanavalan: done

config/kube.py
renukamanavalan: done

config/kube.py
renukamanavalan: done

config/kube.py
renukamanavalan: done

config/kube.py
renukamanavalan: done

config/kube.py
renukamanavalan: done

config/kube.py
renukamanavalan: done

config/kube.py
renukamanavalan: done

config/main.py
renukamanavalan: done

show/main.py
renukamanavalan: done

config/main.py
```diff
@@ -929,6 +930,8 @@ def config(ctx):
 config.add_command(nat.nat)
 config.add_command(vlan.vlan)
 
+config.add_command(kube.kubernetes)
```
jleveque, line 933: Suggest adding this to the group above, in alphabetical order, adding a comment above and removing the NAT comment, like so:

```python
# Add groups from other modules
config.add_command(aaa.aaa)
config.add_command(aaa.tacacs)
config.add_command(feature.feature)
config.add_command(kube.kubernetes)
config.add_command(nat.nat)
config.add_command(vlan.vlan)
```

show/kube.py
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/python -u
+# -*- coding: utf-8 -*-
+
+import os
+
+import click
+
```
jleveque, line 7: Remove this blank line. `click`, `sonic_py_common` and `utilities_common` are all technically "third-party" libraries relative to `show`.

show/main.py
```diff
@@ -2424,5 +2425,7 @@ def tunnel():
 
     click.echo(tabulate(table, header))
 
+cli.add_command(kube.kubernetes)
```
jleveque, line 2428: Suggest to add a comment above this line:

```python
# Add groups from other modules
```

config/main.py
renukamanavalan: Done. Thanks for the 'comments'.

show/kube.py
renukamanavalan: Done

show/main.py
renukamanavalan: done

abdosi:
@renukamanavalan Do we need this in 201911 ? This seems to be enhancement.
If needed please create PR for 201911 as cherry-pick has conflict.

@rlhui

briantopping:
Apologies for the dumb question, is there a conversation about this change somewhere?

renukamanavalan:
> Apologies for the dumb question, is there a conversation about this change somewhere?
Good question. There is a HLD review on this Tuesday (9/22). Please join.
Here is work in progress. [link](https://github.com/renukamanavalan/SONiC/blob/kube_systemd/doc/kubernetes/Kuberenetes-support.md)
Soon I will create a PR for this doc.

briantopping:
Thanks for the link. I only found SONiC a few days ago, rather innocently in a search for an OSS ONIE NOS and choosing SONiC by project momentum. When I realized it ran Docker, I was blown away. So I started searching for whether anyone had tried putting Kubelet on it. I am ecstatic to find your effort in this regard.

My initial impressions is that "legacy SONiC" needs to be supported with `systemd` for those that will never join their switch into a control fabric such as with Kubernetes. I also noted a team somewhere that was using Docker Swarm to create a switch control fabric, so it's presumptive to assume Kubernetes will be the only control plane ever used.

The unifying aspect of these different control planes (ie systemd, Docker Swarm, Kubernetes) is they all use the Docker daemon socket for control. The `systemd` approach indirectly uses the socket via the Docker CLI app and shell scripts. Thus, if Docker itself on a node became the source of record for the state of a node, control plane fabrics could reasonably coexist. And to that end, it might help reduce interdependencies in the support matrix.

If it's worth connecting over this before the HLD review, please don't hesitate to reach out here or on my email.

renukamanavalan:
Hi @briantopping, 
Thanks for your interest and very helpful comments. 

Yes, there is Docker Swarm and also OpenShift. The reasons, we picked
a) We need to pick one for our first step.
b) Kubernetes is the most widely adopted
c)  It has excellent community support, worldwide.

Said that, the goal is to keep future adaptation to other technology/tool as simple as possible. If you look at the core design, it is still systemd that manages at service level. When it comes to down to docker level, it either runs docker command or instruct the remote system to start/stop/.... In Kubernetes, it controls it through labels and in docker swarm, it could be different.

This is achieved by replacing docker commands with container commands as  `docker start/stop/...` become `container start/stop/...`.  The container start, under the hood calls `docker` or `label create\remove` based on owner, which is local/kube. When we extend to docker-swarm, we would add one more owner type as, say "swarm" and update the container action commands will be updated to handle swarm.

Guess, we don't have much time to connect, before community review, which is at PDT 8AM tomorrow. Please join. We can connect after that for the follow ups.

briantopping:
Thanks for the updates, yes OpenShift and Rancher as well if you want to consider other Kubernetes-based distributions. 

I have a feeling that depending on `systemd` will seem obtuse in a few years, but there's a high probability I am missing requirements, details and nuances. 

As a user, I would like to see Kubernetes (or any orchestrator) run as unencumbered as possible. There is a lot of tribal knowledge baked into Kubernetes (for instance) and it's a nonzero possibility that the SONiC container orchestration team would miss one or more of these nuances as it tries to make k8s work through `systemd` instead of the Docker socket. 

So instead, what if the `systemd` implementation was refactored to act similarly as a client to `dockerd`? For instance, `systemctl status frr` (or whatever the container is called) would work directly with Docker to provide results, allowing other orchestrators to work from the same information source (docker itself). If it is already doing that, then why have the orchestrators talking to `systemd`? This shim seems to be a maintenance liability. 

I am thinking about the overall level of effort to get to a robust and bulletproof implementation that works across all the orchestrators you've named, especially in the face of future implementation changes that we cannot foresee today. If the `systemd` dependency is removed, we are not constantly tracking their version stream, just updating `kubelet` or whatever and moving on.

I regret belaboring this point if you understood what I was talking about from the previous post. Just wanted to save time tomorrow when there is a bigger audience whose valuable time is a consideration.

renukamanavalan:
I have to differ a little bit here. The Kubernetes is robust, but they don't handle *all* possible scenarios, which is improbable and becoming too complex to handle "all" will ultimately fall apart. Our SONiC scenario, where we plan to manage daemons only, is not the mainstream scenario for kubernetes. To tell you a sample, a daemonset does not support restartPolicy or backoff timer.  As you know, each switch is carrying production traffic and its health is of utmost importance, with minimal down time for any dockers in any scenario. The availability of kubernetes servers can't be assured. Plus, there are many scenarios, where a switch may decide not to join. I can go on ... :-)

What we did here is to take a first step, in a non-invasive fashion, but pretty effective in bringing in external support for container management. We are going to learn a lot from this step and this can pave our way to next.

Your vision of complete external control -- Need more than  kubernetes -- one more layer of orchestration that could monitor switches & kubernetes servers as a whole and take appropriate corrective action as and when needed. We are not there yet.

Thank you,
Renuka

briantopping:
Yes! That makes perfect sense, thank you for the explanation!! Grateful for your elaboration before tomorrow's meeting. 🙏🙏🙏

PR #972, opened 2020-06-30
Append default mask field if user has not provided one

prsunny:
**- What I did**
If user configures IP (v4 or v6) address without a mask, append the default `/32` or `/128` mask

**- How I did it**

**- How to verify it**
Redis command

```
config interface ip add Vlan1000 2001::1
config interface ip add Vlan1000 1.1.1.3
127.0.0.1:6379[4]> KEYS *VLAN_INTERFACE*
1) "VLAN_INTERFACE|Vlan1000|2001::1/128"
2) "VLAN_INTERFACE|Vlan1000|1.1.1.3/32"
```

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -2045,7 +2045,12 @@ def add(ctx, interface_name, ip_addr, gw):
             ctx.fail("'interface_name' is None!")
 
     try:
-        ipaddress.ip_network(unicode(ip_addr), strict=False)
+        net = ipaddress.ip_network(unicode(ip_addr), strict=False)
+        if '/' not in ip_addr:
+            if net.version == 4:
+                ip_addr = "%s/32" % ip_addr
+            else:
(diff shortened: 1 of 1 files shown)


config/main.py
```diff
@@ -2045,7 +2045,12 @@ def add(ctx, interface_name, ip_addr, gw):
             ctx.fail("'interface_name' is None!")
 
     try:
-        ipaddress.ip_network(unicode(ip_addr), strict=False)
+        net = ipaddress.ip_network(unicode(ip_addr), strict=False)
+        if '/' not in ip_addr:
```
pavel-shirshov, line 2049: I checked this code and found that `str(net)` will give correct ip address with the correct mask. So this could be changed to
`ip_addr = str(net)`

prsunny: Good point. Changed it!

PR #957, opened 2020-06-22
[filter-fdb] Check VLAN Presence When Filter FDB

tahmed-dev:
**- What I did**
FTOS fast conversion script generates bogus vlan that does not exist.
This PR uses config_db in order to verify that provided vlans exist
in the switch configuration.

signed-off-by: Tamer Ahmed <tamer.ahmed@microsoft.com>

**- How I did it**
Added new code to parse config_db.json and used this to filter out arp entries

**- How to verify it**
```
sonic-utilities$ sudo pytest sonic-utilities-tests/filter_fdb_entries_test.py  -vvvv
=================================================================================================================================================== test session starts ===================================================================================================================================================
platform linux2 -- Python 2.7.13, pytest-3.0.6, py-1.4.32, pluggy-0.4.0 -- /usr/bin/python
cachedir: sonic-utilities-tests/.cache
rootdir: /sonic/sonic-utilities/sonic-utilities-tests, inifile: pytest.ini
plugins: cov-2.4.0
collected 10 items 

sonic-utilities-tests/filter_fdb_entries_test.py::TestFilterFdbEntries::testFilterFdbEntries[testData0] PASSED
sonic-utilities-tests/filter_fdb_entries_test.py::TestFilterFdbEntries::testFilterFdbEntries[testData1] PASSED
sonic-utilities-tests/filter_fdb_entries_test.py::TestFilterFdbEntries::testFilterFdbEntries[testData2] PASSED
sonic-utilities-tests/filter_fdb_entries_test.py::TestFilterFdbEntries::testFilterFdbEntries[testData3] PASSED
sonic-utilities-tests/filter_fdb_entries_test.py::TestFilterFdbEntries::testFilterFdbEntries[testData4] PASSED
sonic-utilities-tests/filter_fdb_entries_test.py::TestFilterFdbEntries::testFilterFdbEntries[testData5] PASSED
sonic-utilities-tests/filter_fdb_entries_test.py::TestFilterFdbEntries::testFilterFdbEntries[testData6] PASSED
sonic-utilities-tests/filter_fdb_entries_test.py::TestFilterFdbEntries::testFilterFdbEntries[testData7] PASSED
sonic-utilities-tests/filter_fdb_entries_test.py::TestFilterFdbEntries::testFilterFdbEntries[testData8] PASSED
sonic-utilities-tests/filter_fdb_entries_test.py::TestFilterFdbEntries::testFilterFdbEntries[testData9] PASSED

[...]

PR #975, opened 2020-07-08
[filter-fdb] Check VLAN Presence When Filter FDB (#957)

tahmed-dev:
* [filter-fdb] Check VLAN Presence When Filter FDB

FTOS fast conversion script generates bogus vlan that does not exist.
This PR uses config_db in order to verify that provided vlans exist
in the switch configuration.

* review comments
making lgtm happy
Added two more test cases

* Update existing test case and adding new one

* adding support for filter ou based on vlan ip network

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/scripts/filter_fdb_entries.py
+++ b/scripts/filter_fdb_entries.py
@@ -8,9 +8,36 @@ import syslog
 import traceback
 import time
 
+from ipaddress import ip_address, ip_network, ip_interface
 from collections import defaultdict
 
-def get_arp_entries_map(filename):
+def get_vlan_cidr_map(filename):
+    """
+        Generate Vlan CIDR information from Config DB file
+
+        fdb entries could be contaminated with foreigh Vlan entries as seen in the case of 
+        FTOS fast conversion. SONiC Vlan CIDR configuration will be used to filter out
+        those invalid Vlan entries out.
+
+        Args:
+            filename(str): Config DB data file
+
+        Returns:
+            vlan_cidr(dict) map of Vlan CIDR configuration for SONiC device
+    """
+    with open(filename, 'r') as fp:
+        config_db_entries = json.load(fp)
+
+    vlan_cidr = defaultdict()
+    if "VLAN_INTERFACE" in config_db_entries.keys() and "VLAN" in config_db_entries.keys():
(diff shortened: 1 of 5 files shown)


scripts/filter_fdb_entries.py
```diff
@@ -8,33 +8,67 @@
...
+        Returns:
+            vlan_cidr(dict) map of Vlan CIDR configuration for SONiC device
+    """
+    with open(filename, 'r') as fp:
+        config_db_entries = json.load(fp)
+
+    vlan_cidr = defaultdict()
+    if "VLAN_INTERFACE" in config_db_entries.keys() and "VLAN" in config_db_entries.keys():
+        for vlan_key in config_db_entries["VLAN_INTERFACE"].keys():
+            vlan, cidr = tuple(vlan_key.split('|'))
```
prsunny, line 34: There could be entry without cidr in config_db. You may want to skip such entries or it may fail. Can you also put those entries in test_vector and verify. 

            "VLAN_INTERFACE": {
                "Vlan1000": {}
            },

tahmed-dev: Good catch. did not know such entries exists. @yxieca FYI... we might want to hold off on .81.

prsunny: Actually, these entries exists in 201911 and master. 201811 doesn't have such entries FYI

tahmed-dev: @prsunny, is this a bug in these branches! I did not recall seeing them, however when you alerted me, it happened that I had master image on my DUT and I saw it! I created a fix for it. No, I am inclined to know why they exist on master and on 201911. Sounds like a bug there that needs to be addressed.

prsunny: it is not a bug, but a feature introduced as part of VRF in 201911 and above

abdosi: @tahmed-dev is above comment has been addressed ?

tahmed-dev: @abdosi, yes it is fixed on master. please cherry-pick this [PR](https://github.com/Azure/sonic-utilities/pull/976) as well.

PR #979, opened 2020-07-09
Support to verify reboot for secure boot

xumia:
**- What I did**
Add a command "sonic_installer verify_reboot" to verify the image before reboot.
The command will be called when reboot/fast-reboot/warm-reboot.

**- How I did it**

**- How to verify it**

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -33,7 +33,6 @@ EXIT_SYNCD_SHUTDOWN=11
 EXIT_FAST_REBOOT_DUMP_FAILURE=12
 EXIT_FILTER_FDB_ENTRIES_FAILURE=13
 EXIT_NO_CONTROL_PLANE_ASSISTANT=20
-EXIT_SONIC_INSTALLER_VERIFY_REBOOT=21
 
 function error()
 {
@@ -307,11 +306,7 @@ function reboot_pre_check()
     fi
 
     # Verify the next image by sonic_installer
-    local message=$(sonic_installer verify_next_image 2>&1)
-    if [ $? -ne 0 ]; then
-        debug "Failed to verify next image: ${message}"
-        exit ${EXIT_SONIC_INSTALLER_VERIFY_REBOOT}
-    fi
+    sonic_installer verify-next-image
     
     # Make sure ASIC configuration has not changed between images
     ASIC_CONFIG_CHECK_SCRIPT="/usr/bin/asic_config_check"
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -307,9 +307,9 @@ function reboot_pre_check()
     fi
 
     # Verify the next image by sonic_installer
-    local message=$(sonic_installer verify-next-image 2>&1)
+    local message=$(sonic_installer verify_next_image 2>&1)
     if [ $? -ne 0 ]; then
-        debug "Failed to verify next image by running command: sonic_installer verify-next-image, result message: ${message}"
+        debug "Failed to verify next image by running command: sonic_installer verify_next_image, result message: ${message}"
         exit ${EXIT_SONIC_INSTALLER_VERIFY_REBOOT}
     fi
     
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -309,7 +309,7 @@ function reboot_pre_check()
     # Verify the next image by sonic_installer
     local message=$(sonic_installer verify_next_image 2>&1)
     if [ $? -ne 0 ]; then
-        debug "Failed to verify next image by running command: sonic_installer verify_next_image, result message: ${message}"
+        debug "Failed to verify next image: ${message}"
         exit ${EXIT_SONIC_INSTALLER_VERIFY_REBOOT}
     fi
     
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -32,6 +32,7 @@ EXIT_SYNCD_SHUTDOWN=11
 EXIT_FAST_REBOOT_DUMP_FAILURE=12
 EXIT_FILTER_FDB_ENTRIES_FAILURE=13
 EXIT_NO_CONTROL_PLANE_ASSISTANT=20
+EXIT_SONIC_INSTALLER_VERIFY_REBOOT=21
 
 function error()
 {
@@ -304,10 +305,11 @@ function reboot_pre_check()
         exit ${EXIT_FILE_SYSTEM_FULL}
     fi
 
-    # Make sure that the next image exists
-    if [[ ! -d ${IMAGE_PATH} ]]; then
-        debug "Next image ${NEXT_SONIC_IMAGE} doesn't exist ..."
-        exit ${EXIT_NEXT_IMAGE_NOT_EXISTS}
+    # Verify reboot by sonic_installer
+    local message=$(sonic_installer verify-reboot)
+    if [ $? -ne 0 ]; then
+        debug "Failed to verify next reboot by running command: sonic_installer verify-reboot, result message: ${message}"
+        exit ${EXIT_SONIC_INSTALLER_VERIFY_REBOOT}
     fi
     
     # Make sure ASIC configuration has not changed between images
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -33,6 +33,7 @@ EXIT_SYNCD_SHUTDOWN=11
 EXIT_FAST_REBOOT_DUMP_FAILURE=12
 EXIT_FILTER_FDB_ENTRIES_FAILURE=13
 EXIT_NO_CONTROL_PLANE_ASSISTANT=20
+EXIT_SONIC_INSTALLER_VERIFY_REBOOT=21
 
 function error()
 {
@@ -306,7 +307,12 @@ function reboot_pre_check()
     fi
 
     # Verify the next image by sonic_installer
-    sonic_installer verify-next-image
+    INSTALLER_VERIFY_RC=0
+    sonic_installer verify-next-image > /dev/null || INSTALLER_VERIFY_RC=$?
+    if [[ INSTALLER_VERIFY_RC -ne 0 ]]; then
+        error "Failed to verify next image. Exit code: $INSTALLER_VERIFY_RC"
+        exit ${EXIT_SONIC_INSTALLER_VERIFY_REBOOT}
+    fi
     
     # Make sure ASIC configuration has not changed between images
     ASIC_CONFIG_CHECK_SCRIPT="/usr/bin/asic_config_check"
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -306,10 +306,10 @@ function reboot_pre_check()
         exit ${EXIT_FILE_SYSTEM_FULL}
     fi
 
-    # Verify reboot by sonic_installer
-    local message=$(sonic_installer verify-reboot)
+    # Verify the next image by sonic_installer
+    local message=$(sonic_installer verify-next-image 2>&1)
     if [ $? -ne 0 ]; then
-        debug "Failed to verify next reboot by running command: sonic_installer verify-reboot, result message: ${message}"
+        debug "Failed to verify next image by running command: sonic_installer verify-next-image, result message: ${message}"
         exit ${EXIT_SONIC_INSTALLER_VERIFY_REBOOT}
     fi
     
--- a/scripts/reboot
+++ b/scripts/reboot
@@ -81,7 +81,7 @@ function reboot_pre_check()
     rm ${filename}
 
     # Verify the next image by sonic_installer
-    local message=$(sonic_installer verify_next_image 2>&1)
+    local message=$(sonic_installer verify-next-image 2>&1)
     if [ $? -ne 0 ]; then
         VERBOSE=yes debug "Failed to verify next image: ${message}"
         exit ${EXIT_SONIC_INSTALLER_VERIFY_REBOOT}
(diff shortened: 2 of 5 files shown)


qiluo-msft:
@Staphylo Could you please help review ?

sonic_installer/bootloader/bootloader.py
```diff
@@ -43,8 +52,23 @@ def verify_binary_image(self, image_path):
         """verify that the image is supported by the bootloader"""
         raise NotImplementedError
 
+    def verify_reboot(self):
```
qiluo-msft, line 55: >verify_reboot [](start = 8, length = 13)

Better name: verify_next_image

sonic_installer/bootloader/bootloader.py
```diff
@@ -43,8 +52,23 @@ def verify_binary_image(self, image_path):
...
+
     @classmethod
     def detect(cls):
         """returns True if the bootloader is in use"""
         return False
 
+    @classmethod
+    def get_image_path(cls, image):
+        """returns the image path"""
+        prefix = HOST_PATH + '/' + IMAGE_DIR_PREFIX
```
qiluo-msft, line 72: >'/' [](start = 29, length = 3)

os.join()

sonic_installer/main.py
```diff
@@ -476,5 +476,15 @@ def rollback_docker(container_name):
 
     click.echo('Done')
 
+# verify reboot
+@cli.command('verify_reboot')
+def verify_reboot():
```
jleveque, line 481: Along with Qi's comment above, this command should also be named form clearly, e.g., `verify-next-image`.

qiluo-msft:
>         subprocess.call(['rm','-rf', os.path.join(HOST_PATH, image_dir)])

Seem you function `get_image_path` is useful for many use cases. Could you refactor thoroughly?

---
Refers to: sonic_installer/bootloader/aboot.py:115 in 3a374f3. [](commit_id = 3a374f34b218d8a863e8a0d0a69dc78bcd2e1bd2, deletion_comment = False)

sonic_installer/bootloader/aboot.py
```diff
@@ -129,6 +129,13 @@ def verify_binary_image(self, image_path):
         except subprocess.CalledProcessError:
             return False
 
+    def verify_reboot(self):
+        if not super(AbootBootloader, self).verify_reboot():
+            return False
+        image = self.get_next_image()
+        image_path = self.get_image_path(image) + '/sonic.swi'
```
qiluo-msft, line 136: >sonic.swi [](start = 52, length = 9)

Magic string in aboot.py

scripts/fast-reboot
```diff
@@ -304,10 +305,11 @@ function reboot_pre_check()
         exit ${EXIT_FILE_SYSTEM_FULL}
     fi
 
-    # Make sure that the next image exists
-    if [[ ! -d ${IMAGE_PATH} ]]; then
-        debug "Next image ${NEXT_SONIC_IMAGE} doesn't exist ..."
-        exit ${EXIT_NEXT_IMAGE_NOT_EXISTS}
+    # Verify reboot by sonic_installer
+    local message=$(sonic_installer verify-reboot)
```
qiluo-msft, line 309: >message [](start = 10, length = 7)

Did you only capture stdout but stderr?

sonic_installer/bootloader/aboot.py
Staphylo: and `os.path.join`

sonic_installer/bootloader/bootloader.py
```diff
@@ -43,8 +52,23 @@ def verify_binary_image(self, image_path):
         """verify that the image is supported by the bootloader"""
         raise NotImplementedError
 
+    def verify_reboot(self):
+        """verify the image for reboot"""
+        image = self.get_next_image()
```
Staphylo, line 57: @jleveque pointed out on my review that this class should not implement function bodies.
`OnieInstallerBootloader` might be a better candidate.

qiluo-msft: @jleveque Could you share your thought? It looks good to me because it is general/safe/default implementation for SONiC on any platform, not only ONIE but also Aboot.

It is like C++ virtual function vs pure virtual function

---
In reply to: [452473189](https://github.com/Azure/sonic-utilities/pull/979#discussion_r452473189) [](ancestors = 452473189)

jleveque: The idea was to keep the Bootloader class as an abstract base class; all implementation will be in the subclasses.

xumia: @jleveque Do you think it is ok to implement the checking image path existing in all the subclasses?

jleveque: I think it keeps in line with the current design, and allows different bootloaders to implement different logic.

sonic_installer/bootloader/bootloader.py
```diff
@@ -43,8 +52,23 @@ def verify_binary_image(self, image_path):
...
+        image_path = self.get_image_path(image)
+        if not path.exists(image_path):
+            sys.stderr.write('Next image {0} doesn\'t exist ...\n'.format(image))
```
jleveque, line 60: This function shouldn't write to stderr. I think the calling application should be responsible for outputting to stderr or osyslog based on the return value.

sonic_installer/bootloader/aboot.py
```diff
@@ -129,6 +130,13 @@ def verify_binary_image(self, image_path):
         except subprocess.CalledProcessError:
             return False
 
+    def verify-next-image(self):
```
qiluo-msft, line 133: >verify-next-image [](start = 8, length = 17)

Make sure you lint code before commit.

sonic_installer/bootloader/bootloader.py
qiluo-msft: Thanks @jleveque sharing the though flow!
I vote for leave it here for reasons:
1. all bootloaders will share this code
2. Aboot or any other subclass may override it, but still call this base class implementation first, then implement more verification logic.

---
In reply to: [452504725](https://github.com/Azure/sonic-utilities/pull/979#discussion_r452504725) [](ancestors = 452504725,452473189)

sonic_installer/bootloader/bootloader.py
xumia: Fixed.

sonic_installer/bootloader/bootloader.py
xumia: Fixed.

sonic_installer/main.py
xumia: Fixed.

sonic_installer/bootloader/aboot.py
xumia: Fixed.

scripts/fast-reboot
xumia: Added the stderr message.

sonic_installer/bootloader/aboot.py
xumia: Fixed

scripts/fast-reboot
```diff
@@ -305,10 +306,11 @@ function reboot_pre_check()
...
     fi
 
-    # Make sure that the next image exists
-    if [[ ! -d ${IMAGE_PATH} ]]; then
-        debug "Next image ${NEXT_SONIC_IMAGE} doesn't exist ..."
-        exit ${EXIT_NEXT_IMAGE_NOT_EXISTS}
+    # Verify the next image by sonic_installer
+    local message=$(sonic_installer verify_next_image 2>&1)
+    if [ $? -ne 0 ]; then
+        debug "Failed to verify next image by running command: sonic_installer verify_next_image, result message: ${message}"
```
jleveque, line 312: No need to output the full command in the message. Suggest shortening message to:

```
"Failed to verify next image: ${message}"
```

scripts/reboot
```diff
@@ -79,10 +80,11 @@ function reboot_pre_check()
...
     rm ${filename}
 
-    # Make sure that the next image exists
-    if [[ ! -d ${IMAGE_PATH} ]]; then
-        VERBOSE=yes debug "Next image ${NEXT_SONIC_IMAGE} doesn't exist ..."
-        exit ${EXIT_NEXT_IMAGE_NOT_EXISTS}
+    # Verify the next image by sonic_installer
+    local message=$(sonic_installer verify_next_image 2>&1)
+    if [ $? -ne 0 ]; then
+        VERBOSE=yes debug "Failed to verify the next image by running command: sonic_installer verify_next_image, result message: ${message}"
```
jleveque, line 86: No need to output the full command in the message. Suggest shortening message to:

```
"Failed to verify next image: ${message}"
```

sonic_installer/main.py
```diff
@@ -476,5 +476,15 @@ def rollback_docker(container_name):
 
     click.echo('Done')
 
+# verify the next image
+@cli.command('verify_next_image')
+def verify_reboot():
```
jleveque, line 481: Please change to:

```suggestion
@cli.command('verify-next-image')
def verify_next_image():
```

xumia: @jleveque , do you mean to change 'verify_next_image' to 'verify-next-image'? I notice that the sub-commands are using "_", for example "binary_version".

sonic_installer/bootloader/bootloader.py
xumia: Fixed.

scripts/fast-reboot
xumia: Fixed.

sonic_installer/main.py
xumia: @jleveque , do you mean to change 'verify_next_image' to 'verify-next-image'? I notice that the sub-commands are using "_", for example "binary_version".

scripts/reboot
xumia: Fixed.

sonic_installer/bootloader/bootloader.py
qiluo-msft: I agree. Should not print to either stderr or stdout.
This function is responsible to verify and return Boolean.
The caller is responsible to interpret the Boolean return value.

---
In reply to: [452556977](https://github.com/Azure/sonic-utilities/pull/979#discussion_r452556977) [](ancestors = 452556977)

xumia: The message removed.

xumia:
retest default please

qiluo-msft:
Retest this please

scripts/fast-reboot
```diff
@@ -305,10 +306,11 @@ function reboot_pre_check()
         exit ${EXIT_FILE_SYSTEM_FULL}
     fi
 
-    # Make sure that the next image exists
-    if [[ ! -d ${IMAGE_PATH} ]]; then
-        debug "Next image ${NEXT_SONIC_IMAGE} doesn't exist ..."
-        exit ${EXIT_NEXT_IMAGE_NOT_EXISTS}
+    # Verify the next image by sonic_installer
+    local message=$(sonic_installer verify_next_image 2>&1)
+    if [ $? -ne 0 ]; then
```
yxieca, line 311: This part of the reboot script is executed under -e option, meaning that the failure will trigger script to exit. Please take examples in the script to catch error code.

sonic_installer/main.py
jleveque: Yes, currently all subcommands use underscores, but we are transitioning to using hyphens. I suggest that all new subcommands in all CLI utilities use hyphens, and we will slowly migrate all existing commands, most likely using aliases at first.

xumia: Got it, thanks for the info, Fixed.

scripts/fast-reboot
xumia: Thanks, the command can be simplified, fixed.

qiluo-msft: If you simplified this way, the `sonic_installer verify-next-image` will only output `Failed` or `Done`, which has no context in `fast-reboot` output.

As Ying suggested: Please take examples in the script to catch error code.

Your original `debug` command is still useful.

---
In reply to: [453135626](https://github.com/Azure/sonic-utilities/pull/979#discussion_r453135626) [](ancestors = 453135626)

xumia: Thanks Qi, fixed.

qiluo-msft: Thanks @yxieca actually

sonic_installer/main.py
```diff
@@ -476,5 +476,15 @@ def rollback_docker(container_name):
 
     click.echo('Done')
 
+# verify the next image
+@cli.command('verify-next-image')
+def verify_next_image():
+    """ Verify the next image for reboot"""
+    bootloader = get_bootloader()
+    if not bootloader.verify_next_image():
+        click.echo('Failed')
```
jleveque, line 485: Suggest a more descriptive message like, `Image verification failed`

sonic_installer/main.py
```diff
@@ -476,5 +476,15 @@ def rollback_docker(container_name):
...
+        sys.exit(1)
+    click.echo('Done')
```
jleveque, line 487: Suggest a more descriptive message like, `Image successfully verified`

sonic_installer/main.py
xumia: Fixed.

sonic_installer/main.py
xumia: Thanks, fixed.

PR #981, opened 2020-07-10
[config] Add systemctl stop service check function

kuanyu99:
**- What I did**

Add a check function for "systemctl stop $SERVICE" if it returns with error

**- How I did it**

Use "systemctl is-active $SERVICE" to check if the service is dead

**- How to verify it**

Execute "reload config" inside the sonic console

### DIFF
--- a/show/main.py
+++ b/show/main.py
@@ -19,6 +19,7 @@ from utilities_common.db import Db
 from datetime import datetime
 import utilities_common.constants as constants
 from utilities_common.general import load_db_config
+from json.decoder import JSONDecodeError
 
 # mock the redis for unit test purposes #
 try:
@@ -128,6 +129,10 @@ def run_command(command, display_cmd=False, return_cmd=False):
     rc = proc.poll()
     if rc != 0:
         sys.exit(rc)
+
+def get_cmd_output(cmd):
+    proc = subprocess.Popen(cmd, text=True, stdout=subprocess.PIPE)
+    return proc.communicate()[0], proc.returncode
 
 # Lazy global class instance for SONiC interface name to alias conversion
 iface_alias_converter = lazy_object_proxy.Proxy(lambda: clicommon.InterfaceAliasConverter())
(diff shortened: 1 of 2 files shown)


config/main.py
```diff
@@ -516,6 +526,19 @@ def _get_sonic_generated_services(num_asic):
                 generated_services_list.append(line.rstrip('\n'))
     return generated_services_list, generated_multi_instance_services
 
+def _chk_service_stopped(service_name):
```
jleveque, line 529: Suggest renaming to `_wait_for_service_stop`

config/main.py
```diff
@@ -516,6 +526,19 @@ def _get_sonic_generated_services(num_asic):
...
+    # Check the service for 10 times with 5s interval
+    for i in range(10):
+        (srv_active, err) = run_command("systemctl is-active {}".format(service_name))
+        if "inactive" in srv_active:
+            return True
+        time.sleep(5)
```
jleveque, line 535: This will wait up to 50 seconds. Can you please explain the choice of 50 seconds?

config/main.py
```diff
@@ -147,8 +147,16 @@ def execute_systemctl(list_of_services, action):
                 click.echo("Executing {} of service {}...".format(action, service))
                 run_command("systemctl {} {}".format(action, service))
             except SystemExit as e:
-                log_error("Failed to execute {} of service {} with error {}".format(action, service, e))
-                raise
+                if action == SYSTEMCTL_ACTION_STOP:
+                    log_warning("Get error {} when stopping {}. Try to wait it.".format(e, service))
```
jleveque, line 151: Suggest rewording message as follows:

```
"Received error {} when stopping service '{}'. Waiting to see if it stops ...".format(e, service)
```

config/main.py
```diff
@@ -185,6 +193,8 @@ def run_command(command, display_cmd=False, ignore_error=False):
 
     if proc.returncode != 0 and not ignore_error:
         sys.exit(proc.returncode)
+    else:
+        return (out, err)
```
jleveque, line 197: Please update the docstring to note that upon success the function returns a tuple containing the contents of stdout and stderr.

config/main.py
kuanyu99: It takes at most 50 seconds when I try to stop each of the services

kuanyu99:
Retest this please

config/main.py
jleveque: Maybe we can increase it to an even 60 seconds to add a buffer, and decrease the sleep to 1 second?

kuanyu99: Sure. Good point.

jleveque:
Retest this please

kuanyu99:
Retest this please

Issue #982, opened 2020-07-10
Replace underscores with hyphens in all CLI subcommands/options


jleveque:
The idea is to unify all SONiC CLI subcommands to use hyphens as separators instead of underscores to align more with the Linux standard. This will require a period of transition where we support both formats to allow users to migrate to using the hypenated versions, so we will need to temporarily create aliases for all affected commands so that both the hyphenated and underscore version work. Then we can set a date to deprecate the versions with underscores.

jleveque:
PR to replace underscores with hyphens in sonic-installer: https://github.com/Azure/sonic-utilities/pull/983

PR #983, opened 2020-07-11
[sonic-installer] Hyphens instead of underscores in command and subcommands

jleveque:
Replace underscores with hyphens in all sonic-installer subcommands as well as the root command itself.

- Install two entry points for sonic-installer: `sonic-installer` and `sonic_installer`. This allows for backward compatibility and time for users to transition from using the underscore versions to the hyphenated versions.

- Replace underscores with hyphens in the following subcommands:
    ```
    set-default
    set-next-boot
    binary-version
    upgrade-docker
    rollback-docker
    ```

    - Also add aliases from the underscore versions of these subcommnds to the new hyphenated subcommands to allow for backward compatibility and time for users to transition from using the underscore versions to the hyphenated versions.

- Print deprecation warnings to stderr if issued command line contains any deprecated versions with underscores

- Bonus: Addition of the AliasedGroup base class also adds support for abbreviated subcommands

- Fixed some style issues which did not align with PEP8 standards

- Eliminate duplicate code by adding DOCKER_CONTAINER_LIST; Remove unknown 'amon' container from the list

This is work towards resolving issue https://github.com/Azure/sonic-utilities/issues/982. We can set a deprecation date for the underscore versions of the commands, and when that date arrives, we can simply delete all aliases and deprecation warnings.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -135,7 +135,8 @@ setup(
             'pddf_ledutil = pddf_ledutil.main:cli',
             'show = show.main:cli',
             'sonic-clear = clear.main:cli',
-            'sonic_installer = sonic_installer.main:cli',
+            'sonic-installer = sonic_installer.main:sonic_installer',
+            'sonic_installer = sonic_installer.main:sonic_installer',
             'undebug = undebug.main:cli',
             'watchdogutil = watchdogutil.main:watchdogutil',
         ]
--- a/setup.py
+++ b/setup.py
@@ -136,7 +136,7 @@ setup(
             'show = show.main:cli',
             'sonic-clear = clear.main:cli',
             'sonic-installer = sonic_installer.main:sonic_installer',
-            'sonic_installer = sonic_installer.main:sonic_installer',
+            'sonic_installer = sonic_installer.main:sonic_installer',  # Deprecated
             'undebug = undebug.main:cli',
             'watchdogutil = watchdogutil.main:watchdogutil',
         ]
--- a/sonic_installer/main.py
+++ b/sonic_installer/main.py
@@ -12,6 +12,50 @@ from swsssdk import SonicV2Connector
 from .bootloader import get_bootloader
 from .common import run_command
 
+
+# Global Config object
+_config = None
+
+
+class AliasedGroup(click.Group):
+    """This subclass of click.Group supports abbreviations and
+       looking up aliases in a config file with a bit of magic.
+    """
+
+    def get_command(self, ctx, cmd_name):
+        global _config
+
+        # If we haven't instantiated our global config, do it now and load current config
+        if _config is None:
+            _config = Config()
+
+            # Load our config file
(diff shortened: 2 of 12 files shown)


setup.py
```diff
@@ -135,7 +135,8 @@
             'pddf_ledutil = pddf_ledutil.main:cli',
             'show = show.main:cli',
             'sonic-clear = clear.main:cli',
-            'sonic_installer = sonic_installer.main:cli',
+            'sonic-installer = sonic_installer.main:sonic_installer',
+            'sonic_installer = sonic_installer.main:sonic_installer',
```
qiluo-msft, line 139: How about adding a depreciation warning error message to stderr if the underscore binary is called?

jleveque: Great idea! I added a deprecation warning in the latest commit for the root command, but I think it would also be good to do it for the subcommands, as well.

jleveque:
Retest this please

jleveque:
Retest this please

jleveque:
Retest this please

jleveque:
Retest this please

sonic_installer/main.py
```diff
@@ -111,36 +185,44 @@ def get_container_image_id_all(image_name):
...
+@click.group(cls=AliasedGroup)
+def sonic_installer():
     """ SONiC image installation manager """
     if os.geteuid() != 0:
         exit("Root privileges required for this operation")
 
+    # Warn the user if they are calling the deprecated version of the command (with an underscore instead of a hyphen)
+    if os.path.basename(sys.argv[0]) == "sonic_installer":
+        click.secho("Warning: 'sonic_installer' command is deprecated and will be removed in the future", fg="red", err=True)
+        click.secho("Please use 'sonic-installer' instead", fg="red", err=True)
```
qiluo-msft, line 215: Extracted as a function and reuse.

jleveque: Done.

sonic_installer/main.py
```diff
@@ -111,36 +185,50 @@ def get_container_image_id_all(image_name):
...
     db = SonicV2Connector()
     db.connect(db_name, False)
     _hash = table_name + db.get_db_separator(db_name) + warm_app_name
     client = db.get_redis_client(db_name)
-    return  client.hdel(_hash, key)
+    return client.hdel(_hash, key)
+
+
+def print_deprecation_warning(deprecated_cmd_or_subcmd, new_cmd_or_subcmd):
+    click.secho("Warning: '{}' {}command is deprecated and will be removed in the future"
```
qiluo-msft, line 206: >{} [](start = 31, length = 2)

extra pair?

qiluo-msft: Oh, I see it.

---
In reply to: [453998034](https://github.com/Azure/sonic-utilities/pull/983#discussion_r453998034) [](ancestors = 453998034)

jleveque: No. It's intentional.

jleveque:
Retest this please

PR #985, opened 2020-07-12
Use /var/tmp instead of /tmp for sonic_installer and generate_dump

olivier-singla:
sonic_installer and generate_dump utilities use currently /tmp directory
to store temporary files. Since the amount of data stored can be rather
large (especially in the case of the sonic_installer), we are switching to
use /var/tmp  instead.

We are doing these changes in anticipation that at some point /tmp will being
mounted over TMPFS

/tmp is typically mounted on TMPFS, which means it's fast and does not produce
wear on SSD or Hard-Drive. /var/tmp is mounted on the rootfs partition,
therefore not in the TMPFS partition.

We also noticed that both sonic_installer and generate_dump produce temporary
files which are not deleted once thse programes are done. We are then making
changes to peform a clean-up and remove the temporary files not needed anymore.

# Please enter the commit message for your changes. Lines starting
# with '#' will be ignored, and an empty message aborts the commit.
#
# Date:      Sat Jul 11 19:58:59 2020 -0400
#
# On branch var_tmp
# Changes to be committed:
#	modified:   scripts/generate_dump
#	modified:   sonic_installer/bootloader/aboot.py
#	modified:   sonic_installer/bootloader/grub.py
#	modified:   sonic_installer/bootloader/onie.py
#	modified:   sonic_installer/bootloader/uboot.py
#	modified:   sonic_installer/common.py
#	modified:   sonic_installer/main.py
#

**- What I did**

**- How I did it**

**- How to verify it**


jleveque:
>We are doing these changes in anticipation that at some point /tmp will being
mounted over TMPFS

>/tmp is typically mounted on TMPFS, which means it's fast and does not produce
wear on SSD or Hard-Drive. /var/tmp is mounted on the rootfs partition,
therefore not in the TMPFS partition.

I'm not clear on what the benefit of this PR is. I can see benefit in cleaning up the temporary files once the applications are done with them, but why change the location to /var/tmp? As you mentioned if /tmp is eventually mounted in tmpfs, there will be less wear and tear on the hard disk, so I believe tmpfs is actually the better solution here. Plus, on devices with small hard disks, it will be beneficial, if not critical, to keep temporary files off the hard disk.

olivier-singla:
> > We are doing these changes in anticipation that at some point /tmp will being
> > mounted over TMPFS
> 
> > /tmp is typically mounted on TMPFS, which means it's fast and does not produce
> > wear on SSD or Hard-Drive. /var/tmp is mounted on the rootfs partition,
> > therefore not in the TMPFS partition.
> 
> I'm not clear on what the benefit of this PR is. I can see benefit in cleaning up the temporary files once the applications are done with them, but why change the location to /var/tmp? As you mentioned if /tmp is eventually mounted in tmpfs, there will be less wear and tear on the hard disk, so I believe tmpfs is actually the better solution here. Plus, on devices with small hard disks, it will be beneficial, if not critical, to keep temporary files off the hard disk, thus keeping the location as /tmp and mounting /tmp as tmpfs should be the ultimate goal.

The idea to use /var/tmp was to minimize the impact of moving /tmp to tmpfs, and have the two utilities, sonic_installer and generate_dump, to perform as currently (/tmp over SSD or HD). This said, I can revert to use /tmp instead of using /var/tmp

rajendra-dendukuri:
> > We are doing these changes in anticipation that at some point /tmp will being
> > mounted over TMPFS
> 
> > /tmp is typically mounted on TMPFS, which means it's fast and does not produce
> > wear on SSD or Hard-Drive. /var/tmp is mounted on the rootfs partition,
> > therefore not in the TMPFS partition.
> 
> I'm not clear on what the benefit of this PR is. I can see benefit in cleaning up the temporary files once the applications are done with them, but why change the location to /var/tmp? As you mentioned if /tmp is eventually mounted in tmpfs, there will be less wear and tear on the hard disk, so I believe tmpfs is actually the better solution here. Plus, on devices with small hard disks, it will be beneficial, if not critical, to keep temporary files off the hard disk, thus keeping the location as /tmp and mounting /tmp as tmpfs should be the ultimate goal.

It is typical to have systems with larger hard disks relative to the RAM they have.  For certain applications like sonic_installer, we need to store the temporary file in persistent storage while we extract it. If we keep the temporary file in RAM and then extract it, we will need more free space in RAM. So it is ideal that large temporary files be stored on disk and manageable sized files can go in /tmp which is proposed to be changed to a tmpfs. That is the reason for changing sonic_installer to /var/tmp.

jleveque:
> It is typical to have systems with larger hard disks relative to the RAM they have.

Typically, yes. However, there are a handful of SONiC-supported platforms where the RAM is 4 GB and the SDD is also 4 GB or even smaller (e.g., 2 GB). These platforms would actually be harmed by this change rather than helped.

I'd like @lguohan's to provide his opinion here, as well.

rajendra-dendukuri:
> > It is typical to have systems with larger hard disks relative to the RAM they have.
> 
> Typically, yes. However, there are a handful of SONiC-supported platforms where the RAM is 4 GB and the SDD is also 4 GB or even smaller (e.g., 2 GB). These platforms would actually be harmed by this change rather than helped.

No *new* harm will be done due this PR. The switches were already using /tmp which is currently mounted on disk. We are just moving it to /var/tmp which is also on disk. Essentially it is a no-op.

olivier-singla:
> I'd like @lguohan's to provide his opinion here, as well.

I was wondering, did you had a chance to review this change?

rajendra-dendukuri:
@lguohan can you please review this change.

PR #986, opened 2020-07-13
 Fix the None Type Exception when Interface Table does not exist (cold boot) as part of db migration

abdosi:
Fix the None Type Exception when Interface Table does not exist (cold boot) as part of db migration
Because of this exception database service does not come up.

After fix all docker and services cam up fine.


prsunny:
retest this please

PR #968, opened 2020-06-28
Display large number of FDB entries fast from state_db and appl_db

JiangboHe:
**- What I did**
I change the mothod by showing mac table from state FDB_TABLE to display large number of fdb tables faster.
The time cost to display change from 50 seconds to 2 seconds for 8000 fdb tables or 20 seconds to 1 seconds for 5000 fdb tables. 

**- How I did it**
The key search and judge of fdb tables in asic table is too complex. It will cost 20 seconds for 5000 fdb tables or 50 seconds for 8000 fdb tables. It is not good for users or script running.

**- How to verify it**
Learn thounds of fdb tables and show them with fdbshow

**- Previous command output (if the output of a command-line utility has changed)**

**- New command output (if the output of a command-line utility has changed)**

PR #988, opened 2020-07-13
[show] Fix for 'show mac' command during DB flush

shlomibitton:
Signed-off-by: Shlomi Bitton <shlomibi@mellanox.com>

**- What I did**
When running 'show mac' command while there are many MAC entries in the DB, an exception can occur if during the execution of the command the DB is being flushed.

**- How I did it**

There are two checks in the fix, the first is to check if the entry still valid and present in the DB, if not continue, if yes try to fetch it. the 'try except' wrap is to make sure that right after the first check the entry not flushed from the DB to make sure this issue won't happen again.
the first check is necessary because a needed blocking flag is set to 'True' on the function call to redis, if there are many DB fetch misses the execution of the command can take a lot of time which is not reasonable. 

**- How to verify it**

1) Inject ~10000 MAC entried into the FDB.
2) Wait until aging time of the FDB.
3) Issue 'show mac' command.

Key 'ASIC_STATE:SAI_OBJECT_TYPE_FDB_ENTRY:{"bvid":"oid:0x26000000000acb","mac":"00:11:00:00:00:01","switch_id":"oid:0x21000000000000"}' unavailable in database '1'

Normal 'show mac' command output.

### DIFF
--- a/scripts/fdbshow
+++ b/scripts/fdbshow
@@ -32,14 +32,18 @@ import sys
 from natsort import natsorted
 from swsssdk import SonicV2Connector, port_util
 from tabulate import tabulate
+from sonic_daemon_base.daemon_base import Logger
 
 class FdbShow(object):
 
     HEADER = ['No.', 'Vlan', 'MacAddress', 'Port', 'Type']
     FDB_COUNT = 0
+    DB_MISS_THRESHOLD = 10
 
     def __init__(self):
         super(FdbShow,self).__init__()
+        self.logger = Logger("FDB")
+        self.db_miss_counter = 0
         self.db = SonicV2Connector(host="127.0.0.1")
         self.if_name_map, \
         self.if_oid_map = port_util.get_interface_oid_map(self.db)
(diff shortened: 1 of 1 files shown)


scripts/fdbshow
```diff
@@ -68,8 +72,14 @@ class FdbShow(object):
             fdb = json.loads(fdb_entry .split(":", 2)[-1])
             if not fdb:
                 continue
-
-            ent = self.db.get_all('ASIC_DB', s, blocking=True)
+            if self.db.exists('ASIC_DB',s):
+                try:
+                    ent = self.db.get_all('ASIC_DB', s, blocking=True)
+                except:
```
prsunny, line 78: Can we just look for an exception with a log saying some entries are aged-out?. Apparently, this is not a frequent use-case and the user can re-issue the command? Also there is an improvement on 'show mac' to be fetched from STATE_DB

shlomibitton: @prsunny yes it is not a frequent use-case.
you are suggesting to print as an output when this case occurs that some MAC aged out?
re-issue the command will probably work, but this is expected behaviour?
can you please point to the improvement you are referring to?

prsunny: yes, i was referring to this PR - https://github.com/Azure/sonic-utilities/pull/968

shlomibitton: @prsunny looks like the PR you mentioned will improve and fix the issue.
Is it going to be merged soon? If it will take a long time to review and merge all PR's needed for this improvement maybe we should merge this fix for now, what do you think?

prsunny: I believe, the other one would be updated and merged soon.

shlomibitton: OK, thanks

PR #991, opened 2020-07-16
[acl-loader] Fix bugs in acl-loader

daall:
- Fix issue where protocol == 0 would be ignored
- Emit warning if --table_name not found in Config DB

**- What I did/How I did it**
I fixed a bug in acl-loader where rules with protocol=0 would have the protocol field ignored due to python's implicit boolean conversion. I fixed this by explicitly adding the `== 0` check to the ip protocol field.

I also fixed/added a warning when the provided `--table_name` can't be found in Config DB. Previously this case would fail silently, now the user will receive a warning to point them in the right direction.

**- How to verify it**
Run `acl-loader` with a non-existent `--table_name`:
```
admin@sonic:~$ sudo acl-loader update full test.json --table_name yeehaw
Warning: Table "yeehaw" not found
```

Run `acl-loader` with a rule that includes `"protocol": 0`:
```
admin@sonic:~$ sudo acl-loader update full test.json --session_name test
admin@sonic:~$ show acl rule
Table       Rule      Priority  Action                Match
----------  ------  ----------  --------------------  --------------
EVERFLOW    RULE_1        9999  MIRROR INGRESS: test  IP_PROTOCOL: 0
```

### DIFF
--- a/acl_loader/main.py
+++ b/acl_loader/main.py
@@ -244,6 +244,9 @@ class AclLoader(object):
         :param table_name: Table name
         :return:
         """
+        if not self.is_table_valid(table_name):
+            warning("Table \"%s\" not found" % table_name)
+
         self.current_table = table_name
 
     def set_session_name(self, session_name):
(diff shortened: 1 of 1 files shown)


daall:
retest this please

daall:
retest this please

PR #994, opened 2020-07-17
Add secure fast/warm-reboot support for Aboot

Staphylo:
Instead of having multiple implementation of preparing a SWI image for
secureboot, fast-reboot now reuses boot0.
SWI images booting in regular mode will keep using the old behavior.

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -271,14 +271,23 @@ function teardown_control_plane_assistant()
     fi
 }
 
+function is_secureboot() {
+    grep -Eq 'secure_boot_enable=[1y]' /proc/cmdline
+}
+
 function setup_reboot_variables()
 {
     # Kernel and initrd image
     NEXT_SONIC_IMAGE=$(sonic-installer list | grep "Next: " | cut -d ' ' -f 2)
     IMAGE_PATH="/host/image-${NEXT_SONIC_IMAGE#SONiC-OS-}"
     if grep -q aboot_platform= /host/machine.conf; then
-        KERNEL_IMAGE="$(ls $IMAGE_PATH/boot/vmlinuz-*)"
-        BOOT_OPTIONS="$(cat "$IMAGE_PATH/kernel-cmdline" | tr '\n' ' ') SONIC_BOOT_TYPE=${BOOT_TYPE_ARG}"
+        if is_secureboot; then
+            KERNEL_IMAGE=""
+            BOOT_OPTIONS="SONIC_BOOT_TYPE=${BOOT_TYPE_ARG} secure_boot_enable=1"
+        else
+            KERNEL_IMAGE="$(ls $IMAGE_PATH/boot/vmlinuz-*)"
+            BOOT_OPTIONS="$(cat "$IMAGE_PATH/kernel-cmdline" | tr '\n' ' ') SONIC_BOOT_TYPE=${BOOT_TYPE_ARG}"
+        fi
     elif grep -q onie_platform= /host/machine.conf; then
         KERNEL_OPTIONS=$(cat /host/grub/grub.cfg | sed "/$NEXT_SONIC_IMAGE'/,/}/"'!'"g" | grep linux)
         KERNEL_IMAGE="/host$(echo $KERNEL_OPTIONS | cut -d ' ' -f 2)"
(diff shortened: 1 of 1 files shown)


qiluo-msft:
Could you change the title to "Add secure fast-reboot and warm-reboot support for Aboot" if this is true?

scripts/fast-reboot
```diff
@@ -271,14 +271,23 @@ function teardown_control_plane_assistant()
...
 {
     # Kernel and initrd image
     NEXT_SONIC_IMAGE=$(sonic-installer list | grep "Next: " | cut -d ' ' -f 2)
     IMAGE_PATH="/host/image-${NEXT_SONIC_IMAGE#SONiC-OS-}"
     if grep -q aboot_platform= /host/machine.conf; then
-        KERNEL_IMAGE="$(ls $IMAGE_PATH/boot/vmlinuz-*)"
-        BOOT_OPTIONS="$(cat "$IMAGE_PATH/kernel-cmdline" | tr '\n' ' ') SONIC_BOOT_TYPE=${BOOT_TYPE_ARG}"
+        if is_secureboot; then
+            KERNEL_IMAGE=""
+            BOOT_OPTIONS="SONIC_BOOT_TYPE=${BOOT_TYPE_ARG}"
```
qiluo-msft, line 286: >" [](start = 58, length = 1)

Will you add `secure_boot_enable=1` into `BOOT_OPTIONS`?

This is also applied to non-aboot secure boot flow.

scripts/fast-reboot
```diff
@@ -332,6 +341,21 @@ function reboot_pre_check()
     fi
 }
 
+function load_aboot_secureboot_kernel() {
+    local next_image="$IMAGE_PATH/sonic.swi"
+    echo "Loading next image from $next_image"
+    (
+        cd /tmp
```
qiluo-msft, line 348: >cd /tmp [](start = 8, length = 7)

Use `mktemp -d` to prevent file confliction?

scripts/fast-reboot
```diff
@@ -332,6 +341,21 @@ function reboot_pre_check()
...
+        unzip -qp "$next_image" boot0 | \
+            swipath=$next_image kexec=true testonly=true ENV_EXTRA_CMDLINE="$BOOT_OPTIONS" bash -
```
qiluo-msft, line 350: >testonly [](start = 43, length = 8)

why `testonly`?
The name seems weird.

renukamanavalan:
retest this please

scripts/fast-reboot
Staphylo: Because the loading of the kernel and initramfs is done through boot0 the `secure_boot_enable=1` option is added automatically. Adding it here would be redundant but harmless.
Let me know if you still want me to add it here.

scripts/fast-reboot
Staphylo: Actually I could remove the block that goes into /tmp since boot0 does not dirty the current directory.
I just did this as a safety mechanism but `mktemp -d` is not useful here.

scripts/fast-reboot
Staphylo: This is a setting that existed for a long time in boot0 and that basically stops after `kexec -l` and right before `kexec -e`
I agree that the setting name is confusing, I will add another one called `loadonly=true` that will have the exact same behavior.
I originally just reused the old setting name as we can't easily remove it for legacy reasons.
I will make the change in https://github.com/Azure/sonic-buildimage/pull/4994

Staphylo:
retest this please

scripts/fast-reboot
```diff
@@ -271,14 +271,23 @@ function teardown_control_plane_assistant()
     fi
 }
 
+function is_secureboot() {
+    grep -Eq 'secure_boot_enable=[1y]' /proc/cmdline
```
yxieca, line 275: This implies the current image is boot up with secure boot.

What if a device is booting from none-secure-boot image into secure boot image? Does fast-reboot work there?

Staphylo: You would need to go in Aboot and reconfigure the secureboot setting via the console.
It is therefore not possible to fast-reboot from non-secure to secure mode. Which is why it hasn't been handled here.

jleveque:
Retest this please

jleveque:
Retest this please

qiluo-msft:
Retest this please

scripts/fast-reboot
qiluo-msft: Please help added it. This code will be shared to non-aboot secure boot flow which using `kexec`.

I guess the aboot has the intelligence to keep single `secure_boot_enable=1` if there is already one. There is no need to duplicate it on cmdline.

---
In reply to: [456649028](https://github.com/Azure/sonic-utilities/pull/994#discussion_r456649028) [](ancestors = 456649028)

qiluo-msft:
Does this PR works for warm-reboot, in theory?

---
In reply to: [660299771](https://github.com/Azure/sonic-utilities/pull/994#issuecomment-660299771) [](ancestors = 660299771)

Staphylo:
Yes this PR works for both fast-reboot and warm-reboot.

qiluo-msft:
Retest this please

qiluo-msft:
Retest this please

PR #1014, opened 2020-07-31
[fanshow] Display other fan status, such as Updating

Junchao-Mellanox:
**- What I did**

The original fan status can be one of "OK", "Not OK", "N/A". This PR allow a new fan status "Updating" 

**- How I did it**

If fan status is not "true" or "false", display the status field value in CLI output.

**- How to verify it**

Manually test

Not changed.

Not changed.


lguohan:
need unit test

PR #1021, opened 2020-07-31
[tests] Add unit tests for 'show platform ...' commands

jleveque:
Add unit tests for 'show platform ...' commands


tests/show_platform_test.py
```diff
@@ -0,0 +1,60 @@
+import os
+import sys
+import textwrap
+
+import mock
+from click.testing import CliRunner
+from unittest import TestCase
```
lguohan, line 7: probably should no longer to unittest framework, should move to pytest

PR #1023, opened 2020-08-03
[tests]: fix drops_group_test failure on second run

lguohan:

**- What I did**
a few tests in drops_group_test fails on second run.
The reason is that /tmp/dropstat is not cleaned, the first
run leave some state in the folder which cause the subsequent
run to fail.

The fix is the always clean up the folder.

**- How I did it**
The fix is the always clean up the folder.

**- How to verify it**
run python setup.py test multiple times.


lguohan:
test failure is not related this pr.

PR #1026, opened 2020-08-04
Fix show interface status display of Interface to Port-channel mapping.

abdosi:
**- What I did**
Fixed show interface status to get Interface to
Port-Channel mapping  information from APP DB instead of Config DB.
Reason was when we add/remove memeber from Port-Channel using command
teamdctl (eg: teamdctl PortChannel0002 port remove Ethernet0) show
interface status does not get updated and provide wrong information.

Some of ansible/pytest (eg: ansible/roles/test/tasks/arpall.yml ) 
uses teamdctl command to add/remove interface from port-channel
so without this change debugging becomes difficult as information is not valid.

**- How I did it**
Updated to get information from APP_DB instead of CONFIG_DB

**- How to verify it**
Manual Verification:

a) Initial Display:
admin@str-s6000-on-2:/$ show interface status
      Interface            Lanes    Speed    MTU           Alias             Vlan    Oper    Admin    Type    Asym PFC
---------------  ---------------  -------  -----  --------------  ---------------  ------  -------  ------  ----------
      Ethernet0      29,30,31,32      40G   9100    fortyGigE0/0  PortChannel0002      up       up   QSFP+         off

b) Now Remove Ethernet from Port channel

admin@str-s6000-on-2:/$ teamdctl PortChannel0002 port remove Ethernet0

Before Fix:
========
c) Display again (No change ) Same as Initial Display
admin@str-s6000-on-2:/$ show interface status
      Interface            Lanes    Speed    MTU           Alias             Vlan    Oper    Admin    Type    Asym PFC
---------------  ---------------  -------  -----  --------------  ---------------  ------  -------  ------  ----------
      Ethernet0      29,30,31,32      40G   9100    fortyGigE0/0  PortChannel0002      up       up   QSFP+         off

After Fix:
=======
admin@str-s6000-on-2:/$ show interface status
      Interface            Lanes    Speed    MTU           Alias             Vlan    Oper    Admin    Type    Asym PFC
---------------  ---------------  -------  -----  --------------  ---------------  ------  -------  ------  ----------
      Ethernet0      29,30,31,32      40G   9100    fortyGigE0/0           routed      up       up   QSFP+         off


prsunny:
@abdosi, I think we should not use teamdctl command but "config portchannel member del <portchannel_name> <port_name>". "teamdctl" would have been used as legacy when the config command was not available.

abdosi:
> @abdosi, I think we should not use teamdctl command but "config portchannel member del <portchannel_name> <port_name>". "teamdctl" would have been used as legacy when the config command was not available.

@prsunny this change is compatible with config command also.  I feel app db is better since show interface port-channel also gets information from app db and that way both command show interface status and port-channel will be consistent.

prsunny:
> > @abdosi, I think we should not use teamdctl command but "config portchannel member del <portchannel_name> <port_name>". "teamdctl" would have been used as legacy when the config command was not available.
> 
> @prsunny this change is compatible with config command also. I feel app db is better since show interface port-channel also gets information from app db and that way both command show interface status and port-channel will be consistent.

Not really, "show int port-channel" shows the current state of members from app_db. But the config state (from config_db) is now only in "show int status". Basically that shows the user that, the particular interface is bind to portchannel in configuration.

abdosi:
As mentioned in above comments closing this PR.

PR #1028, opened 2020-08-04
[config load]: do not stop/reset/reload service if it is disabled

lguohan:
add config_test unit test to test this scenario

**- What I did**
do not stop/reset/reload service if it is disabled

**- How I did it**
check feature table, if the service is in disabled state, then remove the service form the list

**- How to verify it**
unit test


tests/config_test.py
```diff
@@ -0,0 +1,177 @@
...
+Executing restart of service bgp...
+Executing restart of service pmon...
+Executing restart of service lldp...
+Executing restart of service hostcfgd...
+Executing restart of service nat...
+Please note setting loaded from minigraph will be lost after system reboot. To preserve setting, run `config save`.
+"""
+
+@pytest.fixture
+def setup_config_db():
```
lguohan, line 135: i will add the conftest.py to put all common fixture there.

config/main.py
```diff
@@ -337,6 +340,20 @@ def run_command(command, display_cmd=False, ignore_error=False):
     if proc.returncode != 0 and not ignore_error:
         sys.exit(proc.returncode)
 
+def _get_device_type():
```
jleveque, line 343: This is a good candidate to move to sonic-py-common

config/main.py
```diff
@@ -328,6 +328,9 @@ def run_command(command, display_cmd=False, ignore_error=False):
     if display_cmd == True:
         click.echo(click.style("Running command: ", fg='cyan') + click.style(command, fg='green'))
 
+    if os.environ["UTILITIES_UNIT_TESTING"] == "1":
+        return
```
jleveque, line 332: Is is best to add this here? Should we instead mock this function in any tests which should override this function?

lguohan: let me try that.

config/main.py
lguohan: will add.

tests/conftest.py
```diff
@@ -0,0 +1,82 @@
...
+    'nat.service',
+    'teamd.service',
+    'syncd.service',
+    'snmp.timer',
+    'telemetry.timer']
+
+
+def run_command(command, display_cmd=False, return_cmd=False):
+    if display_cmd == True:
+        click.echo(click.style("Running command: ", fg='cyan') + click.style(command, fg='green'))
```
jleveque, line 54: Is it possible to create a fixture to do this?
Also, can you use `mock` to fake the execution of the command instead of redefining it?

lguohan: mock is trying to mock the input, I cannot find the mock support mock function.

jleveque: Your new method of reassigning the function is more clear, and I'd be OK with it. But I believe you should be able to use `mock.patch` like this: https://github.com/Azure/sonic-utilities/pull/1021/files#diff-ec4cac544e7af7e612e78e8480cd701dR48

lguohan: the mock is used to get specific return value from the mocked function.

jleveque: It doesn't actually call the function, it just provides the return value as if the function returned it. I think you should just be able to set `return_value=None` and the function will basically become a no-op which returns None.

But I'm OK with your current approach, also.

Also, just a note, the `mock` library is now a part of `unittest` as of Python 3.3. So if we continue using `mock`, we will need to import `unittest` anyway.

lguohan: mock.patch is still used for mock the return value. In this case, we do not need to mock the return value.

jleveque: True. I just suggested making the function a no-op using `mock.patch` also. But your method is clear, so it's fine.

lguohan: no-op won't work for this case, since we still want to the function to output the command so that we can compare with the expected output.

config/main.py
```diff
@@ -715,6 +730,12 @@ def _stop_services():
     if asic_type == 'mellanox' and 'pmon' in services_to_stop:
         services_to_stop.remove('pmon')
 
+    disable_services = _get_disabled_services_list()
```
jleveque, line 733: This was a typo on my part and now you've copied and pasted it. Sorry. would you mind changing `disable_services` to `disabled_services` wherever it's used as part of this PR, please?

tests/conftest.py
jleveque: True. For the commands which set --verbose by default. But that's only a few commands. If only there was an easy way to capture the acutal commands being issued, rather than the stdout...

tests/conftest.py
```diff
@@ -0,0 +1,81 @@
...
+    'nat.service',
+    'teamd.service',
+    'syncd.service',
+    'snmp.timer',
+    'telemetry.timer']
+
+
+def _dummy_run_command(command, display_cmd=False, return_cmd=False):
+    if display_cmd == True:
+        click.echo(click.style("Running command: ", fg='cyan') + click.style(command, fg='green'))
```
jleveque, line 54: Suggest simply having the dummy command always print the command. Then this will be useful to other subcommands which don't set `display_cmd=True`.

```
def _dummy_run_command(command, display_cmd=False, return_cmd=False):
    click.echo(command)
```

lguohan: I am not really to make this change as some output are explicitly hidden in the show command and user won't see all commands. if we show all the commands, then it will not be the same as what user see in the real box.

jleveque: Yes, my point was with this method, we can verify that the all the underlying commands are run correctly, even the ones that are hidden.

lguohan: it is better to change the show command its self to have a verbose mode where we can check all commands. But I think part of the unit test to verify we are seeing exactly the same as user see so that we won't have surprise in terms of what user sees.

lguohan: we can have a verbose option in load_minigraph to show all commands, and we can validate that in the unit test. but I think that is going to be a separate pr.

jleveque: Point taken. All commands *should* already support a `verbose` flag which gets passed to `run_command`. I guess for those commands we can test the function with the flag set to both True and False.

PR #1029, opened 2020-08-05
pass db connector as click context

lguohan:
it is better not let every command to have its
own connector and connect to db.

it is also not good to have a global db variable.

Here, the idea is to have a single db connector
and pass the connector as a click context

**- What I did**
it is better not let every command to have its
own connector and connect to db.

it is also not good to have a global db variable.

Here, the idea is to have a single db connector
and pass the connector as a click context.

This will allow us to easily split the big config file into multiple 
small files without maintaining a global db variable.

**- How I did it**
use make_pass_decorator to pass the db object.

**- How to verify it**
unit test.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -19,6 +19,7 @@ from minigraph import parse_device_desc_xml
 from config_mgmt import ConfigMgmtDPB
 from utilities_common.intf_filter import parse_interface_in_filter
 from utilities_common.util_base import UtilHelper
+from utilities_common.db import Db
 from portconfig import get_child_ports, get_port_config_file_name
 
 import aaa
@@ -50,7 +51,6 @@ CFG_LOOPBACK_ID_MAX_VAL = 999
 CFG_LOOPBACK_NO="<0-999>"
 
 asic_type = None
-config_db = None
 
 # ========================== Syslog wrappers ==========================
 
@@ -691,7 +691,7 @@ def _abort_if_false(ctx, param, value):
         ctx.abort()
 
 
-def _get_disabled_services_list():
+def _get_disabled_services_list(config_db):
     disabled_services_list = []
 
     feature_table = config_db.get_table('FEATURE')
@@ -713,7 +713,7 @@ def _get_disabled_services_list():
 
     return disabled_services_list
 
-def _stop_services():
+def _stop_services(config_db):
     # This list is order-dependent. Please add services in the order they should be stopped
     # on Mellanox platform pmon is stopped by syncd
     services_to_stop = [
@@ -730,7 +730,7 @@ def _stop_services():
     if asic_type == 'mellanox' and 'pmon' in services_to_stop:
         services_to_stop.remove('pmon')
 
-    disabled_services = _get_disabled_services_list()
+    disabled_services = _get_disabled_services_list(config_db)
 
     for service in disabled_services:
         if service in services_to_stop:
@@ -739,7 +739,7 @@ def _stop_services():
     execute_systemctl(services_to_stop, SYSTEMCTL_ACTION_STOP)
 
 
-def _reset_failed_services():
+def _reset_failed_services(config_db):
     # This list is order-independent. Please keep list in alphabetical order
     services_to_reset = [
         'bgp',
@@ -762,7 +762,7 @@ def _reset_failed_services():
         'telemetry'
     ]
 
-    disabled_services = _get_disabled_services_list()
+    disabled_services = _get_disabled_services_list(config_db)
 
     for service in disabled_services:
         if service in services_to_reset:
@@ -771,7 +771,7 @@ def _reset_failed_services():
     execute_systemctl(services_to_reset, SYSTEMCTL_ACTION_RESET_FAILED)
 
 
-def _restart_services():
+def _restart_services(config_db):
     # This list is order-dependent. Please add services in the order they should be started
     # on Mellanox platform pmon is started by syncd
     services_to_restart = [
@@ -790,7 +790,7 @@ def _restart_services():
         'telemetry'
     ]
 
-    disabled_services = _get_disabled_services_list()
+    disabled_services = _get_disabled_services_list(config_db)
 
     for service in disabled_services:
         if service in services_to_restart:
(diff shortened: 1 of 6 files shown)


tests/config_test.py
```diff
@@ -52,16 +56,20 @@ def test_load_minigraph(self, get_cmd_module, setup_single_broacom_asic):
         result = runner.invoke(config.config.commands["load_minigraph"], ["-y"])
         print result.exit_code
         print result.output
+        traceback.print_tb(result.exc_info[2])
         assert result.exit_code == 0
         assert "\n".join([ l.rstrip() for l in result.output.split('\n')]) == load_minigraph_command_output
 
     def test_load_minigraph_with_disabled_telemetry(self, get_cmd_module, setup_single_broacom_asic):
         (config, show) = get_cmd_module
+        db = Db()
```
lguohan, line 65: here we create single db object and share the object for both config and show command so that the changes in the mockredis can preserve.

if we use multiple db connector, then everytime a new db connection is made, previous changes are not preserved.

lguohan: in case we do not need to share the db, then we do not need to pass the object explicitly. the click framework will automatically create one.

show/main.py
```diff
@@ -557,12 +556,13 @@ def get_bgp_neighbor_ip_to_name(ip, static_neighbors, dynamic_neighbors):
...
+@click.pass_context
+def cli(ctx):
     """SONiC command line - 'show' command"""
-    global config_db
 
-    config_db = ConfigDBConnector()
-    config_db.connect()
+    ctx.obj = Db()
+
+pass_db = click.make_pass_decorator(Db, ensure=True)
```
lguohan, line 565: this is the decorator. https://click.palletsprojects.com/en/7.x/complex/?highlight=make_pass_decorator

config/main.py
```diff
@@ -3810,25 +3812,27 @@ def feature():
 @feature.command('state', short_help="Enable/disable a feature")
 @click.argument('name', metavar='<feature-name>', required=True)
 @click.argument('state', metavar='<state>', required=True, type=click.Choice(["enabled", "disabled"]))
-def feature_state(name, state):
+@pass_db
+def feature_state(db, name, state):
     """Enable/disable a feature"""
-    state_data = config_db.get_entry('FEATURE', name)
+    state_data = db.cdb.get_entry('FEATURE', name)
```
lguohan, line 3818: previously, it is using the global config_db. now we are using the db object from the context. 

I think this change is necessary, as in subsequent pr, i will split this big config file into multiple py files, each contain one click group. Then, we need to pass the db around using this context.

utilities_common/db.py
```diff
@@ -0,0 +1,6 @@
+from swsssdk import ConfigDBConnector
+
+class Db(object):
+    def __init__(self):
+        self.cdb = ConfigDBConnector()
+        self.cdb.connect()
```
lguohan, line 6: later we will need to expand this object to support various DB, state db, and also multi-asic support.

utilities_common/db.py
```diff
@@ -0,0 +1,6 @@
+from swsssdk import ConfigDBConnector
+
+class Db(object):
```
jleveque, line 3: Should we call this `ConfigDb`? Maybe we will add other DB connectors here in the future?

utilities_common/db.py
jleveque: Should we rename `cdb` to be more descriptive, like `config_db` or `configdb`

lguohan: i am mostly follow the convention in swss vs tests. 

https://github.com/Azure/sonic-swss/blob/82d36c4f7ded1eceb832ab737fc13e6ad145090c/tests/test_port_mac_learn.py#L8

i mostly prefer abbrev as the code is shorter.

jleveque: I see. Shorter is convenient, but it can get confusing, especially as more DBs are added . I see there is now `cntdb` in the swss vs tests. Maybe we should consider transitioning to 3/4-letter abbreviations (`appdb`, cfgdb`, `cntdb`, `asicdb`, etc.)?

lguohan: sure. changed to cfgdb

lguohan:
retest this please

config/main.py
```diff
@@ -1060,7 +1060,8 @@ def load(filename, yes):
 @click.option('-l', '--load-sysinfo', is_flag=True, help='load system default information (mac, portmap etc) first.')
 @click.option('-n', '--no_service_restart', default=False, is_flag=True, help='Do not restart docker services')
 @click.argument('filename', required=False)
-def reload(filename, yes, load_sysinfo, no_service_restart):
+@pass_db
```
judyjoseph, line 1063: Most of the CLI handlers have the @click.pass_context decorator .. where we pass the ctx to the command handler. I see here too we store the Db() object in the context ctx.obj = Db() -- So could we pass the pass_context decorator itself ....something like 
ctx.obj['globalDB'] = Db()
and use the @click.pass_context decorator itself ? 

Would this approach also help than creating a separate decorator for db object ? would like to get your opinion. It could be a generic approach to add various objects into the ctx and use it.

lguohan: check out here. the ctx is a linked list, since we created the ctx at the very top level, it could be the case some group will create its own ctx, and we would need to use find_object to find the top level ctx. therefore, click introduce this explicit pass decorator which allow you to get the ctx by object type. 

https://click.palletsprojects.com/en/7.x/complex/?highlight=make_pass_decorator

at any level, you can directly access this ctx.

lguohan:
@jleveque, any update? I have a few other pr lined up to break the file?

config/main.py
judyjoseph: Thanks @lguohan for the pointer for pass_decorator. I had one question though .. on the config command handlers which already have a  ctx being passed  from parent -- will we be able to use this new @pass_db which is nothing but a context created based on object_type. 
eg:
@config.group(cls=AbbreviationGroup)
@click.pass_context
def portchannel(ctx):
    config_db.connect()
    ctx.obj = {'db': config_db}

lguohan: currently, that code still works, but to unified the logic, we need to refactor that piece of code.

lguohan:
retest this please

lguohan:
retest this please

lguohan:
failure not related to the pr.

PR #1031, opened 2020-08-06
[show] Moving some py imports inside specific show cmds

Sabareesh-Kumar-Anandan:
Some py imports are taking more time causing delay in show cmds.
Moved some imports into specific cmds to improve cli exec time.

 Signed-off: Sabareesh-Kumar-Anandan <sanandan@marvll.com>

**- What I did**
Some py imports in show/main.py script are taking more time on arm cpus.
modules - pkg_resources, netaddr, portconfig, collections are taking around 800ms causing delay for all show cmds.
These modules are used by only specific cmds. So, I have included these imports only for the commands that are using them.

**- How I did it**

**- How to verify it**
Test all show commands. Execution time of show commands is better than before.


lguohan:
i feel this is the probably the right way to approach the problem. we should split the file into multiple small files and only load corresponding file when needed.

https://click.palletsprojects.com/en/7.x/commands/?highlight=plugin#custom-multi-commands

PR #1035, opened 2020-08-07
fanshow: Keep some backward compatibility before FanDrawer introduction

Staphylo:
The new FanDrawer API is meant to supersede the direct Fan declaration
in the platform API.
However, some vendors might have not yet implemented it.
For those, the `show platform fan` cli got broken when the DRAWER_NAME
column was added to the output.

To preserve some backward compatibility with the platform API before the
introduction of the FanDrawer, if the DRAWER_NAME is not defined in the
database, `Unknown` will be displayed.

This simple change avoids breaking `show platform fans`

Reformat the tuple vertically to avoid a long line and improve clarity.

### DIFF
--- a/scripts/fanshow
+++ b/scripts/fanshow
@@ -60,9 +60,17 @@ class FanShow(object):
             else:
                 status = 'N/A'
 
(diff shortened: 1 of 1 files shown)


sheiun:
This commit is quite significant. Even if 'FanDrawer' is proposed, the 'Chassis Fan' and 'PSU Fan' still won't have the 'drawer_name' attribute. Would using 'N/A' be better than 'Unknown' in this case?

PR #1036, opened 2020-08-07
Change fast-reboot script to use swss and radv service script

vaibhavhd:
**- What I did**
Special handling is needed for services during fast, warm and cold reboot paths. This procedural handling is better suited in the respective service scripts to:
1) keep fast-reboot script simpler    2) avoid code duplication.

Call swss and radv service scripts to do special handling of reboot process behavior.

**_This PR should be merged only after service scripts changes are merged in following PRs:_**
RADV - https://github.com/Azure/sonic-buildimage/pull/5108
SWSS - https://github.com/Azure/sonic-buildimage/pull/5070

**- How I did it**
Created and modified service scripts to handle reboot procedure depending on the type. Fast-reboot script now simply calls these service scripts.
**- How to verify it**
Tested service scripts independently.
Tested Fast, warm and cold reboot procedures are tested.

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -503,8 +503,7 @@ systemctl stop nat
 debug "Stopped nat ..."
 
 # Kill radv before stopping BGP service to prevent annoucing our departure.
-debug "Stopping radv ..."
-docker kill radv &>/dev/null || [ $? == 1 ]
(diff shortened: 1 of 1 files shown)


lguohan:
retest this please

PR #1043, opened 2020-08-11
Python 3 compliance

jleveque:
As part of the migration from Python 2 to Python 3, this PR ensures that all Python files in sonic-utilities are Python 3-compliant, and work with both Python 2 and Python 3.

Once this is merged, we can begin building a Python 3-based version of the sonic-utilities package.

### DIFF
--- a/pfc/main.py
+++ b/pfc/main.py
@@ -45,9 +45,9 @@ def showPfcAsym(interface):
 
     sorted_table = natsorted(table)
 
-    click.echo('\n')
+    click.echo()
     click.echo(tabulate(sorted_table, headers=header, tablefmt="simple", missingval=""))
-    click.echo('\n')
+    click.echo()
 
 def configPfcPrio(status, interface, priority):
     configdb = swsssdk.ConfigDBConnector()
@@ -110,9 +110,9 @@ def showPfcPrio(interface):
         table.append([intf, entry.get('pfc_enable', 'N/A')])
     
     sorted_table = natsorted(table)
-    click.echo('\n')
+    click.echo()
     click.echo(tabulate(sorted_table, headers=header, tablefmt="simple", missingval=""))
-    click.echo('\n')
+    click.echo()
     
 @click.group()
 def cli():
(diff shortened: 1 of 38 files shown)


lguohan:
not sure why the build fails here.

pfc/main.py
```diff
@@ -110,9 +110,9 @@ def showPfcPrio(interface):
         table.append([intf, entry.get('pfc_enable', 'N/A')])
     
     sorted_table = natsorted(table)
-    print '\n'
-    print tabulate(sorted_table, headers=header, tablefmt="simple", missingval="")
-    print '\n'
+    click.echo('\n')
```
qiluo-msft, line 113:
```suggestion
    click.echo()
```

scripts/watermarkstat
```diff
@@ -56,7 +56,7 @@ class Watermarkstat(object):
         def get_queue_type(table_id):
             queue_type = self.counters_db.get(self.counters_db.COUNTERS_DB, COUNTERS_QUEUE_TYPE_MAP, table_id)
             if queue_type is None:
-                print >> sys.stderr, "Queue Type is not available!", table_id
+                sys.stderr.write("Queue Type is not available in table '{}'".format(table_id))
```
qiluo-msft, line 59: How about `print(*args, file=sys.stderr)`

jleveque: I considered it, but this method results in a shorter line.

jleveque:
@lguohan:

> not sure why the build fails here.

It looks like `config reload` is failing because the new port breakout functionality is trying to obtain the hwsku, but the database isn't populated: https://github.com/Azure/sonic-utilities/blob/master/config/main.py#L63

Not sure why we just started seeing this failure on this PR, though; it should have existed.

pfc/main.py
jleveque: Changed all occurrences.

scripts/watermarkstat
jleveque: Changed all instsances.

PR #1045, opened 2020-08-11
[config] Eliminate port breakout-related globals

jleveque:
Port breakout-related data should be gathered on-demand, not every time `config` is executed. If the ConfigDB is not ready, the call to get hwsku will hang indefinitely, which will occur when loading config for the first time. Also, if it fails to retrieve the port breakout globals, it will abort, even if the executed command was unrelated to port breakout. This is not desirable behavior.

This PR eliminates port breakout-related global variables, and encapsulates them in functions to be called on-demand, only when commands which require the data are executed. It is currently only accessed in two places. If we feel the need to cache it in the future for efficiency, we can look into adding it to the Click context.

Also rename `_get_option()` to `_get_breakout_cfg_file_name()` to add more detail.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -56,22 +56,6 @@ CFG_LOOPBACK_NO="<0-999>"
 asic_type = None
 
 #
-# Load breakout config file for Dynamic Port Breakout
-#
-
-try:
-    (platform, hwsku) = device_info.get_platform_and_hwsku()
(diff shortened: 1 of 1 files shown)


jleveque:
@samaity: Please review.

samaity:
LGTM 👍  , @jleveque

PR #1046, opened 2020-08-11
Add ip_prefix len based on proxy_arp status

sumukhatv:
**- What I did**
Add ip_prefix length data to ip_mappings based on the proxy_arp status
**- How I did it**
Added the logic in construct_neighbor_advertiser_slice method
**- How to verify it**


scripts/neighbor_advertiser
```diff
@@ -281,28 +285,53 @@ def construct_neighbor_advertiser_slice():
...
-            ipv4_mappings.append(mapping)
-
         ipv6_mappings = []
-        for addr in ipv6_addr:
-            mapping = {
-                'ipAddr': addr,
-                'macAddr': mac_addr
-            }
-            ipv6_mappings.append(mapping)
+        if vlan_intf_table[vlan_interface]['proxy_arp'] == 'enabled':
```
prsunny, line 299: Lets not repeat the code here, just update the prefixlen based on the 'proxy_arp' setting.

scripts/neighbor_advertiser
```diff
@@ -281,28 +285,53 @@ def construct_neighbor_advertiser_slice():
...
+                    'ipAddr': addr,
+                    'ipPrefixLen': 32,
+                    'macAddr': mac_addr
+                }
+                ipv4_mappings.append(mapping)
+
+            for addr in ipv6_addr:
+                mapping = {
+                    'ipAddr': addr,
+                    'ipPrefixLen': 32,
```
prsunny, line 331: This is not 32, it should be 128.

scripts/neighbor_advertiser
```diff
@@ -287,6 +296,8 @@ def construct_neighbor_advertiser_slice():
 
     vlan_intf_table = config_db.get_table('VLAN_INTERFACE')
 
+    vxlanPort = appl_db.get_table('SWITCH_TABLE:switch')['vxlan_port']
```
prsunny, line 299: You may want to check if 'vxlan_port' exists..

scripts/neighbor_advertiser
```diff
@@ -281,32 +294,58 @@ def construct_neighbor_advertiser_slice():
...
             log.log_warning('Cannot find mac addr of vlan interface {}'.format(vlan_interface))
             continue
 
         ipv4_mappings = []
+        ipv6_mappings = []
+        
+        ctr = 0
         for addr in ipv4_addr:
+            if 'proxy_arp' in vlan_intf_table[vlan_interface]:
+                if vlan_intf_table[vlan_interface]['proxy_arp'] == 'enabled':
```
prsunny, line 314: is it better to check with an "and" operator? Just a suggestion to avoid extra "else"

lguohan:
wonder if we can add unit test for this script? examples, under tests/ folder.

scripts/neighbor_advertiser
```diff
@@ -243,8 +254,10 @@ def get_vlan_addresses(vlan_interface):
             keys=line.strip().split(' ')
             if keys[0] == 'inet':
                 ipv4_addr.append(keys[1].split('/')[0])
+                ipv4_prefix.append(keys[1].split('/')[1])
```
yxieca, line 257: This code assume that all input come in '/' format. Is that the expectation? 260 assumes the same.

sumukhatv: Line 256 and 259 are obtaining ipv4 and ipv6 addresses respectively based on the same assumption. The sample output also shows the same representation

lguohan:
resurfacing this qeustion: wonder if we can add unit test for this script? examples, under tests/ folder.

current the unit test coverage for this file is zero.

https://sonic-jenkins.westus2.cloudapp.azure.com/job/common/job/sonic-utilities-build/RCov_20Report/

sumukhatv:
I will write some unittests

lguohan:
thanks, you can find the coverage report under htmlcov

https://sonic-jenkins.westus2.cloudapp.azure.com/job/common/job/sonic-utilities-build/RCov_20Report/

scripts/neighbor_advertiser
```diff
@@ -236,42 +236,22 @@ def get_vlan_addresses(vlan_interface):
...
-                ipv6_addr.append(keys[1].split('/')[0])
-                ipv6_prefix.append(keys[1].split('/')[1])
-            elif keys[0] == 'link/ether':
-                mac_addr = keys[1]
-    except Exception:
-        log.log_error('failed to get %s addresses from o.s.' % vlan_interface)
+    vlan_intfs = config_db.get_table('VLAN_INTERFACE')
+    for item in vlan_intfs.keys():
+        if vlan_interface in item[0]:
+            ipaddr = IPNetwork(item[1])
```
prsunny, line 242: there would be cases where there is no item[1], so we would have to skip those. Like just this "VLAN_INTERFACE|Vlan2000". I would suggest to just use `get_vlan_addr ` function as it already handles that case

scripts/neighbor_advertiser
```diff
@@ -204,62 +208,37 @@ def get_vlan_interface_vxlan_id(vlan_intf_name):
 def get_vlan_addr(vlan_intf_name, ip_ver):
```
yxieca, line 208: Rename this function to get_vlan_addr_prefix?

scripts/neighbor_advertiser
```diff
@@ -281,28 +260,48 @@ def construct_neighbor_advertiser_slice():
 
     all_vlan_interfaces = get_vlan_interfaces()
 
+    vlan_intf_table = config_db.get_table('VLAN_INTERFACE')
+
+    vxlanPort = None
+    if 'vxlan_port' in appl_db.get_all(appl_db.APPL_DB, 'SWITCH_TABLE:switch'):
+        vxlanPort = appl_db.get(appl_db.APPL_DB, 'SWITCH_TABLE:switch', 'vxlan_port')
```
yxieca, line 267: This code block looks not efficient. Would appl_db.get() return None when key not existing?

abdosi:
@sumukhatv Please create PR for 201911

abdosi:
@sumukhatv / @prsunny  i have cherry-picked in 201911. Test file could not get cherry-picked. Please see if things are fine with python2 and there is no exception.

PR #1048, opened 2020-08-12
[watermarkstat] Import print_function from __future__

jleveque:
To support the `file=` argument in `print()` function in Python 2.7.


qiluo-msft:
Is it only for python3 compatibility?

jleveque:
It's for backwards-compatibility with Python 2.7.

PR #1049, opened 2020-08-12
[teamshow]: refactor teamshow to use state db information

lguohan:
**- What I did**
previously, teamshow needs to read data from teamdctl command
output. As teamdmon daemon has been developed to put data into
state db. Now, teamshow can get data from state db.

Also remove teamshow command and incoporate as part of show
command.

Also add extensive unit tests

**- How I did it**
let teamshow read data from state db

**- How to verify it**
unit test and dut test

coverage
```
show/portchannel.py                           92      5    95%
```


show/portchannel.py
```diff
@@ -1,4 +1,9 @@
-#!/usr/bin/python
```
jleveque, line 1: This file is now "show/portchannel.py". If we put all the files directly under show/ we could wind up with naming collisions. I think it would be best to create a directory hierarchy for groups, such that the directory structure mirrors the group/subcommand nesting. E.g, show/interfaces/portchannel.py.

PR #1051, opened 2020-08-13
[filter-fdb] Call Filter FDB Main From Within Test Code

tahmed-dev:
Code coverage requires that python code be run with the same process.
Current test code was invoking filter fdb via shell which launches
new process and so coverage is not available. This PR calls
the main method from within test code.

**- What I did**
Change how main function is invoked

**- How I did it**
code refactor

**- How to verify it**
unit test passes

N/A
N/A


lguohan:
can you fix the build error?

lguohan:
in this case, it probably better to move this script as a console script.

lguohan:
looks like build failure.

PR #1052, opened 2020-08-14
[config] Reduce Calls to SONiC Cfggen

tahmed-dev:
Calls to sonic-cfggen is CPU expensive. This PR reduces calls to
sonic-cfggen during config-setup when configuring buffer/qos.

singed-off-by: Tamer Ahmed <tamer.ahmed@microsoft.com>

**- What I did**
Reduce time required to start config-setup service

**- How I did it**
Used template batch mode to batch together 4 calls into one call to sonic-cfggen

**- How to verify it**
flame graph showed gain of ~1.5% sonic-cfggen CPU util.

N/A

N/A

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1562,40 +1562,35 @@ def reload():
                 click.secho(
                     "Command 'qos reload' failed with invalid namespace '{}'".
                         format(ns),
-                    fg='yellow'
+                    fg="yellow"
                 )
                 raise click.Abort()
             asic_id_suffix = str(asic_id)
 
-        buffer_template_file = os.path.join(hwsku_path, asic_id_suffix, 'buffers.json.j2')
+        buffer_template_file = os.path.join(hwsku_path, asic_id_suffix, "buffers.json.j2")
         if os.path.isfile(buffer_template_file):
-            cmd_ns = "" if ns is DEFAULT_NAMESPACE else "-n {}".format(ns)
-            command = "{} {} -d -t {},config-db".format(
-                SONIC_CFGGEN_PATH,
-                cmd_ns,
-                buffer_template_file,
-            )
-
-            qos_template_file = os.path.join(hwsku_path, asic_id_suffix, 'qos.json.j2')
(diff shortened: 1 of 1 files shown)


tahmed-dev:
Will remove commit from the other branch.

scripts/__init__.py
```diff
@@ -0,0 +1 @@
+#!/usr/bin/env python
```
jleveque, line 1: This line should not be necessary, as this file is not executable.

jleveque:
> Will remove commit from the other branch.

Which commit and which branch are you referring to?

scripts/__init__.py
tahmed-dev: Sorry, I had a couple of commits from another branch since I did not branch off master. removed those commits which had this file. The file is not part of this commit

tahmed-dev:
> > Will remove commit from the other branch.
> 
> Which commit and which branch are you referring to?

I branched off another branch and so there commit that is not in master nor part of this particular PR. Removed those now. Sorry about this inconvenience.

tahmed-dev:
retest default please

tahmed-dev:
retest this please

jleveque:
@tahmed-dev: FYI, "retest default please" will not work. If there is only one test named "default", you must use "retest this please".

tahmed-dev:
retest this please

config/main.py
```diff
@@ -1567,54 +1567,31 @@ def reload():
...
-                clicommon.run_command(command, display_cmd=True)
             else:
                 click.secho('QoS definition template not found at {}'.format(
                     qos_template_file
                 ), fg='yellow')
+
+            command += " --write-to-db"
+            # Apply the configurations only when both buffer and qos
+            # configuration files are presented
+            clicommon.run_command(command, display_cmd=True)
```
neethajohn, line 1594: This section has to move up inside the if block checking the presence of qos template file.

tahmed-dev: @neethajohn, this is delayed run. The if block collected the required parameter already in line 1582 above.

tahmed-dev: Good catch! Will fix.

PR #1055, opened 2020-08-15
[201911] fix show version 

abdosi:
what I did:
Fix show version on 201911. When using new sonic_py-common package
we have to use get_platform() instead of get_platform_info()

How i Verified:
show version is fine after this change

### DIFF
--- a/scripts/decode-syseeprom
+++ b/scripts/decode-syseeprom
@@ -26,7 +26,7 @@ def main():
         raise RuntimeError("must be root to run")
 
     # Get platform name
-    platform = device_info.get_platform_info()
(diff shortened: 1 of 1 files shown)


jleveque:
@abdosi: It also appears as though the same fix is needed in the `get_npu_device_id()` function in sonic_py_common/device_info.py in the 201911 branch.

abdosi:
> @abdosi: It also appears as though the same fix is needed in the `get_npu_device_id()` function in sonic_py_common/device_info.py in the 201911 branch.

Yes, this fix and one more change will be needed. Creating PR for same.

PR #1069, opened 2020-08-24
[config] Remove _get_breakout_cfg_file_name helper function

vdahiya12:
this PR addresses replacing _get_breakout_cfg_file_name
function with device_info.get_path_to_port_config_file().
This is part of a larger clean up of functions which return
the port_config_file which reside outside of
sonic_py_common packages. This is motivated due to
redundancy of code as well as making it easier to change
the default behaviour of helper utility function
port_config_file unified and in one place.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -12,7 +12,7 @@ import sys
 import threading
 import time
 
-from minigraph import earse_device_desc_xml
+from minigraph import parse_device_desc_xml
 from portconfig import get_child_ports, get_port_config_file_name
 from sonic_py_common import device_info
 from swsssdk import ConfigDBConnector, SonicV2Connector, SonicDBConfig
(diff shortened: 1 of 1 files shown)


config/main.py
```diff
@@ -12,7 +12,7 @@
 import threading
 import time
 
-from minigraph import parse_device_desc_xml
+from minigraph import earse_device_desc_xml
```
jleveque, line 15: Looks like an accidental change.

config/main.py
```diff
@@ -12,7 +12,7 @@
...
 from portconfig import get_child_ports, get_port_config_file_name
```
jleveque, line 16: It appears that `get_port_config_file_name` is no longer needed now, either.

config/main.py
vdahiya12: fixed

config/main.py
vdahiya12: fixed

jleveque:
@samaity: Please review.

samaity:
@jleveque @vdahiya12 , LGTM!

PR #1103, opened 2020-09-09
[drop counters] Fix configuration for counters with lowercase names

daall:

**- What I did**
I fixed in a bug in the drop counters CLI where counters that weren't provided in uppercase were not being installed properly into Config DB.

Fixes https://github.com/Azure/sonic-buildimage/issues/5303

**- How I did it**
I figured out that `swssdk` normalizes all table names to uppercase. Since I was providing the drop counter name as part of the table name instead of part of the key, the drop counter name was being set to all uppercase. This is problematic because the counter was being installed into config DB in whatever case the user provided since for that table it is treated as a key.

So, I updated the parts of the code that touch the `DROP_REASONS` table to include the counter name as part of the key rather than as part of the table name.

**- How to verify it**
I've updated the unit tests to include a `lowercase_counter` to check for this behavior. I also validated that all four `config` functions work properly on a physical DUT.


jleveque (approved):

prsunny:
retest this please

daall:
retest this please

prsunny (approved):

daall:
retest this please

daall:
retest this please

PR #1107, opened 2020-09-11
Skip default route present in ASIC-DB but not in APP-DB.

renukamanavalan:
As this is not a bug to report.


prsunny (approved):
lgtm, just to confirm, this will report error if default route is present in app_db but not in asic_db, right?

prsunny:
retest this please

jleveque (approved):

PR #1119, opened 2020-09-17
[vrf] fix check state_db error when vrf moving

tylerlinp:
**- What I did**
Fix a bug that config an interface to a new vrf failed.

**- How I did it**
Check the interface KEY existence instead of KFV in state_db, because of there is no `state` field in new schema.

**- How to verify it**
Bind an interface to a new vrf and check interface master right.


prsunny (approved):

prsunny:
retest this please

PR #1121, opened 2020-09-18
Add vrf naming restrictions

liron-ze:
**- What I did**
Add vrf naming restrictions

**- How I did it**
Limit the characters that can be used in naming

**- Why I did it**
When VRF is named with some special symbols, the vrfmgrd process will be terminated
For example:  sudo config vrf add Vrf:a

**- How to verify it**
sudo config vrf add Vrf:a
ps -ef|grep vrfmgrd


jleveque:
Retest this please

PR #1122, opened 2020-09-19
Support building sonic-utilities as a Python wheel package instead of a Debian package

jleveque:
**- What I did**

Support building sonic-utilities as a Python wheel package rather than a Debian package.

Notes:
- Wheel packaging restricts the installation of files to within the dist-packages directory. Thus, we cannot install data files to system directories. Therefore, I am building a separate 'sonic-utilities-data' package for installing the data files (bash_completion, templates)
- Wheel packages install scripts/entrypoints to the /usr/local/bin directory, whereas Debian packages install to /usr/bin

**- How I did it**

- Specify all Python wheel dependencies in setup.py
- Remove `data_files` section from setup.py
- Build an extra 'sonic-utilities-data' Debian package which will handle installing data files to system directories
- Update all references to sonic-utilities scripts/entrypoints to either reference the new /usr/local/bin/ location or remove absolute path entirely
- Update .gitignore file as necessary

**- How to verify it**

Ensure all CLI-related functionality continues to work as expected

NOTE: PR tests will fail until the build job is updated to build as a Python wheel. PR here: https://github.com/Azure/sonic-build-tools/pull/160


scripts/fast-reboot
```diff
@@ -324,7 +324,7 @@ function reboot_pre_check()
     fi
     
     # Make sure ASIC configuration has not changed between images
-    ASIC_CONFIG_CHECK_SCRIPT="/usr/bin/asic_config_check"
+    ASIC_CONFIG_CHECK_SCRIPT="/usr/local/bin/asic_config_check"
```
lguohan, line 327: is it doable not to change the path?

jleveque: Unfortunately, no. Wheels install to `/usr/local/bin`, whereas .debs install to `/usr/bin`. There's no way to change this behavior.

jleveque:
Retest this please

jleveque:
Retest this please

lguohan (approved):

jleveque:
After reviewing the check build logs, it is apparent there is a chicken/egg problem, as the build copies artifacts from the last successful build of buildimage-vs-all. Therefore, the new sonic-utilities wheel package built here will not even get installed in the VS container before running the swss tests, making this test moot. I will merge this PR and update the submodule in https://github.com/Azure/sonic-buildimage/pull/5409 and iterate on that PR if necessary.

PR #1123, opened 2020-09-22
Fixed config load_minigrpah not working for Multi-asic platfroms.

abdosi:
Why I did:
config load_minigraph not working on Multi-asic platforms. 
Command config pfcwd start default now runs on all asic so if we using
it with ip netns exec asic0 it creates loop where asic0 can not reach redis server of other asic.

How I did:
pfcwd start default should be called only once.

Verified:

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1182,13 +1182,10 @@ def load_minigraph(db, no_service_restart):
         clicommon.run_command(command, display_cmd=True)
         client.set(config_db.INIT_INDICATOR, 1)
 
-        # get the device type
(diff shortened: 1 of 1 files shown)


smaheshm:
retest this please

smaheshm:
Looks good, lets wait for the test results.

jleveque:
Retest this please

smaheshm (approved):
Looks good. Please wait for tests to pass.

abdosi:
Retest this please

PR #1124, opened 2020-09-22
[201911] Fix load minigraph for multi-asic platforms

abdosi:
Why I did:
config load_minigraph not working on Multi-asic platforms.
Command config qos reload now runs on all asic so if we using
it with ip netns exec asic0 it creates loop where asic0 can not reach redis server of other asic.

We need this PR https://github.com/Azure/sonic-utilities/pull/1123 in 201911 for complete fix.

How I did:
config qos reload should be called only once.

Verified:

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -994,7 +994,7 @@ def load_minigraph(no_service_restart):
         run_command("acl-loader update full /etc/sonic/acl.json", display_cmd=True)
 
     # generate QoS and Buffer configs
-    run_command("{}config qos reload".format(ns_cmd_prefix), display_cmd=True)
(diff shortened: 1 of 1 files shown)


smaheshm (approved):
Looks good.

smaheshm:
why don't fix the pfcwd start too, or you prefer cherry picking from master?

abdosi:
> why don't fix the pfcwd start too, or you prefer cherry picking from master?

yes cherry-pick

abdosi:
retest this please

jleveque:
Retest this please

jleveque:
Retest this please

PR #1125, opened 2020-09-22
[setup.py] Update dependencies

jleveque:
Add more dependencies to setup.py. Now that we're building as a wheel, we can add all dependent packages to make building and installation more robust, as there will no longer be a need to install them explicitly.

Move sonic-config-engine from `install_requires` to `tests_require`, as it is only needed by the unit tests.

Remove code to install fastentrypoints package if not installed. When building a Python 3 wheel using python3 setup.py bdist_wheel, we would receive a permissions error when calling easy_install.main(['fastentrypoints']), as with Python 3 it appears elevated permissions are required to install the package. This appears to be a behavior change between the Python 2 and Python 3 versions of setuptools. This needs to be explicitly installed before building the package. We are already installing in the SONiC slave container. Will update the sonic-utilities README to specify, also.


jleveque:
Ignoring LTGM alerts. One is from a file which is not modified in this PR; the other is in fastentrypoints.py which is taken verbatim from https://github.com/ninjaaron/fast-entry_points. I prefer to not modify that file, but rather open an issue against that repo.

lguohan (approved):

lguohan:
@jleveque , can you check the build failure?

jleveque:
Build failure should be fixed via https://github.com/Azure/sonic-build-tools/pull/160

jleveque:
Retest this please

jleveque:
Retest this please

jleveque:
Retest this please

PR #1132, opened 2020-09-26
[config] Reload Monit configuration after potential hostname changes

jleveque:
**- What I did**

After hostname changes, we need to reload the Monit configuration in order for Monit to pick up the new system hostname. Otherwise, it will retain the previous hostname.

**- How I did it**

Call `sudo monit reload` after restarting hostname-config.service.

**- How to verify it**

1. Modify hostname in config_db.json, run `config reload`, then run `monit status` and ensure hostname is updated
2. Change hostname using `config hostname` command, then run `monit status` and ensure hostname is updated


jleveque:
Retest this please

lguohan (approved):

PR #1142, opened 2020-09-30
[acl-loader] Revert fix for IP protocol == 0

daall:

**- What I did**
Fixes https://github.com/Azure/sonic-buildimage/issues/5469

It appears that the openconfig model includes a 0 default value, so there's no way to differentiate between a user passing in 0 and a user passing in no value at all, so this will add a "protocol == 0" qualifier to any ACL without a protocol explicitly defined.

**- How I did it**
I removed the `protocol == 0` check.

**- How to verify it**
Load an acl.json (like the one from the issue) with a rule with no IP protocol defined. Verify that no IP protocol qualifier is added to this rule.

### DIFF
--- a/acl_loader/main.py
+++ b/acl_loader/main.py
@@ -413,7 +413,9 @@ class AclLoader(object):
     def convert_ip(self, table_name, rule_idx, rule):
         rule_props = {}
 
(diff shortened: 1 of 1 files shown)


bratashX (approved):

bratashX:
@lguohan @yxieca, could you please review this PR

jleveque (approved):

PR #1148, opened 2020-10-06
Fix fast-reboot-dump.py ipaddress error

kuanyu99:
**- What I did**

Fix weird behavior when using ipaddress to parse the ipv4 and ipv6 address.

**- How I did it**

Instead of using ipaddress.ip_interface(IP_ADDR), I change to use ipaddress.IPv4Address to check if the ip address is ipv4.
The same method is also used inside sonic-utilities/config/nat.py

**- How to verify it**

When there are some mac learned by SONIC, execute the fast-reboot. It should not return error.

```
Sep 28 08:26:16.418276 as8000-3 ERR fast-reboot-dump: Got an exception '192.168.0.208' does not appear to be an IPv4 or IPv6 interface:
Traceback:
  File "/usr/bin/fast-reboot-dump.py", line 298, in <module>
    res = main()
  File "/usr/bin/fast-reboot-dump.py", line 289, in main
    neighbor_entries = generate_neighbor_entries(root_dir + '/arp.json', all_available_macs)
  File "/usr/bin/fast-reboot-dump.py", line 42, in generate_neighbor_entries
    if ipaddress.ip_interface(ip_addr).ip.version != 4:
  File "/usr/local/lib/python2.7/dist-packages/ipaddress.py", line 239, in ip_interface
    address)
ValueError: '192.168.0.208' does not appear to be an IPv4 or IPv6 interface
```

### DIFF
--- a/scripts/fast-reboot-dump.py
+++ b/scripts/fast-reboot-dump.py
@@ -39,9 +39,12 @@ def generate_neighbor_entries(filename, all_available_macs):
         arp_output.append(obj)
 
         ip_addr = key.split(':')[2]
-        if ipaddress.ip_interface(ip_addr).ip.version != 4:
+        try:
+            ipaddress.IPv4Address(ip_addr)
+        except ipaddress.AddressValueError:
             #This is ipv6 address
             ip_addr = key.replace(key.split(':')[0] + ':' + key.split(':')[1] + ':', '')
+
         neighbor_entries.append((vlan_name, mac, ip_addr))
         syslog.syslog(syslog.LOG_INFO, "Neighbor entry: [Vlan: %s, Mac: %s, Ip: %s]" % (vlan_name, mac, ip_addr))
 
--- a/scripts/fast-reboot-dump.py
+++ b/scripts/fast-reboot-dump.py
@@ -40,7 +40,7 @@ def generate_neighbor_entries(filename, all_available_macs):
 
         ip_addr = key.split(':')[2]
         try:
-            ipaddress.IPv4Address(ip_addr)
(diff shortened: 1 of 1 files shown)


kuanyu99:
Retest this please

kuanyu99:
Retest this please

tahmed-dev:
@kuanyu99 Thanks for efforts. the issue could be fixed in Python 2 bu decoding the string as unicode:
```
print("IP version: %s" % ipaddress.ip_interface('192.168.0.208'.decode()).ip.version)
IP version: 4
```

In Python 3 this above is not required.

kuanyu99:
@tahmed-dev Thank you very much! This information help me a lot.

But there are still have other problems.
The ipv6 version fdb items in the redis-db looks like "NEIGH_TABLE:Vlan1000:fe80::XXXXXX". After parsing, the ip_addr equal to "fe80" which will return AddressValueError during calling ipaddress.

lguohan:
is there anything missing in the unit test?

bingwang-ms:
I found the same issue and created another PR #1164 to do a fix. Sorry for not founding this PR. Both PR address the issue, but I think #1164 is more elegant in both python 2 and python 3.

kuanyu99:
> I found the same issue and created another PR #1164 to do a fix. Sorry for not founding this PR. Both PR address the issue, but I think #1164 is more elegant in both python 2 and python 3.

No problem. I didn't consider the python3 case. I also think your solution is better.
I will close this ticket once that PR is merged.

kuanyu99:
Due to #1164 has been merged, close this PR.

PR #1150, opened 2020-10-06
[pfcstat] Create new copies of Rx and Tx dictionaries

neethajohn:
**- What I did**
Fixes Azure/sonic-buildimage#5540

The issue was due to pfcstat rx and tx dict values copied using '=' and hence the Rx counts were always getting overwritten by the Tx counts
Unit tests didn't catch the problem because the Tx and Rx counts were set to the same value.

**- How I did it**
create new copies of pfcstat Rx and Tx dictionaries
Updated unit tests to use different values for Tx and Rx counts

**- How to verify it**
Without the change, modified the mock counter values for Rx and Tx to be different. The pfcstat tests were failing.
With the changes, send PFC pause frames from the fanout and from the DUT and ensured that values were getting reflected correctly for Rx and Tx.


smaheshm:
Looks good. Thanks for working on the fix!!

smaheshm (approved):
:shipit:

lguohan (approved):

PR #1152, opened 2020-10-09
[Multi-asic] Enhanced Feature Table configuration for multi-asic platforms

abdosi:
**- What I did**
Enhanced Feature Table configuration for multi-asic platforms
to programmed for all config db's.

**- How I did**

Following changes were done:-

- Feature Table config command check all the DB's values are same else it will fail

- Feature Table show command always display from Host/Global Config DB. 

- Also made sure some of other test cases (crm/bgp route) that import mock_*_asic to do using reload otherwise we can see failure. 

**- How to verify it**
- Unit test added specific for Multiasic.
- Test case verfiy if DB's content are not same config command fails
- Test case verify if if DB's content are same config and show command passes.


jleveque:
Retest this please

lguohan:
can you enhance the unit test to cover multi asic ones?

abdosi:
@jleveque can you please review. I have updated PR with new set of changes.

abdosi:
FYI @judyjoseph and @gechiang 
Please see the changes done in crm and bgp test for using reload() the  module

tests/crm_test.py
```diff
@@ -1,3 +1,4 @@
+from importlib import reload
```
jleveque, line 1: Suggest to organize `from ...` imports _after_ `import ...` imports within the same group, i.e.:

```python
import os
import sys
from importlib import reload
```

jleveque (changes requested):

tests/crm_test.py
abdosi: @jleveque updated

jleveque (dismissed):

utilities_common/multi_asic.py
```diff
@@ -148,3 +148,13 @@ def multi_asic_args(parser=None):
     parser.add_argument('-n', '--namespace', default=None,
                         help='Display interfaces for specific namespace')
     return parser
+
+class MultiAsicDb(object):
```
judyjoseph, line 152: @abdosi do we really need this new class, I see that the Db object itself is enhanced to supports multi-asic ? https://github.com/Azure/sonic-utilities/blob/master/utilities_common/db.py#L7

abdosi: @judyjoseph as part of next PR i will be removing that file and will be using MultiAsicDb so that code remains consistent in multiasic.py

abdosi:
retest this please

utilities_common/multi_asic.py
abdosi: @judyjoseph On second thought I will close my PR#https://github.com/Azure/sonic-utilities/pull/1312 as I feel that will create conflict for other outstanding PR's and use existing Db class and update this PR accordingly.

Thanks for pointing out.

judyjoseph (dismissed):
LGTM

tests/ip_show_routes_multi_asic_test.py
```diff
@@ -1,5 +1,6 @@
 import os
 
+from importlib import reload
```
jleveque, line 3: importlib is part of the standard library. These imports should be grouped as follows:

```python3
import os
from importlib import reload

import pytest
```

jleveque (changes requested):

tests/ip_show_routes_multi_asic_test.py
abdosi: @jleveque Updated.. Remove blank Line,

jleveque (approved):

abdosi:
retest this please

abdosi:
retest this please

PR #1154, opened 2020-10-09
Add `show reboot-history`

sujinmkang:
**- What I did**
To make the reboot cause history information available for telemetry service, during processing the reboot-cause on boot, the information is saved in state-db.
The cli command `show reboot-cause` is changed to support the new file format and file name.
New command `show reboot-cause history` is added to display the reboot-cause information which is stored in state-db.
**- How I did it**
Add the 'show reboot-cause history' command to read up to  the10 previous reboot-cause information from state-db and display them.
Read the last reboot-cause from the new file "previous-reboot-cause.json" and process the data to display for `show reboot-cause` 

**- How to verify it**
Unit test added.
Verified also in DUT

```
admin@sonic:~$ show reboot-cause history
name                 cause        time                          user    comment
-------------------  -----------  ----------------------------  ------  ---------
2020_10_09_04_53_58  warm-reboot  Fri Oct  9 04:51:47 UTC 2020  admin
2020_10_09_02_33_06  reboot       Fri Oct  9 02:29:44 UTC 2020  admin
2020_10_09_02_00_53  fast-reboot  Fri Oct  9 01:58:04 UTC 2020  admin
2020_10_09_01_56_59  reboot       Fri Oct  9 01:53:49 UTC 2020  admin
```

HLD : https://github.com/Azure/SONiC/pull/669


jleveque (changes requested):
I feel like this command should be in the same group as `show reboot-cause`.

Either:

```
show reboot-cause
show reboot-cause history
```
or
```
show reboot cause
show reboot history
```

However, if we go with the latter, we would also need to add an alias from `show reboot-cause` to `show reboot cause` to retain backward-compatibility.

lguohan:
@sujinmkang , please add unit test.

lguohan (changes requested):
please add unit tests

show/main.py
```diff
@@ -1803,20 +1803,74 @@ def mmu():
...
-    # the contents of the 'reboot cause' file as it was left when the device
-    # went down for reboot. This file should always be created at boot,
-    # but check first just in case it's not present.
-    if not os.path.isfile(PREVIOUS_REBOOT_CAUSE_FILE):
-        click.echo("Unable to determine cause of previous reboot\n")
+@click.argument('history', required=False, type=click.Choice(["history"]))
+def reboot_cause(history):
+    """Show cause of reboot"""
+    REBOOT_CAUSE_DIR = "/host/reboot-cause/"
+    PREVIOUS_REBOOT_CAUSE_FILE = REBOOT_CAUSE_DIR + "last-reboot-cause.json"
```
jleveque, line 1810: Prefer sticking with "previous-" rather than "last-", as it's a bit more specific, whereas "last" can have a few meanings.

jleveque (changes requested):

show/aliases.ini
```diff
@@ -3,3 +3,4 @@ running-configuration=runningconfiguration
 running-config=runningconfiguration
 startup-configuration=startupconfiguration
 startup-config=startupconfiguration
+reboot-cause=reboot cause
```
jleveque, line 6: Is this tested? I haven't attempted to alias a command to a command + subcommand before.

show/main.py
```diff
@@ -1803,20 +1803,74 @@ def mmu():
...
-    # went down for reboot. This file should always be created at boot,
-    # but check first just in case it's not present.
-    if not os.path.isfile(PREVIOUS_REBOOT_CAUSE_FILE):
-        click.echo("Unable to determine cause of previous reboot\n")
+@click.argument('history', required=False, type=click.Choice(["history"]))
+def reboot_cause(history):
+    """Show cause of reboot"""
+    REBOOT_CAUSE_DIR = "/host/reboot-cause/"
+    PREVIOUS_REBOOT_CAUSE_FILE = REBOOT_CAUSE_DIR + "previous-reboot-cause.json"
+    USER_ISSUE_REBOOT_CAUSE_REGEX ="User issued \'{}\' command [User: {}, Time: {}]"
```
jleveque, line 1811: s/USER_ISSUE_REBOOT_CAUSE_REGEX/USER_ISSUED_REBOOT_CAUSE_REGEX/

jleveque (changes requested):

show/aliases.ini
sujinmkang: good catch! need to be removed.

jleveque: So it didn't work?

show/main.py
```diff
@@ -1803,20 +1803,74 @@ def mmu():
...
-    # the contents of the 'reboot cause' file as it was left when the device
-    # went down for reboot. This file should always be created at boot,
-    # but check first just in case it's not present.
-    if not os.path.isfile(PREVIOUS_REBOOT_CAUSE_FILE):
-        click.echo("Unable to determine cause of previous reboot\n")
+@click.argument('history', required=False, type=click.Choice(["history"]))
+def reboot_cause(history):
+    """Show cause of reboot"""
+    REBOOT_CAUSE_DIR = "/host/reboot-cause/"
+    PREVIOUS_REBOOT_CAUSE_FILE = REBOOT_CAUSE_DIR + "previous-reboot-cause.json"
```
jleveque, line 1810: Suggest using `os.path.join()`

jleveque (changes requested):

show/aliases.ini
sujinmkang: no need and it didn't work.

jleveque: No need, but it could have cleaned up the command syntax a bit. Good to know it doesn't work for the future. If we ever need this type of aliasing, we will have to do more work.

tests/show_reboot_cause_test.py
```diff
@@ -0,0 +1,58 @@
...
+test_path = os.path.dirname(os.path.abspath(__file__))
+modules_path = os.path.dirname(test_path)
+sys.path.insert(0, modules_path)
+
+import show.main as show
+
+
+TEST_REBOOT_CAUSE = "warm-reboot"
+TEST_USER = "admin"
+TEST_REBOOT_TIME = "Fri Oct  9 04:51:47 UTC 2020"
```
jleveque, line 17: Where are these used?

show/main.py
```diff
@@ -1803,20 +1803,74 @@ def mmu():
...
-    PREVIOUS_REBOOT_CAUSE_FILE = "/host/reboot-cause/previous-reboot-cause.txt"
-
-    # At boot time, PREVIOUS_REBOOT_CAUSE_FILE is generated based on
-    # the contents of the 'reboot cause' file as it was left when the device
-    # went down for reboot. This file should always be created at boot,
-    # but check first just in case it's not present.
-    if not os.path.isfile(PREVIOUS_REBOOT_CAUSE_FILE):
-        click.echo("Unable to determine cause of previous reboot\n")
+@click.argument('history', required=False, type=click.Choice(["history"]))
+def reboot_cause(history):
```
jleveque, line 1807: It seems like "reboot_cause" should now become a group and "history" should be a subcommand of that group. The "reboot_cause" group should be defined with `invoke-without-command=True` and the default implementation should be to read and display the previous reboot cause.

jleveque (changes requested):

show/main.py
```diff
@@ -1821,56 +1821,55 @@ def read_last_reboot_cause():
...
-                r.append("")
-            else:
-                r.append(entry['cause'])
+        click.echo(last_reboot_cause)
 
-            if 'time' not in entry:
-                r.append("")
-            else:
-                r.append(entry['time'])
+# 'history' subcommand ("show reboot-cause hostory")
```
jleveque, line 1827: Typo: s/hostory/history/

jleveque (changes requested):

show/main.py
```diff
@@ -1800,23 +1800,76 @@ def mmu():
...
+            return text[len(prefix):]
+        return text
+
+    table = []
+    for tk in table_keys:
+        entry = db.get_all(db.STATE_DB, tk)
+        r = []
+        r.append(remove_prefix(tk, prefix))
+        if 'cause' not in entry:
+            r.append("")
```
jleveque, line 1851: Should we append the empty string, or "N/A"? Same question for all below.

show/main.py
```diff
@@ -1800,23 +1800,76 @@ def mmu():
...
+        else:
+            r.append(entry['user'])
+
+        if 'comment' not in entry:
+            r.append("")
+        else:
+            r.append(entry['comment'])
+        table.append(r)
+
+    header = ['name', 'cause', 'time', 'user', 'comment']
```
jleveque, line 1871: Suggest capitalizing the first letter of each header.

jleveque (changes requested):

show/main.py
sujinmkang: it's already add "N/A" before storing into state-db. so it should return something valid otherwise, there must be something wrong while it's saved into state-db.

jleveque: So this is just a failsafe in case something went wrong saving to State DB?

sujinmkang: Yes.

jleveque: If you feel it's necessary, then maybe we can just shorten the code by doing the following:

```python
r.append(entry['cause'] if 'cause' in entry else "")
r.append(entry['time'] if 'time' in entry else "")
r.append(entry['user'] if 'user' in entry else "")
r.append(entry['comment'] if 'comment' in entry else "")
```

doc/Command-Reference.md
```diff
@@ -541,6 +541,26 @@ This command displays the cause of the previous reboot
...
+
+- Usage:
+  ```
+  show reboot-cause history
+  ```
+
+- Example:
+  ```
+  admin@sonic:~$ show reboot-cause history
+  name                 cause        time                          user    comment
```
jleveque, line 556: Update headers to reflect change to capitalize the first letter of each

jleveque (changes requested):

doc/Command-Reference.md
sujinmkang: I was working on it already :)

jleveque (approved):

sujinmkang:
@lguohan can you please review and approve this PR? The change request has been addressed.

jleveque (changes requested):

sujinmkang:
Please review the new PR : https://github.com/Azure/sonic-utilities/pull/1210. Closing this PR since my rebase attempt messed up the PR.

PR #1158, opened 2020-10-10
show tech with platform dump option

sujinmkang:
**- What I did**
Add an option to force the hardware register dump with -f for `show techsupport --allow-process-stop`
**- How I did it**
Add an option '--allow-process-stop' to `show techsupport` cli command and '-a' to generate_dump script
**- How to verify it**
Perform `show techsupport --allow-process-stop` to collect extra hardware registers information which may require system interruption.
N/A
N/A


show/main.py
```diff
@@ -1227,10 +1227,21 @@ def users(verbose):
...
     cmd = "sudo generate_dump -v"
+    if force:
+        if yes:
+           cmd += " -f" 
+        else:
+            yes = raw_input('Proceed with any system interruption? (y/n) ')
+            if yes.lower() == 'y':
+                cmd += " -f"
+            if yes.lower() == 'n':
+                exit("Please do \'show techsupport\' without \'--force\'")
```
jleveque, line 1244: What happens if someone enters a character other than 'y' or 'n'?

sujinmkang: good point!

show/main.py
```diff
@@ -1240,8 +1240,11 @@ def techsupport(since, verbose, force, yes):
             yes = raw_input('Proceed with any system interruption? (y/n) ')
             if yes.lower() == 'y':
                 cmd += " -f"
-            if yes.lower() == 'n':
+            elif yes.lower() == 'n':
                 exit("Please do \'show techsupport\' without \'--force\'")
+            else:
+                exit("Please try again")
```
jleveque: Why not use `click.confirm()`? https://click.palletsprojects.com/en/7.x/api/?highlight=exit#click.confirm

sujinmkang: Good information. I didn't know the `click.confirm()`.  I will change to use it.

show/main.py
```diff
@@ -1227,10 +1227,19 @@ def users(verbose):
...
 @click.option('--verbose', is_flag=True, help="Enable verbose output")
-def techsupport(since, verbose):
+def techsupport(since, verbose, force, yes):
     """Gather information for troubleshooting"""
     cmd = "sudo generate_dump -v"
+    if force:
+        if yes:
+            cmd += " -f"
+        else:
+            if click.confirm('Proceed with any system interruption?', abort=True):
```
jleveque, line 1240: With `abort=True`, this does not need to be an `if` statement. If the user declines, the command will abort. If you want to print a message and exit, you can remove `abort=True` and add an `else` statement.

jleveque (changes requested):

show/main.py
sujinmkang: Aborting when the user declines, is ok but we can add a message before exit for this case.

show/main.py
```diff
@@ -1237,8 +1237,12 @@ def techsupport(since, verbose, force, yes):
         if yes:
             cmd += " -f"
         else:
-            if click.confirm('Proceed with any system interruption?', abort=True):
+            if click.confirm('Proceed with any system interruption for HW register dump?'):
```
jleveque, line 1240: I believe you should replace "any" with "potential" to make it clearer. Am I understanding the intention correctly?

show/main.py
```diff
@@ -1227,10 +1227,23 @@ def users(verbose):
 
 @cli.command()
 @click.option('--since', required=False, help="Collect logs and core files since given date")
+@click.option('--force', is_flag=True, help="Enable hardware register dump even with system interruption")
+@click.option('--yes', is_flag=True, help="Enable force generating dump option")
```
jleveque, line 1231: Suggest changing help message to: `Assume "yes" as answer to all prompts and
    run non-interactively`

jleveque (changes requested):

show/main.py
```diff
@@ -1227,10 +1227,23 @@ def users(verbose):
...
+        if yes:
+            cmd += " -f"
+        else:
+            if click.confirm('WARNING: Proceed with any potential system interruption for HW register dump?'):
+                cmd += " -f"
+            else:
+                click.echo("Please use --force only for disruptive HW register dump!")
+                system.exit(1)
+
+
```
jleveque, line 1246: Remove one blank line here.

jleveque (changes requested):

sujinmkang:
@jleveque Can you review this change again?

show/main.py
```diff
@@ -1227,10 +1227,22 @@ def users(verbose):
 
 @cli.command()
 @click.option('--since', required=False, help="Collect logs and core files since given date")
+@click.option('--force', is_flag=True, help="Enable hardware register dump even with system interruption")
```
jleveque, line 1230: I wonder if we should choose a different name other than "force" for this. Because it doesn't really force anything; it enables disruptive HW register dump. Any other suggestions?

sujinmkang: I think it will be better to use more specific option like --stop_pmon or --stop_process. Let me check with Dell what exact action needs to be done for the disruptive hw register dump case.

sujinmkang: @jleveque I think "--allow_process_stop" might be better for this kind of dump case and since it's obvious for user to know the intention of this flag, we can also remove the "--yes" option from the option check. What do you think?

jleveque: I think `--allow-process-stop` is fine. When you say "remove the `--yes` option", do you mean remove the prompt entirely?

sujinmkang: correct!

jleveque: Sure. I think that's fine.

scripts/generate_dump
```diff
@@ -827,6 +828,8 @@ OPTIONS
         passed to them
     -n
         Noop mode. Don't actually create anything, just echo what would happen
+    -f
+        Force mode. This allows any system interruption to correct any platform HW registers status.
```
jleveque, line 832: Please update this to match the new option name

show/main.py
```diff
@@ -1228,11 +1228,16 @@ def users(verbose):
 @cli.command()
 @click.option('--since', required=False, help="Collect logs and core files since given date")
 @click.option('--verbose', is_flag=True, help="Enable verbose output")
-def techsupport(since, verbose):
+@click.option('--allow-process-stop', is_flag=True, help="Allow hardware register dump even with system interruption")
```
jleveque, line 1231: Suggest changing help message to:

```
Also perform hardware register dump which may cause system interruption
```

jleveque (changes requested):

show/main.py
jleveque: Can you change the message to what I provided above? You changed some, but not all.

scripts/generate_dump
```diff
@@ -827,6 +828,8 @@ OPTIONS
         passed to them
     -n
         Noop mode. Don't actually create anything, just echo what would happen
+    -a
+        Allow any process stop. This allows any system interruption to collect any platform HW register status.
```
jleveque, line 832: Too many "any"s. Please reword as follows:

```
Allow any process stop. This allows for collection of platform HW register status, which may require potential system interruption.
```

jleveque (changes requested):

show/main.py
jleveque: @sujinmkang: Can you please update this message to the message I provided above? (Note that I also just edited the message again).

sujinmkang: hardware registers dump can be done even without this option. I will just update the message with "Dump additional data which may require system interruption"

scripts/generate_dump
```diff
@@ -868,6 +872,9 @@ while getopts ":xnvhzs:" opt; do
         z)
             DO_COMPRESS=false
             ;;
+        a)
+            ALLOW_PROCESS_STOP="-a"
```
jleveque, line 876: Should this read `ALLOW_PROCESS_STOP="-f"`? Because this is the flag which gets passed to `/usr/bin/hw-management-generate-dump.sh` on line 767. I assume `/usr/bin/hw-management-generate-dump.sh` is still expecting `-f`.

sujinmkang: This is not yet finalized. we can change it to any letter at this point. Once we decide, we can ask vendors to follow.

jleveque (approved):

jleveque:
Retest this please

PR #1180, opened 2020-10-17
Fixes the issue with show interface counters and for pfc and queue counters.

abdosi:
Why I did:
Fix the issues as define https://github.com/Azure/sonic-buildimage/issues/5655

How I did:
For Issue 1 : Updated to pass port_speed in ns_util

For Issue 2: Updated ns_diff to return new_value if valid and old_value is not valid.

How I Verify:

For Issue 2:
Manual Verification
After doing sonic-clear during init 
show interface counters were fine.


jleveque (dismissed):

abdosi:
retest this please

abdosi:
@jleveque Can you review again. I had to update for PR for two more changes:

a) Fix the Unit-test failure
b) Fix the issue#2 for queuestat and pfcstat as they were using local definition of ns_diff instead from netstat library.

jleveque (approved):

Issue #1068, opened 2020-08-24
"show interfaces neighbor expected <interface>" does not list details for all interfaces in a port-channel 


ArunSaravananBalachandran:
**Description**

In a port-channel with multiple interfaces to a neighbor, "show interface neighbor expected <interface>" command lists neighbor information output for only one interface.

**Steps to reproduce the issue**
1. Load configuration with port-channel interfaces with more than one member.
2. Execute "show interfaces neighbor expected" command.

**Describe the results you received**

```

root@sonic:/home/admin# show interfaces neighbor expected Ethernet0
No neighbor information available for interface Ethernet0
root@sonic:/home/admin# show interfaces neighbor expected Ethernet1
LocalPort    Neighbor    NeighborPort    NeighborLoopback    NeighborMgmt    NeighborType
-----------  ----------  --------------  ------------------  --------------  --------------
Ethernet1    ARISTA01T2  Ethernet2       None                100.94.110.214  SpineRouter
root@sonic:/home/admin#
root@sonic:/home/admin# show interfaces neighbor expected
LocalPort    Neighbor    NeighborPort    NeighborLoopback    NeighborMgmt    NeighborType
-----------  ----------  --------------  ------------------  --------------  --------------
Ethernet34   ARISTA01T0  Ethernet1       None                100.94.110.218  ToRRouter
Ethernet1    ARISTA01T2  Ethernet2       None                100.94.110.214  SpineRouter
Ethernet36   ARISTA02T0  Ethernet1       None                100.94.110.219  ToRRouter
Ethernet37   ARISTA03T0  Ethernet1       None                100.94.110.220  ToRRouter
Ethernet5    ARISTA03T2  Ethernet2       None                100.94.110.215  SpineRouter
Ethernet38   ARISTA04T0  Ethernet1       None                100.94.110.221  ToRRouter
Ethernet39   ARISTA05T0  Ethernet1       None                100.94.110.222  ToRRouter
Ethernet17   ARISTA05T2  Ethernet2       None                100.94.110.216  SpineRouter
Ethernet42   ARISTA06T0  Ethernet1       None                100.94.110.223  ToRRouter
Ethernet44   ARISTA07T0  Ethernet1       None                100.94.110.224  ToRRouter
Ethernet21   ARISTA07T2  Ethernet2       None                100.94.110.217  SpineRouter
Ethernet45   ARISTA08T0  Ethernet1       None                100.94.110.225  ToRRouter
[...]

PR #1181, opened 2020-10-19
Fix the issue of the show interfaces neighbor expected

chiourung:
**- What I did**
      Fix the issue #1068 

**- How I did it**
      Use interface as the keys to find the neighbors expected.

**- How to verify it**
    1. Load configuration with port-channel interfaces with more than one member.
    2. Execute "show interfaces neighbor expected" command.


chiourung:
This has been fixed in https://github.com/Azure/sonic-utilities/commit/b6af9f48e29852985137a57ff5396591af7f107c

PR #1202, opened 2020-10-30
[show] Moving some py imports inside specific cmds

Sabareesh-Kumar-Anandan:
Signed-off-by: Sabareesh Kumar Anandan <sanandan@marvell.com>

**- What I did**
Some py imports in show/main.py script are taking more time on arm cpu.
modules - pkg_resources, netaddr are taking around 500ms causing delay for all show cmds.
These modules are used by only specific cmds. So, I have included these imports only for the commands that are using them.

**- How I did it**

**- How to verify it**
Test all show commands. Execution time of show commands is better than before.

### DIFF
--- a/show/main.py
+++ b/show/main.py
@@ -1,5 +1,4 @@
 import json
-#import netaddr
 import os
 import subprocess
 import sys
@@ -55,7 +54,7 @@ def get_routing_stack():
 
 
 # Global Routing-Stack variable
-routing_stack = "frr"
+routing_stack = get_routing_stack()
 
 # Read given JSON file
 def readJsonFile(fileName):
--- a/show/main.py
+++ b/show/main.py
@@ -1,5 +1,5 @@
 import json
-import netaddr
+#import netaddr
 import os
 import subprocess
 import sys
@@ -9,7 +9,6 @@ import netifaces
 import utilities_common.cli as clicommon
 import utilities_common.multi_asic as multi_asic_util
 from natsort import natsorted
-from pkg_resources import parse_version
 from sonic_py_common import device_info, multi_asic
 from swsssdk import ConfigDBConnector
 from swsscommon.swsscommon import SonicV2Connector
@@ -56,7 +55,7 @@ def get_routing_stack():
 
 
 # Global Routing-Stack variable
-routing_stack = get_routing_stack()
+routing_stack = "frr"
 
 # Read given JSON file
 def readJsonFile(fileName):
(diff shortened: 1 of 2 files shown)


show/main.py
```diff
@@ -56,7 +56,7 @@ def get_routing_stack():
 
 
 # Global Routing-Stack variable
-routing_stack = get_routing_stack()
+routing_stack = "frr"
```
lguohan, line 59: need to think about this.

show/main.py
```diff
@@ -1,20 +1,20 @@
...
+#import netaddr
 import os
 import subprocess
 import sys
 
 import click
 from natsort import natsorted
 import netifaces
-from pkg_resources import parse_version
+#from pkg_resources import parse_version
```
jleveque, line 12: Why comment out?

show/main.py
```diff
@@ -1,20 +1,20 @@
...
 
 import feature
 import interfaces
 import kube
-import mlnx
+#import mlnx
```
jleveque, line 17: I believe commenting this out completely will break Mellanox platforms.

show/main.py
Sabareesh-Kumar-Anandan: This module import is taking more time(~300ms) on arm cpu. As this is required only for "show ntp" cmd, I have added it inside ntp command. see below

show/main.py
Sabareesh-Kumar-Anandan: this is included after checking for mellanox platform. see below (line 1008)

show/main.py
lguohan: this reduce the function, can you exclude this in this pr? also it does not fit in the pr title.

utilities_common/cli.py
```diff
@@ -5,7 +5,7 @@
 
 import click
 import json
-import netaddr
+#import netaddr
```
lguohan, line 8: how much delay is this one? btw, we moved to python3, is this still an issue?

show/main.py
```diff
@@ -1,20 +1,20 @@
 #! /usr/bin/python -u
 
 import json
-import netaddr
+#import netaddr
```
lguohan, line 4: do not comment, please remove.

utilities_common/cli.py
Sabareesh-Kumar-Anandan: takes around 200ms

show/main.py
Sabareesh-Kumar-Anandan: removed it. will think of better way of doing this

show/main.py
Sabareesh-Kumar-Anandan: done

Sabareesh-Kumar-Anandan:
Hi @lguohan, 
I see 3 tests have failed. I don't think failure is caused by this change. Is it possible to retrigger the tests?

lguohan:
retest this please

lguohan (approved):

lguohan:
retest this please

PR #1205, opened 2020-11-02
[MultiDB] add MultiDB warmboot support - backing up database

dzhangalibaba:
* lua script to backup all database into one database and create rdb file - following earlier multiDB warmboot design at https://github.com/Azure/SONiC/blob/master/doc/database/multi_database_instances.md
* copy this rdb file as before to WARM_DIR
* restoring database part is in another PR at sonic-buildimage  https://github.com/Azure/sonic-buildimage/pull/5773, they depend on each other

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -247,10 +247,14 @@ function backup_database()
             end
         end
     " 0 > /dev/null
-    sonic-db-cli SAVE > /dev/null
-    #TODO : need a script to copy all rdb files if there is multiple db instances config
-    docker cp database:/var/lib/redis/$REDIS_FILE $WARM_DIR
-    docker exec -i database rm /var/lib/redis/$REDIS_FILE
+
+    # move all db data into instances where APPL_DB locates
+    migrate_warmboot_database APPL_DB
+
+    inst=`/usr/bin/python -c "import swsssdk; print(swsssdk.SonicDBConfig.get_instancename('APPL_DB'))"`
+    # Dump redis content to a file 'dump.rdb' in warmboot directory
+    docker cp database:/var/lib/$inst/$REDIS_FILE $WARM_DIR
+    docker exec -i database rm /var/lib/$inst/$REDIS_FILE
 }
 
 function setup_control_plane_assistant()
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -248,13 +248,12 @@ function backup_database()
         end
     " 0 > /dev/null
 
-    # move all db data into instances where APPL_DB locates
-    migrate_warmboot_database APPL_DB
+    # move all db data into the instance where APPL_DB locates
+    target_db_inst=`centralize_database APPL_DB`
 
-    inst=`/usr/bin/python -c "import swsssdk; print(swsssdk.SonicDBConfig.get_instancename('APPL_DB'))"`
     # Dump redis content to a file 'dump.rdb' in warmboot directory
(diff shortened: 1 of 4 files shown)


dzhangalibaba:
@qiluo-msft

scripts/migrate_warmboot_database
```diff
@@ -0,0 +1,61 @@
...
+                                     formatter_class=argparse.RawTextHelpFormatter,
+                                     epilog=
+"""
+Example : migrate_warmboot_database APPL_DB
+""")
+    parser.add_argument('target_db', type=str, help='move all db data into the instance where target db locates')
+    args = parser.parse_args()
+
+    if args.target_db:
+            try:
```
qiluo-msft, line 50: >             [](start = 0, length = 12)

too much indentation.

scripts/fast-reboot
```diff
@@ -247,10 +247,14 @@ function backup_database()
             end
         end
     " 0 > /dev/null
-    sonic-db-cli SAVE > /dev/null
-    #TODO : need a script to copy all rdb files if there is multiple db instances config
-    docker cp database:/var/lib/redis/$REDIS_FILE $WARM_DIR
-    docker exec -i database rm /var/lib/redis/$REDIS_FILE
+
+    # move all db data into instances where APPL_DB locates
+    migrate_warmboot_database APPL_DB
```
qiluo-msft, line 252: >migrate_warmboot_database [](start = 4, length = 25)

The name is misleading. It is not directly related to warmboot.
Suggestion `centralize_database`

scripts/migrate_warmboot_database
```diff
@@ -0,0 +1,61 @@
...
+def migrate(target_dbname):
+    target_dbport = swsssdk.SonicDBConfig.get_port(target_dbname)
+    target_dbhost = swsssdk.SonicDBConfig.get_hostname(target_dbname)
+
+    dblists = swsssdk.SonicDBConfig.get_dblist()
+    for dbname in dblists:
+        dbport = swsssdk.SonicDBConfig.get_port(dbname)
+        # if the db is on the same instance, no need to migrate
+        if dbport == target_dbport:
+            continue
```
qiluo-msft, line 17: Is it enough to determine "the same instance" ?

scripts/fast-reboot
```diff
@@ -247,10 +247,14 @@ function backup_database()
...
+
+    inst=`/usr/bin/python -c "import swsssdk; print(swsssdk.SonicDBConfig.get_instancename('APPL_DB'))"`
```
qiluo-msft, line 254: >inst [](start = 4, length = 4)

Instead of using a new process to locate APPL_DB, you may just print it to stdout in above command.

qiluo-msft (changes requested):
As comments

scripts/migrate_warmboot_database
dzhangalibaba: fixed

scripts/fast-reboot
dzhangalibaba: updated

scripts/migrate_warmboot_database
dzhangalibaba: good catch. hostname plus port  is enough to determine the "unique instance" . Updated

scripts/fast-reboot
dzhangalibaba: updated

qiluo-msft (approved):

yxieca (approved):
Can you point me to the code where the centralized DB are redistributed?

PR #1215, opened 2020-11-04
[route_check]Add an option for route_check to control whether log is writen to syslog

bingwang-ms:
Fix https://github.com/Azure/sonic-buildimage/issues/5804
**- What I did**
By default, route_check.py will report an ERROR in syslog if route mismatch is detected, which is out of the control of monit config file and cause LogAnalyzer to report test error. This commit add an option (-s) to control this behavior. And log is not writen to syslog in default.

**- How I did it**
Add an option (-s) for ```route_checker.py```.

**- How to verify it**
The update is verified on Arista-7260.
1. Add a static route whose nexthop is not reachable.
```
ip route add 1.1.1.1 via 192.168.1.101
```
2. Run ```route_check.py```, and error msg is only printed on stdout. Nothing is writen to syslog
3. Run ```route_check.py -s```. and error msg is writen to both stdout and syslog
4. Wait for 15 minutes, and confirm that monit will report the error
```
Nov  4 09:30:36.917367 str-7260cx3-acs-2 ERR monit[631]: 'routeCheck' status failed (255) -- results: { {#012    "missed_ROUTE_TABLE_routes": [#012        "1.1.1.1/32"#012    ]#012} }#012 Failed. Look at reported mismatches above
```

Output is not changed.

Output is not changed.

### DIFF
--- a/scripts/route_check.py
+++ b/scripts/route_check.py
@@ -30,11 +30,13 @@ class Level(Enum):
 
 
 report_level = syslog.LOG_ERR
+write_to_syslog = False
 
-
-def set_level(lvl):
+def set_level(lvl, log_to_syslog):
     global report_level
+    global write_to_syslog
 
+    write_to_syslog = log_to_syslog
     if (lvl == Level.INFO):
         report_level = syslog.LOG_INFO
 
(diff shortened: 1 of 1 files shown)


renukamanavalan (approved):
Thank you!

PR #1226, opened 2020-11-09
Remove SKU create pytest output directory before execution of the script

madhanmellanox:
**- What I did**
I removed the output directory of the SKU create pytest before execution of the script, otherwise the pytest execution will fail when run for the second time.

**- How I did it**
I checked for existence of the output directory and if it exists, I remove it.

**- How to verify it**
Run SKU creator pytest for the second time and ensure it runs successfully.

None

None

### DIFF
--- a/tests/sku_create_test.py
+++ b/tests/sku_create_test.py
@@ -8,6 +8,7 @@ test_path = os.path.dirname(os.path.abspath(__file__))
 modules_path = os.path.dirname(test_path)
 scripts_path = os.path.join(modules_path, "scripts")
 input_path = os.path.join(modules_path, "tests/sku_create_input")
+output_dir_path = os.path.join(modules_path, "tests/sku_create_input/Mellanox-SN2700-D48C8_NEW")
 sku_def_file = os.path.join(input_path, "Mellanox-SN2700-D48C8.xml")
 sku_create_script = os.path.join(scripts_path, "sonic_sku_create.py")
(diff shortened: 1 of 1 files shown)


madhanmellanox:
@jleveque can you please review this? This PR is to avoid the pytest failures seen in sonic-utilities submodule update.

tests/sku_create_test.py
```diff
@@ -49,6 +50,9 @@ def are_file_contents_same(self,fname1,fname2):
         return True
     
     def test_no_param(self):
+        if (os.path.exists(output_dir_path)):
+            shutil.rmtree(output_dir_path)
```
jleveque, line 54: Do you need to `import shutil`?

madhanmellanox: Yes, I have added it in the next commit.

tests/sku_create_test.py
```diff
@@ -3,6 +3,7 @@
 import pytest
 import subprocess
 import re
+import shutil
```
jleveque, line 6: Can you please organize these per [PEP8 standards](https://www.python.org/dev/peps/pep-0008/#imports), and alphabetize in each group? E.g.,
```python
import os
import re
import shutil
import subprocess
import sys

import pytest
```

jleveque (changes requested):

tests/sku_create_test.py
madhanmellanox: Addressed in next commit.

jleveque (approved):

PR #1228, opened 2020-11-11
[load_minigraph]: Add support to load default_config.json file

mlorrillere:
This change add support for a per-device/per-npu default_config.json
file that contains platform-specific or hwsku-specific configuration to
be used when generating config_db*.json

An example of such file in a VoQ chassis could be:
```json
{
    "DEVICE_METADATA": {
        "localhost": {
            "asic_id": "06:00.0"
        }
    }
}
```

**- What I did**

Added support for a default_config.json file present in platform directory for each NPU and for the host. This file will be loaded when `config load_minigraph` is called.

**- Why I did it**

Some VoQ devices require per-platform/per-NPU configurations to be present in CONFIG_DB. `default_config.json` provides a way for a vendor to give the required initial configuration of a device.

**- How to verify it**

Verified manually on a single and multi-NPU platform with and without a `default_config.json` present in the platform directory.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -242,6 +242,22 @@ def _get_device_type():
 
     return device_type
 
+def _get_hwsku(config_db):
+    """
+    Get hwsku from CONFIG_DB
+
+    TODO: move to sonic-py-common
+    """
+
+    metadata = config_db.get_table('DEVICE_METADATA')
+    if metadata and 'localhost' in metadata and 'hwsku' in metadata['localhost']:
+       return metadata['localhost']['hwsku']
+
+    click.echo("Could not get the hwsku from CONFIG_DB, setting hwsku to Unknown")
+    hwsku = 'Unknown'
+
+    return hwsku
+
 def interface_alias_to_name(config_db, interface_alias):
     """Return default interface name if alias name is given as argument
     """
@@ -1284,11 +1300,13 @@ def load_minigraph(db, no_service_restart):
         log.log_info("'load_minigraph' stopping services...")
         _stop_services()
 
+    platform = device_info.get_platform()
+
     # For Single Asic platform the namespace list has the empty string
     # for mulit Asic platform the empty string to generate the config
     # for host
     namespace_list = [DEFAULT_NAMESPACE]
-    num_npus = multi_asic.get_num_asics()
+    num_npus = device_info.get_num_npus()
     if num_npus > 1:
         namespace_list += multi_asic.get_namespaces_from_linux()
 
@@ -1309,6 +1327,25 @@ def load_minigraph(db, no_service_restart):
         else:
             command = "{} -H -m --write-to-db {}".format(SONIC_CFGGEN_PATH, cfggen_namespace_option)
         clicommon.run_command(command, display_cmd=True)
+
+        if namespace is DEFAULT_NAMESPACE:
+            npu_id = ''
+        else:
+            npu_id = device_info.get_npu_id_from_name(namespace)
+
+        # load default_config.json file from platform directory
+        default_config_file = os.path.join('/usr/share/sonic/device/', platform, npu_id, 'default_config.json')
+        if os.path.isfile(default_config_file):
+            command = "{} -j {} {} --write-to-db".format(SONIC_CFGGEN_PATH, default_config_file, cfggen_namespace_option)
+            clicommon.run_command(command, display_cmd=True)
+
+        hwsku = _get_hwsku(config_db)
+        # load default_config.json file from hwsku directory
+        default_config_file = os.path.join('/usr/share/sonic/device/', platform, hwsku, npu_id, 'default_config.json')
+        if os.path.isfile(default_config_file):
+            command = "{} -j {} {} --write-to-db".format(SONIC_CFGGEN_PATH, default_config_file, cfggen_namespace_option)
+            clicommon.run_command(command, display_cmd=True)
+
         client.set(config_db.INIT_INDICATOR, 1)
 
     # get the device type
--- a/tests/config_test.py
+++ b/tests/config_test.py
@@ -18,15 +18,31 @@ from generic_config_updater.generic_updater import ConfigFormat
 
 import config.main as config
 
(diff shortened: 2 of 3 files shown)


jleveque:
Retest this please

anshuv-mfst:
Veda (Nokia) , @arlakshm - could you please review, thanks.

config/main.py
```diff
@@ -294,6 +294,22 @@ def _get_device_type():
 
     return device_type
 
+def _get_hwsku(config_db):
```
arlakshm, line 297: We can remove this function here. Same function available in sonic_py_common  [get_hwsku](https://github.com/Azure/sonic-buildimage/blob/85a631442418582d1a2f77bf0ca1b7b5addee12a/src/sonic-py-common/sonic_py_common/device_info.py#L110)

config/main.py
```diff
@@ -1247,6 +1265,25 @@ def load_minigraph(db, no_service_restart):
...
             command = "{} -H -m --write-to-db {}".format(SONIC_CFGGEN_PATH, cfggen_namespace_option)
         clicommon.run_command(command, display_cmd=True)
+
+        if namespace is DEFAULT_NAMESPACE:
+            npu_id = ''
+        else:
+            npu_id = device_info.get_npu_id_from_name(namespace)
+
+        # load default_config.json file from platform directory
+        default_config_file = os.path.join('/usr/share/sonic/device/', platform, npu_id, 'default_config.json')
```
arlakshm, line 1275: can you give an example what  is present in 'default_config.json'

arlakshm (changes requested):
Please add unit test for this

config/main.py
mlorrillere: `get_hwsku` in `sonic_py_common` does not support namespaces, it will always connect to the host redis instance.

This can be revisited when `sonic_py_common` provides support for namespaces.

config/main.py
mlorrillere: I added the following example in the description:
```json
{
    "DEVICE_METADATA": {
        "localhost": {
            "instance_name": "ASIC0",
            "chassis_db_address" : "240.127.1.1",
            "start_chassis_db" : "1",
            "asic_id": "06:00.0"
        }
    }
}
```

config/main.py
arlakshm: hwsku is for the whole device, wouldn't getting the hwsku form the host redis work will it change for each namespace/asic ?

mlorrillere: This would be the same hwsku for all namespace/asic. However we can't get it before minigraph.xml is parsed and CONFIG_DB is populated, so calling `get_hwsku` from `sonic_py_common` beforehand is not possible since CONFIG_DB will be empty.

arlakshm: if config_db is empty then `_get_hwsku`  won't return the correct hwsku  also. I am trying to understand what the difference between `_get_hwsku`  define here and `get_hwsku ` in `sonic_py_common `

mlorrillere: `_get_hwsku` is called after minigraph.xml is parsed and CONFIG_DB is populated for the corresponding namespace CONFIG_DB initialized at line 1274, _get_hwsku is called at line 1287).

The difference with `get_hwsku` from `sonic_py_common` is that `_get_hwsku` here reuse the existing connection to CONFIG_DB. If we cann `get_hwsku` from `sonic_py_common` instead it will get stuck because we can't open multiple connections the same time.

mlorrillere:
> Please add unit test for this

I've updated `config_test.py` to get coverage for `default_config.json`. I also added similar test for `init_cfg.json` since coverage was missing.

sanmalho-git:
@vganesan-nokia - please review.

vganesan-nokia (changes requested):

config/main.py
mlorrillere: > As per PR [Azure/sonic-buildimage#5991](https://github.com/Azure/sonic-buildimage/pull/5991), for VOQ chassis systems the "asic_id" and "instance_name" are configured via minigraph.xml. Will these values from default_config.json be preferred over what is configured through minigraph?

Both are still valid. If the platform has configuration in `default_config.json` it has to be static. If it is present in both files then `default_config.json` will override the value configured via `minigraph.xml`, which is fine.

The example above is probably incorrect once VoQ config using minigraph.xml is complete ([PR5991](https://github.com/Azure/sonic-buildimage/pull/5991)), however `asic_id` is still required in this file as it is platform-dependent.

mlorrillere: > For VOQ chassis systems, we get "start_chassis_db" and "chassis_db_address" from chassisdb.conf from device specific directory (for example: https://github.com/Azure/sonic-buildimage/blob/master/device/arista/x86_64-arista_7800_sup/chassisdb.conf). These are needed before the database service is started in supervisor card. Having these in config_db is of no use. Also, if these are not used in multi-asic we should not have these in default_config.json.

Right. Actually in the above example only `asic_id` is required as it is tied to each asic physically on the hardware. Other configurations can easily be set using minigraph.xml

mlorrillere:
Hi @arlakshm and @vganesan-nokia ,

Can you approve this review if there is no further comments?

vganesan-nokia (approved):

arlakshm (approved):

vganesan-nokia (approved):

arlakshm:
@mlorrillere, @vganesan-nokia This PR https://github.com/Azure/sonic-buildimage/pull/7513 adds support to populate the `DEVICE_METADATA.asic_id` from the Pci_id present in asic.conf. Do we still need `default_config.json`

mlorrillere:
> @mlorrillere, @vganesan-nokia This PR [Azure/sonic-buildimage#7513](https://github.com/Azure/sonic-buildimage/pull/7513) adds support to populate the `DEVICE_METADATA.asic_id` from the Pci_id present in asic.conf. Do we still need `default_config.json`

I found out about this PR yesterday. We might not need `default_config.json` after all if `asic.conf` includes the PCI addresses of the ASICs.

PR #1233, opened 2020-11-12
[tests] Do not trim from PATH if we did not append to it; Clean up/fix shebangs in scripts

jleveque:
**- What I did**

Some tests were trimming the last item from system PATH, but they never appended anything to it, thus these tests were trimming valid entries from the tail of the system path, which in turn caused scripts with shebangs like `#!/usr/bin/env python` to fail, because once `/usr/bin` was removed from PATH, `python` could not be found.

**- How I did it**

- Remove lines which trimmed the last entry from system PATH in the teardown function if nothing was appended to PATH in the setup method.
- Remove shebangs from non-executable Python files
- Fix shebangs which had a space in them
- Replace `#!/usr/bin/python` shebangs with the preferred, more portable `#!/usr/bin/env python` in executable scripts

**- How to verify it**

Ensure scripts with `#!/usr/bin/env python` shebangs execute properly in pytests.


neethajohn:
Thank you so much Joe for root causing this and fixing!

neethajohn (dismissed):

jleveque:
You're welcome, @neethajohn! If the tests pass, I think I will update this PR again and change all shebangs in the repo. Not only is the the preferred shebang, if all scripts use it and this issue arises again, we might be able to catch it sooner.

neethajohn (approved):

lguohan (approved):

PR #1238, opened 2020-11-13
Fix mock SonicV2Connector in python3: use decode_responses mode so caller code will be the same as python2

qiluo-msft:
The python2 behavior is not changed.

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/tests/mock_tables/dbconnector.py
+++ b/tests/mock_tables/dbconnector.py
@@ -1,6 +1,7 @@
 # MONKEY PATCH!!!
 import json
 import os
+import sys
 
 import mock
 import mockredis
@@ -9,6 +10,13 @@ import swsssdk
 from sonic_py_common import multi_asic
 from swsssdk import SonicDBConfig, SonicV2Connector
 from swsscommon import swsscommon
+
+
+if sys.version_info >= (3, 0):
+    long = int
+    xrange = range
+    basestring = str
+    from functools import reduce
 
 topo = None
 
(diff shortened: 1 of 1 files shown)


jleveque (approved):
Looks good. Tested with Python 3.

tests/mock_tables/dbconnector.py
```diff
@@ -86,6 +95,7 @@ def __init__(self, *args, **kwargs):
         topo = kwargs.pop('topo')
         namespace = kwargs.pop('namespace')
         db_name = kwargs.pop('db_name')
+        self.decode_responses = kwargs.pop('decode_responses', False) == True
```
tahmed-dev, line 98: is `== True` necessary here?

tahmed-dev (approved):

tests/mock_tables/dbconnector.py
jleveque: I asked this on Qi's earlier similar PR for swsssdk :)

It normalizes `decode_responses` to always be a Bool. E.g., if `kwargs.pop('decode_responses', False)` returns `None` or some other unexpected value.

tahmed-dev: I see, so this is similar to `!!<bool>` trick in C. It appears to me that is is a bug in the downstream rather that some we should work around it??

qiluo-msft: There is no bug in downstream. Here I just program conservatively.

PR #1240, opened 2020-11-15
Add 'requests' as install dependency in setup.py

jleveque:
The `requests` package is used by a couple modules (config/kube.py and scripts/neighbor_advertiser), but it was not specified as an install-time dependency. Now that the package is built as Python 3, some commands are crashing with `ModuleNotFoundError: No module named 'requests'`.


jleveque:
Retest this please

lguohan (approved):

PR #1253, opened 2020-11-19
[dropcounters] Fix clear for non-root users

daall:

**- What I did**
Fixes https://github.com/Azure/sonic-buildimage/issues/5602

**- How I did it**
I followed the same pattern from `portstat` and gave each user their own `dropstat` directory instead of indexing the files by uid.

**- How to verify it**
Validated that non-root user can clear after a `sudo sonic-clear dropcounters`. Also verified that each uid has their own `/tmp` directory. Also updated the unit tests.

### DIFF
--- a/scripts/dropstat
+++ b/scripts/dropstat
@@ -30,7 +30,8 @@ try:
         sys.path.insert(0, modules_path)
         sys.path.insert(0, test_path)
         import mock_tables.dbconnector
-        socket.gethostname = lambda : 'sonic_drops_test'
+        socket.gethostname = lambda: 'sonic_drops_test'
+        os.getuid = lambda: 27
 except KeyError:
     pass
 
@@ -75,8 +76,10 @@ std_port_headers_map = {
 # Standard Switch-Level Headers
 std_switch_description_header = ['DEVICE']
 
-# Bookkeeping Files
-dropstat_dir = '/tmp/dropstat/'
+
+def get_dropstat_dir():
+    dropstat_dir_prefix = '/tmp/dropstat'
+    return "{}-{}/".format(dropstat_dir_prefix, os.getuid())
 
 
 class DropStat(object):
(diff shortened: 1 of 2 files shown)


scripts/dropstat
```diff
@@ -30,7 +30,8 @@ try:
         sys.path.insert(0, modules_path)
         sys.path.insert(0, test_path)
         import mock_tables.dbconnector
-        socket.gethostname = lambda : 'sonic_drops_test'
+        socket.gethostname = lambda: 'sonic_drops_test'
+        os.getuid = lambda: 27
```
lguohan, line 34: what is the 27?

daall: It's a dummy value to ensure that the uid for the tests is deterministic. Should I make this a variable somewhere or is a comment OK?

bingwang-ms: I'm confused. Is it always ```27``` ? What will happen if I ran ```dropstat``` in pytest or from shell directly?

daall: This code will only run during the unit tests. It is always 27 so that the unit tests can always check the same number when verifying the output and always delete the same number when it's cleaning up for future tests.

bingwang-ms: Sorry for late reply. What the uid will be if I call ```show dropcounters counts``` in ansible?  ```show dropcounters counts``` will call ```dropstat -c show``` inside.

daall: `ansible` has root access so I believe it would just be `0`. Unless the unit test flag is set, which should only be when sonic-utilities unit tests are running, `os.getuid` will not be overwritten.

bingwang-ms: What's the uid if I call ```show dropcounters counts``` from shell?

daall: Whatever admin's uid is.

bingwang-ms: Thanks. I got it. The updated code will create an individual dir named ```/tmp/dropstat-{uid}``` for each user instead of using the same dir.  Make sense to me.

bingwang-ms (approved):

abdosi:
@daall Please create for PR for 201911

PR #1255, opened 2020-11-20
Remove unnecessary calls to str.decode() now that the package is Python 3

jleveque:
Now that the package is built as Python 3, all strings are unicode by default and `str.decode()` no longer exists.


jleveque:
Retest this please

lguohan (approved):

qiluo-msft:
LGTM

Issue #1241, opened 2020-11-16
console connect/clear commands' exit code not aligned with consutil commands


Blueve:
**Description**

The console connect/clear commands' exit code are not aligned with consutil commands.
This is an example:
```
admin@sonic:~$ consutil connect 1
admin@sonic:~$ echo $?
1
admin@sonic:~$ connect line 1
admin@sonic:~$ echo $?
0
```

This issue is blocking https://github.com/Azure/sonic-utilities/issues/1241 as well

**Steps to reproduce the issue**
1. Run `consutil connect 1` which the console switch feature is not enabled
1. Run `echo $?`, you will get 1 which means the console switch feature is not enabled
1. Run `connect line 1` which should be same with `consutil connect 1`
1. Run `echo $?`, you will get 0 instead of 1

**Describe the results you received**
The `connect line`  `sonic-clear line` will always return 0 no matter what happens.

**Describe the results you expected**
The exit code should be aligned with consutil.

**Additional information you deem important (e.g. issue happens only occasionally)**

**Output of `show version`**

```
(paste your output here)
```

PR #1256, opened 2020-11-20
[connect][sonic-clear] Align the exit code with consutil for line commands

Blueve:
**- What I did**

Align following commands' exit code with consutil commands since there are aliasing commands:
- `connect line`
- `sonic-clear line`

This is a fix for this issue: https://github.com/Azure/sonic-utilities/issues/1241 
The issue will also cause the reverse ssh script couldn't response as expect for console switch feature disabled user(https://github.com/Azure/sonic-buildimage/pull/5438)

**- How I did it**
Bump the inner command's exit code to user.

**- How to verify it**
Installed the script on a Physical DUT and test it.

```
admin@sonic:~$ consutil connect 2
Cannot connect: line [2] has no baud rate
admin@sonic:~$ echo $?
0
```

```
admin@sonic:~$ consutil connect 2
Cannot connect: line [2] has no baud rate
admin@sonic:~$ echo $?
4
```

### DIFF
--- a/connect/main.py
+++ b/connect/main.py
@@ -106,7 +106,7 @@ def line(target, devicename):
 def device(devicename):
     """Connect to device DEVICENAME via serial connection"""
     cmd = "consutil connect -d " + devicename
-    run_command(cmd)
+    sys.exit(run_command(cmd))
 
 if __name__ == '__main__':
     connect()
--- a/connect/main.py
+++ b/connect/main.py
@@ -1,6 +1,7 @@
 import configparser
 import os
 import pexpect
+import sys
 
 import click
 
@@ -71,6 +72,8 @@ def run_command(command, display_cmd=False):
 
     proc = pexpect.spawn(command)
     proc.interact()
+    proc.close()
+    return proc.exitstatus
 
 CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help', '-?'])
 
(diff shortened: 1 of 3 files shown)


lguohan:
can you list all exit code?

Blueve:
> can you list all exit code?

Sure. We defined them in consutil's lib.py(https://github.com/Azure/sonic-utilities/blob/master/consutil/lib.py)

```
ERR_DISABLE = 1
ERR_CMD = 2
ERR_DEV = 3
ERR_CFG = 4
ERR_BUSY = 5
```

connect/main.py
```diff
@@ -93,7 +96,7 @@ def connect():
 def line(target, devicename):
     """Connect to line LINENUM via serial connection"""
     cmd = "consutil connect {}".format("--devicename " if devicename else "") + str(target)
-    run_command(cmd)
+    sys.exit(run_command(cmd))
```
lguohan, line 99: why not change line 109 as well?

Blueve: This is not console command, I think we need change it in another PR if need.
Currently, I think we only leverage console related command's exit code, I think it is OK to keep other command no change.

lguohan: what do you mean 109 is not console command? it is also consutil.

Blueve: ahh... i see, I will update it

Blueve: This is a undocumented command. I fixed it and updated the document

lguohan (approved):

PR #1257, opened 2020-11-20
[acl-loader] Make list sorting compliant with Python 3

jleveque:
In Python 3, `sort()` takes no positional arguments, only keyword arguments. Fix acl-loader to comply.

Fixes vsimage build failure in https://github.com/Azure/sonic-buildimage/pull/5926


lguohan (approved):

lguohan:
i am glad to see kvm test prevent regression.

PR #1260, opened 2020-11-21
Remove unnecessary calls to str.encode() now that the package is Python 3; Fix deprecation warning

jleveque:
In Python 3, all strings are unicode by default. There is no need to encode strings to bytes.

Fixes vsimage check build failure in Azure/sonic-buildimage#5926, due to caclmgrd test failing because acl-loader was encoding the table names to bytes and comparing them against strings, thus failing to match table names and therefore failing to load ACL rules.

Also fix Python 3 deprecation warning in decode-syseeprom, as the `imp` module is deprecated in favor of `importlib`

### DIFF
--- a/acl_loader/main.py
+++ b/acl_loader/main.py
@@ -429,14 +429,14 @@ class AclLoader(object):
                 rule_props["IP_PROTOCOL"] = rule.ip.config.protocol
 
         if rule.ip.config.source_ip_address:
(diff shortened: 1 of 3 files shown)


jleveque:
Retest this please

lguohan (approved):

qiluo-msft:
LGTM

PR #1262, opened 2020-11-23
Fix counters of portstat are abnormal after config reload

Aster-ZhangBen:
Signed-off-by: Aster-zhangben <zhangben@asterfusion.com>

**- What I did**
Fix packets&bytes counters of 'portstat' or 'show interfaces counters' are huge numbers or negative numbers after config reload.

**- How I did it**
Delete the data file of 'portstat' when 'config reload' is executed. The path of 'portstat' data file is '/tmp/portstat-1000'

**- How to verify it**
step1: Injection flow into the port of DUT before reload
step2: Execute 'portstat -c' and then 'config reload'
step3: After restarting, check whether the 'portstat' data file still exists and whether the packets&bytes counters display is normal.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1037,6 +1037,9 @@ def reload(db, filename, yes, load_sysinfo, no_service_restart):
         else:
             cfg_hwsku = cfg_hwsku.strip()
 
+    log_info("Delete portstat data file to ensure accuracy after reload")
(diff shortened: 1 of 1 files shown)


tengfei-astfs:
@stcheng @lguohan Please review

PR #1246, opened 2020-11-17
[config_mgmt.py]: Remove sort_data call, we do not need it now.

praveen-li:
[config_mgmt.py]: Remove sort_data call, we do not need it now.

Backend handles unsorted Port creation.

Signed-off-by: Praveen Chaudhary<pchaudhary@linkedin.com>

**- What I did**
[config_mgmt.py]: Remove sort_data call, we do not need it now.

**- How I did it**
[config_mgmt.py]: Remove sort_data call, we do not need it now.

**- How to verify it**
Build the PKG.

**- Previous command output (if the output of a command-line utility has changed)**

**- New command output (if the output of a command-line utility has changed)**

PR #1264, opened 2020-11-23
[mgmt] Relocate sort_data Method to Config Mgmt

tahmed-dev:
sort_data was deprecatred from sonic-cfggen and so it is being
relocated to the only module that is usingl config_mgmt.

resolves: Azure/sonic-buildimage#6008

**- What I did**
Relocated sort_data to config_mgmt module

**- How I did it**
code change

**- How to verify it**

N/A

N/A

### DIFF
--- a/config/config_mgmt.py
+++ b/config/config_mgmt.py
@@ -6,11 +6,13 @@ try:
     import re
     import syslog
 
-    from json import load
-    from time import sleep as tsleep
+    from collections import OrderedDict
     from imp import load_source
+    from json import load
     from jsondiff import diff
+    from natsort import natsorted
     from sys import flags
+    from time import sleep as tsleep
 
     # SONiC specific imports
     import sonic_yang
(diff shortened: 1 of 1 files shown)


jleveque (approved):

config/config_mgmt.py
```diff
@@ -198,6 +200,12 @@ def readConfigDB(self):
 
         return
 
+    def sortData(self, data):
```
praveen-li, line 203: @tahmed-dev : We do not need it now, https://github.com/Azure/sonic-utilities/pull/1246

I thought --print-data for sonic-cfggen must print sorted value, but if that is not true, We can remove this function all together now.

praveen-li (changes requested):
Similar PR: https://github.com/Azure/sonic-utilities/pull/1246

config/config_mgmt.py
tahmed-dev: @praveen-li I was not aware of [PR:1246](https://github.com/Azure/sonic-utilities/pull/1246). Closing this one in favor of [PR:1246](https://github.com/Azure/sonic-utilities/pull/1246)

PR #1271, opened 2020-11-25
Feature table cli command update

abdosi:
What I did:

- added new value "always_enabled" for the state and auto-restart field of Feature Table

- Once the field value is "always_enabled" it can be changes from config feature .. command

- Updated Command Reference Manual

How I Verify:

Added the unit test case to change state/auto-restart field value of feature that has always_enabled and it should not update the field value.


jleveque (changes requested):
LGTM. Can you also please update the Command Line Reference guide as part of this PR? I realized that the guide did not get updated once we changed the command from `show/config container feature autorestart` to `show/config feature autorestart`, and `show/config feature state` were never added. See: https://github.com/Azure/sonic-utilities/blob/master/doc/Command-Reference.md#container-auto-restart.

Thus can you please do the following:
1. Rename (and relocate alphabetically) the sections for `show/config container feature autorestart` to `show/config feature autorestart` and update the example output to display the "always_enabled" states
2. Add documentation for `show/config container feature state` commands

abdosi:
> LGTM. Can you also please update the Command Line Reference guide as part of this PR? I realized that the guide did not get updated once we changed the command from `show/config container feature autorestart` to `show/config feature autorestart`, and `show/config feature state` were never added. See: https://github.com/Azure/sonic-utilities/blob/master/doc/Command-Reference.md#container-auto-restart.
> 
> Thus can you please do the following:
> 
> 1. Rename (and relocate alphabetically) the sections for `show/config container feature autorestart` to `show/config feature autorestart` and update the example output to display the "always_enabled" states
> 2. Add documentation for `show/config container feature state` commands

Thanks @jleveque for pointing this out. I have updated the document accordingly.

doc/Command-Reference.md
```diff
@@ -48,6 +45,12 @@
 * [ECN](#ecn)
   * [ECN show commands](#ecn-show-commands)
   * [ECN config commands](#ecn-config-commands)
+* [Feature State](#feature-state)
+  * [Feature State show commands](#feature-state-show-commands)
+  * [Feature State config command](#feature-state-config-command)
+* [Feature Auto-restart](#feature-auto-restart)
+  * [Feature Auto-restart show commands](#feature-auto-restart-show-commands)
+  * [Feature Auto-restart config command](#feature-auto-restart-config-command)
```
jleveque, line 53: I think we should combine these into one section (i.e., one section per top-level group), otherwise this table of contents will get extremely long. E.g.,

```
* [Feature](#feature)
  * [Feature show commands](#feature-show-commands)
  * [Feature config commands](#feature-config-commands)
```

Then list all the commands under each group. This is how the other sections are currently laid out.

jleveque (changes requested):

doc/Command-Reference.md
abdosi: @jleveque updated accordingly

jleveque (approved):

abdosi:
@daall Seems some issue with DVS test case for Policer. 
These failure are un-related to this PR.

daall:
> @daall Seems some issue with DVS test case for Policer.
> These failure are un-related to this PR.

ack

abdosi:
retest this please

PR #1279, opened 2020-12-02
[nbrshow] Convert the output of 'arp -n' from bytes to str.

bingwang-ms:
Signed-off-by: bingwang <bingwang@microsoft.com>

**- What I did**
Fix https://github.com/Azure/sonic-buildimage/issues/6085

The output of 'arp -n' is bytes in python3, as a result, the search of 'ether' or other str failed. This commit addresses the issue.

I guess there are similar issues when upgrading to python3. I think we need to come up with a universal method to address these issues.

**- How I did it**
Convert the output of 'arp -n' from bytes to str.

**- How to verify it**
Verified on dx010-6, running master image.
```
admin@str2-dx010-acs-6:~$ show arp
Address       MacAddress         Iface            Vlan
------------  -----------------  ---------------  ------
10.0.0.1      52:54:00:b7:a8:31  PortChannel0002  -
10.0.0.5      52:54:00:b9:b9:23  PortChannel0005  -
......
10.64.246.1   00:e0:ec:83:b8:0f  eth0             -
10.64.247.30  80:3f:5d:08:0d:8e  eth0             -
Total number of entries 26 
```
```
admin@str2-dx010-acs-6:~$ show ndp
Address                    MacAddress         Iface            Vlan    Status
-------------------------  -----------------  ---------------  ------  ---------
fc00::1a                   52:54:00:a8:5b:e7  PortChannel0011  -       REACHABLE
fc00::2                    52:54:00:b7:a8:31  PortChannel0002  -       REACHABLE
......
fe80::9a03:9bff:fe03:2289  98:03:9b:03:22:89  PortChannel0005  -       STALE
fe80::e42:a1ff:fed9:26fe   1c:34:da:eb:bc:80  eth0             -       STALE
Total number of entries 49 
```

### DIFF
--- a/scripts/nbrshow
+++ b/scripts/nbrshow
@@ -106,7 +106,7 @@ class NbrBase(object):
         """
             Fetch Neighbor data (ARP/IPv6 Neigh) from kernel.
         """
-        p = subprocess.Popen(self.cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        p = subprocess.Popen(self.cmd, shell=True, text=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         (output, err) = p.communicate()
         rc = p.wait()
 


scripts/nbrshow
```diff
@@ -111,7 +111,10 @@ class NbrBase(object):
         rc = p.wait()
 
         if rc == 0:
-            rawdata = output
+            if isinstance(output, bytes):
+                rawdata = output.decode()
+            else:
+                rawdata = output
```
jleveque, line 117: Thanks for catching this. I must have missed it. However, rather than this change, please add the parameter `text=True` to the `Popen` call on line 109 instead. E.g.,

```python3
p = subprocess.Popen(self.cmd, shell=True, text=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```

This will force the output to be strings, not bytes.


p = subprocess.Popen(self.cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)

jleveque (changes requested):

scripts/nbrshow
bingwang-ms: I considered this method too. But the ```text=True``` is only available on python 3. I'm afraid we might run this script on python 2 and the call will trigger another exception. So I select current way.

jleveque: sonic-utilities has been converted to Python 3 in the master branch. We are no longer supporting Python 2 moving forward. You can make this change instead.

bingwang-ms: Sure. Updated. Thanks

jleveque (approved):

PR #1281, opened 2020-12-02
[acl_loader] Fix default DENY rule for V6 dataplane ACLs

daall:

**- What I did**
I updated the default DENY rule for V6 to use IP type rather than Ethertype. This is necessary because L3V6 tables do not support Ethertype as a qualifier.

**- How I did it**
Updated the deny_rule method to use the new field.

**- How to verify it**
1. Run acl-loader on a table of type L3V6
2. Verify there are no errors in the logs
3. Verify that IPv6 traffic that does not hit any of the ALLOW rules is dropped.
4. Verify that `aclshow` shows these drops under the DEFAULT_RULE.

### DIFF
--- a/acl_loader/main.py
+++ b/acl_loader/main.py
@@ -539,7 +539,7 @@ class AclLoader(object):
         rule_props["PRIORITY"] = str(self.min_priority)
         rule_props["PACKET_ACTION"] = "DROP"
         if 'v6' in table_name.lower():
-            rule_props["ETHER_TYPE"] = str(self.ethertype_map["ETHERTYPE_IPV6"])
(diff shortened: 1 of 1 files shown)


jleveque (approved):
LGTM. Please wait for other reviewers.

prsunny (approved):
lgtm, just to confirm, is the qualifier already available in existing platforms?

daall:
> lgtm, just to confirm, is the qualifier already available in existing platforms?

@prsunny I've evaluated a few platforms and it seems well supported, let me chat with you a bit more offline to make sure I have my bases covered

lguohan (approved):

daall:
retest default please

daall:
retest this please

daall:
retest this please

PR #1290, opened 2020-12-04
[config] Support for running `config reload` in background

anand-kumar-subramanian:
Config reload enhancements

- What I did
Config reload enhancements

Support for running the config reload command in background mode. This is needed when we issue config reload in an SSH session and disconnect the SSH connection, it should continue the config reload without terminating it.
Exceptions are captured and logged.

- How I did it

Added support for running the config reload command in background mode, so that even if the session timesout on an SSH session, the command will still run to completion.
Exceptions capture and logging on all the operations inside config reload 

- How to verify it
config reload

- Previous command output (if the output of a command-line utility has changed)

- New command output (if the output of a command-line utility has changed)


config/main.py
```diff
@@ -744,11 +749,17 @@ def _restart_services(config_db):
...
+        execute_systemctl(services_to_restart, SYSTEMCTL_ACTION_RESTART)
 
-    # Reload Monit configuration to pick up new hostname in case it changed
-    click.echo("Reloading Monit configuration ...")
-    clicommon.run_command("sudo monit reload")
+        # Reload Monit configuration to pick up new hostname in case it changed
+        click.echo("Reloading Monit configuration ...")
+        clicommon.run_command("sudo monit reload")
+    except Exception as e:
+        log.log_error("'_restart_verify_services' start services failed, error: {}".format(e))
```
jleveque, line 759: This log message is strange. `_restart_verify_services` is not the name of the function. Probably should read:

```python3
log.log_error("'_restart_services' failed, error: {}".format(e))
```

config/main.py
```diff
@@ -995,6 +1006,30 @@ def load(filename, yes):
         log.log_info("'load' executing...")
         clicommon.run_command(command, display_cmd=True)
 
+def handle_signal(signum, frame):
+    cmd= "command execution is aborted, received signal {}".format(signum)
```
jleveque, line 1010: Add space before `=`

config/main.py
```diff
@@ -995,6 +1006,30 @@ def load(filename, yes):
...
+    log.log_info(cmd)
+    if child_pid != 0 :
+       os.kill(child_pid, signal.SIGTERM)
+    sys.exit()
+
+def ignore_signal(signum, frame):
+    infd = open(os.devnull, 'r')
+    os.dup2(infd.fileno(), sys.stdin.fileno())
+
+    #outfd = open(os.devnull, 'a+')
```
jleveque, line 1021: Remove commented code

config/main.py
```diff
@@ -995,6 +1006,30 @@ def load(filename, yes):
...
+    outfd = open(PIPE_FILE, 'a+')
+    os.dup2(outfd.fileno(), sys.stderr.fileno())
+    os.dup2(outfd.fileno(), sys.stdout.fileno())
+
+    cmd= "command execution is continued, received signal {}".format(signum)
```
jleveque, line 1026: Add space before `=`

config/main.py
```diff
@@ -995,6 +1006,30 @@ def load(filename, yes):
...
+    log.log_info(cmd)
+    return
+
+# tail the output of reload command
+def config_reload_log_thread(pid, outfd):
+    os.system("tail -f --pid={} {}".format(pid,PIPE_FILE))
```
jleveque, line 1032: Add space after `,`

config/main.py
```diff
@@ -1031,93 +1067,155 @@ def reload(db, filename, yes, load_sysinfo, no_service_restart):
...
-            click.echo("Could not get the HWSKU from config file, exiting")
-            sys.exit(1)
-        else:
-            cfg_hwsku = cfg_hwsku.strip()
+    if os.path.isfile(PIPE_FILE):
+        os.remove(PIPE_FILE)
+    outfd = open(PIPE_FILE, 'w+')
+
+    child_pid=os.fork()
+        #Parent: wait for child to exit
```
jleveque, line 1075: Add space after `#`; Line also indented too far

config/main.py
```diff
@@ -1031,93 +1067,155 @@ def reload(db, filename, yes, load_sysinfo, no_service_restart):
...
+        if os.path.isfile(PIPE_FILE):
+            os.remove(PIPE_FILE)
+        thread_service_event.join()
+        if pid:
+            sys.exit(status >> 8)
+        return
+    else:
+        # Decouple from parent environment
+        #os.chdir(os.getcwd())
+        #os.chdir('/')
```
jleveque, line 1093: Remove commented code

config/main.py
```diff
@@ -1031,93 +1067,155 @@ def reload(db, filename, yes, load_sysinfo, no_service_restart):
...
+        os.setsid( )
+        os.umask(0)
+        infd = open(os.devnull, 'r')
+        #outfd = open(os.devnull, 'a+')
```
jleveque, line 1097: Remove commented code

config/main.py
```diff
@@ -1031,93 +1067,155 @@ def reload(db, filename, yes, load_sysinfo, no_service_restart):
...
+                os.remove(PIPE_FILE)
+                sys.exit(1)
+            else:
+                cfg_hwsku = cfg_hwsku.strip()
 
-    #Stop services before config push
-    if not no_service_restart:
-        log.log_info("'reload' stopping services...")
-        _stop_services(db.cfgdb)
+        #disable warm-boot configuration
```
jleveque, line 1117: Add space after `#`

config/main.py
```diff
@@ -1031,93 +1067,155 @@ def reload(db, filename, yes, load_sysinfo, no_service_restart):
...
-    # service running in the host + DB services running in each ASIC namespace created per ASIC.
-    # In the below logic, we get all namespaces in this platform and add an empty namespace ''
-    # denoting the current namespace which we are in ( the linux host )
-    for inst in range(-1, num_cfg_file-1):
-        # Get the namespace name, for linux host it is None
-        if inst == -1:
-            namespace = None
-        else:
-            namespace = "{}{}".format(NAMESPACE_PREFIX, inst)
+        #Stop services before config push
```
jleveque, line 1120: Add space after `#`

jleveque (changes requested):
As comments.

Also, just curious if this change is truly necessary. Have you tried simply running `sudo config reload -y &` to run in the background?

config/main.py
anand-kumar-subramanian: Modified.

config/main.py
anand-kumar-subramanian: Removed.

config/main.py
anand-kumar-subramanian: Added space.

config/main.py
anand-kumar-subramanian: Added space.

config/main.py
anand-kumar-subramanian: Added space.

config/main.py
anand-kumar-subramanian: Removed.

config/main.py
anand-kumar-subramanian: Removed.

config/main.py
anand-kumar-subramanian: Added space.

config/main.py
anand-kumar-subramanian: Added space.

anand-kumar-subramanian:
This is needed as we cannot ask the user the do that every time. Config reload command should automatically take care of it instead of asking the user to run the command in the background.

config/main.py
```diff
@@ -58,9 +59,11 @@
 CFG_LOOPBACK_NAME_TOTAL_LEN_MAX = 11
 CFG_LOOPBACK_ID_MAX_VAL = 999
 CFG_LOOPBACK_NO="<0-999>"
+PIPE_FILE = "/tmp/reload.txt"
 
 
 asic_type = None
+child_pid=0
```
jleveque, line 66: I'm not too keen on making this a global, since it is only used by `config reload`

config/main.py
```diff
@@ -58,9 +59,11 @@
 CFG_LOOPBACK_NAME_TOTAL_LEN_MAX = 11
 CFG_LOOPBACK_ID_MAX_VAL = 999
 CFG_LOOPBACK_NO="<0-999>"
+PIPE_FILE = "/tmp/reload.txt"
```
jleveque, line 62: `PIPE_FILE` is not very descriptive for a global constant. If this needs to be global, it should at least be more descriptive, like `RELOAD_PIE_FILE` or similar.

config/main.py
jleveque: I'm not too keen on making this a global, since it is only used by `config reload`.

Also, should all spaces around `=`

config/main.py
```diff
@@ -1014,7 +1048,8 @@ def reload(db, filename, yes, load_sysinfo, no_service_restart):
     if not yes:
         click.confirm(message, abort=True)
 
-    log.log_info("'reload' executing...")
+    global child_pid
+    global PIPE_FILE
```
jleveque, line 1052: `PIPE_FILE` is a constant. You are not modifying it, so no need to declare it here with `global` keyword.

config/main.py
```diff
@@ -995,6 +1006,29 @@ def load(filename, yes):
...
 
+def handle_signal(signum, frame):
+    cmd = "command execution is aborted, received signal {}".format(signum)
+    print(cmd)
+    log.log_info(cmd)
+    if child_pid != 0 :
+       os.kill(child_pid, signal.SIGTERM)
+    sys.exit()
+
+def ignore_signal(signum, frame):
```
jleveque, line 1017: This function doesn't appear to be used. Remove?

jleveque (changes requested):

config/main.py
```diff
@@ -1048,9 +1036,6 @@ def reload(db, filename, yes, load_sysinfo, no_service_restart):
     if not yes:
         click.confirm(message, abort=True)
 
-    global child_pid
```
jleveque: You _are_ setting `child_pid` in this function, and it is a global, so I think you should have kept this line.

jleveque (changes requested):

lguohan:
this is a behavioral change, can we put it as optional?

ben-gale:
@lguohan, @jleveque - I discussed making this optional with the team. It could be done, but I question the merit of doing this. This change (PR) was done in response to a real issue in a real customer deployment. In that case they were leaving multiple switches in an unstable/indeterminate state on losing network connectivity after issuing a config reload command from a remote SSH session. Basically "config reload" was getting stopped after stopping a bunch of containers but before restarting them, and they were never restarting. A re-boot was required to get out of this situation. So, this is a real problem that needs a real solution. Users who are not seeing this problem wouldn't observe a change in behavior anyway. So, I would argue that making this optional is both undesirable (leaves a problem in place, adds complexity) and unnecessary.  I think we should add this unconditionally,.

Thoughts?

ben-gale:
@lguohan, @jleveque - your comments pls? Thx

ben-gale:
@lguohan, @jleveque - still looking for your comments pls.

fk410167:
retest this please

ben-gale:
@lguohan, @jleveque - PR has been updated to add the "-- background" option as discussed - pls review.

Thanks.

PR #1291, opened 2020-12-04
Parse sonic_platform kernel command line parameter to read the platform identifier string

rajendra-dendukuri:
**- What I did**
The platform identifier string is used in the reboot script to identify and execute any custom platform specific reboot script. 
sonic_platform is used as the kernel parameter to pass the platform identifier string.

**- How I did it**
In the reboot script, under the case where it is executed by the crash kernel, parse /proc/cmdline and parse sonic_platform value.

**- How to verify it**
Enable kdump and reboot to operationally enable it.
Using echo c /proc/sysrq-trigger, crash the kernel and observe that the crash kernel executes platform specific reboot script

### DIFF
--- a/scripts/reboot
+++ b/scripts/reboot
@@ -10,7 +10,7 @@ if [ -e $VMCORE_FILE -a -s $VMCORE_FILE ]; then
         echo "We have a /proc/vmcore, then we just kdump'ed"
         echo "User issued 'kdump' command [User: kdump, Time: ${REBOOT_TIME}]" > ${REBOOT_CAUSE_FILE}
         sync
-        PLATFORM=$(grep -oP 'platform=\K\S+' /proc/cmdline)
+        PLATFORM=$(grep -oP 'sonic_platform=\K\S+' /proc/cmdline)
         if [ ! -z "${PLATFORM}" -a -x ${DEVPATH}/${PLATFORM}/${PLAT_REBOOT} ]; then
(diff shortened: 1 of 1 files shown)


scripts/reboot
```diff
@@ -10,7 +10,7 @@ if [ -e $VMCORE_FILE -a -s $VMCORE_FILE ]; then
         echo "We have a /proc/vmcore, then we just kdump'ed"
         echo "User issued 'kdump' command [User: kdump, Time: ${REBOOT_TIME}]" > ${REBOOT_CAUSE_FILE}
         sync
-        PLATFORM=$(grep -oP 'platform=\K\S+' /proc/cmdline)
+        PLATFORM=$(grep -oP 'sonic_platform=\K\S+' /proc/cmdline)
```
lguohan, line 13: I have curious why we cannot use sonic-cfggen to get the platform as line 24?

rajendra-dendukuri: When we are running in the crash kernel context, we do not want to access databases or libraries or anything more involving. The idea is to do as little work as possible to dump the crash kernel and reboot. For that reason, we are passing the platform name as part of the command line.

rajendra-dendukuri: @lguohan can you please approve and merge these changes. We need to move the submodule pointer in sonic-buildimage for all the changes to work.

lguohan (approved):

PR #1302, opened 2020-12-11
Fix show ip/v6 route summary on non-multi-asic platforms to let FRR handle it completely

gechiang:
**- What I did**
1. Changed the handling of "show ip/v6 route xxxx" for non-multi-asic platforms to be handled by FRR completely instead of using the common code added to support multi-asic platform.  This is because there are some options such as summary and tables are not yet supported by multi-asic platform and we should not penalize non-multi-asic platform by sharing the same common code.
2. Due to this change, had to modify/add additional unit tests to ensure good test coverage of > 95%.
3. Took the opportunity to refactor the expected test data output strings into a standalone file to make the test code easier to read.

**- How to verify it**
Perform "show ip/v6 route summary" and many other options to ensure that both pizzabox and multi-asic platforms are handling them correctly.
Before the fix:
```
admin@str-a7050-acs-2:~$ show ip route summary
```

After the fix:
```
admin@str-a7050-acs-2:~$ show ip route summary
connected            26                   26                   
ebgp                 6426                 6425                 
ibgp                 0                    0                    
------
Totals               6454                 6452                 


admin@str-a7050-acs-2:~$
```

Fix https://github.com/Azure/sonic-buildimage/issues/6153


lguohan (approved):

tests/conftest.py
```diff
@@ -111,12 +96,28 @@ def mock_run_bgp_command(vtysh_cmd, bgp_namespace):
...
+        elif request.param == 'ip_specific_route':
+            return show_ip_route_common.show_specific_ip_route_expected_output 
+        elif request.param == 'ip_special_route':
+            return show_ip_route_common.show_special_ip_route_expected_output
+        elif request.param == 'ipv6_route':
+            return show_ip_route_common.show_ipv6_route_expected_output
+        elif request.param == 'ipv6_specific_route':
+            return show_ip_route_common.show_ipv6_route_single_json_expected_output
+        else:
+            return ""
```
daall, line 113: Thoughts on having this as a dictionary and returning `show_ip_route_common.expected_output.get("ipv6_route_err", "")`?

gechiang: @daall Interesting idea!
Let me try it out.

gechiang: @daall Unfortunately I see cases where it is not a 1 to 1 mapping and it is test dependent...  So I will stick with the current method for now...

daall (approved):

PR #1306, opened 2020-12-12
Fix show ip/v6 route summary non-multi-asic platform to interact with…

gechiang:
**- What I did**
1. Changed the handling of "show ip/v6 route xxxx" for non-multi-asic platforms to be handled by FRR completely instead of using the common code added to support multi-asic platform.  This is because there are some options such as summary and tables are not yet supported by multi-asic platform and we should not penalize non-multi-asic platform by sharing the same common code.
2. Due to this change, had to modify/add additional unit tests to ensure good test coverage of > 95%.
3. Took the opportunity to refactor the expected test data output strings into a standalone file to make the test code easier to read.

**- How to verify it**
Perform "show ip/v6 route summary" and many other options to ensure that both pizzabox and multi-asic platforms are handling them correctly.
Before the fix:
```
admin@str-a7050-acs-2:~$ show ip route summary
```

After the fix:
```
admin@str-a7050-acs-2:~$ show ip route summary
connected            26                   26                   
ebgp                 6426                 6425                 
ibgp                 0                    0                    
------
Totals               6454                 6452                 


admin@str-a7050-acs-2:~$
```

This PR is created to port the same change of the master branch to 201911 branch for PR: 
https://github.com/Azure/sonic-utilities/pull/1302


show/bgp_common.py
```diff
@@ -356,14 +357,22 @@ def show_routes(args, namespace, display, verbose, ipver):
...
                 filter_by_ip = ipaddress.ip_network(arg)
             except ValueError:
                 # Not ip address just ignore it
                 pass
-    if not found_json:
-        arg_strg += "json"
+    # Due to options such as "summary" and "tables" are not yet supported in multi-asic platform
+    # we will let FRR handle all the processing instead of handling it here for non multi-asic platform
+    if multi_asic.is_multi_asic():
+        if found_tables:
```
smaheshm, line 371: how are you checking for "summary" option there?

gechiang: For Non-multi-asic platform the parameter checking is all done by BGP itself.
For Multi-ASIC platform it currently will fail due to the added "json" parm was requested to BGP and BGP will fail the request.  So there is no need to specifically check for this for now.
I will be adding the support for the Multi-ASIC platform for the "summary" option which at that time it will be inspected and handled accordingly.  Not part of this PR though...

abdosi (approved):

PR #1320, opened 2020-12-19
Added show ip/v6 route summary support for multi-asic platform

gechiang:
**- What I did**
Added Multi-ASIC support of handling "show ip/v6 route summary"
This was not supported for multi-ASIC platform previously.
When running this cmd under multi-ASIC platform the following output are expected:
```
admin@SONiC:$ show ip route summary -d all
asic0:
connected            6                    6
ebgp                 6371                 6371
ibgp                 88                   88
------
Totals               6467                 6466


asic1:
connected            6                    6
ebgp                 6371                 6371
ibgp                 88                   88
------
Totals               6467                 6466


asic2:
connected            14                   14
ebgp                 43                   43
ibgp                 6408                 6408
------
Totals               6467                 6466


asic3:
connected            14                   14
ebgp                 42                   42
ibgp                 6409                 6409
------
Totals               6467                 6466


asic4:
connected            6                    6
ebgp                 0                    0
ibgp                 6462                 6458
------
Totals               6470                 6465


asic5:
connected            6                    6
ebgp                 0                    0
ibgp                 6462                 6458
------
Totals               6470                 6465


admin@SONiC:$ show ip route summary -n asic2
asic2:
connected            14                   14
ebgp                 43                   43
ibgp                 6408                 6408
------
Totals               6467                 6466


admin@SONiC:$ show ip route summary
asic0:
connected            6                    6
ebgp                 6371                 6371
ibgp                 88                   88
------
Totals               6467                 6466


asic1:
connected            6                    6
ebgp                 6371                 6371
ibgp                 88                   88
------
Totals               6467                 6466


asic2:
connected            14                   14
ebgp                 43                   43
ibgp                 6408                 6408
------
Totals               6467                 6466


asic3:
connected            14                   14
ebgp                 42                   42
ibgp                 6409                 6409
------
Totals               6467                 6466


admin@SONiC:$
 
```

**- How to verify it**
Perform "show ip/v6 route summary" on multi-asic platform to see that this option is now supported.
Before the change:
```
admin@SONiC:~$ show ip route summary
```

After the added support:
```
admin@SONiC:~$ show ip route summary
asic0:
connected            6                    6
ebgp                 6371                 6371
ibgp                 88                   88
------
Totals               6467                 6466


asic1:
connected            6                    6
ebgp                 6371                 6371
ibgp                 88                   88
------
Totals               6467                 6466


asic2:
connected            14                   14
ebgp                 43                   43
ibgp                 6408                 6408
------
Totals               6467                 6466


asic3:
connected            14                   14
ebgp                 42                   42
ibgp                 6409                 6409
------
Totals               6467                 6466


admin@SONiC:$
```

### DIFF
--- a/show/bgp_common.py
+++ b/show/bgp_common.py
@@ -331,6 +331,7 @@ def show_routes(args, namespace, display, verbose, ipver):
     arg_strg = ""
     found_json = 0
     found_tables = 0
+    found_summary = 0
     ns_l = []
     print_ns_str = False
     filter_by_ip = False
@@ -353,25 +354,29 @@ def show_routes(args, namespace, display, verbose, ipver):
     else:
         back_end_intf_set = None
     # get all the other arguments except json that needs to be the last argument of the cmd if present
+    # Handling of multi-asic show ip/v6 route summary will print directly FRR result of each namespace
+    check_summary_set = set(["sum", "summ", "summa", "summar", "summary"])
     for arg in args:
         arg_strg += str(arg) + " "
         if str(arg) == "json":
             found_json = 1
         elif str(arg) == "tables":
             found_tables = 1
+        elif set([str(arg)]).issubset(check_summary_set):
+            found_summary = 1
         else:
             try:
                 filter_by_ip = ipaddress.ip_network(arg)
             except ValueError:
                 # Not ip address just ignore it
                 pass
-    # Due to options such as "summary" and "tables" are not yet supported in multi-asic platform
+    # Due to option such as "tables" is not yet supported in multi-asic platform
     # we will let FRR handle all the processing instead of handling it here for non multi-asic platform
     if multi_asic.is_multi_asic():
         if found_tables:
             print("% Unknown command: show {} route {}".format(ipver, arg_strg))
             return
-        if not found_json:
+        if not found_json and not found_summary:
             arg_strg += "json"
     combined_route = {}
     for ns in ns_l:
@@ -398,6 +403,13 @@ def show_routes(args, namespace, display, verbose, ipver):
                 error_msg = output
             print(error_msg)
             return
+
+        # Multi-asic show ip route summary is handled by going to FRR directly and get those outputs from each namespace
+        if multi_asic.is_multi_asic() and found_summary:
+            print("{}:".format(ns))
+            print(output)
+            continue
+
         route_info = json.loads(output)
         if filter_back_end or print_ns_str:
             # clean up the dictionary to remove all the nexthops that are back-end interface
--- a/show/bgp_common.py
+++ b/show/bgp_common.py
@@ -330,8 +330,7 @@ def show_routes(args, namespace, display, verbose, ipver):
     device = multi_asic_util.MultiAsic(display, namespace)
     arg_strg = ""
     found_json = 0
-    found_tables = 0
-    found_summary = 0
+    found_other_parms = 0
     ns_l = []
     print_ns_str = False
     filter_by_ip = False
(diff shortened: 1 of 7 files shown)


gechiang:
retest default please

gechiang:
retest this please

show/bgp_common.py
```diff
@@ -398,6 +403,13 @@ def show_routes(args, namespace, display, verbose, ipver):
                 error_msg = output
             print(error_msg)
             return
+
```
arlakshm, line 406: If one of the bgp instances in the middle returns error. Output from the remaining instances wont be displayed ?

show/bgp_common.py
```diff
@@ -353,25 +354,29 @@ def show_routes(args, namespace, display, verbose, ipver):
     else:
         back_end_intf_set = None
     # get all the other arguments except json that needs to be the last argument of the cmd if present
+    # Handling of multi-asic show ip/v6 route summary will print directly FRR result of each namespace
+    check_summary_set = set(["sum", "summ", "summa", "summar", "summary"])
```
arlakshm, line 358: Instead of adding the check for argument  is "summary" here.  Can we check if the args is empty  to aggregate the routes from all namesapces  for all other argument we can display the output as-is for each namespace ?

arlakshm (commented):
as comments

show/bgp_common.py
gechiang: Yes. That is the intention.  Most of the case when the FRR of that namespace reject the cmd is due to the parameter the user specified.  and if one name space rejects it, all other namespace will reject the same way so it is best to stop on first error.  Also, if in case the user thinks the issue is only specific to the failing name space but not on other name space, user can always use -n option to display from which namespace he likes to see.

show/bgp_common.py
gechiang: We also like to aggregate the routes from all namespaces for the case where the user specified specific route or in json format.  So for these cases the argument will not be empty but we still like to combine the output from all namespaces and display it.  Are you saying that other than these cases we should always default to let each specified namespace FRR to handle those additional parameter options similar to the "summary" case?  There will be cases where the "json" option will also be honored and for those we will NOT combine the jason output but have each one displayed based on its own name space which will result to multiple json outputs depends on namespaces involved... Is this acceptable?  Or should we handle this "summary" as designed for now and make additional enhancement as we see there is a need to do so?

show/bgp_common.py
```diff
@@ -353,25 +354,29 @@ def show_routes(args, namespace, display, verbose, ipver):
...
     for arg in args:
         arg_strg += str(arg) + " "
         if str(arg) == "json":
             found_json = 1
         elif str(arg) == "tables":
             found_tables = 1
+        elif set([str(arg)]).issubset(check_summary_set):
```
smaheshm, line 365: Not sure I understand the logic. Why not:
```
if str(arg) in check_summary_set:
```

gechiang: The logic is to allow partial matching so that user need not specify the complete word "summary".
This logic is needed so that if user specify something like "umm" it will not be accepted.  it needs to start with "sum" and maybe additional matching chars after "sum".

show/bgp_common.py
gechiang: @arlakshm I have made the changes to do as you suggested.  So other than the specific route specified and the json option for "show ip/v6 route" command such as "summary, table, tables,...etc" will now be all handled by the corresponding namespace FRR directly.  This should also be easier when SONiC decided to upgrade to higher FRR version in the future.

gechiang:
retest this please

arlakshm (approved):
lgtm

PR #1331, opened 2020-12-23
[filter-fdb]: Filter Out FDB Entries With Switch MAC Address

tahmed-dev:
An entry with switch/router MAC address could be added to the FDB
table via fast-boot FDB reprogramming. This PR extend FDB filter
to detect those entries and remove them.

**- What I did**
Filter out FDB entries that switch/router MAC address

**- How I did it**
new code to filter those entries. The code retrieves the device MAC address and uses this MAC address to filter any FDB entry that has it.

**- How to verify it**
unit test added

N/A

N/A

### DIFF
--- a/fdbutil/filter_fdb_entries.py
+++ b/fdbutil/filter_fdb_entries.py
@@ -9,7 +9,7 @@ import traceback
 from collections import defaultdict
 from ipaddress import ip_address, ip_network, ip_interface
 
-def get_vlan_cidr_map(filename):
+def get_vlan_cidr_map_and_switch_mac(filename):
     """
         Generate Vlan CIDR information from Config DB file
 
@@ -37,9 +37,9 @@ def get_vlan_cidr_map(filename):
                     vlan_cidr[vlan] = {4: ip_address("0.0.0.0"), 6: ip_address("::")}
                 vlan_cidr[vlan][ip_interface(cidr).version] = ip_interface(cidr).network
 
-    return vlan_cidr
+    return vlan_cidr, config_db_entries["DEVICE_METADATA"]["localhost"]["mac"].replace(':', '-').upper()
 
-def get_arp_entries_map(arp_filename, config_db_filename):
+def get_arp_entries_map_and_switch_mac(arp_filename, config_db_filename):
     """
         Generate map for ARP entries
 
@@ -53,7 +53,7 @@ def get_arp_entries_map(arp_filename, config_db_filename):
         Returns:
             arp_map(dict) map of ARP entries using MAC as key.
     """
-    vlan_cidr = get_vlan_cidr_map(config_db_filename)
+    vlan_cidr, switch_mac = get_vlan_cidr_map_and_switch_mac(config_db_filename)
 
     with open(arp_filename, 'r') as fp:
         arp_entries = json.load(fp)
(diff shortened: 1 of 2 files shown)


fdbutil/filter_fdb_entries.py
```diff
@@ -87,15 +87,16 @@ def filter_fdb_entries(fdb_filename, arp_filename, config_db_filename, backup_fi
         Returns:
             None
     """
-    arp_map = get_arp_entries_map(arp_filename, config_db_filename)
+    arp_map, switch_mac = get_arp_entries_map_and_switch_mac(arp_filename, config_db_filename)
```
prsunny, line 90: since switch_mac is used only in this function, can we have it as a separate API to fetch instead of changing multiple existing APIs?

tests/filter_fdb_input/test_vectors.py
```diff
@@ -271,8 +362,71 @@
...
+        "arp":[
+            {
+                "NEIGH_TABLE:Vlan1000:192.168.0.10": {
+                    "neigh": "72:06:00:01:00:08",
+                    "family": "IPv4"
+                },
+                "OP": "SET"
+            },
+            {
+                "NEIGH_TABLE:Vlan1000:192.168.0.10": {
```
lguohan, line 384: is this valid scenario, having one neighbor entry map to two different macs?

lguohan: perhaps, existing code already filter out such mac, i remember previously we did correlate the fdb and arp entries, and will insert fdb entries if they exist. However, in this case, we may have fdb entry of switch mac, but I highly doubt we will also have an arp entry that points to the switch mac. Therefore, the current code should be ok.

meanwhile, I think it is good to have the check, but I think we should assign a different neighbor IP here. I do not think we can have same IP assigned to two different macs.

fdbutil/filter_fdb_entries.py
```diff
@@ -87,15 +87,16 @@ def filter_fdb_entries(fdb_filename, arp_filename, config_db_filename, backup_fi
...
 
     with open(fdb_filename, 'r') as fp:
         fdb_entries = json.load(fp)
 
     def filter_fdb_entry(fdb_entry):
         for key, _ in fdb_entry.items():
             if 'FDB_TABLE' in key:
-                return key.split(':')[-1].upper() in arp_map
+                mac = key.split(':')[-1].upper()
+                return mac != switch_mac and mac in arp_map
```
lguohan, line 99: if we see mac = switch_mac, it is kind of very strange, can we get a WARNING log here? maybe something is wrong.

tests/filter_fdb_input/test_vectors.py
prsunny: Agree with Guohan.. the current code already filters out if the arp is not present and there is no valid scenario we get an arp entry (for nbr) with switch mac.

fdbutil/filter_fdb_entries.py
tahmed-dev: It could, I just did not want to open the json files twice.

fdbutil/filter_fdb_entries.py
tahmed-dev: OK, will add logs

tests/filter_fdb_input/test_vectors.py
tahmed-dev: I invented it as I did not see the arp from the switch. I'll change the IPs. The reason to have this test case is to ensure that condition `mac != switch_mac` is hit. Will update the IP address to a different one

PR #1332, opened 2020-12-25
Use redis-subscriptions for additional validation

renukamanavalan:
Subscribe to ASIC-DB; If APPL-DB & ASIC-DB differ, collect subscription updates for a second to checkout, if missed add/deletes are seen in the subscriptions.

This is not 100% fool-proof, but pretty close with possible false negatives, which is again less likely to extend into subsequent run.

[NOTE: This PR is only for 201911. For master, this script will be moved to Python3 & swsscommon]

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/scripts/route_check.py
+++ b/scripts/route_check.py
@@ -8,6 +8,7 @@ import argparse
 import ipaddress
 import syslog
 import json
+import threading
 import time
 from enum import Enum
 from swsssdk import ConfigDBConnector
@@ -30,6 +31,20 @@ class Level(Enum):
 
 
 report_level = syslog.LOG_ERR
+
+THREAD_INST = "inst"
+THREAD_DB = "db"
+THREAD_SET = "set"
+THREAD_DEL = "del"
+thread_data = {
+        THREAD_INST: None,
+        THREAD_DB: None,
+        THREAD_SET: [],
+        THREAD_DEL: []
+        }
+ASIC_TABLE_SUBS = 'ASIC_STATE'
+ASIC_TABLE_KEYVAL = 'SAI_OBJECT_TYPE_ROUTE_ENTRY:'
+ASIC_TABLE_NAME = 'ASIC_STATE:SAI_OBJECT_TYPE_ROUTE_ENTRY'
 
 
 def set_level(lvl):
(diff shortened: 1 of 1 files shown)


scripts/route_check.py
```diff
@@ -120,22 +135,64 @@ def get_routes():
...
+        else:
+            thread_data[THREAD_DEL].append(strip_ip_asic_route_entry(key))
+        print_message(syslog.LOG_DEBUG, "key={} op={}".format(
+            key, "set" if key in keys else "del"))
+
+
+def subs_thread():
+    db = thread_data[THREAD_DB]
+    db.subscribe(ASIC_TABLE_SUBS, lambda table,
+            key, data: upd_handler(key, data))
```
abdosi, line 167: @renukamanavalan we should swsscommon here so that we can handle the operation in batch. This has been done in caclmgrd where we select and pop all the operation until empty. swsscommon will also provvide op as SET or DEL

scripts/route_check.py
```diff
@@ -211,24 +268,43 @@ def check_routes():
...
+        rt_appl_miss, _ = do_diff(rt_appl_miss, sorted(thread_data[THREAD_SET]))
+
+        # Drop all those for which DEL received
+        rt_asic_miss, _ = do_diff(rt_asic_miss, sorted(thread_data[THREAD_DEL]))
+
+    # Stop collection/thread
+    thread_data[THREAD_DB] = None
+
+    # Wait for thread to terminate.
+    thread_data[THREAD_INST].join(2)
```
abdosi, line 288: @renukamanavalan i am wondering will other thread ever terminate ? It is always blocked on listen() ?

scripts/route_check.py
renukamanavalan: The systems running 20191130.56 does not have swsscommon. Do the later versions support ?

```
remanava@BN6-0101-0123-15T1:~$ python3
Python 3.5.3 (default, Nov 18 2020, 21:09:16) 
[GCC 6.3.0 20170516] on linux
>>> from swsscommon import swsscommon
ImportError: No module named 'swsscommon'
>>> quit()
remanava@BN6-0101-0123-15T1:~$ show version| grep -e "^SO"
SONiC Software Version: SONiC.20191130.56
remanava@BN6-0101-0123-15T1:~$ 

```

scripts/route_check.py
renukamanavalan: Yes in two ways.
1) With "thread_data[THREAD_DB]=None", the thread will exit on next subscription message. 
    So we give join timeout of 2 seconds for graceful exit

2) If there is no graceful exit, this thread will be terminated upon main thread exit, as it is marked as daemon thread.

scripts/route_check.py
renukamanavalan: swsscommon is available in python2
Let me switch to it. We would not need a thread.

renukamanavalan:
As swsscommon is supported in python2, the master checkin can  be cherry-picked. So closing this.

PR #1344, opened 2020-12-31
[route_check.py] -  update includes checks on subscriptions

renukamanavalan:
**- What I did**
    The routes flow from APPL-DB to ASIC-DB, via orchagent.
    This tool's job is to verify that all routes added to APPL-DB do
    get into ASIC-DB and all routes removed from APPL-DB are deleted from ASIC-DB.

**- How I did it**
    NOTE: The flow from APPL-DB to ASIC-DB takes non zero milliseconds.
    1) Initiate subscribe for ASIC-DB updates.
    2) Read APPL-DB & ASIC-DB 
    3) Get the diff.
    4) If any diff, 
        4.1) Collect subscribe messages for a second
        4.2) check diff against the subscribe messages 
    5) Rule out local interfaces & default routes
    6) If still outstanding diffs, report failure.

**- How to verify it**
    Run this tool in SONiC switch and watch the result. In case of failure
    checkout the result to validate the failure.
    To simulate failure:
        Stop Orchagent.
        Run this tool, and likely you would see some failures.
        You could potentially remove / add routes in APPL / ASIC DBs with orchagent
        down to ensure failure.
        Analyze the reported failures to match expected.
    You may use the exit code to verify the result as success or not.

### DIFF
--- a/scripts/route_check.py
+++ b/scripts/route_check.py
@@ -1,6 +1,39 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 
+"""
+What it is:
+    The routes flow from APPL-DB to ASIC-DB, via orchagent.
+    This tool's job is to verify that all routes added to APPL-DB do
+    get into ASIC-DB.
+
+
+How:
+    NOTE: The flow from APPL-DB to ASIC-DB takes non zero milliseconds.
+    1) Initiate subscribe for ASIC-DB updates.
+    2) Read APPL-DB & ASIC-DB 
+    3) Get the diff.
+    4) If any diff, 
+        4.1) Collect subscribe messages for a second
+        4.2) check diff against the subscribe messages 
+    5) Rule out local interfaces & default routes
+    6) If still outstanding diffs, report failure.
+
+To verify:
+    Run this tool in SONiC switch and watch the result. In case of failure
+    checkout the result to validate the failure.
+    To simulate failure:
+        Stop Orchagent.
+        Run this tool, and likely you would see some failures.
+        You could potentially remove / add routes in APPL / ASIC DBs with orchagent
+        down to ensure failure.
+        Analyze the reported failures to match expected.
+    You may use the exit code to verify the result as success or not.
+    
+
+
+"""
+
 import argparse
 from enum import Enum
 import ipaddress
@@ -43,6 +76,12 @@ report_level = syslog.LOG_ERR
 write_to_syslog = False
 
 def set_level(lvl, log_to_syslog):
+    """
+    Sets the log level
+    :param lvl: Log level as ERR/INFO/DEBUG; default: syslog.LOG_ERR
+    :param log_to_syslog; True - write into syslog. False: skip
+    :return None
+    """
     global report_level
     global write_to_syslog
 
@@ -55,6 +94,12 @@ def set_level(lvl, log_to_syslog):
 
 
 def print_message(lvl, *args):
+    """
+    print and log the message for give level.
+    :param lvl: Log level for this message as ERR/INFO/DEBUG
+    :param args: message as list of strings or convertible to string
+    :return None
+    """
     if (lvl <= report_level):
         msg = ""
         for arg in args:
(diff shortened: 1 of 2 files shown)


renukamanavalan:
With the following diff, this can be cherry picked onto 201911.

```
remanava@BN6-0101-0123-15T1:~/try$ zdiff rt.py route_check.py 
1c1
< #!/usr/bin/env python
---
> #!/usr/bin/env python3
80c80
<     t = ipaddress.ip_address(ip.split("/")[0].decode('utf-8'))
---
>     t = ipaddress.ip_address(ip.split("/")[0])
85c85
<     t = ipaddress.ip_address(ip.split("/")[0].decode('utf-8'))
---
>     t = ipaddress.ip_address(ip.split("/")[0])
remanava@BN6-0101-0123-15T1:~/try$ 
```

renukamanavalan:
retest this please

liat-grozovik:
@renukamanavalan can you please update the PR description according to the template?
You may check https://github.com/Azure/sonic-utilities/pull/1345 as a reference.

liat-grozovik:
Please note that many of the PRs under sonic-utilities are failing on the 'default' and it seems to be general issue with the build system.

scripts/route_check.py
```diff
@@ -1,6 +1,39 @@
...
+    get into ASIC-DB.
+
+
+How:
+    NOTE: The flow from APPL-DB to ASIC-DB takes non zero milliseconds.
+    1) Initiate subscribe for ASIC-DB updates.
+    2) Read APPL-DB & ASIC-DB 
+    3) Get the diff.
+    4) If any diff, 
+        4.1) Collect subscribe messages for a second
```
liat-grozovik, line 17: should you verify it persist for 3 consecutive reads instead of 1 read?

scripts/route_check.py
```diff
@@ -95,6 +166,12 @@ def cmps(s1, s2):
 
 
 def do_diff(t1, t2):
+    """
```
liat-grozovik, line 169: any reason why this helper starts with do_ and the previous one is not?

scripts/route_check.py
renukamanavalan: That is another scheme. 3 consecutive reads and look for diff that is consistently same across all to imply that as possible  failures. But this has a hole if a a route was getting removed & added quickly multiple times, then it could be a false positive.

Here we look for SET/DEL in subscription messages to see, if a missed ASIC entry matches with a SET subscribe message, which would have added to ASIC. Similarly look for DEL message for an unaccounted ASIC entry as that would have removed it.

scripts/route_check.py
renukamanavalan: s/do_diff/diff_sorted_lists/

renukamanavalan:
retest this please

liushilongbuaa:
retest this please

prsunny (approved):
lgtm.. minor comments..

scripts/route_check.py
```diff
@@ -193,52 +352,82 @@ def filter_out_default_routes(lst):
...
         results["Unaccounted_ROUTE_ENTRY_TABLE_entries"] = rt_asic_miss
-        err_present = True
 
-    if err_present:
-        print_message(syslog.LOG_ERR, "results: {",  json.dumps(results, indent=4), "}")
+    if results:
+        print_message(syslog.LOG_ERR, "Failure results: {",  json.dumps(results, indent=4), "}")
         print_message(syslog.LOG_ERR, "Failed. Look at reported mismatches above")
-        return -1
+        print_message(syslog.LOG_ERR, "add: {", json.dumps(adds, indent=4), "}")
```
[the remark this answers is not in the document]
renukamanavalan: When failure is reported, the log of add/del during that one sec period, will be very useful for looking in-depth.

prsunny (approved):

abdosi:
@renukamanavalan Please create PR for 201911. Cherry-pick has conflcit.

@rlhui

abdosi:
@renukamanavalan 

> @renukamanavalan Please create PR for 201911. Cherry-pick has conflcit.
> 
> @rlhui

@renukamanavalan please create pr for 201911

PR #1349, opened 2021-01-06
[storyteller] adding a grep wrapper with predefined scenarios

yxieca:
**- What I did**

Introduce storyteller tool to grep syslog (or any other logs) for
a story line of a certain scenario.

Defined scenarios are:
- reboot  : device reboot related events.
- service : service start/stop events.
- link    : link up/down events.
- lag     : LAG up/down events.
- bgp     : BGP config change events.
- crash   : process/service crash events.

Unmatched cateory is used as grepping regex directly.

**- How to verify it**
Used this tool in some recent investigations.

### DIFF
--- a/scripts/storyteller
+++ b/scripts/storyteller
@@ -11,7 +11,7 @@ import subprocess
 
 regex_dict = {
                 'bgp'       : 'bgpcfgd',
-                'crash'     : 'what\|unexpected exception\|notify_OA_about_syncd_exception\|SIG',
+                'crash'     : 'what\|unexpected exception\|notify_OA_about_syncd_exception\|SIG\|not expected',
                 'interface' : 'updatePortOperStatus\|Configure .* to',
                 'lag'       : 'link becomes\|addLag',
                 'reboot'    : 'BOOT\|rc.local\|old_config\|minigraph.xml\|Rebooting\|reboot\|executeOperationsOnAsic\|getAsicView\|dumpVidToAsicOperatioId',
--- a/scripts/storyteller
+++ b/scripts/storyteller
@@ -20,9 +20,9 @@ regex_dict = {
 
 
 def exec_cmd(cmd):
-    out = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, shell=True)
+    out = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, shell=True, text=True)
     stdout, stderr = out.communicate()
-    return out.returncode, stdout.decode(encoding='UTF-8'), stderr.decode(encoding='UTF-8')
+    return out.returncode, stdout, stderr
 
 
 def build_options(after=0, before=0, context=0):
--- a/scripts/storyteller
+++ b/scripts/storyteller
@@ -1,0 +1,86 @@
+#!/usr/bin/env python
+
+'''
+Story Teller:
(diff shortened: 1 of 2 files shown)


scripts/storyteller
```diff
@@ -0,0 +1,86 @@
...
+    Utility to help analyze log for certain sequence of events.
+    e.g.: reboot (including warm/fast reboot), interface flapping, etc.
+'''
+
+import argparse
+import subprocess
+
+regex_dict = {
+                'bgp'       : 'bgpcfgd',
+                'crash'     : 'what\|unexpected exception\|notify_OA_about_syncd_exception\|SIG',
```
prsunny, line 14: should we also add "not expected" ?

yxieca: Glad to get some inputs :-)

yxieca: Done. Thanks Prince!

yxieca:
retest this please

scripts/storyteller
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env python
```
jleveque, line 1: sonic-utilities is now only installed as a Python 3 package, so this shebang *should* be fine. However, until we completely remove Python 2 from the image, I suggest changing this to `#!/usr/bin/env python3`, just to be sure.

jleveque (changes requested):

scripts/storyteller
yxieca: Thanks! Done!

scripts/storyteller
```diff
@@ -0,0 +1,86 @@
...
+                'crash'     : 'what\|unexpected exception\|notify_OA_about_syncd_exception\|SIG\|not expected',
+                'interface' : 'updatePortOperStatus\|Configure .* to',
+                'lag'       : 'link becomes\|addLag',
+                'reboot'    : 'BOOT\|rc.local\|old_config\|minigraph.xml\|Rebooting\|reboot\|executeOperationsOnAsic\|getAsicView\|dumpVidToAsicOperatioId',
+                'service'   : 'Starting\|Stopping\|Started\|Stopped',
+             }
+
+
+def exec_cmd(cmd):
+    out = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, shell=True)
```
jleveque, line 23: Suggest adding `text=True` parameter (this causes the return values to be string type in Python 3), then you can remove the `.decode(encoding='UTF-8')`s on line 25.

jleveque (changes requested):

scripts/storyteller
yxieca: This is so much better! Thanks!

scripts/storyteller
```diff
@@ -0,0 +1,86 @@
...
+                        type=int, required=False, default=0)
+    parser.add_argument('-C', '--context', help='Show N lines before and after match',
+                        type=int, required=False, default=0)
+
+    args = parser.parse_args()
+
+    log = args.log
+    reg = build_regex(args.category)
+
+
```
jleveque, line 81: Suggest to delete one of these blank lines, as it is inside the function. Two blank lines should be reserved for between functions. Sorry for the very iterative review.

jleveque (changes requested):

scripts/storyteller
yxieca: sure. done.

jleveque (approved):

PR #1359, opened 2021-01-15
Drop explict 3 seconds pause between two object updates/deletes.

renukamanavalan:
**- What I did**
Drop explicitly introduced 3 seconds pause between object updates using configlet.

**- How I did it**
Remove the added time.sleep(3)

**- How to verify it**
Add 10 objects using old code and see it takes 30+ seconds. Repeat the same with this updated code which would complete in small fraction of a second.

Verified ASIC-DB contents to the extent possible in both scenarios of applying objects using old code with 3 seconds pause and new code w/o any explicit pause, to be the same.

### DIFF
--- a/scripts/configlet
+++ b/scripts/configlet
@@ -77,7 +77,6 @@ A sample for update:
 
 import argparse
 import json
-import time
 
 from swsssdk import ConfigDBConnector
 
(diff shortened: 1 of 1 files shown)


qiluo-msft (approved):

renukamanavalan:
retest default please

renukamanavalan:
retest this please

PR #1363, opened 2021-01-18
[voq/inbandif] Voq inbandif port

vganesan-nokia:
**- What I did**

Inband port can be made available in PORT table. But regular port handlngs are
not applicable for Inband port. This PR has change to avoid regular port handling for inband port

Ref: VOQ HLD https://github.com/Azure/SONiC/blob/master/doc/voq/voq_hld.md

**- How I did it**

The inband port is considered similar to back plane port. An inband_prefix = "Ethernet-IB" and this is checked in all places to avoid regular port handling.

**- How to verify it**

- Expose a port with name prefixed with "Ethernet-IB" to SONiC via port_config.ini
- Have an entry in the PORT table for this port with name prefixed with "Ethernet-IB"
- When the system is booted, there should not be errors related to Inband interface

Note: This PR depends on sonic-buildimage PR https://github.com/Azure/sonic-buildimage/pull/6477

### DIFF
--- a/scripts/sfpshow
+++ b/scripts/sfpshow
@@ -12,7 +12,7 @@ import sys
 
 import click
 from natsort import natsorted
-from sonic_py_common.interface import front_panel_prefix, backplane_prefix
+from sonic_py_common.interface import front_panel_prefix, backplane_prefix, inband_prefix
 from sonic_py_common import multi_asic
 from tabulate import tabulate
 from utilities_common import multi_asic as multi_asic_util
@@ -411,7 +411,7 @@ class SFPShow(object):
             sorted_table_keys = natsorted(port_table_keys)
             for i in sorted_table_keys:
                 interface = re.split(':', i, maxsplit=1)[-1].strip()
-                if interface and interface.startswith(front_panel_prefix()) and not interface.startswith(backplane_prefix()):
+                if interface and interface.startswith(front_panel_prefix()) and not interface.startswith(backplane_prefix()) and not interface.startswith(inband_prefix()):
                     presence = self.db.exists(self.db.STATE_DB, 'TRANSCEIVER_INFO|{}'.format(interface))
                     if presence:
                         self.output += self.convert_interface_sfp_info_to_cli_output_string(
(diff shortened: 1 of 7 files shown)


anshuv-mfst:
@eswaran (Arista), @abdosi  - could you please review the PR, Thanks.

scripts/sfpshow
```diff
@@ -322,7 +322,7 @@ class SFPShow(object):
             sorted_table_keys = natsorted(port_table_keys)
             for i in sorted_table_keys:
                 interface = re.split(':', i, maxsplit=1)[-1].strip()
-                if interface and interface.startswith(front_panel_prefix()) and not interface.startswith(backplane_prefix()):
+                if interface and interface.startswith(front_panel_prefix()) and not interface.startswith(backplane_prefix()) and not interface.startswith(inband_prefix()):
```
ngoc-do, line 325: I think right now, there is no need to check `not interface.startswith(backplane_prefix()) and not interface.startswith(inband_prefix())` since `front_panel_prefix` is different from the other prefixes. But it's safer in case in future the prefixes are changed. So approved.

ngoc-do (dismissed):

scripts/sfpshow
vganesan-nokia: Thanks for approving. But we changed the prefix. As per the review comments during Recycle port HLD, we changed the Inband port prefix to "Ethernet-IB". With this, it is required to handle inband_prefix similar to backplane_prefix.

abdosi (dismissed):
LGTM

anshuv-mfst:
@ngoc-do - please re-approve

anshuv-mfst:
@ngoc-do, @eswaranb - please re-approve.

ngoc-do (approved):

abdosi:
can we add unit test case to check port is skip.

vganesan-nokia:
> can we add unit test case to check port is skip.

The existing system neigh test has been enhanced to add inband port if. Without inband interface, the neighbor will not be synced.  So in virtual chassis setup I use one of the front panel ports as the inband port. The system neigh test implicitly tests for inband if configuration, oper up and host if creation for the inband.

vganesan-nokia:
> can we add unit test case to check port is skip.

Yes. I'll add test case to check skipping Ethernet-IB.

vganesan-nokia:
Capturing approvals before rebasing

image

vganesan-nokia:
> > can we add unit test case to check port is skip.
> 
> Yes. I'll add test case to check skipping Ethernet-IB.

Fixed. Added test case for route check to unit test changes done in route_check.py. Updated the mock appl db to include inband interface so that the code changes in sfpshow will be unit tested (the verification of skipping inband port).

abdosi (approved):

renukamanavalan:
@abdosi, should this get into 202012 ?

PR #1366, opened 2021-01-18
Fix sonic-installer install for Arista secureboot

Staphylo:
**- What I did**

Abstracted away the access to the path of the rootfs via a contextmanager.
In Arista secureboot, the rootfs is never extracted on the flash. Instead it's mounted directly from within the signed SWI.
The `update_sonic_environment` function however always assume that the rootfs to be at the same place. 

**- How I did it**

To alleviate this restriction, a new context manager to obtain the rootfs is introduced.
The choice of a context manager rather than a function is entirely based on error management and cleanup.
Mounting a squashfs from a swi file requires the use of `losetup` which makes the rootfs available under `/dev/loopX`
Once done or on error, we need to free this resource which becomes free when using a contextmanager.

**- How to verify it**

Use `sonic-installer install` on any switch that is an Arista switch with secureboot enabled.
For backward compatibility, test the same command on any other switch.

### DIFF
--- a/sonic_installer/bootloader/aboot.py
+++ b/sonic_installer/bootloader/aboot.py
@@ -9,6 +9,7 @@ import re
 import subprocess
 import sys
 import zipfile
+from contextlib import contextmanager
 
 import click
 
@@ -18,7 +19,9 @@ from ..common import (
    HOST_PATH,
    IMAGE_DIR_PREFIX,
    IMAGE_PREFIX,
+   ROOTFS_NAME,
    run_command,
+   run_command_or_raise,
 )
 from .bootloader import Bootloader
 
@@ -34,7 +37,7 @@ def isSecureboot():
     global _secureboot
     if _secureboot is None:
         with open('/proc/cmdline') as f:
-           m  = re.search(r"secure_boot_enable=[y1]", f.read())
+            m = re.search(r"secure_boot_enable=[y1]", f.read())
         _secureboot = bool(m)
     return _secureboot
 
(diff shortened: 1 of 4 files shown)


Staphylo:
@jleveque @qiluo-msft

qiluo-msft:
>            m  = re.search(r"secure_boot_enable=[y1]", f.read())

remove extra blank char

---
Refers to: sonic_installer/bootloader/aboot.py:40 in c0b9b56. [](commit_id = c0b9b56d0d5748c6671314ebc00c12f901be8549, deletion_comment = False)

sonic_installer/bootloader/aboot.py
```diff
@@ -179,3 +182,25 @@ def base64Decode(cls, text):
...
+
+    def _get_swi_file_offset(self, swipath, filename):
+        with zipfile.ZipFile(swipath) as swi:
+            with swi.open(filename) as f:
+                return f._fileobj.tell() # pylint: disable=protected-access
+
+    @contextmanager
+    def get_rootfs_path(self, image_path):
+        rootfs_path = os.path.join(image_path, ROOTFS_NAME)
+        if os.path.exists(rootfs_path):
```
qiluo-msft, line 194: >if os.path.exists(rootfs_path): [](start = 8, length = 31)

Could you just check isSecureboot() ?

sonic_installer/bootloader/aboot.py
```diff
@@ -3,6 +3,7 @@
 """
 
 import base64
+from contextlib import contextmanager
```
jleveque, line 6: Please organize `from <x> import ...` imports after all `import ...` imports in the appropriate group (without a blank line between), and alphabetize them separately. E.g.,

```
import base64
import collections
import os
import re
import subprocess
import sys
import zipfile
from contextlib import contextmanager
```

jleveque (changes requested):

sonic_installer/bootloader/aboot.py
Staphylo: Will do

sonic_installer/bootloader/aboot.py
Staphylo: I could have but there are some future use cases where `isSecureboot` will have to be changed to what I wrote in this review. It was a conscious decision.

qiluo-msft: My point is that the path exists is not the ground truth for secure boot. So if you need to check secure boot, just check secure boot.
Let's assume under secure boot, someone intentionally or unintentionally create rootfs_path folder, is your logic happy?

---
In reply to: [562283015](https://github.com/Azure/sonic-utilities/pull/1366#discussion_r562283015) [](ancestors = 562283015)

Staphylo: I see your point. I'll change this codepath to use `isSecureboot()`

jleveque (approved):

qiluo-msft (dismissed):

Staphylo:
Just rebased to resolve conflict with `echo_and_log` refactor.

qiluo-msft (approved):

jleveque:
Retest this please

Issue #894, opened 2020-04-29
show mac display wrong information for tagged vlan 


Junchao-Mellanox:
**Description**

**Steps to reproduce the issue**
1. Add a tagged vlan
2. Add a vlan member to it
3. Save config and reboot
4. Issue "show mac"

**Describe the results you received**

```
admin@sonic:~$ show mac
Failed to get Vlan id for bvid oid:0x26000000000013

invalid literal for int() with base 16: 'oid:0x26000000000013'
```

**Describe the results you expected**

It should show the correct fdb entry.

**Additional information you deem important (e.g. issue happens only occasionally)**

This is always reproducible. And here is the related data in ASIC db:
```
admin@sonic:~$ redis-cli -n 1
127.0.0.1:6379[1]> keys ASIC_STATE:SAI_OBJECT_TYPE_FDB_ENTRY:*
1) "ASIC_STATE:SAI_OBJECT_TYPE_FDB_ENTRY:{\"bvid\":\"oid:0x26000000000013\",\"mac\":\"EC:0D:9A:C3:B1:00\",\"switch_id\":\"oid:0x21000000000000\"}"
127.0.0.1:6379[1]> keys ASIC_STATE:SAI_OBJECT_TYPE_VLAN:*
1) "ASIC_STATE:SAI_OBJECT_TYPE_VLAN:oid:0x260000000005c1"
2) "ASIC_STATE:SAI_OBJECT_TYPE_VLAN:oid:0x26000000000647"
3) "ASIC_STATE:SAI_OBJECT_TYPE_VLAN:oid:0x26000000000013"
4) "ASIC_STATE:SAI_OBJECT_TYPE_VLAN:oid:0x260000000005c0"
127.0.0.1:6379[1]> hgetall ASIC_STATE:SAI_OBJECT_TYPE_VLAN:oid:0x26000000000013
1) "NULL"
2) "NULL"
```

According to the code in fanshow, I think we might have a few problems:
1. bvid value "oid:0x26000000000013" cannot be converted to an int value, we might need do it like:
```python
int(fdb["bvid"][oid_pfx:], 16)
``` 
2. Even if we convert it as above, we will display the integer value of an oid to the "Vlan" column. Not sure if it is a good idea.
3. I am not sure why there is a vlan entry in ASIC DB whose key/value is "NULL"/"NULL". Maybe is for the default vlan?

**Output of `show version`**

```
SONiC Software Version: SONiC.201911.80-827aab79
[...]

PR #1368, opened 2021-01-19
[show] Fix 'show mac' output, when FDB entry with Vlan 1 is present

maksymbelei95:
* Skip records of FDB entries, which are linked to default Vlan 1,
  to prevent exception throwing while performing
  command 'show mac' or 'fdbshow'.

**- What I did**
Resolves https://github.com/Azure/sonic-utilities/issues/894
Fixed "show mac" command execution failure in case, when the system has an FDB entry, which is linked to default Vlan 1.
The failure is caused by throwing exception, while trying to get int from None type object.
**- How I did it**
The condition has added to src/sonic-utilities/scripts/fdbshow script to handle and skip FDB entries, for which the system can not get Vlan ID.

**- How to verify it**
Configure your system to receive both tagged and untagged traffic. For example, you could use the next steps:
1. Do configuration on DUT
sudo config portchannel add PortChannel0002
sudo config portchannel member add PortChannel0002 Ethernet68
sudo config vlan add 40
sudo config vlan member add 40 PortChannel0002
sudo config interface ip add Vlan40 40.0.0.1/24
2. Do configuration on Linux host
sudo ip link add bond0 type bond
sudo ip link set dev bond0 type bond mode 4
sudo ip link set enp5s0f1 down
sudo ip link set enp5s0f1 master bond0
sudo ip link set enp5s0f1 up
sudo ip link set bond0 up
sudo ip link add link bond0 name bond0.40 type vlan id 40
sudo ip link set bond0.40 up
sudo ip addr add 40.0.0.3/24 dev bond0.40
3. Do ping from linux host to DUT IP 40.0.0.1
4. Do command "show mac" on DUT
"show mac" command should not be finished with the next message: `int() argument must be a string, a bytes-like object or a number, not 'NoneType'`. Instead, the normal output of the command should be shown.

### DIFF
--- a/scripts/fdbshow
+++ b/scripts/fdbshow
@@ -87,6 +87,10 @@ class FdbShow(object):
             elif 'bvid' in fdb:
                 try:
                     vlan_id = port_util.get_vlan_id_from_bvid(self.db, fdb["bvid"])
+                    if vlan_id is None:
+                        # the situation could be faced if the system has an FDB entries,
+                        # which are linked to default Vlan(caused by untagged trafic)
+                        continue
                 except Exception:
                     vlan_id = fdb["bvid"]
                     print("Failed to get Vlan id for bvid {}\n".format(fdb["bvid"]))


maksymbelei95:
Dear @prsunny,

The issue https://github.com/Azure/sonic-buildimage/issues/894 is pretty similar to https://github.com/Azure/sonic-buildimage/issues/6367. Could I ask you to check could we solve the issue with this changes?

prsunny:
Based on your configuration, IMO, the FDB should be on the 'untagged' vlan, i.e Vlan40 as a port can be untagged in only one Vlan. Displaying as "untagged" looks to be incomplete.

maksymbelei95:
Dear @prsunny,

Actually, the system has two different FDB entries for the same MAC address in case with tagged Vlan: entry, linked to Vlan40 and entry with link to default Vlan(I have updated PR overview comment. Now, there are 3 FDB entries, 2 of which belongs to the same MAC address). 

>Displaying as "untagged" looks to be incomplete.
I am sorry, could you clarify what did you mean with this comment? Are you proposing to mark FDB entries with default Vlan somehow like this: 
```
  No.  Vlan          MacAddress         Port             Type
-----  ------------  -----------------  ---------------  -------
    1  (40)Untagged  24:8A:07:EE:4D:E1  PortChannel0002  Dynamic
```

prsunny:
In the first place, how is fdb learnt on a non-configured Vlan? During initialization all ports are removed from default vlan. Secondly, providing the info as "untagged" can have implications. For. e.g, how can user flush this entry as there is no Vlan associated.

maksymbelei95:
@prsunny, I was wrong in my understanding of the situation. In fact, when you link Ethernet port with tagged Vlan, as in my case, the port could either receive tagged and untagged traffic. In tagged case, the traffic routes to the Vlan, but in untagged case the traffic will be automatically assigned to Native Vlan and routed there. For this traffic, the FDB entry with default Vlan is being created(and marked as Untagged in `show mac` with my solution). I could not see any configurational mechanisms for native Vlans in the system and I presume that the untagged traffic in this case is currently being dropped by the system. Could you correct me if it not so?
Maybe, it would be better not to show FDB entries with default Vlan assigned, as the system behave with the traffic as with trash?

Answering on your questions:
> how is fdb learnt on a non-configured Vlan? During initialization all ports are removed from default vlan.:

The system does not create FDB entries till you link the port with a Vlan.
> providing the info as "untagged" can have implications. For. e.g, how can user flush this entry as there is no Vlan associated.

I agree with you. According to my proposition above, maybe, we could skip such FDB entries and not to show them with `show mac`, or, we could mark them not as `Untagged`, but, for example, as `Native `or something similar? But, IMO, skipping would be better.

maksymbelei95:
@prsunny, could I ask you to pay some attention on my last comment?

gechiang:
@maksymbelei95 When You add a port to a vlan as untagged, It means that any untagged packet packets entering that port should be part of that VLAN (Bridge domain).  Any interface can be an "untagged" member of only 1 Bridging domain (VLAN).
For example, you configure a new VLAN 50 and then add  port1 to it as "untagged"  and also add port2 as tagged.
If now you send an untagged packet with unknown destination MAC through port1, it will be flooded to port2 only.  If you add port3 to VLAN 50 and send the same packet, it will now be flooded to both port2 and port3. This is because port1, 2, and 3 are all part of the same "bridging domain" that was configured under VLAN 50.
If you send an unknown tagged packet through port2, it will be flooded through port1 and port3.  When the packet exit out of port1, it should be untagged.
So it does not make sense to have a "show mac" to display "untagged" under the VLAN column.  As @prsunny mentioned, it would not help the user to understand which VLAN (bridging domain) that mac belongs to.  The purpose of show mac is to list out MACs learnt for each of the VLAN.  In my opinion "show mac" output should not care whether it was learnt over an "untagged" or "tagged" interface.  This is because we already have the interface displayed as part of the "show mac" and we can find out if that interface was part of the "tagged"/"untagged" membership of that VLAN already...
Also, if a VLAN is not configured and thus no port can be part of it, there should not be any MAC learnt as there are no "bridging domain" to classify it to.  In fact, SAI Shall not be announcing such MAC learning event.  If it does, then it is a SAI issue that needs to be fixed.

maksymbelei95:
@gechiang, @prsunny, sorry for delayed response. Thank you for explanation. I have updated the PR. Now, FDB entries, linked to default Vlan, just will not be displayed for user.

prsunny (approved):

gechiang (approved):

gechiang:
@maksymbelei95 
Looks like I overlooked one thing... All changes require unit test code to have test coverage over the new changes.
Can you raise a separate PR to include the missing unit test for this?
Thanks!

qiluo-msft:
@maksymbelei95 Could you also fix the 201911 branch? I tried cherry-pick and it does not work, since there are some other code changes happen on master.

PR #1369, opened 2021-01-19
[sonic-installer] Add information to syslog

d-dashkov:
**- What I did**
Add information related to updating firmware image from sonic-installer.
fix https://github.com/Azure/sonic-buildimage/issues/6264

**- How I did it**
Replace click.echo with a function that combines echo and logger functions

**- How to verify it**
Start the sonic-installer process and then grep syslog
$ sonic-installer install http://not_exist.com
$ tail -n 100 /var/log/syslog | grep "sonic-installer"


sonic_installer/main.py
```diff
@@ -256,6 +265,7 @@ def umount_next_image_fs(mount_point):
     except SonicRuntimeException as ex:
         click.secho("Warning: SONiC environment variables are not supported for this image: {0}".format(str(ex)),
                     fg="red", err=True)
+        write_msg_cli_syslog("Warning: SONiC environment variables are not supported for this image: {0}".format(str(ex)), "error")
```
qiluo-msft, line 268: >write_msg_cli_syslog [](start = 8, length = 20)

Will you remove above line? I guess you need to support `fg`, `err`, etc in your new function

sonic_installer/main.py
```diff
@@ -480,12 +490,12 @@ def upgrade_docker(container_name, url, cleanup_image, skip_check, tag, warm):
 
     DEFAULT_IMAGE_PATH = os.path.join("/tmp/", image_name)
     if url.startswith('http://') or url.startswith('https://'):
-        click.echo('Downloading image...')
+        write_msg_cli_syslog('Downloading image...')
         validate_url_or_abort(url)
         try:
-            urllib.urlretrieve(url, DEFAULT_IMAGE_PATH, reporthook)
+            urlretrieve(url, DEFAULT_IMAGE_PATH, reporthook)
```
qiluo-msft, line 496: >urlretrieve [](start = 12, length = 11)

Why change this?

sonic_installer/main.py
```diff
@@ -129,6 +129,14 @@ def get_docker_tag_name(image):
     return tag
 
 
+def write_msg_cli_syslog(msg, msg_type="notice"):
```
qiluo-msft, line 132: >msg_type="notice" [](start = 30, length = 17)

Could you mimic the click.echo paramters, such as `err`

sonic_installer/main.py
```diff
@@ -129,6 +129,14 @@ def get_docker_tag_name(image):
...
+    click.echo(msg)
+    if msg_type == "notice":
+        log.log_notice(msg)
```
qiluo-msft, line 135: Not used, will you remove this branch?

qiluo-msft (commented):
As comments

sonic_installer/main.py
jleveque: Suggest changing name of function to something like `echo_and_log()`

Also, instead of passing a string with the log level, suggest passing constants from sonic_py_common: https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-py-common/sonic_py_common/logger.py#L20-L24

Then you don't need the branches below. You can instead call `log.log()` and pass the parameter. See here: https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-py-common/sonic_py_common/logger.py#L92

d-dashkov:
Dear @ qiluo-msft,
I refactored the function according to your suggestion, added fg support and replaced the log level with constants.
In this case it needs to change its color https://github.com/Azure/sonic-utilities/pull/1369#discussion_r560581783 and output stream,
but in general, it might make sense to use the log function with the flag "also_print_to_console = True" and just print the log message to the cli.
Also returned urllib import, it was an mistake when merging with master.

lguohan (dismissed):

sonic_installer/main.py
```diff
@@ -15,6 +15,8 @@
 from .exception import SonicRuntimeException
 
 SYSLOG_IDENTIFIER = "sonic-installer"
+LOG_ERR = logger.Logger.LOG_PRIORITY_NOTICE
```
qiluo-msft, line 18: >LOG_PRIORITY_NOTICE [](start = 24, length = 19)

LOG_PRIORITY_ERROR ?

sonic_installer/main.py
```diff
@@ -123,6 +125,15 @@ def get_docker_tag_name(image):
     return tag
 
 
+def echo_and_log(msg, priority=LOG_NOTICE, fg='white'):
+    if priority == LOG_ERR:
+        #print to stderr if priority is error
```
qiluo-msft, line 130: >#print [](start = 8, length = 6)

Please follow the blank and capitalization convention in the same file

sonic_installer/main.py
```diff
@@ -123,6 +125,15 @@ def get_docker_tag_name(image):
     return tag
 
 
+def echo_and_log(msg, priority=LOG_NOTICE, fg='white'):
+    if priority == LOG_ERR:
```
qiluo-msft, line 129: >== [](start = 16, length = 2)

Better to use `>=`

sonic_installer/main.py
```diff
@@ -123,6 +125,15 @@ def get_docker_tag_name(image):
     return tag
 
 
+def echo_and_log(msg, priority=LOG_NOTICE, fg='white'):
```
qiluo-msft, line 128: >'white' [](start = 46, length = 7)

default to None? same as click.secho.

qiluo-msft (changes requested):
As comments

qiluo-msft (approved):

jleveque (approved):

PR #1371, opened 2021-01-20
[show]: Use TCP Connection For Muxcable Commands

tahmed-dev:
Unix domain sockets require privileged access while show command
could be run by ro user. Switch muxcable show command to use TCP
connection to work with ro user.

fixes Azure/sonic-buildimage#6456
resolves Azure/sonic-buildimage#6456

**- What I did**
Fix show command for ro users

**- How I did it**
Change use_unix_socket attribute to False

**- How to verify it**
run `show muxcable config` from ro user account

N/A

N/A (same)

### DIFF
--- a/show/muxcable.py
+++ b/show/muxcable.py
@@ -131,7 +131,7 @@ def status(port, json_output):
     namespaces = multi_asic.get_front_end_namespaces()
     for namespace in namespaces:
         asic_id = multi_asic.get_asic_index_from_namespace(namespace)
-        per_npu_statedb[asic_id] = swsscommon.SonicV2Connector(use_unix_socket_path=True, namespace=namespace)
+        per_npu_statedb[asic_id] = swsscommon.SonicV2Connector(use_unix_socket_path=False, namespace=namespace)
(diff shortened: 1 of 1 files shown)


vdahiya12 (approved):

show/muxcable.py
```diff
@@ -131,7 +131,7 @@ def status(port, json_output):
     namespaces = multi_asic.get_front_end_namespaces()
     for namespace in namespaces:
         asic_id = multi_asic.get_asic_index_from_namespace(namespace)
-        per_npu_statedb[asic_id] = swsscommon.SonicV2Connector(use_unix_socket_path=True, namespace=namespace)
+        per_npu_statedb[asic_id] = swsscommon.SonicV2Connector(use_unix_socket_path=False, namespace=namespace)
```
vdahiya12, line 134: there should also be a similar change on line 236

tahmed-dev: Thanks! done.

vdahiya12 (approved):

jleveque (approved):

PR #1382, opened 2021-01-25
Prevent user from adding more then a single untagged VLAN to an interface

deran1980:
Fix for https://github.com/Azure/sonic-buildimage/issues/6421

**- What I did**
User was able to add an interface to multiple VLANs as untagged. Added a validation to a single untagged member.

**- How I did it**
Added a validation when adding a port as untagged to check in DB if is already member as untagged in a different vlan.

**- How to verify it**
Add vlan, add interface as untagged member in this vlan, add another vlan, try to add the same interface as untagged member in the new vlan.

Command was successful.

Fail in VLAN membership.
Error: Ethernet8 is already untagged member!


liat-grozovik (changes requested):
Missing unit test to cover the flow (positive and negative flow)

liat-grozovik (approved):

PR #1384, opened 2021-01-25
Fix portconfig portchannel 201911

ghost:
**- What I did**
Resolves https://github.com/Azure/sonic-buildimage/issues/6430
Added support of PortChannels to portconfig script.

**- How I did it**
Added checking of prefix in name of port, passed to the script. If it is `EthernetXX`, the script uses table name `PORT `to operate with DB, if it is `PortChannelXX`, then the script will use table name `PORTCHANNEL`, otherwise it will throw an exception to notify that port type is wrong.

**- How to verify it**
1. sudo config portchannel add PortChannel0001
2. sudo config interface mtu PortChannel0001 1510
3. see current MTU of the PortChannel with
`ip link show PortChannel0001`

if to try to pass some trash as port argument, like this `sudo config interface mtu Unknown0002 9100`, you will receive error in console output: `Invalid port type specified`

### DIFF
--- a/scripts/portconfig
+++ b/scripts/portconfig
@@ -32,10 +32,12 @@ PORT_SPEED_CONFIG_FIELD_NAME = "speed"
 PORT_FEC_CONFIG_FIELD_NAME = "fec"
 PORT_MTU_CONFIG_FIELD_NAME = "mtu"
 
+
 class portconfig(object):
     """
-    Process aclstat
+    Process portconfig
     """
+
     def __init__(self, verbose, table_name, port, namespace):
         self.verbose = verbose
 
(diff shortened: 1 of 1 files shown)


gechiang (changes requested):
Changes looks good.  Can you add the unit test test cases to handle the scenarios to ensure MTU can be changed for port channels as well as error cases where the prefix of the port/portchannel is not correct and the handler fails the configuration request.
Also, I think this code has other changes already so please rebase to the latest.  I will review and merge once these are done. Thanks!

TafkaMax:
> Changes looks good. Can you add the unit test test cases to handle the scenarios to ensure MTU can be changed for port channels as well as error cases where the prefix of the port/portchannel is not correct and the handler fails the configuration request. Also, I think this code has other changes already so please rebase to the latest. I will review and merge once these are done. Thanks!

What needs to be done here? I would like to this feature to be merged to the community sonic aswell.

TafkaMax:
Currently I had to manually edit `/etc/config/config_db.json` and set MTU to 1500 then `config load`. Like WTF...

PR #1390, opened 2021-01-29
[ecnconfig] Allow ecn unit test to run without sudo

neethajohn:

Allow ecn unit tests to run without root privileges

**- How I did it**
Included the UTILITIES_UNIT_TESTING' env variable also as one of the conditions to determine if the command needs root privileges for execution

**- How to verify it**
Ran utilities test using the command "python3 setup.py test" and ecn_test.py passed. Prior to the fix, most of the testcases were failing with the error 'Root privileged required for this operation'

### DIFF
--- a/scripts/ecnconfig
+++ b/scripts/ecnconfig
@@ -166,7 +166,7 @@ class EcnConfig(object):
         return result
 
     def set_wred_threshold(self, profile, threshold, value):
-        if os.geteuid() != 0:
+        if os.geteuid() != 0 and os.environ.get("UTILITIES_UNIT_TESTING", "0") != "2":
             sys.exit("Root privileges required for this operation")
 
         field = WRED_CONFIG_FIELDS[threshold]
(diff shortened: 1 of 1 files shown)


scripts/ecnconfig
```diff
@@ -166,7 +166,7 @@ class EcnConfig(object):
         return result
 
     def set_wred_threshold(self, profile, threshold, value):
-        if os.geteuid() != 0:
+        if os.geteuid() != 0 and os.environ.get("UTILITIES_UNIT_TESTING", "0") != "2":
```
qiluo-msft, line 169: Extract lines into a function?
```python
        if os.geteuid() != 0 and os.environ.get("UTILITIES_UNIT_TESTING", "0") != "2":
            sys.exit("Root privileges required for this operation")
```

neethajohn: Done

qiluo-msft (approved):

qiluo-msft:
Retest this please

PR #1398, opened 2021-02-03
do not parse stderr to get correct routing stack

lguohan:
**- What I did**
ignore the message printed on stderr

sonic:$ show ip bgp sum
usermod: Permission denied.
usermod: cannot lock /etc/passwd; try again later.
usermod: Permission denied.
usermod: cannot lock /etc/passwd; try again later.

**- How I did it**
ignore stderr

**- How to verify it**
tested on pilot devices manually

### DIFF
--- a/clear/main.py
+++ b/clear/main.py
@@ -78,8 +78,7 @@ def get_routing_stack():
         proc = subprocess.Popen(command,
                                 stdout=subprocess.PIPE,
                                 shell=True,
-                                text=True,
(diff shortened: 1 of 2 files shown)


qiluo-msft (approved):
I am trying to understand the root cause of the stderr message. Does it mean we need to add read only user into `sudo` group?

lguohan:
issue here: https://github.com/Azure/sonic-buildimage/issues/6617

PR #1409, opened 2021-02-08
Fix: initialize SonicDBConfig differently for single or multi_asic

qiluo-msft:
**- What I did**
Fixes https://github.com/Azure/sonic-buildimage/issues/6708

This bug is exposed by https://github.com/Azure/sonic-utilities/pull/1392. Previously the `config` command will call `SonicDBConfig.load_sonic_global_db_config()` even on a single ASIC platform, and it will silently failed. After exposed, it will fail with error syslog message:
```
admin@sonic:~$ sudo config save
Existing files will be overwritten, continue? [y/N]: ^CAborted!

Feb  8 03:23:48.729434 sonic ERR sniffer: :- initializeGlobalConfig: Sonic database config global file doesn't exist at /var/run/redis/sonic-db/database_global.json
```
**- How I did it**

**- How to verify it**
Tested on DUT

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -929,7 +929,11 @@ def config(ctx):
         platform.add_command(mlnx.mlnx)
 
     # Load the global config file database_global.json once.
-    SonicDBConfig.load_sonic_global_db_config()
+    num_asic = multi_asic.get_num_asics()
+    if num_asic > 1:
+        SonicDBConfig.load_sonic_global_db_config()
+    else:
+        SonicDBConfig.initialize()
 
     if os.geteuid() != 0:
         exit("Root privileges are required for this operation")


lguohan:
@qiluo-msft, why are we having test failure? is this the work problem?

lguohan (approved):

lguohan:
retest vs please

lguohan:
retest this please

config/main.py
```diff
@@ -929,7 +929,11 @@ def config(ctx):
         platform.add_command(mlnx.mlnx)
 
     # Load the global config file database_global.json once.
-    SonicDBConfig.load_sonic_global_db_config()
+    num_asic = multi_asic.get_num_asics()
+    if num_asic > 1:
+        SonicDBConfig.load_sonic_global_db_config()
```
judyjoseph, line 934: @Qi, we could call SonicDBConfig.initializeGlobalConfig() as well here ?

qiluo-msft: I translated the names in https://github.com/Azure/sonic-swss-common/blob/9e91e0d891398b468b8087682ae91335791fac51/common/dbconnector.h#L57-L58

judyjoseph: Thanks Qi, I had noticed that .. wondering if you added that because of the "namespace" parameter which is not there in SonicDBConfig.initializeGlobalConfig(), if so we could add that.

judyjoseph (approved):

PR #1414, opened 2021-02-09
[config] Fix config int add incorrect ip

d-dashkov:
**- What I did**
Added checking if the IP address is correct or return error.
fixes https://github.com/Azure/sonic-buildimage/issues/6693
fixes https://github.com/Azure/sonic-buildimage/issues/6776

**- How I did it**
Added function that check correctness of IP address and mask.

**- How to verify it**
```
config interface ip add Ethernet44 1000.000.000.001/24
Error: 'ip_addr' is not valid.

config interface ip add Ethernet40 100.0.0.10/35
Error: ip mask is not valid.

config interface ip add Vlan200 3000::001/200
Error: ip mask is not valid.
```

In some cases it will fix the IP and everything will work fine:
```
config interface ip add Ethernet44 100.000.000.001/24
show runningconfiguration all | grep Ethernet44
"Ethernet44|100.0.0.1/24": {}

config interface ip add Vlan200 3000::001/64
show runningconfiguration all | grep Vlan200
"Vlan200|3001::1/64": {}

config interface ip add Ethernet40 100.0.0.10/024
show runningconfiguration all | grep Ethernet40
"Ethernet40|100.0.0.10/24": {}
```

In case of wrong IP:
```
Error: 'ip_addr' is not valid.
```
In case of wrong mask:
```
Error: ip mask is not valid.
```


config/main.py
```diff
@@ -889,6 +889,25 @@ def validate_mirror_session_config(config_db, session_name, dst_port, src_port,
 
     return True
 
+def validate_ip_address(ctx, ip_addr):
+    mask_range = 32
+    split_ip_mask = ip_addr.split("/")
+
+    try:
+        ip_addr = str(ipaddress.IPv4Address(split_ip_mask[0]))
+    except ipaddress.AddressValueError:
```
prsunny, line 898: This doesn't look good to me (catching as exception for IPv6 address).

config/main.py
```diff
@@ -2656,6 +2675,8 @@ def add(ctx, interface_name, ip_addr, gw):
         if '/' not in ip_addr:
             ip_addr = str(net)
 
+        ip_addr = validate_ip_address(ctx, ip_addr)
```
prsunny, line 2678: Not sure if you want to call this "validate". Since this is for IPv6, why dont we just call if address is v6 and update the leading zeros?

config/main.py
jleveque: I agree. Suggest using `ipaddress.ip_address()` and checking whether the type of the result is `IPv4Address` or `IPv6Address`

config/main.py
d-dashkov: This function is not only for IPv6, it tries to detect the IP type and validate it. According to the [ipaddress doc](https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv4Address), IPv4Address creates an IP address and optionally removes extra zeros if the value is valid, but returns nothing other than an exception if the IP is invalid. Therefore, this is the only way to check the value with this function.

d-dashkov: This feature is not only for IPv6, it checks the range of digits for v4 and v6 IP addresses and additionally removes leading zeros.

config/main.py
d-dashkov: Done

lguohan:
adding unit test for those fixes?

d-dashkov:
@lguohan, ok, I will provide unit tests for this fix

liat-grozovik:
@prsunny  and @jleveque any further comments? it will be nice to have these fixes in 202012 as well. Do you agree to add the label?

config/main.py
liat-grozovik: @d-dashkov it seems to me the validation is only on the mast for ipv4 and ipv6. in this case we can do the following
- rename the function to validate_ip_mask
- add more validation points so it can be generic function.

d-dashkov: I refactored the code, and now all IP checking has been moved to my separate function and the try-except construct has been removed from the add and remove ip interface functions. So now it is a generic function that returns the IP string without leading zeros, or False if something is wrong.

dprital:
@d-dashkov are you about to fix the conflicts so this PR will be able to be approved ?

d-dashkov:
> @d-dashkov are you about to fix the conflicts so this PR will be able to be approved ?

Yep, tomorrow I will rename function to validate_ip_mask as suggested Liat. Will that be enough or do you need something else from my side to resolve conflict

prsunny (approved):

dprital:
Is this PR planned to be cherry picked to 202012 ?

dprital:
@prsunny  - Can you please merge also to 202012 ?

PR #1415, opened 2021-02-09
Import 'mock' module from 'unittest' library

jleveque:
As of Python 3, `mock` is part of `unittest` in the Python standard library. Since sonic-utilities is now Python 3-only, we should import it this way everywhere.

### DIFF
--- a/tests/config_dpb_test.py
+++ b/tests/config_dpb_test.py
@@ -1,8 +1,8 @@
 import json
 import os
 from imp import load_source
+from unittest import mock
 
-import mock
 import pytest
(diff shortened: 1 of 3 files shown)


qiluo-msft:
Why does unit test expose the bug before?

qiluo-msft (approved):

jleveque:
> Why unit test did not catch it before?

I'm not sure. The third-party `mock` library must be getting installed in the environment by some other means. It is technically still [available](https://pypi.org/project/mock/) for Python 3 via PyPI.

PR #1417, opened 2021-02-09
Fix: initialize SonicDBConfig differently for single or multi_asic (continued)

qiluo-msft:
**- What I did**
This bug is exposed by https://github.com/Azure/sonic-utilities/pull/1392. Previously the `config` command will call `SonicDBConfig.load_sonic_global_db_config()` even on a single ASIC platform, and it will silently failed. After exposed, it will fail with error syslog message:
```
Feb  9 05:04:46.462361 vlab-01 ERR python3: :- initializeGlobalConfig: Sonic database config global file doesn't exist at /var/run/redis/sonic-db/database_global.json
```

**- How I did it**

**- How to verify it**

### DIFF
--- a/acl_loader/main.py
+++ b/acl_loader/main.py
@@ -9,7 +9,7 @@ import openconfig_acl
 import tabulate
 import pyangbind.lib.pybindJSON as pybindJSON
 from natsort import natsorted
-from sonic_py_common import device_info
+from sonic_py_common import device_info, multi_asic
 from swsscommon.swsscommon import SonicV2Connector, ConfigDBConnector, SonicDBConfig
 
 
--- a/acl_loader/main.py
+++ b/acl_loader/main.py
@@ -114,8 +114,13 @@ class AclLoader(object):
         self.tables_db_info = {}
         self.rules_db_info = {}
         self.rules_info = {}
-        # Load global db config. This call is no-op in single npu platforms
-        SonicDBConfig.load_sonic_global_db_config()
+        num_asic = multi_asic.get_num_asics()
+        if num_asic > 1:
+            # Load global db config
+            SonicDBConfig.load_sonic_global_db_config()
+        else:
+            SonicDBConfig.initialize()
+
         self.sessions_db_info = {}
         self.configdb = ConfigDBConnector()
         self.configdb.connect()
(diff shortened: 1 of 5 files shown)


lguohan (dismissed):

jleveque (dismissed):

acl_loader/main.py
```diff
@@ -114,8 +114,13 @@ def __init__(self):
         self.tables_db_info = {}
         self.rules_db_info = {}
         self.rules_info = {}
-        # Load global db config. This call is no-op in single npu platforms
-        SonicDBConfig.load_sonic_global_db_config()
+        num_asic = multi_asic.get_num_asics()
```
judyjoseph, line 117: Could use multi_asic.is_multi_asic() check instead.

acl_loader/main.py
```diff
@@ -114,8 +114,13 @@ def __init__(self):
...
+        if num_asic > 1:
+            # Load global db config
+            SonicDBConfig.load_sonic_global_db_config()
+        else:
+            SonicDBConfig.initialize()
```
judyjoseph, line 122: Need not call SonicDBConfig.initialize() explicitly in case of single ASIC platform, as the SonicDBConfig will load the default config automatically. If we load again it will throw an error ( this was the original behavior as well) (Ref: https://github.com/Azure/sonic-swss-common/blob/7afd43d1bf75c95dd1db86a4562d483c0da538fa/common/dbconnector.cpp#L176)

qiluo-msft: It does not harm to explicitly initialize for single ASIC, and I think this is also a better practice than depending on implicit initialization in later function calls.

lguohan (approved):

acl_loader/main.py
judyjoseph: Sure Qi, Agreed !

crm/main.py
```diff
@@ -211,15 +211,20 @@ def cli(ctx):
     # Use the db object if given as input.
     db = None if ctx.obj is None else ctx.obj.cfgdb
 
+    # Note: SonicDBConfig may be already initialized in unit test, then skip
+    if not SonicDBConfig.isInit():
```
judyjoseph, line 215: isInit() is the check for SonicDBConfig.initialize(), good if it could be moved below in else.

judyjoseph (approved):

PR #1421, opened 2021-02-10
[sfputil] Refactor to utilize sonic-platform package

jleveque:
#### What I did

Refactor sfputil to use sonic-platform package in lieu of old, deprecated platform plugins.

The sfputil utility is still useful, as sfpshow only reads and displays SFP data from State DB. However, we still need a utility to write to the transceivers (low power mode, reset, etc.), and the ability to read directly from the transceivers is helpful for debugging.

#### How I did it

- Complete refactor to use sonic-platform package
- Refactor display functions to align with sfpshow utility
- Bump version to 3.0
- Add unit tests

#### How to verify it

Run all `sfputil` commands, ensure they work. Do the same on multiple platforms. Output should be the same as before.

### DIFF
--- a/sfputil/main.py
+++ b/sfputil/main.py
@@ -384,7 +384,7 @@ def get_physical_port_name(logical_port, physical_port, ganged):
     if logical_port == physical_port:
         return logical_port
     elif ganged:
-        return logical_port + ":%d (ganged)" % physical_port
+        return logical_port + ":{} (ganged)".format(physical_port)
     else:
         return logical_port
 
@@ -394,14 +394,14 @@ def logical_port_name_to_physical_port_list(port_name):
         if platform_sfputil.is_logical_port(port_name):
             return platform_sfputil.get_logical_to_physical(port_name)
         else:
-            click.echo("Error: Invalid port '%s'" % port_name)
+            click.echo("Error: Invalid port '{}'".format(port_name))
             return None
     else:
         return [int(port_name)]
 
 
 def print_all_valid_port_values():
-    click.echo("Valid values for port: %s\n" % str(platform_sfputil.logical))
+    click.echo("Valid values for port: {}\n".format(str(platform_sfputil.logical)))
 
 
 # ==================== Methods for initialization ====================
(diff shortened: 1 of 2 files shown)


sfputil/main.py
```diff
@@ -5,118 +5,386 @@
...
+import os
+import sys
 
-    import click
-    from sonic_py_common import device_info, logger, multi_asic
-    from tabulate import tabulate
-except ImportError as e:
-    raise ImportError("%s - required module not found" % str(e))
+import click
+import sonic_platform
```
sanmalho-git, line 13: @jleveque - We tested this file on a multi-asic and single asic box, except we had to change this line to
<pre>
import sonic_platform.platform
</pre>

In common code, we are seeing a lot of places where we have to make the same change as above for python3 to work. Is this the best way to fix this?

jleveque: I have heard other accounts of this. Why do you need to change it? What is the error you encounter with it like this?

sanmalho-git: admin@Linecard2:$ /usr/bin/python3
Python 3.7.3 (default, Jul 25 2020, 13:03:44) 
[GCC 8.3.0] on linux
>>> import sonic_platform
>>> platform_chassis = sonic_platform.platform.Platform().get_chassis()
AttributeError: module 'sonic_platform' has no attribute 'platform'
>>>

jleveque: I don't see this issue on Arista, Dell or Mellanox platforms. Can you see if there is a noticeable difference in the way your sonic-platform packages are structured?

Also, I just pushed a couple new commits. Can you please test the latest changes?

jleveque:
Retest this please

sfputil/main.py
sanmalho-git: @jleveque sorry for the delay.  There was an issue in the packaging of our sonic-platform that we have now resolved and thus don't see the backtrace with the original contents.

Also, I have tested the latest main.py on multi-asic and single asic pizza box, and it works fine.

mlok-nokia:
Should we also change psuutil as well to use the new PMON 2.0 API's?

jleveque:
> Should we also change psuutil as well to use the new PMON 2.0 API's?

I just merged those changes, actually: https://github.com/Azure/sonic-utilities/pull/1434

sfputil/main.py
jleveque: Great to hear! And thanks for testing the changes!

lguohan (approved):

PR #1423, opened 2021-02-11
Fix deprecation warnings

jleveque:
#### What I did

Fix Python deprecation warnings as reported by pytest

#### How I did it

Convert strings which include escape sequences into raw strings

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -640,7 +640,7 @@ def _change_hostname(hostname):
     if current_hostname != hostname:
         clicommon.run_command('echo {} > /etc/hostname'.format(hostname), display_cmd=True)
         clicommon.run_command('hostname -F /etc/hostname', display_cmd=True)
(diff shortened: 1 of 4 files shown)


config/main.py
```diff
@@ -640,7 +640,7 @@ def _change_hostname(hostname):
     if current_hostname != hostname:
         clicommon.run_command('echo {} > /etc/hostname'.format(hostname), display_cmd=True)
         clicommon.run_command('hostname -F /etc/hostname', display_cmd=True)
-        clicommon.run_command('sed -i "/\s{}$/d" /etc/hosts'.format(current_hostname), display_cmd=True)
+        clicommon.run_command(r'sed -i "/\s{}$/d" /etc/hosts'.format(current_hostname), display_cmd=True)
```
qiluo-msft, line 643: Why only this run_command require `r` ?

jleveque: Because there are escape sequences in the string.

qiluo-msft (approved):

PR #1429, opened 2021-02-11
[config] Prevent deleting VLAN with IP addresses.

d-dashkov:
#### What I did
Prevent deleting VLAN if it has IP addresses.
fixes: https://github.com/Azure/sonic-buildimage/issues/6769
Currently, if an IP was added to the vlan, it creates entries in the VLAN_INTERFACE table and increments the reference counter for the vlan:
```
VLAN_INTERFACE|Vlan22|100.0.0.20/24
VLAN_INTERFACE|Vlan22
```
But after deleting the vlan, they get stuck in the table and throw an error:
```
sonic ERR swss#orchagent: :- removeVlan: Failed to remove ref count 1 VLAN Vlan22
```
Therefore, to prevent this error, make sure that the entries do not exist before deleting the VLAN.

#### How I did it
Added check of the VLAN INTERFACE table and if there are records, the delete operation is not available and an error is displayed.

#### How to verify it
1. config vlan add 22
2. config int ip add Vlan22 10.0.0.51/24
3. config vlan del 22
```
Error: Vlan22 can not be removed. First remove IP addresses assigned to this VLAN
```

Signed-off-by: d-dashkov <Dmytro_Dashkov@Jabil.com>

### DIFF
--- a/config/vlan.py
+++ b/config/vlan.py
@@ -48,10 +48,9 @@ def del_vlan(db, vid):
 
     intf_table = db.cfgdb.get_table('VLAN_INTERFACE')
     for intf_key in intf_table:
-        if type(intf_key) is str and intf_key == 'Vlan{}'.format(vid):
-            db.cfgdb.set_entry('VLAN_INTERFACE', intf_key, None)
-        if type(intf_key) is tuple and intf_key[0] == 'Vlan{}'.format(vid):
-            db.cfgdb.set_entry('VLAN_INTERFACE', intf_key, None)
+        if ((type(intf_key) is str and intf_key == 'Vlan{}'.format(vid)) or
+            (type(intf_key) is tuple and intf_key[0] == 'Vlan{}'.format(vid))):
+            ctx.fail("{} can not be removed. First remove IP addresses assigned to this VLAN".format(vlan))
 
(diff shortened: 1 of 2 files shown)


config/vlan.py
```diff
@@ -46,6 +46,13 @@ def del_vlan(db, vid):
     if clicommon.check_if_vlanid_exist(db.cfgdb, vlan) == False:
         ctx.fail("{} does not exist".format(vlan))
 
+    intf_table = db.cfgdb.get_table('VLAN_INTERFACE')
```
prsunny, line 49: One concern is , by removing it in background, user may not be aware that ip address association is removed. Also in the current Sonic image, you can add an ip address to a vlan interface for which vlan is non-existent. 

For e.g: `config interface ip add Vlan200 2.3.4.6/24` <== This would be accepted, even if Vlan200 doesn't exist.

My suggestion would be to log warning to user that Vlan still have IP address and let user remove the ip. What do you think?

d-dashkov: I made an update as per your suggestion to warn the user, but it requires some changes in the tests.
Also is this the correct behavior of adding IP to non-existent entries? There is an issue on same problem with my PR.(https://github.com/Azure/sonic-buildimage/issues/6358).

prsunny (approved):

PR #1431, opened 2021-02-12
[consutil] Look for udevprefix.conf file under platform dir, not plugins

jleveque:
#### What I did

Config files do not belong under `<platform>/plugins/` directory. That directory is intended for (now-deprecated) platform-related Python plugins. The directory will be removed in the near future.

#### How I did it

Look for udevprefix.conf file directly under the respective platform directory instead.
Also clean up and reorganize imports

NOTE: When this submodule is updated in sonic-buildimage, related files will need to be updated. Currently only platform/broadcom/sonic-platform-modules-cel/haliburton/script/udev_prefix.sh

### DIFF
--- a/consutil/lib.py
+++ b/consutil/lib.py
@@ -4,16 +4,14 @@
 # Helper code for CLI for interacting with switches via console device
 #
 
-try:
-    import click
-    import re
-    import subprocess
-    import pexpect
-    import sys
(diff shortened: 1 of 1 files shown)


Blueve (approved):

jleveque:
Retest this please

PR #1440, opened 2021-02-18
[dpb|config] Fix the validation logic of breakout mode

samaity:
Signed-off-by: Sangita Maity <samaity@linkedin.com>

#### What I did
As per latest update in DPB [DOC](https://github.com/Azure/SONiC/pull/749), fixed this [bug](https://github.com/Azure/sonic-buildimage/issues/6645)

previously we had string value in "breakout_modes" key so it was not matching the whole string, But after the update via, now "breakout_modes" contain a dictionary where `key` is the `breakout_mode` and value is the `alias`. So we can easily check whether the key is present or not.

> DEPENDENCY on [Update Dynamic Port Breakout Logic for flexible alias support](https://github.com/Azure/sonic-buildimage/pull/6831) which fixes the [issue](https://github.com/Azure/sonic-buildimage/issues/6024)

#### How I did it
checked whether the user-specified breakout mode key is present under "breakout_modes" in`platform.json`.

#### How to verify it
If you only have 1x100G[50G,40G,25G,10G], the below CLI command should fail:
`config interface breakout Ethernet96 1x100G` as it matches the whole key.

```
admin@lnos-x1-a-csw05:~$ sudo config interface breakout Ethernet40
1x100G[40G]  2x50G        4x25G[10G]

admin@lnos-x1-a-csw05:~$ sudo config interface breakout Ethernet40 2x50G -v -y

Running Breakout Mode : 2x50G
Target Breakout Mode : 2x50G
[WARNING] No action will be taken as current and desired Breakout Mode are same.
```
 **it matches the whole key.**

```
admin@lnos-x1-a-csw05:~$ sudo config interface breakout Ethernet40 4x25G -v -y
[ERROR] Target mode 4x25G is not available for the port Ethernet40
Aborted!
```

### DIFF
--- a/tests/config_dpb_test.py
+++ b/tests/config_dpb_test.py
@@ -21,26 +21,41 @@ breakout_cfg_file_json = {
         "Ethernet0": {
             "index": "1,1,1,1",
             "lanes": "65,66,67,68",
-            "alias_at_lanes": "Eth1/1, Eth1/2, Eth1/3, Eth1/4",
-            "breakout_modes": "1x100G[40G],2x50G,4x25G[10G]"
+            "breakout_modes": {
+                "1x100G[40G]": ["Eth1"],
+                "2x50G": ["Eth1/1", "Eth1/2"],
+                "4x25G[10G]": ["Eth1/1", "Eth1/2", "Eth1/3", "Eth1/4"]
+            }
         },
         "Ethernet4": {
             "index": "2,2,2,2",
             "lanes": "69,70,71,72",
-            "alias_at_lanes": "Eth2/1, Eth2/2, Eth2/3, Eth2/4",
-            "breakout_modes": "1x100G[40G],2x50G,4x25G[10G],1x50G(2)+2x25G(2)"
+            "breakout_modes": {
+                "1x100G[40G]": ["Eth2"],
+                "2x50G": ["Eth2/1", "Eth2/2"],
+                "4x25G[10G]": ["Eth2/1", "Eth2/2", "Eth2/3", "Eth2/4"],
+                "1x50G(2)+2x25G(2)": ["Eth2/1", "Eth2/2", "Eth2/3"]
+            }
         },
         "Ethernet8": {
             "index": "3,3,3,3",
             "lanes": "73,74,75,76",
             "alias_at_lanes": "Eth3/1, Eth3/2, Eth3/3, Eth3/4",
-            "breakout_modes": "1x100G[40G],2x50G,4x25G[10G],1x50G(2)+2x25G(2)"
+            "breakout_modes": {
+                "1x100G[40G]": ["Eth3"],
+                "2x50G": ["Eth3/1", "Eth3/2"],
+                "4x25G[10G]": ["Eth3/1", "Eth3/2", "Eth3/3", "Eth3/4"],
+                "1x50G(2)+2x25G(2)": ["Eth3/1", "Eth3/2", "Eth3/3"]
+            }
         },
         "Ethernet12": {
             "index": "4,4,4,4",
             "lanes": "77,78,79,80",
-            "alias_at_lanes": "Eth4/1, Eth4/2, Eth4/3, Eth4/4",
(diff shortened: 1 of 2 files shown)


zhenggen-xu (dismissed):

tests/config_dpb_test.py
```diff
@@ -21,26 +21,41 @@
...
+                "2x50G": ["Eth2/1", "Eth2/2"],
+                "4x25G[10G]": ["Eth2/1", "Eth2/2", "Eth2/3", "Eth2/4"],
+                "1x50G(2)+2x25G(2)": ["Eth2/1", "Eth2/2", "Eth2/3"]
+            }
         },
         "Ethernet8": {
             "index": "3,3,3,3",
             "lanes": "73,74,75,76",
             "alias_at_lanes": "Eth3/1, Eth3/2, Eth3/3, Eth3/4",
-            "breakout_modes": "1x100G[40G],2x50G,4x25G[10G],1x50G(2)+2x25G(2)"
```
lguohan, line 37: why line 43 is not removed?

zhenggen-xu: good catch, this should be fixed.

samaity: Done! 👍

lguohan:
can you check why unit test is still failling?

tests/config_dpb_test.py
```diff
@@ -21,26 +21,40 @@
...
         },
         "Ethernet12": {
             "index": "4,4,4,4",
             "lanes": "77,78,79,80",
-            "alias_at_lanes": "Eth4/1, Eth4/2, Eth4/3, Eth4/4",
-            "breakout_modes": "1x100G[40G],2x50G,4x25G[10G]"
+            "breakout_modes": {
+                "1x100G[40G]": ["Eth4"],
+                "2x50G": ["Eth4/1", "Eth4/2"],
+                "4x25G[10G]": ["Eth4/1", "Eth1/2", "Eth4/3", "Eth4/4"]
```
lguohan, line 56: this is also wrong, Eth4/2

tests/config_dpb_test.py
```diff
@@ -21,26 +21,40 @@
...
         "Ethernet4": {
             "index": "2,2,2,2",
             "lanes": "69,70,71,72",
-            "alias_at_lanes": "Eth2/1, Eth2/2, Eth2/3, Eth2/4",
-            "breakout_modes": "1x100G[40G],2x50G,4x25G[10G],1x50G(2)+2x25G(2)"
+            "breakout_modes": {
+                "1x100G[40G]": ["Eth2"],
+                "2x50G": ["Eth2/1", "Eth2/2"],
+                "4x25G[10G]": ["Eth2/1", "Eth2/2", "Eth2/3", "Eth2/4"],
+                "1x50G(2)+2x25G(2)": ["Eth2/1", "Eth2/2", "Eth2/3"]
```
lguohan, line 37: this is wrong, Eth2/1, Eth2/3, Eth2/4

tests/config_dpb_test.py
```diff
@@ -21,26 +21,40 @@
...
         "Ethernet8": {
             "index": "3,3,3,3",
             "lanes": "73,74,75,76",
-            "alias_at_lanes": "Eth3/1, Eth3/2, Eth3/3, Eth3/4",
-            "breakout_modes": "1x100G[40G],2x50G,4x25G[10G],1x50G(2)+2x25G(2)"
+            "breakout_modes": {
+                "1x100G[40G]": ["Eth3"],
+                "2x50G": ["Eth3/1", "Eth3/2"],
+                "4x25G[10G]": ["Eth3/1", "Eth3/2", "Eth3/3", "Eth3/4"],
+                "1x50G(2)+2x25G(2)": ["Eth3/1", "Eth3/2", "Eth3/3"]
```
lguohan, line 47: this is also wrong, Eth3/1, Eth3/3, Eth3/4

lguohan (approved):

PR #1416, opened 2021-02-09
[psushow] Add more output columns; Add option to output in JSON format

jleveque:
**- What I did**

- Enhance `psushow -s` output, adding columns to display PSU Model, Serial, Voltage, Current and Power
- Add `-j/--json` option to display output in JSON format
- Update mock State DB PSU_INFO table to provide values which we can ensure floats are padded to two decimal places in tabular display mode
- Add `--json` option to `show platform psustatus` which in turn adds the `-j` flag to the underlying `psushow` call
- Add unit tests for `psushow` in psushow_test.py
- Add more unit tests for `show platform psustatus` and move to show_platform_test.py

**- How to verify it**

Run `psushow -s` variants

**- Previous command output (if the output of a command-line utility has changed)**
```
admin@sonic:~$ psushow -s
PSU    Status    LED
-----  --------  -----
PSU 1  OK        green
PSU 2  OK        green

admin@sonic:~$ psushow -s -i 1
PSU    Status    LED
-----  --------  -----
PSU 1  OK        green

admin@sonic:~$ psushow -s -i 2
PSU    Status    LED
-----  --------  -----
PSU 2  OK        green
```

**- New command output (if the output of a command-line utility has changed)**
```
admin@sonic:~$ psushow -s
PSU    Model    Serial                          Voltage (V)    Current (A)    Power (W)  Status    LED
-----  -------  ----------------------------  -------------  -------------  -----------  --------  -----
PSU 1  0XNMHJ   CN-0XNMHJ-DED00-036-00M3-A00          12.19           8.37       102.10  OK        green
PSU 2  0XNMHJ   CN-0XNMHJ-DED00-036-00H0-A00          12.18          10.07       122.70  OK        green

admin@sonic:~$ psushow -s -i 1
PSU    Model    Serial                          Voltage (V)    Current (A)    Power (W)  Status    LED
-----  -------  ----------------------------  -------------  -------------  -----------  --------  -----
PSU 1  0XNMHJ   CN-0XNMHJ-DED00-036-00M3-A00          12.21           7.25        95.10  OK        green

admin@sonic:~$ psushow -s -i 2
PSU    Model    Serial                          Voltage (V)    Current (A)    Power (W)  Status    LED
-----  -------  ----------------------------  -------------  -------------  -----------  --------  -----
PSU 2  0XNMHJ   CN-0XNMHJ-DED00-036-00H0-A00          12.19           7.74        94.30  OK        green

admin@sonic:~$ psushow -s -j
[
    {
[...]

PR #1448, opened 2021-02-19
[show] cli support for show muxcable cableinfo

vdahiya12:
Summary:
This PR provides the support for adding CLI commands for retrieving cable vendor name and part number information of the muxcable.
In particular these Cli commands are supported:
` show muxcable cableinfo <port>`

### Type of change

- [ ] Bug fix
- [x] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Approach
added the changes in sonic-utilities/show and sonic-utilities/config by changing the muxcable.py 

To add the support for Cli for muxcable to be utilized for retrieving cable vendor name and part number information of the muxcable.

#### How did you do it?
Added the changes inside sonic-utilities and tested it on the testbed

Ran the cli commands on an Arista7050cx3 testbed with Gemini cable

### DIFF
--- a/show/muxcable.py
+++ b/show/muxcable.py
@@ -382,6 +382,7 @@ def eyeinfo(port, target):
     click.echo(tabulate(lane_data, headers=headers))
     sys.exit(EXIT_SUCCESS)
 
+
 @muxcable.command()
 @click.argument('port', required=True, default=None)
 def cableinfo(port):
--- a/show/muxcable.py
+++ b/show/muxcable.py
@@ -381,3 +381,36 @@ def eyeinfo(port, target):
     lane_data.append(res)
     click.echo(tabulate(lane_data, headers=headers))
     sys.exit(EXIT_SUCCESS)
+
+@muxcable.command()
+@click.argument('port', required=True, default=None)
+def cableinfo(port):
+    """Show muxcable cable information"""
+
+    if platform_sfputil is not None:
+        physical_port_list = platform_sfputil_helper.logical_port_name_to_physical_port_list(port)
+
+    if not isinstance(physical_port_list, list):
+        click.echo("ERR: Unable to get a port on muxcable port")
+        sys.exit(EXIT_FAIL)
+        if len(physical_port_list) != 1:
+            click.echo("ERR: Unable to get a single port on muxcable")
+            sys.exit(EXIT_FAIL)
+
+    physical_port = physical_port_list[0]
+    import sonic_y_cable.y_cable
+    res = sonic_y_cable.y_cable.get_pn_number_and_vendor_name(physical_port)
+    if res == False or res == -1:
+        click.echo("ERR: Unable to get cable info")
+        sys.exit(EXIT_FAIL)
+    headers = ['pin_number', 'vendor_name']
+    lane_data = []
+    pin = str(res[0].decode())
+    vendor = str(res[1].decode())
+    temp_list = []
+
+    temp_list.append(pin)
+    temp_list.append(vendor)
+    lane_data.append(temp_list)
+    click.echo(tabulate(lane_data, headers=headers))
+    sys.exit(EXIT_SUCCESS)
--- a/show/muxcable.py
+++ b/show/muxcable.py
@@ -399,18 +399,20 @@ def cableinfo(port):
 
     physical_port = physical_port_list[0]
     import sonic_y_cable.y_cable
-    res = sonic_y_cable.y_cable.get_pn_number_and_vendor_name(physical_port)
-    if res == False or res == -1:
-        click.echo("ERR: Unable to get cable info")
+    part = sonic_y_cable.y_cable.get_part_number(physical_port)
+    if part == False or part == -1:
+        click.echo("ERR: Unable to get cable info part number")
         sys.exit(EXIT_FAIL)
-    headers = ['pin_number', 'vendor_name']
+    vendor = sonic_y_cable.y_cable.get_vendor(physical_port)
+    if vendor == False or vendor == -1:
+        click.echo("ERR: Unable to get cable info vendor name")
+        sys.exit(EXIT_FAIL)
+    headers = ['Vendor', 'Model']
     lane_data = []
-    pin = str(res[0].decode())
-    vendor = str(res[1].decode())
     temp_list = []
 
-    temp_list.append(pin)
     temp_list.append(vendor)
+    temp_list.append(part)
     lane_data.append(temp_list)
     click.echo(tabulate(lane_data, headers=headers))
     sys.exit(EXIT_SUCCESS)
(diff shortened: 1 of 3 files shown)


tests/muxcable_test.py
```diff
@@ -487,6 +490,51 @@ def test_config_muxcable_disable_loopback(self):
...
+    @mock.patch('show.muxcable.platform_sfputil', mock.MagicMock(return_value=1))
+    @mock.patch('utilities_common.platform_sfputil_helper.logical_port_name_to_physical_port_list', mock.MagicMock(return_value=[0]))
+    def test_show_muxcable_cableinfo(self):
+        runner = CliRunner()
+        db = Db()
+
+        result = runner.invoke(show.cli.commands["muxcable"].commands["cableinfo"],
+                               ["Ethernet0"], obj=db)
+
+        assert result.exit_code == 0
```
jleveque, line 503: Suggest also verifying the stdout output. You can use the pytest `capsys` fixture. You can see an example of how I'm using it in [this PR](https://github.com/Azure/sonic-utilities/pull/1416/files#diff-e9018263425d7d7eb008d59325941f6000b66f81f5b2bc411f00a4baacc688a2R61-R73).

jleveque (changes requested):

tests/muxcable_test.py
vdahiya12: fixed.

show/muxcable.py
```diff
@@ -381,3 +381,36 @@ def eyeinfo(port, target):
...
+            click.echo("ERR: Unable to get a single port on muxcable")
+            sys.exit(EXIT_FAIL)
+
+    physical_port = physical_port_list[0]
+    import sonic_y_cable.y_cable
+    res = sonic_y_cable.y_cable.get_pn_number_and_vendor_name(physical_port)
+    if res == False or res == -1:
+        click.echo("ERR: Unable to get cable info")
+        sys.exit(EXIT_FAIL)
+    headers = ['pin_number', 'vendor_name']
```
jleveque, line 406: This is not a pin number, but rather a part number, or model. Also suggest making heading more human-readable, e.g., `headers = ['Vendor', 'Model']`

I also suggest refactoring `get_pn_number_and_vendor_name()` in a future PR to split into two functions, `get_model()` and `get_vendor()`. Maybe we should do this now, before merging this PR?

show/muxcable.py
```diff
@@ -381,3 +381,36 @@ def eyeinfo(port, target):
...
+    lane_data = []
+    pin = str(res[0].decode())
+    vendor = str(res[1].decode())
```
jleveque, line 409: Why do we need `decode()`? The sonic_y_cable function should return strings.

jleveque (changes requested):

show/muxcable.py
vdahiya12: Y cable api `get_pn_number_and_vendor_name()` was returning bytearray for now. This needs to change to a string.
Will have a PR for this.

show/muxcable.py
vdahiya12: Yeah, agree Should call two API's for these. 
will raise another PR for this.

vdahiya12: fixed

show/muxcable.py
vdahiya12: fixed, API returns string now

tests/muxcable_test.py
jleveque: I don't see the change. Did a commit get lost?

vdahiya12: `assert result.output == expected_muxcable_cableinfo_output` on line number 511 checks the output against expected

show/muxcable.py
```diff
@@ -381,3 +381,38 @@ def eyeinfo(port, target):
...
+        click.echo("ERR: Unable to get cable info vendor name")
+        sys.exit(EXIT_FAIL)
+    headers = ['Vendor', 'Model']
+    lane_data = []
+    temp_list = []
+
+    temp_list.append(vendor)
+    temp_list.append(part)
+    lane_data.append(temp_list)
+    click.echo(tabulate(lane_data, headers=headers))
```
jleveque, line 417: `lane_data` is not a good name, as this has nothing to do with lanes. Looks like it was a copy/paste. Also suggest simplifying logic, e.g.,

```python
    body = [vendor, part_num]

    click.echo(tabulate(body, headers=headers))

```

show/muxcable.py
```diff
@@ -381,3 +381,38 @@ def eyeinfo(port, target):
...
+    if not isinstance(physical_port_list, list):
+        click.echo("ERR: Unable to get a port on muxcable port")
+        sys.exit(EXIT_FAIL)
+        if len(physical_port_list) != 1:
+            click.echo("ERR: Unable to get a single port on muxcable")
+            sys.exit(EXIT_FAIL)
+
+    physical_port = physical_port_list[0]
+    import sonic_y_cable.y_cable
+    part = sonic_y_cable.y_cable.get_part_number(physical_port)
```
jleveque, line 402: Suggest renaming from `part` to `part_num`

jleveque (changes requested):

tests/muxcable_test.py
jleveque: Ah, yes. No need for the capsys fixture because this is using Click. :)

show/muxcable.py
```diff
@@ -381,3 +381,38 @@ def eyeinfo(port, target):
...
+        sys.exit(EXIT_FAIL)
+    headers = ['Vendor', 'Model']
+    lane_data = []
+    temp_list = []
+
+    temp_list.append(vendor)
+    temp_list.append(part)
+    lane_data.append(temp_list)
+    click.echo(tabulate(lane_data, headers=headers))
+    sys.exit(EXIT_SUCCESS)
```
jleveque, line 418: Now that we've changed the success code to 0, there's no need to call `sys.exit(EXIT_SUCCESS)` anywhere, because Click will exit 0 by default.

jleveque (changes requested):

show/muxcable.py
vdahiya12: fixed

show/muxcable.py
vdahiya12: fixed

show/muxcable.py
vdahiya12: fixed.

show/muxcable.py
jleveque: Please plan to open a cleanup PR to remove all other `sys.exit(EXIT_SUCCESS)` calls

jleveque (approved):

vdahiya12:
Retest this please

PR #1451, opened 2021-02-20
[pytest] Increase verbosity

jleveque:
#### What I did

Increase verbosity of pytest to allow for full, non-truncated error messages to appear in CI unit test results in order to help facilitate troubleshooting of unit test failures.

#### How I did it

Replace `-v` with `-vv` in pytest options in pytest.ini

#### How to verify it

Ensure unit test failure messages are no longer truncated


jleveque:
Retest this please

PR #1453, opened 2021-02-20
Add soft-reboot reboot type

sujinmkang:
#### What I did
Add a new reboot named as soft-reboot which can be performed by "kexec -e" 

#### How I did it
Replace the platform reboot with "kexec -e" for the cold reboot case.

#### How to verify it
Verified the reboot on DUT and check the reboot-cause

### DIFF
--- a/scripts/soft-reboot
+++ b/scripts/soft-reboot
@@ -206,7 +206,7 @@ if [ -x ${DEVPATH}/${PLATFORM}/${PLATFORM_PLUGIN} ]; then
     ${DEVPATH}/${PLATFORM}/${PLATFORM_PLUGIN}
 fi
 
-# Reboot: explicitly call Linux "kexec -u"
+# Reboot: explicitly call Linux "kexec -e"
 debug "Rebooting with ${REBOOT_METHOD} to ${NEXT_SONIC_IMAGE} ..."
 echo `cat ${REBOOT_CAUSE_FILE}`
 exec ${REBOOT_METHOD}
--- a/scripts/soft-reboot
+++ b/scripts/soft-reboot
@@ -38,6 +38,7 @@ SSD_FW_UPDATE="ssd-fw-upgrade"
 REBOOT_SCRIPT_NAME=$(basename $0)
 REBOOT_TYPE="${REBOOT_SCRIPT_NAME}"
 PLATFORM_PLUGIN="${REBOOT_TYPE}_plugin"
+BOOT_TYPE_ARG="soft"
 TAG_LATEST=yes
 
 function debug()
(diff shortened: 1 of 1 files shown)


scripts/soft-reboot
```diff
@@ -0,0 +1,217 @@
...
+        debug "Stopping syncd process..."
+        docker exec -i syncd /usr/bin/syncd_request_shutdown --cold > /dev/null
+        sleep 3
+    fi
+}
+
+function clear_warm_boot()
+{
+    # If reboot is requested, make sure the outstanding warm-boot is cleared
+    # So the system will come up from a cold boot.
```
yxieca, line 74: you might want to also execute config warm_restart disable here.

scripts/soft-reboot
```diff
@@ -0,0 +1,217 @@
...
+function stop_sonic_services()
+{
+    if [[ x"$ASIC_TYPE" != x"mellanox" ]]; then
+        debug "Stopping syncd process..."
+        docker exec -i syncd /usr/bin/syncd_request_shutdown --cold > /dev/null
+        sleep 3
+    fi
+}
+
+function clear_warm_boot()
```
yxieca, line 71: Function also did kernel unload, please update function name to reflect that. You can name it as clear_lingering_reboot_config

scripts/soft-reboot
```diff
@@ -0,0 +1,217 @@
...
+if [ -x /sbin/hwclock ]; then
+    /sbin/hwclock -w || /bin/true
+fi
+
+# Run platform specific reboot plugin
+if [ -x ${DEVPATH}/${PLATFORM}/${PLATFORM_PLUGIN} ]; then
+    debug "Running ${PLATFORM} specific plugin..."
+    ${DEVPATH}/${PLATFORM}/${PLATFORM_PLUGIN}
+fi
+
```
yxieca, line 208: Do we want to also invoke watchdog here?

sujinmkang: This is a good question. But since this reboot is to prevent from performing the hardware reboot on those specific devices, we would be better to notify the reboot failure and exit this case. 
Or we can add some condition to check the s6100 specific ssd module gpio status before invoking the watchdog. 
What do you think?

sujinmkang: I discussed with Guohan about the watchdog arming before calling platform_plugin and decided to add it here.

sujinmkang:
@yxieca & @jleveque can you please help review the changes again?

yxieca (approved):

jleveque (approved):

antonptashnik:
@sujinmkang Hi. Could you please point me to some soft-reboot mode explanation? I was not able to find any document unfortunately

PR #1457, opened 2021-02-22
[201811] Drop explict 3 seconds pause between two object updates/deletes. 

renukamanavalan:
#### What I did
Drop explicitly introduced 3 seconds pause between object updates using configlet.
Master [PR #1359](https://github.com/Azure/sonic-utilities/pull/1359)

#### How I did it
Remove the added time.sleep(3)

#### How to verify it
Loaded original minigraph that has all T0s
Get redis-dump of Asic-DB.
Manipulate minigraph to remove a T0
Load that minigraph
Add that removed T0 via configlet
Get the red-dump of ASIC-DB

Compared the two ASIC dumps using saiasiccmp tool to match.

### DIFF
--- a/scripts/configlet
+++ b/scripts/configlet
@@ -208,11 +208,6 @@ def main():
             if parse_only == False:
                 for i in data:
                     process_entry (do_update, i)
-                    # Artificial sleep to give a pause between two entries
-                    # so as to ensure that all internal daemons have digested the 
-                    # previous update, before the next one arrives.
-                    #
-                    time.sleep(3)
             else:
                 print("Parsed:")
                 print(data)


qiluo-msft (approved):

yxieca:
Why was it added? Why is no longer needed? Can you clarify the reason?
(Not that I don't like the change, just curious)

renukamanavalan:
> Why was it added? Why is no longer needed? Can you clarify the reason?
> (Not that I don't like the change, just curious)

When we introduced this as new feature (more than a year back), there were some doubts w.r.t quick updates and consumption of the same down the pipeline in control plane. We did not have any concrete doubts. Still added the 3 seconds pause to play it safe.

Now, with help of latest tool saiasiccmp (thanks to Kamil), we did the ASIC dumps comparison across, adding-all-T0s-via-minigraph vs add-one-of-the-T0s via configlet w/o 3 secs pause. This cleared our doubts and hence removed the assumed-but-redundant safety pause of 3 secs.

jleveque:
Retest this please

PR #1460, opened 2021-02-24
[warm-reboot] Check if warm restart flag is set when issuing a warm-reboot

shi-su:
#### What I did
Check if any warm restart flag is set when issuing a warm-reboot. This check avoids starting a warm reboot while another warm restart is in progress. In the scenario where a warm reboot is issued with another warm restart in progress, the warm restart flag may be reset and part of the components have a risk of doing cold reboot. 
Fix: https://github.com/Azure/sonic-buildimage/issues/6509

#### How I did it
Check if any warm restart flag is set when issuing a warm-reboot.

#### How to verify it
Verified that warm-reboot exits when any warm restart flag is set true.

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -400,8 +400,8 @@ function save_counters_folder() {
 
 function check_warm_restart_in_progress() {
     sonic-db-cli STATE_DB keys "WARM_RESTART_ENABLE_TABLE|*" | while read key ; do
-        if [[ "$(sonic-db-cli STATE_DB hget $key enable)" = "true" ]]; then
-            if [[ "${FORCE}" = "yes" ]]; then
+        if [[ x"$(sonic-db-cli STATE_DB hget $key enable)" == x"true" ]]; then
+            if [[ x"${FORCE}" == x"yes" ]]; then
                 debug "Ignoring warm restart flag for ${key#*|}"
             else
                 echo "Warm restart flag for ${key#*|} is set. Please check if a warm restart for ${key#*|} is in progress."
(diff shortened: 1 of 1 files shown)


scripts/fast-reboot
```diff
@@ -419,6 +428,7 @@ case "$REBOOT_TYPE" in
         sonic-db-cli STATE_DB SET "FAST_REBOOT|system" "1" "EX" "180" &>/dev/null
         ;;
     "warm-reboot")
+        check_warm_restart_in_progress
```
vaibhavhd, line 431: Should we also add support for ignoring (forcing through) this check. Just like we have an option (`-f`) to ignore Orchagent RESTARTCHECK, if some components are not reconciled.

shi-su: Thanks for this suggestion. Ignoring the check if `-f` is selected.

vaibhavhd (approved):

scripts/fast-reboot
```diff
@@ -398,6 +398,19 @@ function save_counters_folder() {
     fi
 }
 
+function check_warm_restart_in_progress() {
+    sonic-db-cli STATE_DB keys "WARM_RESTART_ENABLE_TABLE|*" | while read key ; do
+        if [[ "$(sonic-db-cli STATE_DB hget $key enable)" = "true" ]]; then
```
yxieca, line 403: Suggest to use x"$(...)" == x"true"  instead. this is to avoid error when the sonic-db-cli return empty result.

shi-su: Thanks for this comment. Fixed as prescribed.

yxieca (approved):

PR #1463, opened 2021-02-24
Dropped redundant code.

renukamanavalan:
#### What I did
Removed code that had bug, because it is redundant

#### How I did it
The key used is as read from APPL-DB, hence use it as such to get the value,
to help get the interface name, so it can be used for filtering out.


prsunny (approved):
lgtm. May be wait for other reviewers.

qiluo-msft (approved):

PR #1466, opened 2021-02-25
[warm-reboot]: added automated recover for ISSU file

vadymhlushko-mlnx:
Signed-off-by: Vadym Hlushko <vadymh@nvidia.com>

#### What I did
Modified pre-check for file `/host/warmboot/issu_bank.txt` in `warm-reboot` script.
Added automated recovery of  `/host/warmboot/issu_bank.txt`

#### How I did it
Modified the `warm-reboot` script.

#### How to verify it
Need to somehow corrupt the `/host/warmboot/issu_bank.txt` (list below) and then run the `warm-reboot` command.
For testing I have modified the `warm-reboot` script - added corruption command before `check_issu_bank_file` function call.

1. Remove `issu_bank.txt`
2. Clear `issu_bank.txt`
3. Change characters count in `issu_bank.txt`
4. Write invalid content into `issu_bank.txt`

```
root@arc-switch1041:~# warm-reboot
(/host/warmboot/issu_bank.txt) does NOT exist or empty ...
To recover (/host/warmboot/issu_bank.txt) file, do the following:
$ docker exec -it syncd sx_api_dbg_generate_dump.py
$ docker exec -it syncd cat /tmp/sdkdump | grep 'ISSU Bank'
Command above will print the VALUE of ISSU BANK - 0 or 1, use this VALUE in the next command
$ printf VALUE > /host/warmboot/issu_bank.txt
```

```
root@arc-switch1041:~# warm-reboot
(/host/warmboot/issu_bank.txt) does NOT exist or empty ...
Recovering the (/host/warmboot/issu_bank.txt) file
```

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -170,26 +170,21 @@ function request_pre_shutdown()
     }
 }
 
-function recover_issu_bank_file_instruction()
+function recover_issu_bank_file()
 {
-    debug "To recover (${ISSU_BANK_FILE}) file, do the following:"
-    debug "$ docker exec -it syncd sx_api_dbg_generate_dump.py"
-    debug "$ docker exec -it syncd cat /tmp/sdkdump | grep 'ISSU Bank'"
-    debug "Command above will print the VALUE of ISSU BANK - 0 or 1, use this VALUE in the next command"
-    debug "$ printf VALUE > /host/warmboot/issu_bank.txt"
+    debug "Recovering the (${ISSU_BANK_FILE}) file"
+    docker exec -it syncd sx_api_dbg_generate_dump.py
+    issu_bank_value=`docker exec -it syncd cat /tmp/sdkdump | grep 'ISSU Bank' | grep -o -E '[0-9]+'`
+    printf $issu_bank_value > /host/warmboot/issu_bank.txt
 }
 
 function check_issu_bank_file()
 {
     ISSU_BANK_FILE=/host/warmboot/issu_bank.txt
-    MLNX_ISSU_BANK_BROKEN=102
 
     if [[ ! -s "$ISSU_BANK_FILE" ]]; then
         error "(${ISSU_BANK_FILE}) does NOT exist or empty ..."
-        recover_issu_bank_file_instruction
-        if [[ "$1" = true ]]; then
-            exit "${MLNX_ISSU_BANK_BROKEN}"
-        fi
+        recover_issu_bank_file
         return
     fi
 
@@ -199,10 +194,7 @@ function check_issu_bank_file()
     if [[ $issu_file_chars_count != 1 ]] ||
         [[ "$issu_file_content" != "0" && "$issu_file_content" != "1" ]]; then
         error "(${ISSU_BANK_FILE}) is broken ..."
-        recover_issu_bank_file_instruction
-        if [[ "$1" = true ]]; then
-            exit "${MLNX_ISSU_BANK_BROKEN}"
-        fi
+        recover_issu_bank_file
     fi
 }
 
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -173,8 +173,8 @@ function request_pre_shutdown()
 function recover_issu_bank_file()
 {
     debug "Recovering the (${ISSU_BANK_FILE}) file"
-    docker exec -it syncd sx_api_dbg_generate_dump.py
-    issu_bank_value=`docker exec -it syncd cat /tmp/sdkdump | grep 'ISSU Bank' | grep -o -E '[0-9]+'`
+    docker exec -i syncd sx_api_dbg_generate_dump.py
+    issu_bank_value=`docker exec -i syncd cat /tmp/sdkdump | grep 'ISSU Bank' | grep -o -E '[0-9]+'`
     printf $issu_bank_value > /host/warmboot/issu_bank.txt
 }
 
@@ -577,10 +577,8 @@ if [[ "$REBOOT_TYPE" = "warm-reboot" || "$REBOOT_TYPE" = "fastfast-reboot" ]]; t
     # Pre-shutdown syncd
     initialize_pre_shutdown
 
-    BEFORE_PRE_SHUTDOWN=true
-
     if [[ "x$sonic_asic_type" == x"mellanox" ]]; then
-        check_issu_bank_file "$BEFORE_PRE_SHUTDOWN"
+        check_issu_bank_file
     fi
 
     request_pre_shutdown


vadymhlushko-mlnx:
@qiluo-msft, @yxieca, @stepanblyschak  could you please review?

scripts/fast-reboot
```diff
@@ -170,26 +170,21 @@ function request_pre_shutdown()
...
 {
-    debug "To recover (${ISSU_BANK_FILE}) file, do the following:"
-    debug "$ docker exec -it syncd sx_api_dbg_generate_dump.py"
-    debug "$ docker exec -it syncd cat /tmp/sdkdump | grep 'ISSU Bank'"
-    debug "Command above will print the VALUE of ISSU BANK - 0 or 1, use this VALUE in the next command"
-    debug "$ printf VALUE > /host/warmboot/issu_bank.txt"
+    debug "Recovering the (${ISSU_BANK_FILE}) file"
+    docker exec -it syncd sx_api_dbg_generate_dump.py
+    issu_bank_value=`docker exec -it syncd cat /tmp/sdkdump | grep 'ISSU Bank' | grep -o -E '[0-9]+'`
+    printf $issu_bank_value > /host/warmboot/issu_bank.txt
```
qiluo-msft, line 178: There command could fail. For example, if syncd container is stopped, the dump.py will fail.

And in the context of `check_issu_bank_file`, you need to check `$1` and exit conditionally.

qiluo-msft (dismissed):
As comment

scripts/fast-reboot
vadymhlushko-mlnx: @qiluo-msft I guess, if `syncd` crashed this is very critical, and there is no possibility to proceed with the `warm-reboot` or restoration of `issu_bank.txt`. Could you please clarify about it - how you see the way to proceed?

About the conditional exit, we have email thread with @yxieca and @stepanblyschak, where we discuss to not stop the `warm-reboot` if `issu_bank.txt` is broken.

qiluo-msft: This is my understanding of old behavior:
Based on $BEFORE_PRE_SHUTDOWN
1. before, if anything unexpected happens, immediately stop the reboot
2. after, the `pre-shutdown` already happened, and there is no other choice if anything unepxected happens.

I see there is `#!/bin/bash -e` at the start of the file, so you change both behavior to exit immediately, which is a behavior change.

---
In reply to: [583106643](https://github.com/Azure/sonic-utilities/pull/1466#discussion_r583106643) [](ancestors = 583106643)

yxieca: -e is applicable to the shutdown preparation stage. after we paused swss, -e was removed and we have to proceed no matter what.

yxieca: At this time, we've already stopped swss. The 2 checks are before and after preshutdwon respectively. But at this point, we have to proceed with reboot.

yxieca (approved):

scripts/fast-reboot
yxieca: Regarding Qi's concern. I think we did enough check before we reach the point to check ISSU file. So the syncd docker should be running (otherwise swss docker won't be in good state for proceeding).

However, it is a fair question that after preshutdown, would sx_api_dbg_generate_dump.py continue to work?

qiluo-msft: @yxieca Could you explain "-e was removed and we have to proceed no matter what." ? Which line removing it?

---
In reply to: [583862133](https://github.com/Azure/sonic-utilities/pull/1466#discussion_r583862133) [](ancestors = 583862133,583106643)

yxieca: Here: https://github.com/Azure/sonic-utilities/pull/1466/files#diff-ce99c55499028be5fd2d0d3247c051eb453e98725c645c395ed29ef9d9e1ad21R528. After this point, we would have shutdown radv/bgp/swss, we cannot revert these, so the only choice is to finish the reboot. Even it becomes a cold one. Of course we are doing all the we can to keep it warm/fast :-)

qiluo-msft: Thanks, understood.

---
In reply to: [585116888](https://github.com/Azure/sonic-utilities/pull/1466#discussion_r585116888) [](ancestors = 585116888,583862133,583106643)

qiluo-msft:
>         check_issu_bank_file "$BEFORE_PRE_SHUTDOWN"

Not used, you can remove the argument, and also remove the variable.

---
Refers to: scripts/fast-reboot:583 in 1fd720f. [](commit_id = 1fd720f07baa317422602e0ff1c56cd816a1693e, deletion_comment = False)

qiluo-msft (commented):
Unblock, minor issue

scripts/fast-reboot
```diff
@@ -170,26 +170,21 @@ function request_pre_shutdown()
...
-function recover_issu_bank_file_instruction()
+function recover_issu_bank_file()
 {
-    debug "To recover (${ISSU_BANK_FILE}) file, do the following:"
-    debug "$ docker exec -it syncd sx_api_dbg_generate_dump.py"
-    debug "$ docker exec -it syncd cat /tmp/sdkdump | grep 'ISSU Bank'"
-    debug "Command above will print the VALUE of ISSU BANK - 0 or 1, use this VALUE in the next command"
-    debug "$ printf VALUE > /host/warmboot/issu_bank.txt"
+    debug "Recovering the (${ISSU_BANK_FILE}) file"
+    docker exec -it syncd sx_api_dbg_generate_dump.py
```
yxieca, line 176: Can you change -it to -i. I think this change is needed if warm-reboot command was called from an ssh session.

vadymhlushko-mlnx:
> > ```
> >     check_issu_bank_file "$BEFORE_PRE_SHUTDOWN"
> > ```
> 
> Not used, you can remove the argument, and also remove the variable.
> 
> Refers to: scripts/fast-reboot:583 in [1fd720f](https://github.com/Azure/sonic-utilities/commit/1fd720f07baa317422602e0ff1c56cd816a1693e). [](commit_id = [1fd720f](https://github.com/Azure/sonic-utilities/commit/1fd720f07baa317422602e0ff1c56cd816a1693e), deletion_comment = False)

Done

scripts/fast-reboot
vadymhlushko-mlnx: Done

qiluo-msft (approved):

dprital:
@qiluo-msft  - Who can merge it ? we need it also for 202012. Thanks.

prsunny:
@vaibhavhd for viz

PR #1469, opened 2021-02-28
[acl-loader] Add support for matching on ICMP and VLAN info

daall:
- Add ICMP and VLAN fields
- Add new unit test cases

#### What I did
I added support for loading rules that match on VLAN ID and ICMP type/code in acl-loader.

#### How I did it
These fields were added to the model in https://github.com/Azure/sonic-buildimage/pull/6896. This PR extracts those fields so that users can use them in their rules.

#### How to verify it
Updated unit tests. Can also verify locally that existing acl.json files can still be loaded, and those using the new fields add the correct output to config DB.

### DIFF
--- a/acl_loader/main.py
+++ b/acl_loader/main.py
@@ -290,6 +290,14 @@ class AclLoader(object):
         """
         return self.tables_db_info[tname]['type'].upper().startswith(self.ACL_TABLE_TYPE_MIRROR)
 
+    def is_table_ipv6(self, tname):
+        """
+        Check if ACL table type is IPv6 (L3V6 or MIRRORV6)
+        :param tname: ACL table name
+        :return: True if table type is IPv6 else False
+        """
+        return "V6" in self.tables_db_info[tname]["type"].upper()
+
     def is_table_control_plane(self, tname):
         """
         Check if ACL table type is ACL_TABLE_TYPE_CTRLPLANE
@@ -409,9 +417,16 @@ class AclLoader(object):
             else:
                 try:
                     rule_props["ETHER_TYPE"] = int(rule.l2.config.ethertype)
-                except:
-                    raise AclLoaderException("Failed to convert ethertype %s table %s rule %s" % (
+                except Exception:
+                    raise AclLoaderException("Failed to convert ethertype %s; table %s rule %s" % (
                         rule.l2.config.ethertype, table_name, rule_idx))
+
+        if rule.l2.config.vlan_id:
+            try:
+                rule_props["VLAN_ID"] = int(rule.l2.config.vlan_id)
+            except Exception:
+                raise AclLoaderException("Failed to convert VLAN ID %s; table %s rule %s" % (
+                    rule.l2.config.vlan_id, table_name, rule_idx))
 
         return rule_props
 
@@ -450,6 +465,27 @@ class AclLoader(object):
         if self.is_table_mirror(table_name):
             if rule.ip.config.dscp:
                 rule_props["DSCP"] = rule.ip.config.dscp
+
+        return rule_props
+
+    def convert_icmp(self, table_name, rule_idx, rule):
+        rule_props = {}
+
+        is_table_v6 = self.is_table_ipv6(table_name)
+        type_key = "ICMPV6_TYPE" if is_table_v6 else "ICMP_TYPE"
+        code_key = "ICMPV6_CODE" if is_table_v6 else "ICMP_CODE"
+
+        if rule.icmp.config.type:
+            try:
+                rule_props[type_key] = int(rule.icmp.config.type)
+            except Exception:
+                raise AclLoaderException("Failed to convert %s; table %s, rule %s" % (type_key, table_name, rule_idx))
+
+        if rule.icmp.config.code:
+            try:
+                rule_props[code_key] = int(rule.icmp.config.code)
+            except Exception:
+                raise AclLoaderException("Failed to convert %s; table %s, rule %s" % (code_key, table_name, rule_idx))
 
         return rule_props
 
@@ -527,6 +563,7 @@ class AclLoader(object):
         deep_update(rule_props, self.convert_action(table_name, rule_idx, rule))
         deep_update(rule_props, self.convert_l2(table_name, rule_idx, rule))
         deep_update(rule_props, self.convert_ip(table_name, rule_idx, rule))
+        deep_update(rule_props, self.convert_icmp(table_name, rule_idx, rule))
         deep_update(rule_props, self.convert_transport(table_name, rule_idx, rule))
         deep_update(rule_props, self.convert_input_interface(table_name, rule_idx, rule))
 
--- a/acl_loader/main.py
+++ b/acl_loader/main.py
@@ -482,7 +482,7 @@ class AclLoader(object):
         type_key = "ICMPV6_TYPE" if is_table_v6 else "ICMP_TYPE"
         code_key = "ICMPV6_CODE" if is_table_v6 else "ICMP_CODE"
 
-        if rule.icmp.config.type:
+        if rule.icmp.config.type or rule.icmp.config.type == 0:
             try:
                 icmp_type = int(rule.icmp.config.type)
 
@@ -495,7 +495,7 @@ class AclLoader(object):
                     "Failed to convert %s; table %s, rule %s; exception=%s" %
                     (type_key, table_name, rule_idx, str(e)))
 
-        if rule.icmp.config.code:
+        if rule.icmp.config.code or rule.icmp.config.code == 0:
             try:
                 icmp_code = int(rule.icmp.config.code)
 
--- a/acl_loader/main.py
+++ b/acl_loader/main.py
@@ -423,18 +423,13 @@ class AclLoader(object):
                         "Failed to convert ethertype %s; table %s rule %s; exception=%s" %
                         (rule.l2.config.ethertype, table_name, rule_idx, str(e)))
 
-        if rule.l2.config.vlan_id:
-            try:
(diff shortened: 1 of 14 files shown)


daall:
depends on https://sonic-jenkins.westus2.cloudapp.azure.com/job/vs/job/buildimage-vs-all/596/console

acl_loader/main.py
```diff
@@ -409,10 +417,17 @@ def convert_l2(self, table_name, rule_idx, rule):
             else:
                 try:
                     rule_props["ETHER_TYPE"] = int(rule.l2.config.ethertype)
-                except:
-                    raise AclLoaderException("Failed to convert ethertype %s table %s rule %s" % (
+                except Exception:
```
qiluo-msft, line 420: >Exception [](start = 23, length = 9)

Not your fault. It ignores the original exception, we need at least log it or print to stderr.

acl_loader/main.py
```diff
@@ -409,10 +417,17 @@ def convert_l2(self, table_name, rule_idx, rule):
...
+                    raise AclLoaderException("Failed to convert ethertype %s; table %s rule %s" % (
                         rule.l2.config.ethertype, table_name, rule_idx))
 
+        if rule.l2.config.vlan_id:
```
qiluo-msft, line 424: >vlan_id [](start = 26, length = 7)

Do we need to validate input for:
1. vlan_id<=0
2. valn_id too large

acl_loader/main.py
```diff
@@ -409,10 +417,17 @@ def convert_l2(self, table_name, rule_idx, rule):
...
+            try:
+                rule_props["VLAN_ID"] = int(rule.l2.config.vlan_id)
+            except Exception:
```
qiluo-msft, line 427: >Exception [](start = 19, length = 9)

It ignores the original exception, we need at least log it or print to stderr.

acl_loader/main.py
daall: Orchagent does this, but I think it's probably a good idea to check here too (it'll be more obvious to the user!).

acl_loader/main.py
```diff
@@ -453,6 +468,27 @@ def convert_ip(self, table_name, rule_idx, rule):
...
         return rule_props
 
+    def convert_icmp(self, table_name, rule_idx, rule):
+        rule_props = {}
+
+        is_table_v6 = self.is_table_ipv6(table_name)
+        type_key = "ICMPV6_TYPE" if is_table_v6 else "ICMP_TYPE"
+        code_key = "ICMPV6_CODE" if is_table_v6 else "ICMP_CODE"
+
+        if rule.icmp.config.type:
```
qiluo-msft, line 478: >type [](start = 28, length = 4)

Is 0 a valid type?

acl_loader/main.py
```diff
@@ -290,6 +290,14 @@ def is_table_mirror(self, tname):
         """
         return self.tables_db_info[tname]['type'].upper().startswith(self.ACL_TABLE_TYPE_MIRROR)
 
+    def is_table_ipv6(self, tname):
+        """
+        Check if ACL table type is IPv6 (L3V6 or MIRRORV6)
+        :param tname: ACL table name
+        :return: True if table type is IPv6 else False
+        """
+        return "V6" in self.tables_db_info[tname]["type"].upper()
```
qiluo-msft, line 299: >"V6" [](start = 15, length = 4)

Determine from table name seems not robust. Is there a better way?

tests/acl_input/acl1.json
```diff
@@ -141,6 +141,57 @@
...
+								"ip": {
+									"config": {
+										"protocol": "IP_ICMP",
+										"source-ip-address": "::1/128",
+										"destination-ip-address": "::1/128"
+									}
+								},
+								"icmp": {
+									"config": {
+										"type": "4",
```
qiluo-msft, line 166: >type [](start = 11, length = 4)

How did you test?

acl_loader/main.py
daall: This is based on the table type, not the name.

acl_loader/main.py
daall: I'll add this to the test cases but "0" should work. openconfig uses 0 as a default value, so checking `== 0` will actually add ICMP_TYPE = 0 to all rules.

tests/acl_input/acl1.json
daall: The tests try to load various files, with `acl1.json` being a valid file. If the change is correct, this file should be successfully loaded and we should count 6 rules, as others have done previously.

acl_loader/main.py
qiluo-msft: If my understanding is correct, the table types are from
[https://github.com/Azure/sonic-buildimage/blob/95ec75e24ed2c95186c9ab8367bc17dfadc0d44b/src/sonic-config-engine/minigraph.py#L450-L455](https://github.com/Azure/sonic-buildimage/blob/95ec75e24ed2c95186c9ab8367bc17dfadc0d44b/src/sonic-config-engine/minigraph.py#L450-L455)

Is it better to check `L3V6` or ends with `V6` to determine icmp/icmpv6?

---
In reply to: [584383685](https://github.com/Azure/sonic-utilities/pull/1469#discussion_r584383685) [](ancestors = 584383685)

tests/acl_input/acl1.json
qiluo-msft: Agree you add check for `6`. Better to check more deeper details.

---
In reply to: [584383894](https://github.com/Azure/sonic-utilities/pull/1469#discussion_r584383894) [](ancestors = 584383894)

acl_loader/main.py
daall: Ack, I updated it to check for a specific list of table types (in case we add some new table in the future that is IPv6 but does not have V6 as a suffix).

acl_loader/main.py
daall: Update: this was a very good catch! The sonic-acl-extension defines ICMP as a uint, so we have this default value == 0 issue. I need to patch it to treat ICMP type and code as strings so that we can use `0` as a valid code/type.

acl_loader/main.py
daall: Done

acl_loader/main.py
daall: done

acl_loader/main.py
daall: added!

qiluo-msft (approved):

daall:
retest default please

daall:
retest this please

acl_loader/main.py
```diff
@@ -453,6 +481,41 @@ def convert_ip(self, table_name, rule_idx, rule):
...
         return rule_props
 
+    def convert_icmp(self, table_name, rule_idx, rule):
+        rule_props = {}
+
+        is_table_v6 = self.is_table_ipv6(table_name)
+        type_key = "ICMPV6_TYPE" if is_table_v6 else "ICMP_TYPE"
+        code_key = "ICMPV6_CODE" if is_table_v6 else "ICMP_CODE"
+
+        if rule.icmp.config.type or rule.icmp.config.type == 0:
```
qiluo-msft, line 491: >rule.icmp.config.type [](start = 11, length = 21)

Is it possible that rule.icmp.config.type == "null"?

acl_loader/main.py
```diff
@@ -453,6 +481,41 @@ def convert_ip(self, table_name, rule_idx, rule):
...
+            try:
+                icmp_type = int(rule.icmp.config.type)
```
qiluo-msft, line 493: >rule.icmp.config.type [](start = 32, length = 21)

Is it possible rule.icmp.config.type is an integer? if yes, you don't need cast to int.

acl_loader/main.py
```diff
@@ -422,7 +445,12 @@ def convert_ip(self, table_name, rule_idx, rule):
         # so there isn't currently a good way to check if the user defined proto=0 or not.
         if rule.ip.config.protocol:
             if rule.ip.config.protocol in self.ip_protocol_map:
-                rule_props["IP_PROTOCOL"] = self.ip_protocol_map[rule.ip.config.protocol]
+                # Special case: "1" is the wrong protocol number for ICMPV6, we need to use "58" instead if
```
qiluo-msft, line 448: >Special case: "1" is the wrong protocol number for ICMPV6 [](start = 18, length = 57)

Do you mean input acl data could be wrong?
Is it better to exit with error instead of correcting it?

acl_loader/main.py
daall: "null" is a string so it will give an error when we try to cast it. The user shouldn't ever be passing null explicitly, we just have the value defined in the yang model so that the user is able to pass 0 as a valid ICMP type/code.

acl_loader/main.py
daall: No. ICMP has a different protocol number for IPv4 and IPv6. Since user passes a string describing the protocol they want to match, we need to do a conversion. I thought this was slightly cleaner than having an entire separate case for `rule.ip.config.protocol` == "IP_ICMP"

acl_loader/main.py
qiluo-msft: Understood. To make code more readable, you may check `rule.icmp.config.type == "null"` first, if not equal, `rule.icmp.config.type` must be an integer, no need to cast.

---
In reply to: [586069013](https://github.com/Azure/sonic-utilities/pull/1469#discussion_r586069013) [](ancestors = 586069013)

daall: OK, that sounds good to me.

qiluo-msft (approved):

PR #1470, opened 2021-03-01
[Mellanox] Change mellanox buffer migrator for new SKU Mellanox-SN2700-D40C8S8

Junchao-Mellanox:
**- What I did**

Change mellanox buffer migrator for new SKU Mellanox-SN2700-D40C8S8

**- How I did it**

New SKU Mellanox-SN2700-D40C8S8 will reuse buffer configuration of Mellanox-SN2700-D48C8

**- How to verify it**

Run sonic-mgmt qos test and pass


stephenxs (approved):

qiluo-msft (approved):

neethajohn (commented):
Have a comment on Azure/sonic-buildimage#6926. The mock config db tables probably will need to be updated if that holds good

stephenxs:
> Have a comment on [Azure/sonic-buildimage#6926](https://github.com/Azure/sonic-buildimage/pull/6926). The mock config db tables probably will need to be updated if that holds good

I’ll look into it a bit more. But we don’t need to change the db tables because the interfaces in config db is to hold the various lossless buffer profiles in buffer pg table, rather than the interfaces themselves.

> Have a comment on [Azure/sonic-buildimage#6926](https://github.com/Azure/sonic-buildimage/pull/6926). The mock config db tables probably will need to be updated if that holds good

Hi
I’ll look into it a bit more. But I think we don’t need to change the db tables because the interfaces in config db is to hold the various lossless buffer profiles in buffer pg table, rather than the interfaces themselves.

neethajohn:
> > Have a comment on [Azure/sonic-buildimage#6926](https://github.com/Azure/sonic-buildimage/pull/6926). The mock config db tables probably will need to be updated if that holds good
> 
> I’ll look into it a bit more. But we don’t need to change the db tables because the interfaces in config db is to hold the various lossless buffer profiles in buffer pg table, rather than the interfaces themselves.
> 
> > Have a comment on [Azure/sonic-buildimage#6926](https://github.com/Azure/sonic-buildimage/pull/6926). The mock config db tables probably will need to be updated if that holds good
> 
> Hi
> I’ll look into it a bit more. But I think we don’t need to change the db tables because the interfaces in config db is to hold the various lossless buffer profiles in buffer pg table, rather than the interfaces themselves.

Wouldn't the interfaces in the BUFFER_PG table change?

stephenxs:
> > Hi
> > I’ll look into it a bit more. But I think we don’t need to change the db tables because the interfaces in config db is to hold the various lossless buffer profiles in buffer pg table, rather than the interfaces themselves.
> 
> Wouldn't the interfaces in the BUFFER_PG table change?

Hi @neethajohn ,
Not really.
The DB migrator has a generic logic to check whether the buffer configuration aligns with the default one. For `BUFFER_PG` table, it checks whether:
- the PGs are default ones: PG 0 for lossy and PG 3-4 for lossless
- the `profile` aligns with the port's speed and cable length

But it doesn't check whether the port configuration, including its speed, cable length, etc., aligns with the default value according to the SKU configuration. In addition, we configured the interfaces with different (`speed`, `cable length`) tuple in order to verify all possible compositions in `pg_profile_lookup.ini`.
To make it simple, we have some config DB files for different SKUs sharing the same port configuration.

neethajohn (approved):

abdosi:
@Junchao-Mellanox  Please create PR for 201911.

@qiluo-msft

Junchao-Mellanox:
Hi @abdosi and @qiluo-msft , I suppose this PR is not necessary for 201911. @stephenxs, please correct me if i am wrong.

stephenxs:
> Hi @abdosi and @qiluo-msft , I suppose this PR is not necessary for 201911. @stephenxs, please correct me if i am wrong.

You are correct.
@abdosi @qiluo-msft @liat-grozovik 
We don't need to back-port this PR to 201911 because:
 - The motivation of this PR is to support upgrading from older versions to the 202012/master for D40C8S8
 - There isn't a flow in which the device is upgraded from a version older than 201911 to 201911 since 201911 is the first version supporting D40C8S8
Hence, we don't need to back-port it to 201911.

PR #1473, opened 2021-03-03
[config]: Create portchannel with LACP key

DavidZagury:
This PR depends on:

- https://github.com/Azure/sonic-utilities/pull/1566
- https://github.com/Azure/sonic-swss/pull/1660

#### What I did
Fix issue - https://github.com/Azure/sonic-buildimage/issues/4009
Change the LACP key to be generated from the Port Channel name instead of always being 0.
When upgrading without warm-reboot update old port channels to use the new default.

#### How I did it
When adding a new port-channel add by default the key lacp_key with the value 'auto' to the port-channel table, this is done to change the port-channel LACP key to be generated from the Port Channel name instead of always being 0.

When upgrading without warm-reboot, also update old port channels to use the new default.
This is not done on warm-reboot to avoid the link from going down.

#### How to verify it
Create a new port-channel, add a port and run tcpdump on this port.
LACP Key should be the number in the end of the port channel name with leading 1.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1500,7 +1500,8 @@ def add_portchannel(ctx, portchannel_name, min_links, fallback):
         ctx.fail("{} already exists!".format(portchannel_name))
 
     fvs = {'admin_status': 'up',
-           'mtu': '9100'}
+           'mtu': '9100',
+           'lacp_key': 'auto'}
     if min_links != 0:
         fvs['min_links'] = str(min_links)
     if fallback != 'false':
--- a/scripts/db_migrator.py
+++ b/scripts/db_migrator.py
@@ -44,7 +44,7 @@ class DBMigrator():
                      none-zero values.
               build: sequentially increase within a minor version domain.
         """
-        self.CURRENT_VERSION = 'version_2_0_0'
+        self.CURRENT_VERSION = 'version_2_0_2'
 
         self.TABLE_NAME      = 'VERSIONS'
         self.TABLE_KEY       = 'DATABASE'
@@ -525,9 +525,24 @@ class DBMigrator():
 
     def version_2_0_1(self):
         """
-        Current latest version. Nothing to do here.
+        Version 2_0_1.
         """
         log.log_info('Handling version_2_0_1')
+        warmreboot_state = self.stateDB.get(self.stateDB.STATE_DB, 'WARM_RESTART_ENABLE_TABLE|system', 'enable')
+
+        if warmreboot_state != 'true':
+            portchannel_table = self.configDB.get_table('PORTCHANNEL')
+            for name, data in portchannel_table.items():
+                data['lacp_key'] = 'auto'
+                self.configDB.set_entry('PORTCHANNEL', name, data)
+        self.set_version('version_2_0_2')
+        return 'version_2_0_2'
+
+    def version_2_0_2(self):
+        """
+        Current latest version. Nothing to do here.
+        """
+        log.log_info('Handling version_2_0_2')
         return None
 
     def get_version(self):
--- a/tests/db_migrator_input/config_db/portchannel-expected.json
+++ b/tests/db_migrator_input/config_db/portchannel-expected.json
@@ -1,0 +1,39 @@
+{
+    "PORTCHANNEL|PortChannel0": {
+        "admin_status": "up",
+        "members@": "Ethernet0,Ethernet4",
+        "min_links": "2",
+        "mtu": "9100",
+        "lacp_key": "auto"
+    },
+    "PORTCHANNEL|PortChannel1": {
+        "admin_status": "up",
+        "members@": "Ethernet8,Ethernet12",
+        "min_links": "2",
+        "mtu": "9100",
+        "lacp_key": "auto"
+    },
+    "PORTCHANNEL|PortChannel0123": {
+        "admin_status": "up",
+        "members@": "Ethernet16",
+        "min_links": "1",
+        "mtu": "9100",
+        "lacp_key": "auto"
+    },
+    "PORTCHANNEL|PortChannel0011": {
+        "admin_status": "up",
+        "members@": "Ethernet20,Ethernet24",
+        "min_links": "2",
+        "mtu": "9100",
+        "lacp_key": "auto"
+    },
+    "PORTCHANNEL|PortChannel9999": {
+        "admin_status": "up",
+        "mtu": "9100",
+        "lacp_key": "auto"
+    },
+    "VERSIONS|DATABASE": {
+        "VERSION": "version_2_0_2"
+    }
+} 
+
--- a/tests/db_migrator_input/config_db/portchannel-input.json
+++ b/tests/db_migrator_input/config_db/portchannel-input.json
@@ -1,0 +1,33 @@
+{
+    "PORTCHANNEL|PortChannel0": {
+        "admin_status": "up",
+        "members@": "Ethernet0,Ethernet4",
+        "min_links": "2",
+        "mtu": "9100"
+    },
+    "PORTCHANNEL|PortChannel1": {
+        "admin_status": "up",
+        "members@": "Ethernet8,Ethernet12",
+        "min_links": "2",
+        "mtu": "9100"
+    },
+    "PORTCHANNEL|PortChannel0123": {
+        "admin_status": "up",
+        "members@": "Ethernet16",
+        "min_links": "1",
+        "mtu": "9100"
+    },
+    "PORTCHANNEL|PortChannel0011": {
+        "admin_status": "up",
+        "members@": "Ethernet20,Ethernet24",
+        "min_links": "2",
+        "mtu": "9100"
+    },
+    "PORTCHANNEL|PortChannel9999": {
+        "admin_status": "up",
+        "mtu": "9100"
+    },
+    "VERSIONS|DATABASE": {
+        "VERSION": "version_2_0_1"
+    }
+} 
--- a/tests/db_migrator_test.py
+++ b/tests/db_migrator_test.py
@@ -247,3 +247,26 @@ class TestInitConfigMigrator(object):
         assert not diff
 
         assert not expected_db.cfgdb.get_table('CONTAINER_FEATURE')
+
+
+class TestLacpKeyMigrator(object):
+    @classmethod
+    def setup_class(cls):
+        os.environ['UTILITIES_UNIT_TESTING'] = "2"
+
+    @classmethod
+    def teardown_class(cls):
+        os.environ['UTILITIES_UNIT_TESTING'] = "0"
+        dbconnector.dedicated_dbs['CONFIG_DB'] = None
+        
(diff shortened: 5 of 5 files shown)


config/main.py
```diff
@@ -1375,7 +1375,7 @@ def add_portchannel(ctx, portchannel_name, min_links, fallback):
         ctx.fail("{} already exists!".format(portchannel_name))
 
     fvs = {'admin_status': 'up',
-           'mtu': '9100'}
+           'mtu': '9100', 'lacp_key': 'auto'}
```
lguohan, line 1378: can you add unit test for this?

liat-grozovik:
retest this please

scripts/db_migrator.py
```diff
@@ -473,7 +473,22 @@ def version_2_0_0(self):
         Current latest version. Nothing to do here.
         """
         log.log_info('Handling version_2_0_0')
+        warmreboot_state = self.stateDB.get(self.stateDB.STATE_DB, 'WARM_RESTART_ENABLE_TABLE|system', 'enable')
+
+        if warmreboot_state != 'true':
+            portchannel_table = self.configDB.get_table('PORTCHANNEL')
+            for name, data in portchannel_table.items():
+                data['lacp_key'] = 'auto'
```
judyjoseph, line 481: Shouldn't we do a configDB.get_entry() first for that key and update the 'lacp_key' alone as a new field and then set_entry(). Otherwise the other valid fields already present would be removed.

scripts/db_migrator.py
```diff
@@ -473,7 +473,22 @@ def version_2_0_0(self):
         Current latest version. Nothing to do here.
```
judyjoseph, line 473: Remove this comment from here.

scripts/db_migrator.py
```diff
@@ -473,7 +473,22 @@ def version_2_0_0(self):
...
+        if warmreboot_state != 'true':
+            portchannel_table = self.configDB.get_table('PORTCHANNEL')
+            for name, data in portchannel_table.items():
+                data['lacp_key'] = 'auto'
+                self.configDB.set_entry('PORTCHANNEL', name, data)
+            self.set_version('version_2_0_1')
+
+        return 'version_2_0_1'
 
+    def version_2_0_1(self):
```
judyjoseph, line 487: Adding @yxieca too to take a look at this change/approach to increment version in case of db_migrator. Thanks Ying.

scripts/db_migrator.py
prsunny: @judyjoseph , why can't we treat `lacp_key` to `auto `if not specified?.

DavidZagury: We kept the old behavior if lacp_key is not set to make sure that if we are upgrading from an older version and doing warm reboot the lacp key won't change because otherwise it will cause a drop in the connection.

scripts/db_migrator.py
DavidZagury: Fixed

config/main.py
DavidZagury: @lguohan  what tests do you believe should be for this? it only adding another key to the DB.
There are already tests on the sonic-swss PR that check that adding this key will behave as expected.
What should I tests in the sonic-utilities side?

scripts/db_migrator.py
DavidZagury: @judyjoseph we are already getting the keys information when we are calling the portchannel_table.items() and retrieving name and data.
The other field are present in the data variable and there is no need for calling another function to retrieve them.

liat-grozovik:
@yaxia and @judyjoseph any further comments or we are good to cont? tests are also passing :-)

config/main.py
yxieca: Expand the existing test to check 'lacp_key' exists after newly generated or version upgraded?

judyjoseph:
> @yaxia and @judyjoseph any further comments or we are good to cont? tests are also passing :-)

Sry, I missed responding earlier - @DavidZagury could you take care of Ying's comment to verify the config DB indeed has the lacp_key present in case of cold boot ? thanks

DavidZagury:
> > @yaxia and @judyjoseph any further comments or we are good to cont? tests are also passing :-)
> 
> Sry, I missed responding earlier - @DavidZagury could you take care of Ying's comment to verify the config DB indeed has the lacp_key present in case of cold boot ? thanks

I have added a tests that check the db_migrator has added an lacp_key to the db

liat-grozovik:
@yxieca  and @judyjoseph please review following new changes.

tests/db_migrator_input/config_db/portchannel-expected.json
```diff
@@ -0,0 +1,39 @@
...
+        "mtu": "9100",
+        "lacp_key": "auto"
+    },
+    "PORTCHANNEL|PortChannel9999": {
+        "admin_status": "up",
+        "mtu": "9100",
+        "lacp_key": "auto"
+    },
+    "VERSIONS|DATABASE": {
+        "VERSION": "version_2_0_1"
```
stephenxs, line 36:
```suggestion
        "VERSION": "version_2_0_2"
```

tests/db_migrator_input/config_db/portchannel-input.json
```diff
@@ -0,0 +1,33 @@
...
+        "members@": "Ethernet20,Ethernet24",
+        "min_links": "2",
+        "mtu": "9100"
+    },
+    "PORTCHANNEL|PortChannel9999": {
+        "admin_status": "up",
+        "mtu": "9100"
+    },
+    "VERSIONS|DATABASE": {
+        "VERSION": "version_2_0_0"
```
stephenxs, line 31:
```suggestion
        "VERSION": "version_2_0_1"
```

tests/db_migrator_test.py
```diff
@@ -215,3 +215,24 @@ def test_port_autoneg_migrator(self):
...
+        dbconnector.dedicated_dbs['CONFIG_DB'] = None
+
+    def test_lacp_key_migrator(self):
+        dbconnector.dedicated_dbs['CONFIG_DB'] = os.path.join(mock_db_path, 'config_db', 'portchannel-input')
+        import db_migrator
+        dbmgtr = db_migrator.DBMigrator(None)
+        dbmgtr.migrate()
+        dbconnector.dedicated_dbs['CONFIG_DB'] = os.path.join(mock_db_path, 'config_db', 'portchannel-expected')
+        expected_db = Db()
+
```
stephenxs, line 236:
```suggestion

        advance_version_for_expected_database(dbmgtr.configDB, expected_db.cfgdb, 'version_2_0_2')

```
This is to make sure the test won't fail if there are db migrators coming in the future with higher version

stephenxs:
Depends on Azure/sonic-utilities#1614

liat-grozovik:
now that the dependent PR is merged. @yxieca and @judyjoseph can you please review / approve?

liat-grozovik:
kindly reminder @yxieca and @judyjoseph can you please review / approve?

judyjoseph:
@DavidZagury, could you resolve the conflicts, before we can merge in. thanks.

tests/db_migrator_test.py
```diff
@@ -247,3 +247,33 @@ def test_init_config_feature_migration(self):
         assert not diff
 
         assert not expected_db.cfgdb.get_table('CONTAINER_FEATURE')
+
+
+class TestLacpKeyMigrator(object):
```
judyjoseph, line 252: good to add the setup_class/teardown_class class methods.

tests/db_migrator_test.py
```diff
@@ -247,3 +247,33 @@ def test_init_config_feature_migration(self):
...
+    def test_init_config_feature_migration(self):
```
judyjoseph, line 253: do we really need this function in class TestLacpKeyMigrator

DavidZagury: Fixed

tests/db_migrator_test.py
DavidZagury: Fixed

DavidZagury:
@BasimShalata my PR failed here with the same tests as your PR: https://github.com/Azure/sonic-utilities/pull/1657

lguohan:
can you add better description for the pr? for example, can you add description for the db_migrator change?

scripts/db_migrator.py
```diff
@@ -525,9 +525,24 @@ def version_2_0_0(self):
 
     def version_2_0_1(self):
         """
-        Current latest version. Nothing to do here.
+        Version 2_0_1.
         """
         log.log_info('Handling version_2_0_1')
+        warmreboot_state = self.stateDB.get(self.stateDB.STATE_DB, 'WARM_RESTART_ENABLE_TABLE|system', 'enable')
+
+        if warmreboot_state != 'true':
```
lguohan, line 533: what is the behavior of lacp key for warm reboot?

DavidZagury: In case that the upgrade was done using warm-reboot we are not updating the DB.
This is done because changing the lacp key of the existing port channels will cause the link to go down which we can't do during warm reboot.

liat-grozovik:
@lguohan and @tahmed-dev can you please review latest comments? if all is OK please approve so we can merge.

config/main.py
```diff
@@ -1500,7 +1500,8 @@ def add_portchannel(ctx, portchannel_name, min_links, fallback):
         ctx.fail("{} already exists!".format(portchannel_name))
 
     fvs = {'admin_status': 'up',
-           'mtu': '9100'}
+           'mtu': '9100',
+           'lacp_key': 'auto'}
```
tahmed-dev, line 1504: can we read the `lacp_key` from the `cli` in case one would like to give specific key instead of auto generating it?

DavidZagury: lacp_key attribute was not exposed to the user at all and was always set to zero.
in the scope of fixing this issue, we are now setting the lacp_key to a unique ID (xxxx) derived from the Portchannelxxxx.
and this is what the auto mode is doing.

I don't think adding this CLI is part of the scope of this bug fix, it is another feature that we could provide, but we need to fix the bug that started this first (https://github.com/Azure/sonic-buildimage/issues/4009)
This issue was marked as closed, but until this PR will be merged we still haven't fixed the bug, and it still an issue in the system.
If needed we will create a new PR with this improvement.

tahmed-dev (approved):

judyjoseph (approved):

liat-grozovik:
@DavidZagury can you please confirm that all comments were handled include the test? If so, we can ask @lguohan  to review and approve

DavidZagury:
> @DavidZagury can you please confirm that all comments were handled include the test? If so, we can ask @lguohan to review and approve
@lguohan I have handled all the comments and added the checks in the tests that were requested

lguohan (approved):

PR #1475, opened 2021-03-03
[acl] Expand VLAN into VLAN members when creating an ACL table

daall:

#### What I did
I added an option to `config acl add table` to allow users to expand VLANs into their constituent VLAN members when specifying the list of ports for the table.

#### How I did it
I added `expand_vlans` which will signal to the CLI to check which ports are members of the given VLAN and add those to the port binding list. The reason for this is that some platforms may not support binding to the VLAN directly, so it may be helpful to have a convenient option to refer to the Vlan without actually trying to bind the ACL table to the VLAN.

#### How to verify it
Added new unit tests.


doc/Command-Reference.md
```diff
@@ -1371,6 +1371,33 @@ When the optional argument "max_priority"  is specified, each rule’s priority
 
 Go Back To [Beginning of the document](#) or [Beginning of this section](#acl)
 
+**config acl add table**
+
+This command is used to create new ACL tables.
+
+- Usage:
+  ```
+  config acl add table [OPTIONS] <table_name> <table_type> [-d <description>] [-p <ports>] [-s (ingress | egress)] [--expand_vlans]
```
qiluo-msft, line 1380: >--expand_vlans [](start = 116, length = 14)

`expand_vlans` should be an implementation detail, not a CLI option. User could not specify this option freely in current image.
This is similar to minigraph.py.

doc/Command-Reference.md
```diff
@@ -1371,6 +1371,33 @@ When the optional argument "max_priority"  is specified, each rule’s priority
...
+  ```
+
+- Parameters:
+  - table_name: The name of the ACL table to create.
+  - table_type: The type of ACL table to create (e.g. "L3", "L3V6", "MIRROR")
+  - description: A description of the table for the user. (default is the table_name)
+  - ports: A comma-separated list of ports to add to the table. (default is all front-panel ports)
```
qiluo-msft, line 1387: >A comma-separated list of ports to add to the table. (default is all front-panel ports) [](start = 11, length = 87)

Since it support ports, interfaces (single port, portchannel or vlan). Let's make the usage/document clear.
Not easy to describe it in the `[-p <ports>]`, it's okay to keep it as-is.

config/main.py
```diff
@@ -3369,37 +3369,67 @@ def get_acl_bound_ports():
...
+
+    If the provided interface is a VLAN, then this method will return its member ports.
+
+    If the provided interface is not a VLAN, then this method will return a list with only
+    the provided interface in it.
     """
     config_db = ConfigDBConnector()
     config_db.connect()
 
+    vlan_members = config_db.get_keys("VLAN_MEMBER")
```
qiluo-msft, line 3385: >vlan_members [](start = 4, length = 12)

Did you test an empty vlan? an non-existing vlan?

config/main.py
```diff
@@ -3369,37 +3369,67 @@ def get_acl_bound_ports():
...
+
+    members = [v[1] for v in vlan_members if port_name == v[0]]
```
qiluo-msft, line 3387: >v [](start = 24, length = 1)

Maybe more readable:
`for vlan, member in vlan_members`

config/main.py
```diff
@@ -3369,37 +3369,67 @@ def get_acl_bound_ports():
...
+    return members or [port_name]
```
qiluo-msft, line 3388: >[port_name] [](start = 22, length = 11)

Seem you will return `[vlan]` if it is empty. Is it expected?

config/main.py
```diff
@@ -3369,37 +3369,67 @@ def get_acl_bound_ports():
...
-    if ports:
+    if ports and expand_vlans:
+        port_list = list()
+        for port in ports.split(","):
+            port_list += expand_vlan_ports(port)
+        table_info["ports@"] = ",".join(set(port_list))
+    elif ports:
         table_info["ports@"] = ports
     else:
         table_info["ports@"] = ",".join(get_acl_bound_ports())
```
qiluo-msft, line 3407: >",".join(get_acl_bound_ports()) [](start = 31, length = 31)

ConfigDBConector could hanlde the translation between list to string. Maybe it's easier to reuse its functions.

doc/Command-Reference.md
daall: Actually, I guess we don't have that option available in orchagent at this time, so it might be better to handle this internally and document the behavior in the CLI. Thoughts?

qiluo-msft: Agree with "handle this internally and document the behavior in the command reference".

config/main.py
daall: I'm not sure if we can since we have to expand each interface and then de-duplicate the results and put them together.

config/main.py
daall: Nope, good catch. Fixed!

config/main.py
daall: Added.

doc/Command-Reference.md
daall: Let me know if the updated description is clear enough

qiluo-msft (approved):

PR #1477, opened 2021-03-04
[201911][acl] Expand VLAN into VLAN members when creating an ACL table

daall:

#### What I did
Ported #1475 to 201911.

#### How I did it
Moved files from tests to sonic-utilities-tests.

#### How to verify it
Run unit tests + run ACL tests against 201911 image with this change and verify there is no regression.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -2552,6 +2552,62 @@ def get_acl_bound_ports():
 
     return list(ports)
 
+
+def expand_vlan_ports(port_name):
+    """
+    Expands a given VLAN interface into its member ports.
+
+    If the provided interface is a VLAN, then this method will return its member ports.
+
+    If the provided interface is not a VLAN, then this method will return a list with only
+    the provided interface in it.
+    """
+    config_db = ConfigDBConnector()
+    config_db.connect()
+
+    if port_name not in config_db.get_keys("VLAN"):
+        return [port_name]
+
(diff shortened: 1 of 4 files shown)


qiluo-msft (approved):

qiluo-msft:
Why LGTM did not catch the same issue on master PR?

daall:
> Why LGTM did not catch the same issue on master PR?

Great question! I am not sure but I can put in a fix against master for it as well.

daall:
> > Why LGTM did not catch the same issue on master PR?
> 
> Great question! I am not sure but I can put in a fix against master for it as well.

@qiluo we get an import error if we follow the linter's advice. I am wondering if this is some new thing that is allowed in python3 but not python3? I will do some digging, but we need to leave the import as is for now.

qiluo-msft:
The lgtm is somehow correct. It looks confusing to human, what is the meaning of `config`?
```
import config.main as config
from config.main import expand_vlan_ports, parse_acl_table_info
```

daall:
> The lgtm is somehow correct. It looks confusing to human, what is the meaning of `config`?
> 
> ```
> import config.main as config
> from config.main import expand_vlan_ports, parse_acl_table_info
> ```

It's the "config" cli command. I think I can probably import things in a more structured way to 1) make it less confusing and 2) fix the LGTM warning. Let me play around with it some more.

qiluo-msft (approved):

PR #1479, opened 2021-03-04
[acl-loader] Improve input validation for acl_loader

daall:

#### What I did
I fixed two issues:
1) The IPv6 block rule was only looking at the table name, not the table type, when determining what type of default deny rule to add. This isn't very robust.

2) Individual fields are validated, but there are some invalid combinations like mixing certain protocol numbers like TCP (6) with fields that don't make sense for TCP (like ICMP type/code fields).

#### How I did it
1) Used the new `is_table_ipv6` method to verify that the table is a V6 dataacl table.
2) Added a post-check once all qualifiers are validated to ensure that they make sense when put together.

#### How to verify it
Added new test cases, verified that the existing ACL and CACL tests have no regression.

### DIFF
--- a/acl_loader/main.py
+++ b/acl_loader/main.py
@@ -559,6 +559,19 @@ class AclLoader(object):
 
         return rule_props
 
+    def validate_rule_fields(self, rule_props):
+        protocol = rule_props.get("IP_PROTOCOL", 6)
+        if "TCP_FLAGS" in rule_props and protocol != 6:
+            raise AclLoaderException("IP_PROTOCOL={} is not TCP, but TCP flags were provided".format(protocol))
+
+        protocol = rule_props.get("IP_PROTOCOL", 1)
+        if ("ICMP_TYPE" in rule_props or "ICMP_CODE" in rule_props) and protocol != 1:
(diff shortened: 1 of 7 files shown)


acl_loader/main.py
```diff
@@ -559,6 +559,19 @@ def convert_input_interface(self, table_name, rule_idx, rule):
 
         return rule_props
 
+    def validate_rule_fields(self, rule_props):
+        protocol = rule_props.get("IP_PROTOCOL", 6)
```
qiluo-msft, line 563: >6 [](start = 49, length = 1)

I think the code should be okay, I just comment on readability.
At first glance, changing the default value for the same field is confusing. If the field is missing, you just skip all the check.
Suggest get the field value only once.

qiluo-msft (commented):
As comment

daall:
retest this please

qiluo-msft (approved):

abdosi:
@daall please create PR for 201911.

sonic-util test  is very different between master and 201911

@qiluo-msft

daall:
already done: https://github.com/Azure/sonic-utilities/pull/1481

PR #1486, opened 2021-03-10
[reboot] User-friendly reboot cause message for kernel panic

yozhao101:
Signed-off-by: Yong Zhao <yozhao@microsoft.com>

#### What I did
If the rebooting of SONiC device was caused by kernel panic, then the CLI command `show reboot-cause` should show `Kernel Panic`.

#### How I did it
Currently if kernel was panicked, then the device would be rebooted. The `reboot` script wrote a message into `reboot-cause.txt`. I just updated the content of this message.

#### How to verify it
I verified this change on the virtual switch in the following steps:
1. Trigger kernel panic: `echo c > /proc/sysrq-trigger`
2. After device was rebooted, run the CLI `show reboot-cause`:
`admin@vlab-01:~$ show reboot-cause`
`Kernel Panic [Time: Tue 09 Mar 2021 03:03:56 AM UTC]`

`admin@vlab-01:~$ show reboot-cause`
`User issued 'kdump' command [User: kdump, Time: Mon 08 Mar 2021 01:47:43 AM UTC]`

`admin@vlab-01:~$ show reboot-cause`
`Kernel Panic [Time: Tue 09 Mar 2021 03:03:56 AM UTC]`

### DIFF
--- a/scripts/reboot
+++ b/scripts/reboot
@@ -9,7 +9,7 @@ REBOOT_TIME=$(date)
 VMCORE_FILE=/proc/vmcore
 if [ -e $VMCORE_FILE -a -s $VMCORE_FILE ]; then
         echo "We have a /proc/vmcore, then we just kdump'ed"
-        echo "User issued 'kdump' command [User: kdump, Time: ${REBOOT_TIME}]" > ${REBOOT_CAUSE_FILE}
+        echo "Kernel Panic (Time: ${REBOOT_TIME})" > ${REBOOT_CAUSE_FILE}
         sync
         PLATFORM=$(grep -oP 'sonic_platform=\K\S+' /proc/cmdline)
         if [ ! -z "${PLATFORM}" -a -x ${DEVPATH}/${PLATFORM}/${PLAT_REBOOT} ]; then
--- a/scripts/reboot
+++ b/scripts/reboot
@@ -9,7 +9,7 @@ REBOOT_TIME=$(date)
 VMCORE_FILE=/proc/vmcore
 if [ -e $VMCORE_FILE -a -s $VMCORE_FILE ]; then
         echo "We have a /proc/vmcore, then we just kdump'ed"
-        echo "Kernel Panic (Time: ${REBOOT_TIME})" > ${REBOOT_CAUSE_FILE}
+        echo "Kernel Panic [Time: ${REBOOT_TIME}]" > ${REBOOT_CAUSE_FILE}
         sync
         PLATFORM=$(grep -oP 'sonic_platform=\K\S+' /proc/cmdline)
         if [ ! -z "${PLATFORM}" -a -x ${DEVPATH}/${PLATFORM}/${PLAT_REBOOT} ]; then
--- a/show/reboot_cause.py
+++ b/show/reboot_cause.py
@@ -8,19 +8,19 @@ from swsscommon.swsscommon import SonicV2Connector
 import utilities_common.cli as clicommon
 
 
-PREVIOUS_REBOOT_CAUSE_FILE = "/host/reboot-cause/previous-reboot-cause.json"
+PREVIOUS_REBOOT_CAUSE_FILE_PATH = "/host/reboot-cause/previous-reboot-cause.json"
 
 def read_reboot_cause_file():
-    reboot_cause = None
+    reboot_cause_dict = {}
 
-    if os.path.exists(PREVIOUS_REBOOT_CAUSE_FILE):
-        with open(PREVIOUS_REBOOT_CAUSE_FILE) as file_handler:
+    if os.path.exists(PREVIOUS_REBOOT_CAUSE_FILE_PATH):
+        with open(PREVIOUS_REBOOT_CAUSE_FILE_PATH) as prev_reboot_cause_file:
             try:
-                reboot_cause = json.load(file_handler)
+                reboot_cause_dict = json.load(prev_reboot_cause_file)
             except json.JSONDecodeError as err:
-                click.echo("Failed to load JSON file '{}'!".format(PREVIOUS_REBOOT_CAUSE_FILE))
+                click.echo("Failed to load JSON file '{}'!".format(PREVIOUS_REBOOT_CAUSE_FILE_PATH))
 
-    return reboot_cause
+    return reboot_cause_dict
 
 #
 # 'reboot-cause' group ("show reboot-cause")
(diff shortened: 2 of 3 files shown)


jleveque:
@sujinmkang: We want to rephrase this message to be more user-friendly. However, we may need to either adjust the syntax here or in the `process-reboot-cause` parsing logic to parse the fields properly. Please provide your feedback.

sujinmkang:
@jleveque does this PR also wants to add the detail  kernel panic information or just inform the kernel crash as last reboot cause?

jleveque:
> @jleveque does this PR also wants to add the detail kernel panic information or just inform the kernel crash as last reboot cause?

Just inform that kernel panic was the last reboot cause.

sujinmkang:
@jleveque If the kernel panic happens after passing this line, I mean, if it happens during reboot, then there is still having a chance to miss the kernel panic? I think it will be good to have the core file directory or file name which is related to the kernel panic. We can add it from process-reboot-cause.

jleveque:
> @jleveque If the kernel panic happens after passing this line, I mean, if it happens during reboot, then there is still having a chance to miss the kernel panic? I think it will be good to have the core file directory or file name which is related to the kernel panic. We can add it from process-reboot-cause.

I see your point, but if a kernel panic occurs during reboot, it will be difficult (possibly impossible, e.g., if the filesystem is read-only) to leave a breadcrumb that we can use after booting back up to determine that a kernel panic occurred.

Also, technically speaking, if a user issues one of the `reboot` commands and a kernel panic happens during the reboot process, what was the cause of the reboot? Technically the reboot *was* initiated by the user.

sujinmkang:
https://github.com/Azure/sonic-buildimage/blob/28cb43cb42e3223cade2efa9a5f60542d97a89e7/src/sonic-host-services/scripts/determine-reboot-cause#L129
@yozhao101 please add the kernel panic case if you want to change the reboot-cause here.

sujinmkang:
And this too : https://github.com/Azure/sonic-utilities/blob/3a482ac38f1440fa0cbc00f621dc94ebb551757b/show/reboot_cause.py#L26

show/reboot_cause.py
```diff
@@ -30,7 +30,10 @@ def reboot_cause(ctx):
         # Read the previous reboot cause
         data = read_reboot_cause_file()
         if data['user'] == "N/A":
-            reboot_cause = "{}".format(data['cause'])
+            if data['cause'] == "Kernel Panic":
+                reboot_cause = "{} (Time: {})".format(data['cause'], data['time'])
+            else: 
+                reboot_cause = "{}".format(data['cause'])
         else:
             reboot_cause = USER_ISSUED_REBOOT_CAUSE_REGEX.format(data['cause'], data['user'], data['time'])
```
jleveque, line 38: Maybe we can make this more generic such that it appends user and/or time to the output if they are not 'N/A'?

yozhao101: Agree and great suggestion. Will update.

show/reboot_cause.py
```diff
@@ -9,14 +9,18 @@
 
 
 PREVIOUS_REBOOT_CAUSE_FILE = "/host/reboot-cause/previous-reboot-cause.json"
```
jleveque, line 11: Suggest renaming to `PREVIOUS_REBOOT_CAUSE_FILE_PATH`

show/reboot_cause.py
```diff
@@ -9,14 +9,18 @@
...
 
 def read_reboot_cause_file():
-    result = ""
+    reboot_cause = None
+
     if os.path.exists(PREVIOUS_REBOOT_CAUSE_FILE):
-        with open(PREVIOUS_REBOOT_CAUSE_FILE) as f:
-            result = json.load(f)
-    return result
+        with open(PREVIOUS_REBOOT_CAUSE_FILE) as file_handler:
```
jleveque, line 17: Suggest renaming from `file_handler` to `prev_reboot_cause_file`

show/reboot_cause.py
```diff
@@ -26,18 +30,21 @@ def read_reboot_cause_file():
...
-        if data['user'] == "N/A":
-            if data['cause'] == "Kernel Panic":
-                reboot_cause = "{} (Time: {})".format(data['cause'], data['time'])
-            else: 
-                reboot_cause = "{}".format(data['cause'])
-        else:
-            reboot_cause = USER_ISSUED_REBOOT_CAUSE_REGEX.format(data['cause'], data['user'], data['time'])
+        reboot_cause_str = ""
+
+        # Read the previous reboot reason
```
jleveque: s/reason/cause/

show/reboot_cause.py
```diff
@@ -26,18 +30,21 @@ def read_reboot_cause_file():
...
+        reboot_cause = read_reboot_cause_file()
+
+        if reboot_cause and "cause" in reboot_cause.keys():
+            reboot_cause_str = "Cause: {}".format(reboot_cause["cause"])
+
+            if "user" in reboot_cause.keys() and reboot_cause["user"] != "N/A":
+                reboot_cause_str += ", User: {}".format(reboot_cause["user"])
+
+            if "time" in reboot_cause.keys() and reboot_cause["time"] != "N/A":
+                reboot_cause_str += ", Time: {}".format(reboot_cause["time"])
```
jleveque: No need for "Cause:` prefix. Also, we lose the "User issued '<command>' command" syntax here. Instead we just see "<command>".

Suggest keeping format the same as before.

jleveque (changes requested):

scripts/reboot
```diff
@@ -9,7 +9,7 @@ REBOOT_TIME=$(date)
 VMCORE_FILE=/proc/vmcore
 if [ -e $VMCORE_FILE -a -s $VMCORE_FILE ]; then
         echo "We have a /proc/vmcore, then we just kdump'ed"
-        echo "User issued 'kdump' command [User: kdump, Time: ${REBOOT_TIME}]" > ${REBOOT_CAUSE_FILE}
+        echo "Kernel Panic (Time: ${REBOOT_TIME})" > ${REBOOT_CAUSE_FILE}
```
jleveque, line 12: Suggest changing parentheses to square brackets to match the message format produced by the CLI.

jleveque (changes requested):

jleveque (approved):

show/reboot_cause.py
```diff
@@ -8,15 +8,19 @@
...
-            result = json.load(f)
-    return result
+    reboot_cause_dict = {}
+
+    if os.path.exists(PREVIOUS_REBOOT_CAUSE_FILE_PATH):
+        with open(PREVIOUS_REBOOT_CAUSE_FILE_PATH) as prev_reboot_cause_file:
+            try:
+                reboot_cause_dict = json.load(prev_reboot_cause_file)
+            except json.JSONDecodeError as err:
+                click.echo("Failed to load JSON file '{}'!".format(PREVIOUS_REBOOT_CAUSE_FILE_PATH))
```
yozhao101, line 21: @jleveque If we can not read the JSON file successfully, then CLI `show reboot-cause` will output:

    Failed to load JSON file /host/reboot-cause/previous-reboot-cause.json!
    Unknow

I am wondering whether it is appropriate?

jleveque: Under what circumstances are you seeing this? `/host/reboot-cause/previous-reboot-cause.json` should be written by the determine-reboot-cause service which is one of the first services run after boot.

yozhao101: If the service `determine-reboot-cause` can not write the reboot reason into the file `/host/reboot-cause/previous-reboot-cause.json` successfully, then this JSON file will be an empty file. If we run this CLI `show reboot-cause`, this exception will be raised.

jleveque: Under what circumstances will `determine-reboot-cause` fail to write the file?

yozhao101: One possible reason I think is the process `determine-reboot-cause` opened the file, but it crashed before it can write previous reboot reason into the file.

jleveque: In that case, we would need to fix a bug in `determine-reboot-cause`. I think `Unknown` is the proper thing to display in this case, and outputting an error would be helpful to understand why. Maybe we should simply send the error message to stderr:

```python
click.echo("Failed to load JSON file '{}'!".format(PREVIOUS_REBOOT_CAUSE_FILE_PATH), err=True)
```

jleveque (approved):

jleveque (approved):

yozhao101:
Retest this please.

PR #1488, opened 2021-03-10
Add script to block/unblock ICMP traffic on a given ACL table [experimental; do not merge]

daall:

### DIFF
--- a/scripts/set_icmp_state
+++ b/scripts/set_icmp_state
@@ -87,7 +87,7 @@ def verify_valid_ip_address(ip_address, table_ip_version):
         table_ip_version = 6 if table_ip_version == IP_VERSION_V6 else 4
         if destination_ip_version != table_ip_version:
             error("Cannot use an IPv{} rule with an IPv{} table, exiting...".format(destination_ip_version, table_ip_version))
-    except Exception as e:
+    except ValueError as e:
         error("Could not parse {} as a valid IP address; exception={}".format(ip_address, e))
 
 
--- a/scripts/set_icmp_state
+++ b/scripts/set_icmp_state
@@ -1,0 +1,293 @@
+#!/usr/bin/env python
+"""
+Utility for blocking and unblocking ICMP traffic on ACL tables.
+
+The block operation will insert a DENY rule at the top of the table. The unblock operation
+will remove an existing DENY rule that has been created by the block operation (i.e. it does
+NOT insert an ALLOW rule, only removes DENY rules).
+
+There are two scopes: you can block/allow traffic to all destinations, or block/allow traffic
+to specific destinations. The "all" rule will be be given higher priority (9999) than rules to
+specific destinations (9995). SONiC supports inserting multiple rules at the same priority, so
+all destination-specific rules will be given 9995.
+
+You can also specify a VLAN ID to restrict the block to a given VLAN destination. Note that you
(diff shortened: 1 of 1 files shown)


scripts/set_icmp_state
```diff
@@ -0,0 +1,293 @@
+#!/usr/bin/env python
```
qiluo-msft, line 1: >python [](start = 15, length = 6)

Does it also work on python3?

scripts/set_icmp_state
```diff
@@ -0,0 +1,293 @@
...
+
+    This checks that 1) the provided address is valid, and 2) that the IP version of the
+    address matches the IP version of the table this address will be applied to.
+    """
+    try:
+        destination_ip_version = ipaddress.ip_network(ip_address).version
+        table_ip_version = 6 if table_ip_version == IP_VERSION_V6 else 4
+        if destination_ip_version != table_ip_version:
+            error("Cannot use an IPv{} rule with an IPv{} table, exiting...".format(destination_ip_version, table_ip_version))
+    except Exception as e:
```
qiluo-msft, line 90: >Exception [](start = 11, length = 9)

Is there a more specific exception type?

scripts/set_icmp_state
```diff
@@ -0,0 +1,293 @@
...
+        error("Could not parse {} as a valid IP address; exception={}".format(ip_address, e))
```
qiluo-msft, line 91: >exception [](start = 57, length = 9)

also print its stack?

scripts/set_icmp_state
```diff
@@ -0,0 +1,293 @@
...
+    return IP_VERSION_V6 if table_type in ("L3V6", "MIRRORV6") else IP_VERSION_V4
+
+
+def get_block_rule_key(destination_ip_address):
+    """Get the key that will be used to refer to the ACL rule used to block ICMP traffic to a destination.
+
+    Since the rules are all given the same priority in SONiC, we can't identify a rule based on the priority.
+    So, we use the destination IP being blocked to give each rule a unique name in the system.
+    """
+    return BLOCK_KEY_PREFIX + str(destination_ip_address)
```
qiluo-msft, line 132: >destination_ip_address [](start = 34, length = 22)

Since this is user input, and will later used to query db, do you want to validate it?

scripts/set_icmp_state
daall: In 201911 no, because click is not installed. On later versions, yes.

scripts/set_icmp_state
daall: Per the docs it will be a `ValueError`, so we can check for that.

scripts/set_icmp_state
daall: `exception={}` should cover that, I think? Are you looking for something more specific?

scripts/set_icmp_state
daall: This is validated in `verify_valid_ip_address`. Are there some other checks I should do as well?

qiluo-msft: I see. That's good validation!

---
In reply to: [591861189](https://github.com/Azure/sonic-utilities/pull/1488#discussion_r591861189) [](ancestors = 591861189)

qiluo-msft (approved):

PR #1493, opened 2021-03-11
Update load_minigraph to disable unnecessary services (bgp, lldp...) for fabric asic

ngoc-do:
Fabric asics do not need some services like bgp, lldp, teamd.
They just need database, swss, and syncd running.

Signed-off-by: ngocdo <ngocdo@arista.com>

#### How I did it

I disable them with `FEATURE` flag configured in `CONFIG_DB`, which is listened and reacted by hostcfgd. 

This PR will update `CONFIG_DB's FEATURE` for fabric asics. The update is executed in load_minigraph, and saved to config_db.json. So the update will run only once.

If a system has both NPU and fabric, only services running on fabric are disabled. NPU's services are not affected.

One field, `always_disabled_on_asics` is added to `FEATURE`'s services that we want to disable for fabric asics. For example, `bgp@4.service` and `bgp@5.service` will be disabled.
```
    "FEATURE": {
        "bgp": {
            "state": "enabled",
            "has_timer": "False",
            "has_global_scope": "False",
            "has_per_asic_scope": "True",
            "auto_restart": "enabled",
            "high_mem_alert": "disabled",
            "always_disabled_on_asics": "4,5"
        },
   }
```

hostcfg will act on that change and disable the services. This is done in another PR at `sonic-buildimage`,
which is https://github.com/Azure/sonic-buildimage/pull/7026.

#### How to verify it

Tested on our systems with NPU and fabrics.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -238,6 +238,15 @@ def _get_device_type():
 
     return device_type
 
+def _get_switch_type(config_db):
+    """Get switch_type from CONFIG_DB
+    """
+    metadata = config_db.get_table('DEVICE_METADATA')
+    if metadata and 'localhost' in metadata and 'switch_type' in metadata['localhost']:
+       return metadata['localhost']['switch_type']
+    switch_type = 'Unknown'
+    return switch_type
+
 def interface_alias_to_name(config_db, interface_alias):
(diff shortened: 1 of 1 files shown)


lguohan:
need unit test

PR #1474, opened 2021-03-03
[201811] Warmboot script improvements - timeout in exec and disable swss autorestart

vaibhavhd:
#### What I did
Below changes are made to warmboot/fastboot script:

1. Disable swss auto-restart during warm reboot.
2. Add timeout to make sure syncd shutdown request will return in time. 5s
3. Disable trap handler after +e.
4. Make sure that syncd pre-shutdown wait won't take more than 60 seconds.
5. Make sure subsequent docker exec won't stuck for long time
6. Before shutdown, check docker exec on the relevant docker containers still works.

#### How to verify it
Testing on a physical tested - 6100 T0:

Ideal warmboot
```
$ sudo warm-reboot -vvv
Wed Mar 3 21:08:15 UTC 2021 Disabling swss auto restart ...
Wed Mar 3 21:08:16 UTC 2021 Pausing orchagent ...
Wed Mar 3 21:08:16 UTC 2021 Stopping radv ...
Wed Mar 3 21:08:17 UTC 2021 Stopping bgp ...
Wed Mar 3 21:08:17 UTC 2021 Stopped bgp ...
Wed Mar 3 21:08:20 UTC 2021 Initialize pre-shutdown ...
Wed Mar 3 21:08:21 UTC 2021 Requesting pre-shutdown ...
Wed Mar 3 21:08:22 UTC 2021 Waiting for pre-shutdown ...
Wed Mar 3 21:08:24 UTC 2021 Pre-shutdown succeeded, state: pre-shutdown-succeeded ...
Wed Mar 3 21:08:24 UTC 2021 Backing up database ...
Wed Mar 3 21:08:25 UTC 2021 Stopping teamd ...
Wed Mar 3 21:08:25 UTC 2021 Stopped teamd ...
Wed Mar 3 21:08:25 UTC 2021 Stopping syncd ...
Wed Mar 3 21:08:36 UTC 2021 Stopped syncd ...
Wed Mar 3 21:08:36 UTC 2021 Stopping all remaining containers ...
Wed Mar 3 21:08:39 UTC 2021 Stopped all remaining containers ...
Wed Mar 3 21:08:41 UTC 2021 Enabling Watchdog before warm-reboot
Wed Mar 3 21:08:41 UTC 2021 Rebooting with /sbin/kexec -e to SONiC-OS-20181130.70 ...
```

Two attempts captured below
`Initialize pre-shutdown` took ~0.5s/~0.5s
[...]

PR #1495, opened 2021-03-11
Warmboot script improvements - timeout exec, disable swss autorestart, remove trap

vaibhavhd:
#### What I did
Related PR for 201811 branch - https://github.com/Azure/sonic-utilities/pull/1474

Below changes are made to warmboot/fastboot script:

1. Disable swss auto-restart during warm reboot.
2. Add timeout to make sure syncd shutdown request will return in time. 5s
3. Disable trap handler after +e.
4. Make sure that syncd pre-shutdown wait won't take more than 60 seconds.
5. Make sure subsequent docker exec won't stuck for long time
6. Before shutdown, check docker exec on the relevant docker containers still works.

#### How to verify it
Tested warmreboot on a physical testbed.

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -155,7 +155,7 @@ function initialize_pre_shutdown()
 {
     debug "Initialize pre-shutdown ..."
     TABLE="WARM_RESTART_TABLE|warm-shutdown"
-    RESTORE_COUNT=`sonic-db-cli STATE_DB hget "${TABLE}" restore_count`
+    RESTORE_COUNT=$(/usr/bin/redis-cli -n 6 hget "${TABLE}" restore_count)
     if [[ -z "$RESTORE_COUNT" ]]; then
         sonic-db-cli STATE_DB hset "${TABLE}" "restore_count" "0" > /dev/null
     fi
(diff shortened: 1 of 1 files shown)


jleveque (approved):

vaibhavhd:
retest this please

scripts/fast-reboot
```diff
@@ -715,6 +739,9 @@ if [ -x ${DEVPATH}/${PLATFORM}/${PLATFORM_PLUGIN} ]; then
     ${DEVPATH}/${PLATFORM}/${PLATFORM_PLUGIN}
 fi
 
+# Re-enable swss autorestart for the next boot
+config feature autorestart swss enabled
```
vaibhavhd, line 743: @jleveque in this step my intention was to re-enable autorestart of `swss`. However, in my tests, I have noticed that after warm-reboot `swss autorestart` remains disabled.
Do you know what am I doing missing here?

jleveque: I can only assume the saved configuration has it marked as disabled? But I see you're not saving the config after disabling it, so it's not you doing it. But I would start by checking that, and if so, figure out why it's saved as disabled.

vaibhavhd: I checked again, and this was due to database container being killed/stopped before that, therefore it was not able to enable it back
```
  File "/usr/local/lib/python3.7/dist-packages/redis/connection.py", line 559, in connect
    sock = self._connect()
  File "/usr/local/lib/python3.7/dist-packages/redis/connection.py", line 615, in _connect
    raise err
  File "/usr/local/lib/python3.7/dist-packages/redis/connection.py", line 603, in _connect
    sock.connect(socket_address)
ConnectionRefusedError: [Errno 111] Connection refused
```

Further, I found out that after executing `systemctl stop swss`, the container do not automatically restart even if autorestart is enabled for swss at the time.
In comparison, for `201811` branch, after `docker kill swss`  monit/systemctl timers keeps restarting swss

yxieca (approved):

PR #1506, opened 2021-03-16
Fix the python error for handle download image error

chiourung:
#### What I did
Correct python error

#### How to verify it
run sonic_installer install http://XXX
(Use wrong URL to lead download error )


chiourung:
retest this please

PR #1508, opened 2021-03-16
Enable PFCWD only on ports where PFC is enabled

ayurkiv-nvda:
#### What I did
Prevent errors in the log when PFCWD starts on ports that PFC was not previously enabled. 

Example of such errors in the log:
```
Nov 13 17:27:22.878468 arc-switch1038 ERR swss#orchagent: :- createEntry: Failed to start PFC Watchdog on port Ethernet100
Nov 13 17:27:23.878620 arc-switch1038 NOTICE swss#orchagent: :- registerInWdDb: No lossless TC found on port Ethernet100
```
Add unit test to cover the new flow.

#### How I did it
Before enabling PFCWD on a port, verify if PFC was previously enabled. If not, reply with an error to the user and as a warning in the log.
In the case where the command to enable PFCWD is done for all ports, the configuration will be applied for only those that has PCF enabled and for the rest the configuration will be skipped. In this case the output of the requested command will be only with a list of skipped ports.

#### How to verify it
run:
`pfcwd start EthernetX 600` where EthernetX is a port that has not configured PFC (usually admin state 'down' by default)

```
SKIPPED: PFC is not enabled on port: Ethernet0
SKIPPED: PFC is not enabled on port: Ethernet66
```

### DIFF
--- a/pfcwd/main.py
+++ b/pfcwd/main.py
@@ -250,8 +250,7 @@ class PfcwdCli(object):
 
     def verify_pfc_enable_status_per_port(self, port, pfcwd_info):
         pfc_status = self.config_db.get_entry(PORT_QOS_MAP, port).get('pfc_enable')
-
-        if pfc_status == None:
+        if pfc_status is None:
             log.log_warning("SKIPPED: PFC is not enabled on port: {}".format(port), also_print_to_console=True)
             return
 
@@ -385,9 +384,7 @@ class PfcwdCli(object):
         }
 
         for port in active_ports:
-            self.config_db.set_entry(
-                CONFIG_DB_PFC_WD_TABLE_NAME, port, pfcwd_info
-            )
+            self.verify_pfc_enable_status_per_port(port, pfcwd_info)
 
         pfcwd_info = {}
         pfcwd_info['POLL_INTERVAL'] = DEFAULT_POLL_INTERVAL * multiply
--- a/pfcwd/main.py
+++ b/pfcwd/main.py
@@ -11,6 +11,11 @@ from sonic_py_common.multi_asic import get_external_ports
 from tabulate import tabulate
 from utilities_common import multi_asic as multi_asic_util
 from utilities_common import constants
+from sonic_py_common import logger
+
+SYSLOG_IDENTIFIER = "config"
+
+log = logger.Logger(SYSLOG_IDENTIFIER)
 
 # mock the redis for unit test purposes #
 try:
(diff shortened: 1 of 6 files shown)


pfcwd/main.py
```diff
@@ -242,6 +247,21 @@ def start(self, action, restoration_time, ports, detection_time):
...
+
+        if pfc_status == None:
+            log.log_warning("SKIPPED: PFC is not enabled on port: {}".format(port), also_print_to_console=True)
+            return
+
+        self.config_db.mod_entry(
+            CONFIG_DB_PFC_WD_TABLE_NAME, port, None
+        )
+        self.config_db.mod_entry(
+            CONFIG_DB_PFC_WD_TABLE_NAME, port, pfcwd_info
```
smaheshm, line 262: pfcwd_info not initialized!

tests/pfcwd_test.py
```diff
@@ -192,6 +192,37 @@ def test_pfcwd_start_actions(self, mock_os):
...
+        result = runner.invoke(
+            pfcwd.cli.commands["show"].commands["config"],
+            obj=db
+        )
+        print(result.output)
+        assert result.output == pfcwd_show_config_output
+
+        mock_os.geteuid.return_value = 0
+        # remove 'pfc_enabled' entry from Ethernet0 node
+        db.cfgdb.mod_entry("PORT_QOS_MAP", "Ethernet0", None)
```
smaheshm, line 211: One suggestion, instead of modifying here don't add the "PORT_QOS_MAP" entry for one of the ports.

smaheshm:
You should also update the "start_default()", and add a unit test for "config pfcwd start_default"

smaheshm:
Thanks for adding the unit tests. For completeness, please also add a unit test for 'multi ASIC' platform. You already updated the multi ASIC config DBs, just need to add test case.

pfcwd/main.py
ayurkiv-nvda: Done

tests/pfcwd_test.py
ayurkiv-nvda: Yes, we can do not add "pfc_enabled" entry for  "PORT_QOS_MAP|Ethernet0". 
But in this case output of other testcases will be also affected, because "pfc_enabled" entry now become mandatory and all testcases are using the same config_db files
For example: after this change Ethernet0 will be continuously skipped during pfcwd configuration so output of existing testcases will be different

smaheshm: There is no test case that would be affected by not adding. Either way is fine.

ayurkiv-nvda:
> Thanks for adding the unit tests. For completeness, please also add a unit test for 'multi ASIC' platform. You already updated the multi ASIC config DBs, just need to add test case.

Done

ayurkiv-nvda:
> Thanks for adding the unit tests. For completeness, please also add a unit test for 'multi ASIC' platform. You already updated the multi ASIC config DBs, just need to add test case.

Done

ayurkiv-nvda:
> You should also update the "start_default()", and add a unit test for "config pfcwd start_default"

Done

smaheshm:
retest this please

smaheshm:
Nice!! Wish there was a command to show which ports has PFCWD enabled. Lets wait for tests to pass.

smaheshm:
retest this please

smaheshm (approved):
:shipit:

liat-grozovik (approved):

liat-grozovik:
@abdosi i am not sure if this is cleanly can be cherry picked or not but I suggest to consider the fix for 201911 as well. 
If you would like to take it we can check or create new PR.
@daall appreciate if you can take to 202012 with high prio.

PR #1514, opened 2021-03-18
Stop PMON docker before cold and soft reboots

aravindmani-1:
#### What I did
Stopped PMON before cold boot
#### How I did it
Edit reboot script and stop pmon docker and services
#### How to verify it
Check docker ps output just before cold boot is being executed.  
UT:
[cold_boot_UT.txt](https://github.com/Azure/sonic-utilities/files/6161266/cold_boot_UT.txt)


aravindmani-1:
@jleveque Please review this PR.

scripts/reboot
```diff
@@ -178,6 +190,8 @@ if [ -x ${DEVPATH}/${PLATFORM}/${PLATFORM_UPDATE_REBOOT_CAUSE} ]; then
     ${DEVPATH}/${PLATFORM}/${PLATFORM_UPDATE_REBOOT_CAUSE}
 fi
 
+stop_pmon_service
```
jleveque, line 193: I understand you want to stop this at the last possible moment, but does it just make things simpler to add the logic to the existing `stop_sonic_services()` function?

jleveque (changes requested):
Once all comments are addressed, please also add the same logic to the new soft-reboot script as part of this PR: `scripts/soft-reboot`

scripts/reboot
aravindmani-1: If there is any platform API is introduced/invoked in between after stopping PMON, it will fail if PMON is shutdown early.

yxieca: I don't see a problem moving stop_sonic_service to here.

yxieca: @aravindthoram can we consolidate the change in stop_sonic_services and move the call to there?

aravindmani-1: @yxieca Will commit the new changes soon.

aravindmani-1:
@yxieca , @jleveque  can you please review the latest changes?.

scripts/reboot
```diff
@@ -51,16 +51,29 @@ function tag_images()
...
     if [[ x"$SUBTYPE" == x"DualToR" ]]; then
         debug "DualToR detected, stopping mux container before reboot..."
         systemctl stop mux
     fi
+
     if [[ x"$ASIC_TYPE" != x"mellanox" ]]; then
         debug "Stopping syncd process..."
         docker exec -i syncd /usr/bin/syncd_request_shutdown --cold > /dev/null
         sleep 3
+        stop_pmon_service
```
yxieca, line 76: I think you should put stop_pmon_service outside the if block. Why skipping it for Mellanox platforms?

aravindmani-1:
@yxieca Addressed the comments.

yxieca (approved):

jleveque (approved):

PR #1519, opened 2021-03-22
[acl] Use a list instead of a comma-separated string for ACL port list

daall:

#### What I did
Fixes https://github.com/Azure/sonic-buildimage/issues/4934

#### How I did it
I discovered that using the @ syntax to insert the port list as a comma-separated list/array actually causes the port list to be deleted when the same table is `config add`-ed multiple times. Using a list works. (We still need to determine the root cause for why the @ syntax fails).

#### How to verify it
Updated the tests and ran the command locally to verify the fix. Still working on adding a new regression test case for this.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -3349,7 +3349,7 @@ def parse_acl_table_info(table_name, table_type, description, ports, stage):
         if port not in valid_acl_ports:
             raise ValueError("Cannot bind ACL to specified port {}".format(port))
 
-    table_info["ports@"] = ",".join(port_list)
+    table_info["ports"] = list(port_list)
 
     table_info["stage"] = stage
 
--- a/config/main.py
+++ b/config/main.py
@@ -3341,7 +3341,7 @@ def parse_acl_table_info(table_name, table_type, description, ports, stage):
     if ports:
         for port in ports.split(","):
             port_list += expand_vlan_ports(port)
-        port_list = set(port_list)
+        port_list = list(set(port_list))  # convert to set first to remove duplicate ifaces
     else:
         port_list = valid_acl_ports
 
@@ -3349,7 +3349,7 @@ def parse_acl_table_info(table_name, table_type, description, ports, stage):
         if port not in valid_acl_ports:
             raise ValueError("Cannot bind ACL to specified port {}".format(port))
 
-    table_info["ports"] = list(port_list)
+    table_info["ports"] = port_list
 
     table_info["stage"] = stage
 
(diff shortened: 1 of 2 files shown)


config/main.py
```diff
@@ -3349,7 +3349,7 @@ def parse_acl_table_info(table_name, table_type, description, ports, stage):
         if port not in valid_acl_ports:
             raise ValueError("Cannot bind ACL to specified port {}".format(port))
 
-    table_info["ports@"] = ",".join(port_list)
+    table_info["ports"] = list(port_list)
```
qiluo-msft, line 3352: >list [](start = 26, length = 4)

It's already a list, right?

tests/acl_config_test.py
```diff
@@ -36,7 +34,7 @@ def test_parse_table_with_vlan_and_duplicates(self):
         assert table_info["stage"] == "egress"
 
         # Since Ethernet4 is a member of Vlan1000 we should not include it twice in the output
-        port_list = table_info["ports@"].split(",")
+        port_list = set(table_info["ports"])
```
qiluo-msft, line 37: >port_list [](start = 8, length = 9)

port_set?

lguohan:
unit itest?

daall:
> unit itest?

I updated the existing unit tests to reflect this change. I don't think it's possible to unit test this exact scenario b/c the mock config db APIs will overwrite the swsscommon operation that is causing the issue.

qiluo-msft:
Could you add a vs test instead? If that is challenging, a kvm test is the next choice, and finally sonic-mgmt test case.

daall:
> Could you add a vs test instead? If that is challenging, a kvm test is the next choice, and finally sonic-mgmt test case.

I can add a vs test!

tests/acl_config_test.py
qiluo-msft: I mean the variable name is not proper. Not a blocking issue, but you can make it better readable.

---
In reply to: [598998736](https://github.com/Azure/sonic-utilities/pull/1519#discussion_r598998736) [](ancestors = 598998736)

qiluo-msft (approved):

daall:
> > Could you add a vs test instead? If that is challenging, a kvm test is the next choice, and finally sonic-mgmt test case.
> 
> I can add a vs test!

https://github.com/Azure/sonic-swss/pull/1694

qiluo-msft:
@daall, I tried cherry-pick to 201911 but see conflicts in unit test code. Could you submit another PR on 201911 branch?

PR #1520, opened 2021-03-23
route_check: Fix hanging & logging level

renukamanavalan:
#### What I did
**Fix for hanging:**
Enforce max len on string to be printed to stdout.
With huge mismatch, the message to print can be too long to hang python print statement.

**Lower logging level:**
The log for "_details of route mismatch_" is lowered to WARNING. The caller of the route_check may log error, as it sees it fit. The failure of route_check is indicated to caller via exit code.

#### How to verify it
Simulate error (_route-mismatch_), and verify
a) message len does not exceed set MAX
b) log level is WARNING

### DIFF
--- a/scripts/route_check.py
+++ b/scripts/route_check.py
@@ -115,14 +115,13 @@ def print_message(lvl, *args):
     :param args: message as list of strings or convertible to string
     :return None
     """
-    rem_len = PRINT_MSG_LEN_MAX
     msg = ""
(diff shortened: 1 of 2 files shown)


lguohan (dismissed):

renukamanavalan:
Please don't merge yet.
A correction is required.

lguohan (approved):

PR #1523, opened 2021-03-24
cache the bvid to vlan translations

aystarik:
Add lookup table for bvid to vlan translations.
bvid_tlb will store previous successful translations from slow  get_vlan_id_from_bvid()
This patch does not change the output from the command, only speeds up it for case of 10k+ MAC tables.

### DIFF
--- a/scripts/fdbshow
+++ b/scripts/fdbshow
@@ -118,11 +118,11 @@ class FdbShow(object):
                 else:
                     try:
                         vlan_id = port_util.get_vlan_id_from_bvid(self.db, bvid)
+                        bvid_tlb[bvid] = vlan_id
                         if vlan_id is None:
                             # the situation could be faced if the system has an FDB entries,
                             # which are linked to default Vlan(caused by untagged trafic)
                             continue
-                        bvid_tlb[bvid] = vlan_id
                     except Exception:
                         vlan_id = bvid
                         print("Failed to get Vlan id for bvid {}\n".format(bvid))
--- a/scripts/fdbshow
+++ b/scripts/fdbshow
@@ -87,6 +87,7 @@ class FdbShow(object):
         if not fdb_str:
             return
 
+        bvid_tlb = {}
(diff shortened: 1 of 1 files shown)


prsunny (commented):
lgtm, request @qiluo-msft to review

scripts/fdbshow
```diff
@@ -111,15 +112,20 @@ class FdbShow(object):
...
-                except Exception:
-                    vlan_id = fdb["bvid"]
-                    print("Failed to get Vlan id for bvid {}\n".format(fdb["bvid"]))
+                bvid = fdb["bvid"]
+                if bvid in bvid_tlb:
+                    vlan_id = bvid_tlb[bvid]
+                else:
+                    try:
+                        vlan_id = port_util.get_vlan_id_from_bvid(self.db, bvid)
+                        if vlan_id is None:
```
qiluo-msft, line 121: >if vlan_id is None: [](start = 24, length = 19)

Even `if vlan_id is None`, you can still cache it to save future redis access.

qiluo-msft (commented):
As comment

scripts/fdbshow
```diff
@@ -87,6 +87,7 @@ class FdbShow(object):
         if not fdb_str:
             return
 
+        bvid_tlb = {}
```
prsunny, line 90: if you meant bvid table, please rename to `bvid_tbl`

aystarik: I meant Translation lookaside buffer, as in MMU, thus TLB.

scripts/fdbshow
aystarik: Agree, updated patch

qiluo-msft (approved):

qiluo-msft:
Retest this please

aystarik:
@qiluo-msft should I do something to complete those expected checks?

prsunny (approved):

PR #1529, opened 2021-03-26
[warm-reboot] Fix failures of warm reboot on disconnect of ssh session

maksymbelei95:
* Starting the script in background mode and detaching the process
  from terminal session to prevent failures, caused by closing or
  sudden disconnecting of terminal session.
* Redirecting output of the script to the file to prevent failures,
  when the script tries to write an output to file descriptor of
  the nonexistent terminal session.
* Adding new parameter to script to be able to explicitly run the
  script in foreground mode with output to the terminal.
* Updating the command reference doc according to the changes.

#### What I did
Resolves https://github.com/Azure/sonic-buildimage/issues/7127
Fixed failures of warm reboot, when the SSH session is being disconnected.
As the script will now be executed in background mode by default, added parameter to explicitly run it as usual, in foreground mode.
Updated command reference according to the changes.

#### How I did it
By restarting the script in background mode with detaching it from the terminal session. 
All the output has redirected to file /var/log/warm-reboot.txt for warm-reboot case, or /var/log/fast-reboot.txt for fast-reboot, depends on  `REBOOT_TYPE`. This will prevent crashes of the script in case, when it will try to write some data to the file descriptor of the disconnected terminal session.

#### How to verify it
1. Connect to the switch with SSH;
2. Execute `sudo warm-reboot -v`;
3. See the current progress of warm reboot with `cat /var/log/warm-reboot.txt`;
4. Close SSH connection before warm reboot finish;
Warm reboot should finish successfully, in spite of status of the SSH session.

The script will be running in background detached mode with output to the file. The related log will be shown in terminal before restarting in background mode:
```
admin@sonic:~$ sudo warm-reboot
Detaching the process from the terminal session. Redirecting output to /var/log/warm-reboot.txt.
```
All the usual logs will be written to warm-reboot.txt.


maksymbelei95:
retest this please

liat-grozovik:
retest this please

scripts/fast-reboot
```diff
@@ -430,6 +436,13 @@ then
     exit "${EXIT_FAILURE}"
 fi
 
+# re-run the script in background mode with detaching from the terminal session
+if [[ x"${DETACH}" == x"yes" ]]; then
+   echo "Detaching the process from the terminal session. Redirecting output to ${LOG_PATH}."
+   $0 "$@" -d 1>$LOG_PATH 2>$1 & disown %%
```
yxieca, line 442: 2>$1?

maksymbelei95: It means redirect stderr to stdout. stdout in our case is directed to the file.

yxieca: I suspect you wanted to do so, you can use 2>&1 or change the whole thing to 

$0 "$@" -d &>$LOG_PATH & disown %%

scripts/fast-reboot
```diff
@@ -21,6 +21,8 @@ PLATFORM_PLUGIN="${REBOOT_TYPE}_plugin"
 LOG_SSD_HEALTH="/usr/local/bin/log_ssd_health"
 SSD_FW_UPDATE="ssd-fw-upgrade"
 TAG_LATEST=yes
+DETACH=yes
```
yxieca, line 24: I think -d should mean detach and default should be DETACH=no

scripts/fast-reboot
maksymbelei95: Done.

scripts/fast-reboot
maksymbelei95: Sorry for late response. The idea of the changes is to prevent failures of warm-reboot/fast-reboot in cases when the ssh session is broken/closed. I made detached mode as default and leave possibility to run the scripts in non-detached mode with -d if the user wants. If to make as you propose, the user could forget to add -d parameter, what may lead to failure of the reboot. Due to this, I prefer to leave parameter -d as is. By the way, parameter -t made in similar way, so, I thought that there is no problem to let -d to run the process in non-detached mode, not wise versa.

liat-grozovik:
@yxieca can you please further review and see if all comments were addressed and it can be merged?

liat-grozovik:
it LGTM. @yxieca ?

scripts/fast-reboot
yxieca: @maksymbelei95 I don't think changing default behavior is the right thing. This blocking behavior has been used until now in tests and production. Changing the default behavior could have unexpected consequences. For special scenario, it should not be default.

stepanblyschak:
@maksymbelei95 Could you please resolve the conflict?

pshulik:
@yxieca, Could you please review the latest merge

yxieca:
> @yxieca, Could you please review the latest merge

I still see detach becomes default behavior. I think the default behavior should be the original behavior: blocking.

MaratGubaiev:
Could anybody please rerun the tests?
@yxieca, @liat-grozovik

yxieca (approved):

dprital:
@yxieca I see that this PR was approved by you. Can you please merge it ?

stepanblyschak (approved):

PR #1530, opened 2021-03-26
Relax the install_requires, no need to exact version as long as there are no broken changes with future versions

qiluo-msft:
#### What I did
Fixes https://github.com/Azure/sonic-buildimage/issues/7152

#### How I did it
Relax the install_requires


jleveque:
Two questions:

1. What about backporting to the 202012 branch?
2. How can we be certain there are no broken changes with future versions? That was the whole reason for pinning the versions down -- to prevent newer versions getting installed and potentially breaking things.

qiluo-msft:
1. I added 'request' label for old branch.
2. When that happens, we could add new requirement rule to enforce the dependency. On the other hand, pinning down is too conservative, and we need to make sure the host environment or docker already installed the required version, otherwise, there will be upgrade/downgrade, which is confusing.


---
In reply to: [808638168](https://github.com/Azure/sonic-utilities/pull/1530#issuecomment-808638168) [](http://example.com/codeflow?ancestors=808638168,808638168)

setup.py
```diff
@@ -156,24 +156,24 @@
...
-        'netifaces==0.10.7',
-        'pexpect==4.8.0',
-        'pyroute2==0.5.14',
-        'requests==2.25.0',
+        'click>=7.0',
+        'ipaddress>=1.0.23',
+        'jsondiff>=1.2.0',
+        'jsonpatch>=1.32.0',
+        'm2crypto>=0.31.0',
+        'natsort>=6.2.1',  # 6.2.1 is the last version which supports Python 2. Can update once we no longer support Python 2
```
jleveque, line 164: You can remove this comment; sonic-utilities is now Python 3-only. Could also try bumping it up to 7.1.1.

jleveque (changes requested):

setup.py
qiluo-msft: Bumped and removed.

jleveque (approved):

jleveque (changes requested):
Sorry, I misled you. You can revert your last commit. We cannot bump it up just yet, because it conflicts with the sonic-py-common and sonic-platform-common packages, which still install 6.2.1 because they still support Python 2. Thus, we need to remain back on 6.2.1 for now.

```
pkg_resources.ContextualVersionConflict: (natsort 7.1.1 (/__w/1/s/.eggs/natsort-7.1.1-py3.7.egg), Requirement.parse('natsort==6.2.1'), {'sonic-py-common', 'sonic-platform-common'})
```

jleveque (approved):

jleveque (approved):

PR #1531, opened 2021-03-28
[acl-loader]: do not add default deny rule for egress acl

lguohan:

#### What I did
do not add default deny rule for egress acl

#### How to verify it
tbd


acl_loader/main.py
```diff
@@ -636,7 +644,7 @@ def convert_rules(self):
                 except AclLoaderException as ex:
                     error("Error processing rule %s: %s. Skipped." % (acl_entry_name, ex))
 
-            if not self.is_table_mirror(table_name):
+            if not self.is_table_mirror(table_name) and not self.is_table_egress(table):
```
qiluo-msft, line 647: >table [](start = 81, length = 5)

table_name ?

qiluo-msft (approved):

SavchukRomanLv:
@lguohan, @qiluo-msft can you please provide info why these changes happen (HLD, some other doc), as after this test_unmatched_blocked (which expects that all unmatched traffic for egress will be automatically blocked) started failing
Thank you in advance!

qiluo-msft:
@SavchukRomanLv Sorry for the inconvenience. We have use case that one physical port belongs to 2 vlans and there are multiple egress ACL tables bound to the same port, actually each ACL table only have ACL rules matching one vlan ID.

If all unmatched traffic for egress will be automatically dropped, the first table will drop remaining traffic on one vlan and all traffic on the other vlan. To quickly fix this issue, we expected user provide explicit drop rules at the end of each ACL table, and we removed the default drop rule.

Please be aware that we translate the egress ACL table bound to vlan to egress ACL table bound to each port in that vlan in currently implementation.

stepanblyschak:
@qiluo-msft @lguohan is there a plan to change the sonic-mgmt test expectations?

PR #1540, opened 2021-04-01
[acl_loader]: add iptype match to the rules for dataplane acl

lguohan:
#### What I did
dataplane acl has v4 and v6 type. in case the rule does not
specify the iptype, the acl_loader will automatically add
the match for the iptype based on the table type.

for l3 table, it will add ethertype = 0x800
for l3v6 table, it will add iptype = ipv4any

#### How to verify it
unit test

### DIFF
--- a/acl_loader/main.py
+++ b/acl_loader/main.py
@@ -299,6 +299,22 @@ class AclLoader(object):
         """
         return self.tables_db_info[tname]['type'].upper().startswith(self.ACL_TABLE_TYPE_MIRROR)
 
+    def is_table_l3v6(self, tname):
+        """
+        Check if ACL table type is L3V6
+        :param tname: ACL table name
+        :return: True if table type is L3V6 else False
+        """
+        return self.tables_db_info[tname]["type"].upper() == "L3V6"
+
+    def is_table_l3(self, tname):
+        """
+        Check if ACL table type is L3
+        :param tname: ACL table name
+        :return: True if table type is L3 else False
+        """
+        return self.tables_db_info[tname]["type"].upper() == "L3"
+
     def is_table_ipv6(self, tname):
         """
         Check if ACL table type is IPv6 (L3V6 or MIRRORV6)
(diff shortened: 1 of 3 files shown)


tests/acl_loader_test.py
```diff
@@ -82,6 +83,17 @@ def test_vlan_id_not_a_number(self, acl_loader):
             acl_loader.rules_info = {}
             acl_loader.load_rules_from_file(os.path.join(test_path, 'acl_input/illegal_vlan_nan.json'))
 
+    def test_ethertype_translation(self, acl_loader):
+        acl_loader.rules_info = {}
+        acl_loader.load_rules_from_file(os.path.join(test_path, 'acl_input/acl1.json'))
+        assert acl_loader.rules_info[("DATAACL", "RULE_3")]
+        assert acl_loader.rules_info[("DATAACL", "RULE_3")] == {
+            "VLAN_ID": 369,
+            "ETHER_TYPE": 35020,
```
lguohan, line 92: from yang model translation, it is a integer, but our default rule is a string. it is ok as redis db is always string based, but there is a slight difference here

qiluo-msft (dismissed):

acl_loader/main.py
```diff
@@ -593,6 +609,14 @@ def convert_rule_to_db_schema(self, table_name, rule):
 
         rule_props["PRIORITY"] = str(self.max_priority - rule_idx)
 
+        # add ip type match to dataplane acl if not specified in the rule
+        if self.is_table_l3v6(table_name):
+            if "IP_TYPE" not in rule_props:
+                rule_props["IP_TYPE"] = "IPV6ANY"  # ETHERTYPE is not supported for DATAACLV6
+        elif self.is_table_l3(table_name):
+            if "ETHER_TYPE" not in rule_props:
+                rule_props["ETHER_TYPE"] = str(self.ethertype_map["ETHERTYPE_IPV4"])
```
daall, line 618: rule props won't have IP_TYPE or ETHER_TYPE until we do the `convert_action` and `deep_update` step below

daall: however *if* these exist below they will be overwritten. I still think it might be more clear to do this logic after we process all the openconfig fields

daall (dismissed):

qiluo-msft (approved):

daall (dismissed):

qiluo-msft (dismissed):

PR #1544, opened 2021-04-03
Reduce time taken by show commands on multi-asic platforms

abdosi:
What I did:
Optimize/Reduce the time taken by `show interface *` and `show ip bgp *` commands  on Multi-asic platforms.

How I did:
When checking for whether given port/port-channel/bgp-session is internal we used to walk through all namespace and collect data from each of then and then filter out. 
Now as part of this PR we  pass the namespace where port/port-channel/bgp-session belongs and then filter them out thus reducing the time considerably.

How I verify:

```
Before fix:

abdosi@xxxx:~$ time show interface status
real    1m0.347s
user    0m9.928s
sys     0m2.292s

abdosi@xxxx:~$ time show interface counters
real    0m43.750s
user    0m8.520s
sys     0m2.056s

abdosi@xxxx:~$ time show ip bgp summary
real    0m17.255s
user    0m2.476s
sys     0m0.804s

After fix:

abdosi@xxxx:~$ time show interface status
real    0m4.214s
user    0m2.140s
sys     0m0.412s

abdosi@xxxx:~$ time show interface counters
real    0m4.879s
user    0m2.500s
sys     0m0.680s

abdosi@xxxx:~$ time show ip bgp summary
real    0m6.769s
user    0m1.864s
sys     0m0.644s
```


rlhui (approved):

arlakshm (approved):

lguohan:
need #1541

PR #1553, opened 2021-04-09
sonic-installer: fix py3 issues in bootloader.aboot

Staphylo:
These issues are only seen on secureboot enabled platform.

### DIFF
--- a/sonic_installer/bootloader/aboot.py
+++ b/sonic_installer/bootloader/aboot.py
@@ -156,7 +156,7 @@ class AbootBootloader(Bootloader):
                 return None
             with swi.open(sigInfo, 'r') as sigFile:
                 for line in sigFile:
-                    data = line.split(':')
+                    data = line.decode('utf8').split(':')
                     if len(data) == 2:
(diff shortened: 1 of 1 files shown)


Staphylo:
I believe we also want this change to be cherry-picked on 202012

lguohan (approved):

daall:
@staphylo there is a merge conflict cherry-picking to 202012 which I guess is probably because we don't have https://github.com/Azure/sonic-utilities/pull/1366 in 202012. Are those changes needed in the 202012 branch as well?

Staphylo:
@dall I do believe we nee #1366 in 202012 as well.
Let me run some tests on 202012 and I'll confirm this back to you.

I can alternatively make a change against the 202012 branch that coalesce both changes.
Let me know what you prefer.

Staphylo:
@daall I can confirm that both changes have to be backported.
Let me know if you need my help with anything.

daall:
Thanks for confirming @staphylo, I will go ahead and cherry-pick both PRs to 202012! I'll let you know if I run into any problems.

PR #1555, opened 2021-04-11
No more IP validation as it is more likely a URL

renukamanavalan:
#### What I did
Dropped IP validation as server takes URL.

#### How to verify it
Provide a URL for IP in kube server config command.

### DIFF
--- a/config/kube.py
+++ b/config/kube.py
@@ -21,22 +21,6 @@ KUBE_STATE_SERVER_TS = "last_update_ts"
 KUBE_LABEL_TABLE = "KUBE_LABELS"
 KUBE_LABEL_SET_KEY = "SET"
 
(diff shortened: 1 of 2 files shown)


config/kube.py
```diff
@@ -82,9 +65,7 @@ def server():
 @pass_db
 def ip(db, vip):
     """Specify a kubernetes cluster VIP"""
-    if vip and not is_valid_ip4_addr(vip) and not is_valid_ip6_addr(vip):
```
isabelmsft, line 85: Can we assume the URL is valid?

renukamanavalan: The only validation is upon join. At the point of configuration, the URL need not be available. SONiC will periodically try until join.

isabelmsft (approved):

PR #1567, opened 2021-04-16
No more IP validation as it is more likely a URL (#1555)

renukamanavalan:
Dropped IP validation as server takes URL.


renukamanavalan:
PR #1555 is labeled for cherry-pick. Closing this.

PR #1569, opened 2021-04-19
disk_check: Check & mount RO as RW using tmpfs

renukamanavalan:
#### What I did
There is a bug that occasionally turn root-overlay as RO. This makes /etc & /home as RO. This blocks any new remote user login, as that needs to write into /etc & /home.

This tool scans /etc & /home (_or given dirs_) as in RW or RO state. If RO, it could create a writable overlay using tmpfs.
This is transient and stays until next reboot. Any write after the overlay will be lost upon reboot.

But this allows new remote users login.

#### How I did it
Create upper & work dirs in /run/mount (tmpfs). Mount /etc & /home as lowerdirs and use the same name for final merge. This allows anyone opening a file in **/etc or /home to operate on the merged overlay, transparently.**

#### How to verify it
Mount any dir on tmpfs ( mount -t tmpfs tmpfs test_dir)
remount as RO (mount -o remount,ro test_dir)
Pass that dir to this script. (disk_check.py -d ./test_dir)
Now it should be RW

### DIFF
--- a/scripts/disk_check.py
+++ b/scripts/disk_check.py
@@ -1,0 +1,151 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+"""
+What:
+    There have been cases, where disk turns RO due to kernel bug.
+    In RO state, system blocks new remote user login via TACACS.
+    This utility is to check & make transient recovery as needed.
+
+How:
+    check for RW permission. If RO, create writable overlay using tmpfs.
+
+    By default "/etc" & "/home" are checked and if in RO state, make them RW
+    using overlay on top of tmpfs.
+
+    Making /etc/ & /home as writable lets successful new remote user login.
+
+    If in RO state or in RW state with the help of tmpfs overlay,
+    syslog ERR messages are written, to help raise alerts.
+
+    Monit may be used to invoke it periodically, to help scan & fix and
+    report via syslog.
+
+"""
+
+import argparse
+import os
+import sys
+import syslog
+import subprocess
+
+TEST_LOG_FN = None
+UPPER_DIR = "/run/mount/upper"
+WORK_DIR = "/run/mount/work"
+MOUNTS_FILE = "/proc/mounts"
+
+def log_err(m):
+    print("Err: {}".format(m))
+    syslog.syslog(syslog.LOG_ERR, m)
+    if TEST_LOG_FN:
+        TEST_LOG_FN(syslog.LOG_ERR, m)
+
+
+def log_info(m):
+    print("Info: {}".format(m))
+    syslog.syslog(syslog.LOG_INFO, m)
+
+
+def log_debug(m):
+    print("debug: {}".format(m))
+    syslog.syslog(syslog.LOG_DEBUG, m)
+
+
+def test_rw(dirs): 
+    for d in dirs:
+        rw = os.access(d, os.W_OK)
+        if not rw:
+            log_err("{} dir is not RW".format(d))
+            return False
+        else:
+            log_debug("{} dir is RW".format(d))
+    return True
+
+
+def run_cmd(cmd):
+    proc = subprocess.run(cmd, shell=True, text=True, capture_output=True)
+    ret = proc.returncode
+    if ret:
+        log_err("failed: ret={} cmd={}".format(ret, cmd))
+    else:
+        log_info("ret={} cmd: {}".format(ret, cmd))
+
+    if proc.stdout:
+        log_info("stdout: {}".format(str(proc.stdout)))
+    if proc.stderr:
+        log_info("stderr: {}".format(str(proc.stderr)))
+    return ret
+
+
+def get_dname(path_name):
+    return path_name.replace('/', ' ').strip().split()[-1]
+
+
+def do_mnt(dirs):
+    if os.path.exists(UPPER_DIR):
+        log_err("Already mounted")
+        return 1
+
+    for i in (UPPER_DIR, WORK_DIR):
+        ret = run_cmd("mkdir {}".format(i))
+        if ret:
+            break
+
+    for d in dirs:
+        if not ret:
+            ret = run_cmd("mount -t overlay overlay_{} -o lowerdir={},"
+            "upperdir={},workdir={} {}".format(
+                get_dname(d), d, UPPER_DIR, WORK_DIR, d))
+
+    if ret:
+        log_err("Failed to mount {} as RW".format(dirs))
+    else:
+        log_info("{} are mounted as RW".format(dirs))
+    return ret
+
+
+def is_mounted(dirs):
+    if not os.path.exists(UPPER_DIR):
+        return False
+
+    onames = set()
+    for d in dirs:
+        onames.add("overlay_{}".format(get_dname(d)))
+
+    with open(MOUNTS_FILE, "r") as s:
+        for ln in s.readlines():
+            n = ln.strip().split()[0]
+            if n in onames:
+                log_debug("Mount exists for {}".format(n))
+                return True
+    return False
+
+
+def do_check(skip_mount, dirs):
+    ret = 0
+    if not test_rw(dirs):
+        if not skip_mount:
+            ret = do_mnt(dirs)
+
+    # Check if mounted
+    if (not ret) and is_mounted(dirs):
+        log_err("READ-ONLY: Mounted {} to make RW".format(dirs))
+
+    return ret
+
+
+def main():
+    parser=argparse.ArgumentParser(
+            description="check disk for RW and mount etc & home as RW")
+    parser.add_argument('-s', "--skip-mount", action='store_true', default=False,
+            help="Skip mounting /etc & /home as RW")
+    parser.add_argument('-d', "--dirs", default="/etc,/home",
+            help="dirs to mount")
+    args = parser.parse_args()
+
+    ret = do_check(args.skip_mount, args.dirs.split(","))
+    return ret
+
+
+if __name__ == "__main__":
+    sys.exit(main())
--- a/scripts/disk_check.py
+++ b/scripts/disk_check.py
@@ -3,19 +3,19 @@
 
 """
 What:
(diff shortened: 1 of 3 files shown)


renukamanavalan:
Fixing unit test failure.

lguohan:
@qiluo-msft, can you review?

scripts/disk_check.py
```diff
@@ -0,0 +1,151 @@
...
+    print("Info: {}".format(m))
+    syslog.syslog(syslog.LOG_INFO, m)
+
+
+def log_debug(m):
+    print("debug: {}".format(m))
+    syslog.syslog(syslog.LOG_DEBUG, m)
+
+
+def test_rw(dirs):
```
qiluo-msft, line 54: >test_rw [](http://example.com/codeflow?start=4&length=7)

`test_writable`

setup.py
```diff
@@ -80,6 +80,7 @@
         'scripts/db_migrator.py',
         'scripts/decode-syseeprom',
         'scripts/dropcheck',
+        'scripts/disk_check.py',
```
qiluo-msft, line 83: >disk_check [](http://example.com/codeflow?start=17&length=10)

The filename is misleading. It looks only checking, but actually checking and fixing.

scripts/disk_check.py
```diff
@@ -0,0 +1,151 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+"""
+What:
+    There have been cases, where disk turns RO due to kernel bug.
```
qiluo-msft, line 6: >RO [](http://example.com/codeflow?start=44&length=2)

The comment block is very helpful. To make it better, try to explain RO, or just use readonly.

scripts/disk_check.py
```diff
@@ -0,0 +1,151 @@
...
+    In RO state, system blocks new remote user login via TACACS.
+    This utility is to check & make transient recovery as needed.
+
+How:
+    check for RW permission. If RO, create writable overlay using tmpfs.
```
qiluo-msft, line 11: >RW [](http://example.com/codeflow?start=14&length=2)

write

scripts/disk_check.py
```diff
@@ -0,0 +1,151 @@
...
+
+    By default "/etc" & "/home" are checked and if in RO state, make them RW
+    using overlay on top of tmpfs.
+
+    Making /etc/ & /home as writable lets successful new remote user login.
```
qiluo-msft, line 16: >/ [](http://example.com/codeflow?start=15&length=1)

Do you want to remove the trailing `/` to be consistent?

scripts/disk_check.py
```diff
@@ -0,0 +1,151 @@
...
+import syslog
+import subprocess
+
+TEST_LOG_FN = None
+UPPER_DIR = "/run/mount/upper"
+WORK_DIR = "/run/mount/work"
+MOUNTS_FILE = "/proc/mounts"
+
+def log_err(m):
+    print("Err: {}".format(m))
```
qiluo-msft, line 38: >print [](http://example.com/codeflow?start=4&length=5)

print to stderr

scripts/disk_check.py
```diff
@@ -0,0 +1,151 @@
...
+    syslog.syslog(syslog.LOG_ERR, m)
+    if TEST_LOG_FN:
+        TEST_LOG_FN(syslog.LOG_ERR, m)
```
qiluo-msft, line 41: Seem the code is disabled. Will you just remove?

scripts/disk_check.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def log_info(m):
+    print("Info: {}".format(m))
+    syslog.syslog(syslog.LOG_INFO, m)
+
+
+def log_debug(m):
+    print("debug: {}".format(m))
```
qiluo-msft, line 50: >print [](http://example.com/codeflow?start=4&length=5)

It's annoying to print debug messages on stdout.

scripts/disk_check.py
```diff
@@ -0,0 +1,151 @@
...
+    syslog.syslog(syslog.LOG_DEBUG, m)
+
+
+def test_rw(dirs): 
+    for d in dirs:
+        rw = os.access(d, os.W_OK)
+        if not rw:
+            log_err("{} dir is not RW".format(d))
```
qiluo-msft, line 58: >dir [](http://example.com/codeflow?start=24&length=3)

`path` or `directory` is better than `dir`. Actually you can just remove it.

qiluo-msft, line 58: >RW [](http://example.com/codeflow?start=35&length=2)

writable.

scripts/disk_check.py
```diff
@@ -0,0 +1,151 @@
...
+        log_info("ret={} cmd: {}".format(ret, cmd))
+
+    if proc.stdout:
+        log_info("stdout: {}".format(str(proc.stdout)))
+    if proc.stderr:
+        log_info("stderr: {}".format(str(proc.stderr)))
+    return ret
+
+
+def get_dname(path_name):
```
qiluo-msft, line 80: >get_dname [](http://example.com/codeflow?start=4&length=9)

Could you use `os.path.basename(path_name)` ?

scripts/disk_check.py
```diff
@@ -0,0 +1,151 @@
...
+    return path_name.replace('/', ' ').strip().split()[-1]
```
qiluo-msft, line 81: >' ' [](http://example.com/codeflow?start=34&length=3)

Blanks are valid characters in a path, so I think the replacement will hit some corner cases.

scripts/disk_check.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def do_mnt(dirs):
+    if os.path.exists(UPPER_DIR):
+        log_err("Already mounted")
+        return 1
+
+    for i in (UPPER_DIR, WORK_DIR):
+        ret = run_cmd("mkdir {}".format(i))
```
qiluo-msft, line 90: >mkdir [](http://example.com/codeflow?start=23&length=5)

Better use python function `os.mkdir(path)`

scripts/disk_check.py
```diff
@@ -0,0 +1,151 @@
...
+            help="Skip mounting /etc & /home as RW")
+    parser.add_argument('-d', "--dirs", default="/etc,/home",
+            help="dirs to mount")
+    args = parser.parse_args()
+
+    ret = do_check(args.skip_mount, args.dirs.split(","))
+    return ret
+
+
+if __name__ == "__main__":
```
qiluo-msft, line 150: >__main__ [](http://example.com/codeflow?start=16&length=8)

Do you require root privilege to run this script? If yes, could you check at the begining?

qiluo-msft (commented):
As comments

scripts/disk_check.py
renukamanavalan: done

setup.py
renukamanavalan: Not  fixing. A temporary/transient hold in place. If you notice, even after successful overlay, which has /etc as RW *now*, it still reports error.

scripts/disk_check.py
renukamanavalan: done.

scripts/disk_check.py
renukamanavalan: done

scripts/disk_check.py
renukamanavalan: done

scripts/disk_check.py
renukamanavalan: Intentional. When someone runs in fore-ground, I prefer all spew in stdout to help user,

scripts/disk_check.py
renukamanavalan: Used from test, to capture the logs.

scripts/disk_check.py
renukamanavalan: Again, intentional to help with users, who run it in fore-ground

scripts/disk_check.py
renukamanavalan: done.

renukamanavalan: done

scripts/disk_check.py
renukamanavalan: os.path.basename(os.path.normpath(path_name))

scripts/disk_check.py
renukamanavalan: os.path.basename(os.path.normpath(path_name))

scripts/disk_check.py
renukamanavalan: done

scripts/disk_check.py
renukamanavalan: You need privileges when fix is required. Moreover, the main usage is for monit

qiluo-msft: Suggest add user privilege check. Otherwise normal user will try and `mkdir` a directory owned by that user, and fail at `mount`. This is unexpected.

scripts/disk_check.py
qiluo-msft: Remove personal test code. Otherwise implement it as a feature.

scripts/disk_check.py
qiluo-msft: `print to stderr` does not hurt your intention.

renukamanavalan: done

scripts/disk_check.py
renukamanavalan: Mocked

scripts/disk_check.py
renukamanavalan: Unprivileged users can't create dir under /run/mount. They can still run check.

qiluo-msft (approved):

abdosi:
@renukamanavalan please create PR for 201911 as difference in python2 vs python3. Cherry-pick is not recommended here.

PR #1574, opened 2021-04-20
[multi-asic] support show ip bgp neigh/network for multi asic

arlakshm:
Signed-off-by: Arvindsrinivasan Lakshmi Narasimhan <arlakshm@microsoft.com>

#### What I did
This PR is to add support for the commands `show ip bgp neighbor` and `show ip bgp network`
Add unit tests for these commands
#### How I did it
Add change to get the bgp neighbor and network information from each namespace

#### How to verify it
Check the command works on single and multi asic platforms


show/bgp_frr_v4.py
```diff
@@ -24,43 +24,93 @@ def bgp():
...
+            # info_type is only valid if ipaddress is specified
+            if info_type is not None:
+                command += ' {}'.format(info_type)
+        except ValueError as err:
+            ctx = click.get_current_context()
+            ctx.fail("{}\n".format(err))
+
+    ns_list = multi_asic.get_namespace_list(namespace)
+    output = ""
+    for ns in ns_list:
```
gechiang, line 77: Question on the overall design to handle multi-ASIC.  By looking at how this is implemented it looks like it will cycle through all the name spaces and no attempt to merge/filter but will display whatever each namespace has to the output directly?

arlakshm: @gechiang, Yes, this command we just display neighbor from all the namespaces. No filtering done for this commands

show/bgp_frr_v4.py
```diff
@@ -24,43 +24,93 @@ def bgp():
...
+                    ['routes', 'advertised-routes', 'received-routes']),
+                required=False)
+@click.option('--namespace',
+                '-n',
+                'namespace',
+                default=None,
+                type=str,
+                show_default=True,
+                help='Namespace name or all',
+                callback=multi_asic_util.multi_asic_namespace_validation_callback)
```
smaheshm, line 46: why not add '-d' option to display only the external BGP sessions. Would be good to follow the CLI norm we've been following to display the frontend results by default.

smaheshm, line 46: if it's not lot of work can you explore adding the 'json' dump option.

show/bgp_frr_v4.py
```diff
@@ -24,43 +24,93 @@ def bgp():
...
+                metavar='[bestpath|json|longer-prefixes|multipath]',
+                type=click.Choice(
+                    ['bestpath', 'json', 'longer-prefixes', 'multipath']),
+                required=False)
+@click.option('--namespace',
+                '-n',
+                'namespace',
+                type=str,
+                show_default=True,
+                required=True if multi_asic.is_multi_asic is True else False,
```
smaheshm, line 98: on mulit ASIC "namespace" is mandatory.  @gechiang What's the approach 'show ip route' takes? Does it also require namespace to be mandatory.

smaheshm: I guess it's fine, it makes sense to print 'network' from a particular BGP instance.

show/bgp_frr_v4.py
arlakshm: I agree,  this PR is just adding support for multi asic. We can explore the json option in a future PR

show/bgp_frr_v4.py
arlakshm: Each BGP instance in multi asic platform will have all the learnt prefixes. This command will present the information from one specific instance.

tests/show_bgp_neighbor_test.py
```diff
@@ -0,0 +1,125 @@
...
+        print("SETUP")
+        from .mock_tables import mock_single_asic
+        importlib.reload(mock_single_asic)
+        from .mock_tables import dbconnector
+        dbconnector.load_namespace_config()
+
+    @pytest.mark.parametrize('setup_single_bgp_instance, test_vector',
+                             [
+                                ('bgp_v4_neighbors_output', 'bgp_v4_neighbors'),
+                                ('bgp_v4_neighbors_output', 'bgp_v4_neighbors'),
```
smaheshm, line 53: why twice?

smaheshm: never mind, one of them has IP address param. Can you change the name so that if test fails it's easier to know which one.

smaheshm, line 53: problem with parameterizing test cases is it can be confusing to find out which one failed if the variables have same names.

show/bgp_frr_v4.py
arlakshm: @smaheshm, as discussed offline adding `-d` option to this command adds lot of complexity and/or slowness to the cli command.  The command will display all the neighbor information, the `ipaddress` option can be used to display information for specific neighbor.

tests/show_bgp_neighbor_test.py
arlakshm: updated the test param names

gechiang (approved):

smaheshm (approved):
:shipit:

PR #1578, opened 2021-04-23
Provide support to install platform extensions

VenkatCisco:
#### What I did
Provide a generic capability to show/main.py for platform extensions

#### How I did it
invoke a new method called install_extensions() that is easily extensible for any vendor platform. Import the file if available or else move on. 

#### How to verify it
Enable vendor specific commands via the extensions. For instance, the below code extends the capability to support 'show platform idprom'. Logic can be extended per vendor needs. 

@click.command()
@click.option('--raw/--no-raw', default=False, help='Hexdump raw IDPROMs')
@click.option('--all/--no-all', default=False, help='Dump all known IDPROMs')
@click.option('--list/--no-list', default=False, help='List known IDPROMs')
@click.argument('name', nargs=-1)
def idprom(name, raw, all, list):
    """ Show platform IDPROM information """
    args = [ PLATFORM_PY, 'idprom' ]

    if all:
        if len(name) > 0:
            click.echo('?Option --all ignored when given a list of IDPROM names\n')
        elif list:
            click.echo('?Option --all ignored when combined with --list\n')
        else:
            args.append('--all')
    if list:
        if len(name) > 0:
            click.echo('?Option --list ignored when given a list of IDPROM names\n')
        else:
            args.append('--list')

does not exist. 
root@sonic:/home/cisco# show platform idprom 

board: [i2c:IO-MUX-0:0x50]/eeprom: /sys/bus/i2c/devices/17-0050/eeprom
    CHASSIS_SERIAL: FOC2330NCA3
    CLEI: UNASSIGNED
    CONTROLLER_TYPE: 1555
    DEVICE_VALS: 2414074587345059840
    ECI_NUMBER: ECI123
    GROUP_TYPE: 69


show/main.py
```diff
@@ -1470,12 +1475,33 @@ def ztp(status, verbose):
...
+        for cmd in extensions.get(key, []):
+            root.add_command(cmd)
 
 # Load plugins and register them
 helper = util_base.UtilHelper()
 for plugin in helper.load_plugins(plugins):
     helper.register_plugin(plugin, cli)
 
+# invoke method for installing extensions
+install_extensions()
```
judyjoseph, line 1504: Would this cause inconsistent platform CLI's across SONiC running in different vendor devices ?  As as i know we have common platform CLI's and platform API support is added via platform plugins. @jleveque to comment as well.

VenkatCisco: as per review comments, the code is restructured to the below (along with PR1577)

the underlying show_platform/cli/show.py will be moved under the respective plugin path as follows:

   sonic-utilities/show/plugins/cisco-8000.py 
Similar future enhancements for config and clear options will be as:

      sonic-utilities/config/plugins/cisco-8000.py
      sonic-utilities/clear/plugins/cisco-8000.py

VenkatCisco: > Would this cause inconsistent platform CLI's across SONiC running in different vendor devices ? As as i know we have common platform CLI's and platform API support is added via platform plugins. @jleveque to comment as well.

the tests failed after changes in dbp arp and vlan tests which has no intersection to the code changes that were made. Is it possible to kick off a rerun of the tests (i dont have privileges to do so).

anshuv-mfst:
@judyjoseph - could you please re-run the tests.

show/plugins/cisco-8000.py
```diff
@@ -0,0 +1,103 @@
...
+def version(verbose):
+    filespec = '/opt/cisco/etc/build_info.yaml'
+    with open(filespec) as f:
+        build_info = yaml.safe_load(f)
+    for d in build_info.get('sdk_versions', []):
+        for k,v in d.items():
+            click.echo('{}: {}'.format(k,v))
+
+@click.group()
+def npu():
```
judyjoseph, line 40: Since it is a platform specific plugin, it should be under the platform type grouping ( Ref:  https://github.com/Azure/sonic-utilities/blob/master/show/plugins/mlnx.py )

show/plugins/cisco-8000.py
```diff
@@ -0,0 +1,103 @@
...
+
+@npu.command()
+def status():
+    """ Show NPU status information """
+    click.echo('npu status\n')
+
+@click.command()
+def inventory():
+    """ Show platform inventory information """
+    invoke([PYTHON3, '-m', 'cisco.bin.inventory', 'show'])
```
judyjoseph, line 52: Could use this existing API (https://github.com/Azure/sonic-utilities/blob/3ff68c4e5287ab2f5d23c23176ebd75a4f629bf0/utilities_common/cli.py#L498)  instead of this new function invoke()

judyjoseph, line 52: Do we need the explicit invocation with "python3 -m" command, I felt if the script which is called has something like "#!/usr/bin/env python3" it should tell which py interpreter to use ?

show/plugins/cisco-8000.py
```diff
@@ -0,0 +1,103 @@
...
+            idprom,
+        ],
+    }
+
+    groups = {
+        'cli': cli,
+        'platform': platform.platform,
+    }
+
+    for key,root in groups.items():
```
judyjoseph, line 98: Good to add the "command format" and sample output in the PR description ( with this new plugin change )

show/plugins/cisco-8000.py
garigipati31: Fixed the code based on your recommendation

show/plugins/cisco-8000.py
garigipati31: Fixed per your recommendation.

garigipati31: Sure, fixed

garigipati31:
> #### What I did
> 
> Provide a generic capability to show/main.py for platform extensions
> #### How I did it
> 
> invoke a new method called install_extensions() that is easily extensible for any vendor platform. Import the file if available or else move on.
> #### How to verify it
> 
> Enable vendor specific commands via the extensions. For instance, the below code extends the capability to support 'show platform idprom'. Logic can be extended per vendor needs.
> 
> @click.command()
> @click.option('--raw/--no-raw', default=False, help='Hexdump raw IDPROMs')
> @click.option('--all/--no-all', default=False, help='Dump all known IDPROMs')
> @click.option('--list/--no-list', default=False, help='List known IDPROMs')
> @click.argument('name', nargs=-1)
> def idprom(name, raw, all, list):
> """ Show platform IDPROM information """
> args = [ PLATFORM_PY, 'idprom' ]
> 
> ```
> if all:
>     if len(name) > 0:
>         click.echo('?Option --all ignored when given a list of IDPROM names\n')
>     elif list:
>         click.echo('?Option --all ignored when combined with --list\n')
>     else:
>         args.append('--all')
> if list:
>     if len(name) > 0:
>         click.echo('?Option --list ignored when given a list of IDPROM names\n')
>     else:
>         args.append('--list')
> ```
> 
> #### Previous command output (if the output of a command-line utility has changed)
> 
> does not exist.
> #### New command output (if the output of a command-line utility has changed)
> 
> root@sonic:/home/cisco# show platform idprom
> 
> board: [i2c:IO-MUX-0:0x50]/eeprom: /sys/bus/i2c/devices/17-0050/eeprom
> CHASSIS_SERIAL: FOC2330NCA3
> CLEI: UNASSIGNED
> CONTROLLER_TYPE: 1555
> DEVICE_VALS: 2414074587345059840
> ECI_NUMBER: ECI123
> GROUP_TYPE: 69

root@sonic:/home/admin# show platform inventory 
    Name                Product ID      Version         Serial Number   Description

Chassis
    CHASSIS             8101-32H-O      0.10            FHH2446P00A     Cisco 8100 32x100G QSFP28 1RU Fixed System w/o HBM, Open SW

Route Processors
    RP0                 8101-32H-O      0.10            FHH2444000T     Cisco 8100 32x100G QSFP28 1RU Fixed System w/o HBM, Open SW

Power Supplies
    PSU0                PSU650W-ACPE    0.0             LIT24244CKU     650W AC Power Module with Port-side Exhaust
    PSU1                PSU650W-ACPE    0.0             LIT24244CKP     650W AC Power Module with Port-side Exhaust

Cooling Devices
    fantray0                                                            
    fantray1                                                            
    fantray2                                                            
    fantray3                                                            
    fantray4                                                            

FPDs
    RP0/iofpga                          1.2.0-827                       IOFPGA
    RP0/valentine                       1.2.0-1505                      VALENTINE

show/plugins/cisco-8000.py
garigipati31: I have added the output of the 'show platform inventory', which is part of plugin to the description.

show/plugins/cisco-8000.py
```diff
@@ -0,0 +1,83 @@
...
+        for cmd in extensions.get(key, []):
+            root.add_command(cmd)
+
+def register(cli):
```
judyjoseph, line 82: To add check for asic type before adding the plugin. Eg: 
    version_info = device_info.get_sonic_version_info()
    if (version_info and version_info.get('asic_type') == 'cisco'):
        install_extensions(cli)

show/plugins/cisco-8000.py
```diff
@@ -0,0 +1,83 @@
...
+    with open(filespec) as f:
+        build_info = yaml.safe_load(f)
+    for d in build_info.get('sdk_versions', []):
+        for k,v in d.items():
+            click.echo('{}: {}'.format(k,v))
+
+@click.command()
+def inventory():
+    """ Show platform inventory information """
+    cmd = 'inventory.py show'
```
judyjoseph, line 31: don't see these file inventory.py added ? Is it also present in /opt/cisco/bin/ ?

judyjoseph: For the inventory, a better approach will be define something new in sonic_platform, define API's and use those API's here instead of calling platform specific scripts.
Ref: https://github.com/Azure/sonic-utilities/blob/9b7c58b678998c2dbab1f4d5ee9ec6a9294ab666/show/platform.py#L13

show/plugins/cisco-8000.py
```diff
@@ -0,0 +1,83 @@
...
+    import yaml
+    from show import platform
+    import utilities_common.cli as clicommon
+except ImportError as e:
+    raise ImportError("%s - required module not found" % str(e))
+
+PLATFORM_PY = '/opt/cisco/bin/platform.py'
+
+# Platform specific version invoked by generic show.version
+def version(verbose):
```
judyjoseph, line 20: I don't see this function version() being used

show/plugins/cisco-8000.py
VenkatCisco: ack. Moved inventory() method to show/platform.py. A call to inventoryshow will be made from platform.py. For now, the inventoryshow is a python module sitting under /usr/local/bin/inventoryshow. this is built as part of cisco's platform code and placed under the target path (/usr/local/bin).

judyjoseph: My idea was to create a inventory base class in https://github.com/Azure/sonic-platform-common and implement that in Cisco 8800 platform  ( similar to as it is done here https://github.com/Azure/sonic-buildimage/tree/master/platform/broadcom/sonic-platform-modules-dell/s6100/sonic_platform for fan/psu/sfp etc. )

But if that is a bigger task, we should keep this CLI in show/plugins/cisco-8000.py itself as before. The inventoryshow script wont be present for all other platforms.

judyjoseph:
Looks ok, need to revert the "show platform inventory" to cisco plugin file.

VenkatCisco:
> Looks ok, need to revert the "show platform inventory" to cisco plugin file.

I am not clear on this. There was a discussion earlier to move 'show platform' related extensions to 'platform.py. Given 'inventory' is an extension to 'show platform' and is going to be platform (or vendor) agnostic moving forward in a distributed chassis environment, and hence, so just like: 

- show platform temperature
- show platform firmware
- show platform fan etc. 

I believe it makes sense to keep 'show platform inventory' in platform.py.

judyjoseph:
> > Looks ok, need to revert the "show platform inventory" to cisco plugin file.
> 
> I am not clear on this. There was a discussion earlier to move 'show platform' related extensions to 'platform.py. Given 'inventory' is an extension to 'show platform' and is going to be platform (or vendor) agnostic moving forward in a distributed chassis environment, and hence, so just like:
> 
> * show platform temperature
> * show platform firmware
> * show platform fan etc.
> 
> I believe it makes sense to keep 'show platform inventory' in platform.py.

@VenkatCisco, the general approach while adding a new functionality is to add/update the platform base classes defined in sonic-platform-common. This could be implemented by various platform (or vendor ) -- in their respective sonic_platform module ( eg: https://github.com/Azure/sonic-buildimage/tree/master/platform/broadcom/sonic-platform-modules-dell/s6100/sonic_platform). Further we can have common scripts (eg: https://github.com/Azure/sonic-utilities/blob/master/scripts/tempershow), to implement API's to get data from DB which could be populated by a platform daemon.

So the idea I was suggesting was to keep the "inventory" command under cisco plugin until we develop the infra surrounding abstracting/storing/retrieving the inventory data.

VenkatCisco:
> > > Looks ok, need to revert the "show platform inventory" to cisco plugin file.
> > 
> > 
> > I am not clear on this. There was a discussion earlier to move 'show platform' related extensions to 'platform.py. Given 'inventory' is an extension to 'show platform' and is going to be platform (or vendor) agnostic moving forward in a distributed chassis environment, and hence, so just like:
> > 
> > * show platform temperature
> > * show platform firmware
> > * show platform fan etc.
> > 
> > I believe it makes sense to keep 'show platform inventory' in platform.py.
> 
> @VenkatCisco, the general approach while adding a new functionality is to add/update the platform base classes defined in sonic-platform-common. This could be implemented by various platform (or vendor ) -- in their respective sonic_platform module ( eg: https://github.com/Azure/sonic-buildimage/tree/master/platform/broadcom/sonic-platform-modules-dell/s6100/sonic_platform). Further we can have common scripts (eg: https://github.com/Azure/sonic-utilities/blob/master/scripts/tempershow), to implement API's to get data from DB which could be populated by a platform daemon.
> 
> So the idea I was suggesting was to keep the "inventory" command under cisco plugin until we develop the infra surrounding abstracting/storing/retrieving the inventory data.

@judyjoseph,  I have updated the PR, based on your comments. Removed references to inventory code from the platform.py and moved it to cisco-8000.py plugin.

judyjoseph (approved):

PR #1580, opened 2021-04-28
Fixing serial number read to get from DB if it is populated

dgsudharsan:
#### What I did
Modified show version command to pick serial number from STATE_DB if it was populated instead of getting it from EEPROM.

#### How I did it
Check state_db to see if serial number EEPROM section is populated. If yes use the data from DB. If not, read it from the decode-syseeprom

#### How to verify it
Verify running 'show version' with and without pmon running to validate both scenarios, since eepromd in pmon populates the EEPROM table in state_db

### DIFF
--- a/show/main.py
+++ b/show/main.py
@@ -966,8 +966,7 @@ def version(verbose):
     eeprom_table = db.get_all(db.STATE_DB, 'EEPROM_INFO|0x23')
     if "Name" in eeprom_table and eeprom_table["Name"] == "Serial Number" and "Value" in eeprom_table:
         serial_number = eeprom_table["Value"]
-
-    if not serial_number:
+    else:
         serial_number_cmd = "sudo decode-syseeprom -s"
         serial_number = subprocess.Popen(serial_number_cmd, shell=True, text=True, stdout=subprocess.PIPE).stdout.read()
 
--- a/show/main.py
+++ b/show/main.py
@@ -960,7 +960,7 @@ def version(verbose):
     asic_type = version_info['asic_type']
     asic_count = multi_asic.get_num_asics()
 
-    serial_number=''
+    serial_number = None
     db = SonicV2Connector()
     db.connect(db.STATE_DB)
     eeprom_table = db.get_all(db.STATE_DB, 'EEPROM_INFO|0x23')
--- a/show/main.py
+++ b/show/main.py
@@ -960,8 +960,16 @@ def version(verbose):
     asic_type = version_info['asic_type']
     asic_count = multi_asic.get_num_asics()
 
-    serial_number_cmd = "sudo decode-syseeprom -s"
-    serial_number = subprocess.Popen(serial_number_cmd, shell=True, text=True, stdout=subprocess.PIPE)
+    serial_number=''
+    db = SonicV2Connector()
+    db.connect(db.STATE_DB)
+    eeprom_table = db.get_all(db.STATE_DB, 'EEPROM_INFO|0x23')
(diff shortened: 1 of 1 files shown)


show/main.py
```diff
@@ -960,8 +960,16 @@ def version(verbose):
     asic_type = version_info['asic_type']
     asic_count = multi_asic.get_num_asics()
 
-    serial_number_cmd = "sudo decode-syseeprom -s"
-    serial_number = subprocess.Popen(serial_number_cmd, shell=True, text=True, stdout=subprocess.PIPE)
+    serial_number=''
+    db = SonicV2Connector()
+    db.connect(db.STATE_DB)
+    eeprom_table = db.get_all(db.STATE_DB, 'EEPROM_INFO|0x23')
```
jleveque, line 966: One concern is whether this will hang if the DB is unconfigured or dead. `show version` should always work regardless of the state of the DB. Can you please confirm that this works and does not hang on an unconfigured switch?

dgsudharsan: > One concern is whether this will hang if the DB is unconfigured or dead. `show version` should always work regardless of the state of the DB. Can you please confirm that this works and does not hang on an unconfigured switch?

The show version command hangs even without my change when "CONFIG_DB_INITIALIZED" is removed from CONFIG_DB. With no docker database the show version command returns a trace. My change doesn't modify any of the existing behavior. Without state-db eeprom table populated, the actual eeprom is accessed and serial number is fetched. This can be verified by stopping pmon container

jleveque: Thanks for testing. As long as the behavior doesn't change with this patch, I'm OK. If we have to revisit the behavior later we'll cross that bridge then.

show/main.py
```diff
@@ -960,8 +960,16 @@ def version(verbose):
     asic_type = version_info['asic_type']
     asic_count = multi_asic.get_num_asics()
 
-    serial_number_cmd = "sudo decode-syseeprom -s"
-    serial_number = subprocess.Popen(serial_number_cmd, shell=True, text=True, stdout=subprocess.PIPE)
+    serial_number=''
```
jleveque, line 963: Add spaces around `=`. Also suggest assigning to `None` by default rather than empty string.

jleveque (changes requested):

show/main.py
dgsudharsan: > Add spaces around `=`. Also suggest assigning to `None` by default rather than empty string.

Done

show/main.py
```diff
@@ -960,8 +960,16 @@ def version(verbose):
...
-    serial_number_cmd = "sudo decode-syseeprom -s"
-    serial_number = subprocess.Popen(serial_number_cmd, shell=True, text=True, stdout=subprocess.PIPE)
+    serial_number = None
+    db = SonicV2Connector()
+    db.connect(db.STATE_DB)
+    eeprom_table = db.get_all(db.STATE_DB, 'EEPROM_INFO|0x23')
+    if "Name" in eeprom_table and eeprom_table["Name"] == "Serial Number" and "Value" in eeprom_table:
+        serial_number = eeprom_table["Value"]
+
+    if not serial_number:
```
jleveque, line 970: I think this could simply be an `else` case to the above `if`

jleveque (changes requested):

jleveque (approved):

show/main.py
dgsudharsan: Done

liat-grozovik (approved):

liat-grozovik:
@jleveque would you like this enhancement to be in 202012 as well? if so, please add a label.
I believe it can help to run platform API tests when pmon is not running.

PR #1602, opened 2021-05-07
[config]: Use mod_entry when editing VLAN_INTERFACE

theasianpianist:
Signed-off-by: Lawrence Lee <lawlee@microsoft.com>

#### What I did
Replace `set_entry` with `mod_entry` when setting the `proxy_arp` value for a VLAN.

Using `set_entry` will delete any other fields set for the key used, which is not desirable.

#### How to verify it
Add some field to `VLAN_INTERFACE|Vlan1000`, then run `config vlan proxy_arp 1000 enabled`. Confirm with `sonic-db-cli` that the extra field is still present after the command runs.

### DIFF
--- a/config/vlan.py
+++ b/config/vlan.py
@@ -92,7 +92,7 @@ def config_proxy_arp(db, vid, mode):
     if not clicommon.is_valid_vlan_interface(db.cfgdb, vlan):
         ctx.fail("Interface {} does not exist".format(vlan))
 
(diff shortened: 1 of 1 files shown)


jleveque (approved):

prsunny (approved):

jleveque:
@theasianpianist: Please add labels as appropriate if this should be cherry-picked into release branches.

PR #1568, opened 2021-04-17
[sonic-utilities] CLI support for port auto negotiation

Junchao-Mellanox:
#### What I did

1. Add CLI support for port auto negotiation feature.
2. Add db_migrator change for auto negotiation feature
2. Add unit test cases for all changes

#### How I did it

1. Add new subcommands to "config interface" command group to allow user configuring port auto negotiation
2. Add new subcommands to "show interfaces" command group to allow user show auto negotiation status
3. In db_migrator.py, change auto negotiation related DB field to latest one

#### How to verify it

Add unit tests for each new commands and db_migrator

PR #1614, opened 2021-05-14
[test][db_migrator] Make advance_version_for_expected_database available for other db migrator test cases as well

stephenxs:
#### What I did
Originally, the method `advance_version_for_expected_database` was introduced (in Azure/sonic-utilities#1566) to handle the case the latest version in `CONFIG_DB` is greater than the latest version in `mellanox_buffer_migrator`.
Now there are other database migrators whose test cases can also encounter this situation, like port auto-negotiation (Azure/sonic-utilities#1568) and port-channel for LACP key (Azure/sonic-utilities#1473).
So I would like to make the method public, available for all database migrators.
Related database migrator test cases have been updated accordingly.

#### How to verify it
Run the unit test.

### DIFF
--- a/tests/db_migrator_input/config_db/non-default-config-expected.json
+++ b/tests/db_migrator_input/config_db/non-default-config-expected.json
@@ -1115,6 +1115,6 @@
         "speed": "50000"
     },
     "VERSIONS|DATABASE": {
-        "VERSION": "version_2_0_1"
+        "VERSION": "version_2_0_0"
     }
 }
--- a/tests/db_migrator_test.py
+++ b/tests/db_migrator_test.py
@@ -24,6 +24,17 @@ def get_sonic_version_info_mlnx():
     return {'asic_type': 'mellanox'}
 
 
+def advance_version_for_expected_database(migrated_db, expected_db, last_interested_version):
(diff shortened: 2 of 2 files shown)


tests/db_migrator_input/config_db/non-default-config-expected.json
```diff
@@ -1115,6 +1115,6 @@
         "speed": "50000"
     },
     "VERSIONS|DATABASE": {
-        "VERSION": "version_2_0_1"
+        "VERSION": "version_2_0_0"
```
neethajohn, line 1118: w

stephenxs: The logic in `advance_version_for_expected_database` is:
  - to advance the version in the expected database if it matches the latest version defined by its migrator
  - not to do anything otherwise.

For example, for buffer migrator, the latest version is `version_2_0_0` so all the expected files should have this version. However, this was updated by a previous PR by mistake. So we need to correct it here.

liat-grozovik (approved):

liat-grozovik:
@neethajohn can you please review? this one is blocking us from moving forward with another fix we need on master.

neethajohn (approved):

PR #1535, opened 2021-03-30
Fix remove ip rif

maksymbelei95:
#### What I did
Resolves https://github.com/Azure/sonic-buildimage/issues/7028
Added checking of static routes, related to the interface, before deleting of the last IP entry to prevent deleting the RIF if a static route is present in the system.
The related unit tests has added.

#### How I did it
By checking IPv4 and IPv6 routes, received from Zebra.

#### How to verify it
1. `sudo config int ip add Ethernet2 10.0.0.18/31`
2. `sudo config route add prefix 20.0.0.1/32 nexthop Ethernet2`
3. `sudo config int ip rem Ethernet2 10.0.0.18/31`
If `10.0.0.18/31` is the last IP entry, you will receive the next error: `Error: Cannot remove the last IP entry of interface Ethernet2. A static ip route is still bound to the RIF.`
The similar situation is for IPv6: `Error: Cannot remove the last IP entry of interface Ethernet2. A static ipv6 route is still bound to the RIF.`

PR #1619, opened 2021-05-18
[config] Fix adding dev to static route

d-dashkov:
**What I did**
Change rule for adding dev vlan to route
Update unit tests with ethernet and vlan cases
fixes https://github.com/Azure/sonic-buildimage/issues/7643

Also solved https://github.com/Azure/sonic-utilities/pull/1534#discussion_r628463998 with notifying the user.

**How I did it**
Add a rule for checking vlan in the table, which overrides checking by vlan interface. Since vlan does not have interface entry until an IP address is added to it

**How to verify it**
```
sudo config vlan add 20
sudo config route add prefix 1.2.3.4/32 nexthop dev Vlan20

show ip route
S   1.2.3.4/32 [1/0] is directly connected, Vlan20 inactive, weight 1, 00:21:59
```

If user trying to set blackhole on existing route
"Error: this route is already configured on nexthop, remove it before configuring to black hole"

Signed-off-by: d-dashkov <Dmytro_Dashkov@Jabil.com>

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -4008,7 +4008,8 @@ def add_route(ctx, command_str):
 
     # If defined intf name, check if it belongs to interface
     if 'ifname' in route:
-        if (not route['ifname'] in config_db.get_keys('VLAN_INTERFACE') and
+        if (not route['ifname'] in config_db.get_keys('VLAN') and
+            not route['ifname'] in config_db.get_keys('VLAN_INTERFACE') and
             not route['ifname'] in config_db.get_keys('INTERFACE') and
             not route['ifname'] in config_db.get_keys('PORTCHANNEL_INTERFACE') and
             not route['ifname'] == 'null'):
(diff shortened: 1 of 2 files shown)


config/main.py
```diff
@@ -4008,7 +4008,8 @@ def add_route(ctx, command_str):
 
     # If defined intf name, check if it belongs to interface
     if 'ifname' in route:
-        if (not route['ifname'] in config_db.get_keys('VLAN_INTERFACE') and
+        if (not route['ifname'] in config_db.get_keys('VLAN') and
```
prsunny, line 4011: This will break the assumption made as part of PR https://github.com/Azure/sonic-utilities/pull/1535#issuecomment-825330038

d-dashkov: I see that this will allow configuring a route to ifname without ip, but this seems to be the only way to solve this problem.

dprital:
@d-dashkov , @prsunny  - What is the status with this PR ? Do you have plans to approve it ?

prsunny:
> @d-dashkov , @prsunny - What is the status with this PR ? Do you have plans to approve it ?

As mentioned, this poses another challenge in the delete flow. Hence, not planning to merge

d-dashkov:
@prsunny could you check this PR, I returned the rules for vlan and just left a warning about a black hole

PR #1625, opened 2021-05-19
Fix Aboot breakage sonic-installer due to package migration

Staphylo:
#### Failure cause

Fixes https://github.com/Azure/sonic-buildimage/issues/7566

The `get_rootfs_path` contextmanager was recently repurposed to implement
`get_file_in_image` . It then got used as a function by leveraging some
python complexity to bypass the restrictions coming with the contextmanager construct.
This was a conscious decision to prevent exactly what happened.
The method was then called multiple times to compute relative paths to the image directory.
Simple path joins using `new_image_dir` could have been performed instead.

The `get_rootfs_path` implementation for Aboot bootloaders behaves differently
when a rootfs is extracted or kept within the SWI image. It also behaves differently on secureboot systems.
The updated method was used by providing parameters it was never meant to process.

#### Context around the failure

Over time, the installation and boot process on Aboot has slightly diverged from
the ONIE one. There are 3 scenarios to consider.

1) Regular boot similar to ONIE
This one should just work the same as the filesystem layout is unchanged.

2) docker_inram where dockerfs.tar.gz is extracted in tmpfs at boot
In this scenario the boot is similar to the regular one beside that
`dockerfs.tar.gz` is preserved intact on the flash and not extracted.
By design this does not fit the sonic-package-manager requierements and
the migration should be skipped which is what I did in this review.
In the coming months this boot mode will look closer to `3)` below.

3) Secureboot on Arista devices
In this scenario the SWI image is kept intact and nothing extracted
from it. By ensuring the integrity of the SWI we can guarantee that no
code/data has been tampered with. This mode also relies on
`docker_inram` at the moment.
It could be enhanced when sonic-package-manager can guarantee the
integrity of code and data that is both installed and migrated.

#### Solution provided

The method `get_file_in_image` was reverted to its original meaning
`get_rootfs_path` as there is no point in keeping the new one.
It doesn't have the necessary logic to handle more than the rootfs and
the logic for which it was repurposed can easily be achieved by the following line.
`os.path.join(bootloader.get_image_path(binary_image_name), 'something')`
If this is a desirable helper it should become a method of its own.

A new `Bootloader` method called `support_package_migration` is
introduced to allow the bootloader to skip the package migration based
on the image (e.g docker_inram) or its own configuration (e.g secureboot).
By default all bootloaders enable the package migration.

That change leads to `1)` above running package migration while `2)` and `3)`
skip it.

The temporary error message now becomes a warning message for platforms that do not
support package migration.

```
Command: losetup -d /dev/loop2
Warning: SONiC package migration is not supported for this bootloader/image
Command: sync;sync;sync
```

### DIFF
--- a/sonic_installer/bootloader/aboot.py
+++ b/sonic_installer/bootloader/aboot.py
@@ -19,6 +19,7 @@ from ..common import (
    HOST_PATH,
    IMAGE_DIR_PREFIX,
    IMAGE_PREFIX,
+   ROOTFS_NAME,
    run_command,
    run_command_or_raise,
 )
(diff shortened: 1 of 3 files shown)


jleveque:
@stepanblyschak: Please review.

stepanblyschak (approved):

yxieca (approved):

PR #1631, opened 2021-05-24
[config][tacacs+] Change tacacs+ minimum timeout value base on specification.

steven-guo-ec:
[config][tacacs+] Change tacacs+ minimum timeout value base on specification.

- What I did
    Change the minimum timeout value of TACACS+ to follow the value in specification.

- Why I did it
    The minimum timeout value of tacacs CLI is not the same as
    specification, so modify the CLI to make it follow it.

- How I verified it
   Verified the minimum value is 1 on switch.

Signed-off-by: Po-hong Guo <steven_guo@edge-core.com>

### DIFF
--- a/config/aaa.py
+++ b/config/aaa.py
@@ -94,10 +94,10 @@ tacacs.add_command(default)
 
 
 @click.command()
-@click.argument('second', metavar='<time_second>', type=click.IntRange(0, 60), required=False)
+@click.argument('second', metavar='<time_second>', type=click.IntRange(1, 60), required=False)
 @click.pass_context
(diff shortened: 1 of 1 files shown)


renukamanavalan:
Curious! 
"_minimum timeout value base on specification._" -- Which specification is being referred here ?

steven-guo-ec:
Refer to `config tacacs add` section, parameter `timout` in [Command Reference](https://github.com/Azure/sonic-utilities/blob/master/doc/Command-Reference.md#tacacs-config-commands).

PR #1572, opened 2021-04-20
Change to use rvtysh when calling the show commands

xumia:
#### What I did
Change to use rvtysh when calling the show commands

PR #1646, opened 2021-05-31
Change to use rvtysh when calling the show commands

xumia:
#### What I did
Change to use rvtysh when calling the show commands
Merge code from master and fix code conflict, see https://github.com/Azure/sonic-utilities/pull/1572


qiluo-msft (approved):

qiluo-msft:
Can we enable the build checker and let this PR pass?

xumia:
> Can we enable the build checker and let this PR pass?

Sure
How about the other branches?

PR #1647, opened 2021-05-31
[201911] Change to use rvtysh when calling the show commands (#1572)

xumia:
Change to use rvtysh when calling the show commands

#### What I did
Change to use rvtysh when calling the show commands
Merge code from master and fix code conflict, see #1572


qiluo-msft (approved):

qiluo-msft (changes requested):
We don't have `rvtysh` command on 201911 branch

xumia:
> We don't have `rvtysh` command on 201911 branch

We have merged the command 'vtysh' to 201911 branch, see https://github.com/Azure/sonic-buildimage/pull/7572

qiluo-msft (approved):

PR #1657, opened 2021-06-07
[LLDP] Fix lldpshow script to display multiple MAC addresses on the same remote physical interface

BasimShalata:
#### What I did

Fix an issue when having the following scenario:
1- remote interface has 2 MACs on the same physical interface.
2- "show lldp table" command displays one entry for only one MAC address

#### How I did it

Root cause:
"show lldp table" command uses lldpshow script to get, parse and display data from the lldp open source package (lldpctl script).
lldpctl script returns a proper info about the 2 MACs but the issue is with the lldpshow script parser where it built a dictionary which its key is the local physical interface.
Therefore when having 2 MACs, lldpctl will return 2 entries but the lldpshow parser will overwrite the first entry.

Fix:
Change the key to be a string of "interface#MAC".
This will enable having 2 entries for 2 different MAC addresses.
   
In addition:
- update display_sum()-->get_summary_output() to return a string instead of printing it directly. This to allow checking the returned value inside the new unit test.

#### How to verify it

- Test with the above scenario and check existence of the 2 MAC addresses in the lldp table.
- Add a new unit test for this scenario.

Output of "show lldp table" remains the same but after the fix will be able to see all MAC addresses learnt on the same physical interface:

# show lldp table
Capability codes: (R) Router, (B) Bridge, (O) Other
LocalPort    RemoteDevice           RemotePortID       Capability    RemotePortDescr
-----------  ---------------------    -----------------     ------------  -----------------
EthernetX                                   xx:xx:xx:xx:xx:xx
EthernetX                                   yy:yy:yy:yy:yy:yy

### DIFF
--- a/scripts/lldpshow
+++ b/scripts/lldpshow
@@ -127,53 +127,60 @@ class Lldpshow(object):
                 l_intf = intf.attrib['name']
                 if l_intf.startswith(BACKEND_ASIC_INTERFACE_NAME_PREFIX):
                     continue
-                self.lldpsum[l_intf] = {}
+                remote_port = intf.find('port')
+                r_portid = remote_port.find('id').text
+                key = l_intf + "#" + r_portid
+                self.lldpsum[key] = {}
+                self.lldpsum[key]['l_intf'] = l_intf
+                self.lldpsum[key]['r_portid'] = r_portid
                 chassis = intf.find('chassis')
                 capabs = chassis.findall('capability')
                 capab = self.parse_cap(capabs)
                 rmt_name = chassis.find('name')
                 if rmt_name is not None:
-                    self.lldpsum[l_intf]['r_name'] = rmt_name.text
+                    self.lldpsum[key]['r_name'] = rmt_name.text
                 else:
-                    self.lldpsum[l_intf]['r_name'] = ''
-                remote_port = intf.find('port')
-                self.lldpsum[l_intf]['r_portid'] = remote_port.find('id').text
+                    self.lldpsum[key]['r_name'] = ''
                 rmt_desc = remote_port.find('descr')
                 if rmt_desc is not None:
-                    self.lldpsum[l_intf]['r_portname'] = rmt_desc.text
+                    self.lldpsum[key]['r_portname'] = rmt_desc.text
                 else:
-                    self.lldpsum[l_intf]['r_portname'] = ''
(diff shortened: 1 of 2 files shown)


lguohan:
no unit test added?

BasimShalata:
I double checked your request and found the following:
1- Currently we don't have any unit test for lldp.
2- "show lldp table" is using the lldpctl script from the lldp package to display lldp entries.
it doesn't use any table from the redis and also we don't have any infrastructure for the lldp unit test to mock the lldpctl.

@lguohan,
do you have any suggestion?

liat-grozovik:
@jleveque kindly reminder to assist with pointers to unit test which can be used for this PR.

jleveque:
> @jleveque kindly reminder to assist with pointers to unit test which can be used for this PR.

You can mock the lines which call other utilities (like https://github.com/Azure/sonic-utilities/blob/bce123c946a348685675a7f399870cce3acee0da/scripts/lldpshow#L84), and return mocked-up output as if it had been returned by the utility. I can't find an example at the moment, but I'm sure there are some in the existing tests.

BasimShalata:
> > @jleveque kindly reminder to assist with pointers to unit test which can be used for this PR.
> 
> You can mock the lines which call other utilities (like
> 
> https://github.com/Azure/sonic-utilities/blob/bce123c946a348685675a7f399870cce3acee0da/scripts/lldpshow#L84
> 
> ), and return mocked-up output as if it had been returned by the utility. I can't find an example at the moment, but I'm sure there are some in the existing tests.

Added a new test which basically doesn't mock the lldpctl as there is more simple way which is:
1- new functionality was added to the parser method and this is the one we need to test.
2- in the unit test, I'm not calling lldp.get_info() which fill lldp.lldpraw DB with an XML output returned from the get_info() which calling lldpctl script, but I'm setting directlly lldp.lldpraw with the info I need to test.
3- in order to test the parser output, I changed the display_sum() method to return a value instead of printing it to the stdout and that's in order to test this returned value.

BasimShalata:
@lguohan I have added the unit test for this change.
can you please review.

lguohan:
@abdosi, can you review this one?

liat-grozovik:
@BasimShalata please handle conflicts.
@abdosi please provide your inputs. This fix is required for 202106.

lguohan (approved):
:shipit:

PR #1660, opened 2021-06-10
[Mellanox] Add different dumps files collected by SAI to techsupport

DavidZagury:
This PR depends on https://github.com/Azure/sonic-buildimage/pull/7844

#### What I did
Enhance the way techsupport collects dumps provided by SAI and do not keep them in the filesystem once the techsupport is created.

#### How I did it
- Create a temporary directory once techsupport script is executed, gather all files SAI dumps created, add them to the tech support and remove the temporary directory and files.
-  Remove direct call to FW trace collection as it is done now by SAI dump

#### How to verify it
Run techsupport, verify newly created files are part of it including mlxtrace which is no longer called directly by the techsupport script.

### DIFF
--- a/scripts/generate_dump
+++ b/scripts/generate_dump
@@ -758,10 +758,20 @@ enable_logrotate() {
 #  None
 ###############################################################################
 collect_mellanox() {
-    local sai_dump_filename="/tmp/sai_sdk_dump_$(date +"%m_%d_%Y_%I_%M_%p")"
+    local sai_dump_folder="/tmp/saisdkdump"
+    local sai_dump_filename="${sai_dump_folder}/sai_sdk_dump_$(date +"%m_%d_%Y_%I_%M_%p")"
+
+    ${CMD_PREFIX}docker exec -it syncd mkdir -p $sai_dump_folder
     ${CMD_PREFIX}docker exec -it syncd saisdkdump -f $sai_dump_filename
-    ${CMD_PREFIX}docker exec syncd tar Ccf $(dirname $sai_dump_filename) - $(basename $sai_dump_filename) | tar Cxf /tmp/ -
-    save_file $sai_dump_filename sai_sdk_dump true
+
+    copy_from_docker syncd $sai_dump_folder $sai_dump_folder
+    echo "$sai_dump_folder"
+    for file in `ls $sai_dump_folder`; do
+        save_file ${sai_dump_folder}/${file} sai_sdk_dump true 
+    done
+
+    ${CMD_PREFIX}rm -rf $sai_dump_folder
+    ${CMD_PREFIX}docker exec -it syncd rm -rf $sai_dump_folder
 
     local mst_dump_filename="/tmp/mstdump"
     local max_dump_count="3"


liat-grozovik (changes requested):
please have single directory for all SAI and below dumps as they comes from different tools.
we should have it also for additional dumps to be added in the future.

DavidZagury:
> please have single directory for all SAI and below dumps as they comes from different tools.
> we should have it also for additional dumps to be added in the future.

@liat-grozovik I have changed the code according to this comment

scripts/generate_dump
```diff
@@ -758,11 +758,28 @@ enable_logrotate() {
...
+        save_file ${sai_dump_folder}/${file} sai_sdk_dump true 
+    done
+
+    ${CMD_PREFIX}rm -rf $sai_dump_folder
+    ${CMD_PREFIX}docker exec -it syncd rm -rf $sai_dump_folder
+
+    local mlxtrace_file=`${CMD_PREFIX}docker exec syncd find /tmp -type f -iname "mt*_pci_cr0_mlxtrace.trc"`
+    if [[ $mlxtrace_file ]]; then
+        copy_from_docker syncd $mlxtrace_file $mlxtrace_file
+        save_file $mlxtrace_file mlxtrace false
```
BasimShalata, line 779: I think these commands are redundant now after collecting all logs from sai dump dir.
@DavidZagury please double check it.

scripts/generate_dump
```diff
@@ -758,11 +758,28 @@ enable_logrotate() {
...
+    copy_from_docker syncd $sai_dump_folder $sai_dump_folder
+    echo "$sai_dump_folder"
+    for file in `ls $sai_dump_folder`; do
+        save_file ${sai_dump_folder}/${file} sai_sdk_dump true 
+    done
+
+    ${CMD_PREFIX}rm -rf $sai_dump_folder
+    ${CMD_PREFIX}docker exec -it syncd rm -rf $sai_dump_folder
+
+    local mlxtrace_file=`${CMD_PREFIX}docker exec syncd find /tmp -type f -iname "mt*_pci_cr0_mlxtrace.trc"`
```
liat-grozovik: why do you need to call it and not SAI?

DavidZagury: It was a leftover from the old implementation that I forgot to remove when I copied the final code from the switch to the commit

scripts/generate_dump
```diff
@@ -758,10 +758,20 @@ enable_logrotate() {
...
+    copy_from_docker syncd $sai_dump_folder $sai_dump_folder
+    echo "$sai_dump_folder"
+    for file in `ls $sai_dump_folder`; do
+        save_file ${sai_dump_folder}/${file} sai_sdk_dump true 
+    done
+
+    ${CMD_PREFIX}rm -rf $sai_dump_folder
+    ${CMD_PREFIX}docker exec -it syncd rm -rf $sai_dump_folder
 
     local mst_dump_filename="/tmp/mstdump"
```
liat-grozovik, line 776: @DavidZagury can you please check if the mstdump is also needed here or already covered by the SAI dump once the extended API is called.

BasimShalata: @liat-grozovik,

As you see the mst dump is not part of the extended SAI dump.
the change that we can do is to create the mst logs into the same dir for the SAI dump but it will only save one code line "save_file "${mst_dump_filename}${i}" mstdump true" with the same performance :)
I don't think its a necessary change.

DavidZagury: Since the mst files are not just gathered here to the tar file but are also being created in this script by SONiC (and not by SAI), it would mean to move the mst code to before the new loop and changing it to this:
```
local max_dump_count="3"
for i in $(seq 1 $max_dump_count); do
        ${CMD_PREFIX}/usr/bin/mstdump /dev/mst/mt*conf0 &gt; "${sai_dump_folder}${i}"
done
```
This would also change the location that the mst files are being saved into.
It wouldn't be a significant change, and would not really make the code any simpler.

DavidZagury: If SAI will add this to the file they are creating we could remove all this code.

liat-grozovik: Agree it can be later enhanced. For now LGTM

liat-grozovik (approved):

PR #1674, opened 2021-06-15
Fixing db_migrator for Feature table

dgsudharsan:
#### What I did
Fixing db_migrator.py for FEATURE table. In case of version_unknown, all versions' migrator APIs would be invoked which will also invoke the feature_table migration. Without the field check for 'status' the current logic will set it 'disabled' and will set 'state' field too to be 'disabled' which will in turn disable all the features bringing down the system.

#### How I did it
Added check to migrate only if 'status' field is present.

#### How to verify it
Install the image through onie and confirm if services are up

### DIFF
--- a/scripts/db_migrator.py
+++ b/scripts/db_migrator.py
@@ -180,9 +180,10 @@ class DBMigrator():
         '''
         feature_table = self.configDB.get_table('FEATURE')
         for feature, config in feature_table.items():
-            state = config.pop('status', 'disabled')
-            config['state'] = state
-            self.configDB.set_entry('FEATURE', feature, config)
+            if 'status' in config:
+                state = config.pop('status', 'disabled')
+                config['state'] = state
+                self.configDB.set_entry('FEATURE', feature, config)
 
         container_feature_table = self.configDB.get_table('CONTAINER_FEATURE')
         for feature, config in container_feature_table.items():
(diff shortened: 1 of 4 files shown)


dgsudharsan:
This needs to be cherry-picked to 202012

liat-grozovik:
@qiluo-msft can you please review and help to merge? this is a show stopper issue for fresh installation from ONIE

scripts/db_migrator.py
```diff
@@ -180,9 +180,10 @@ def migrate_feature_table(self):
         '''
         feature_table = self.configDB.get_table('FEATURE')
         for feature, config in feature_table.items():
-            state = config.pop('status', 'disabled')
-            config['state'] = state
-            self.configDB.set_entry('FEATURE', feature, config)
+            if 'status' in config:
+                state = config.pop('status', 'disabled')
```
qiluo-msft, line 184: Can we fix this bug by change default value to `enabled`?

dgsudharsan: I believe we can remove the default value. The if check will make sure that 'status' is always present. Let me make the change

liat-grozovik:
@qiluo-msft can we have a PR against 202012 and not wait for master sonic-utilities stability? 202012 is currently not working properly and will have production issues.

qiluo-msft (approved):

PR #1675, opened 2021-06-15
[CI] sonic-config-engine now depends on SONiC YANG packages

jleveque:
sonic-config-engine now depends on sonic-yang-mgmt and sonic-yang-models, so we now need to install them before we install sonic-config-engine as part of continuous integration to get the CI working again


jleveque:
This PR fixes the CI, but now the failures appear to be real issues.

lguohan:
test not related to the pr, unblock.

PR #1676, opened 2021-06-15
[202012]Fixing db_migrator for Feature table 

dgsudharsan:
This is a cherrypick of https://github.com/Azure/sonic-utilities/pull/1674 for 202012 since the master checks are failing and we need to unblock the 202012 branch

#### What I did
Fixing db_migrator.py for FEATURE table. In case of version_unknown, all versions' migrator APIs would be invoked which will also invoke the feature_table migration. Without the field check for 'status' the current logic will set it 'disabled' and will set 'state' field too to be 'disabled' which will in turn disable all the features bringing down the system.

#### How I did it
Added check to migrate only if 'status' field is present.

#### How to verify it
Install the image through onie and confirm if services are up

### DIFF
--- a/scripts/db_migrator.py
+++ b/scripts/db_migrator.py
@@ -181,9 +181,10 @@ class DBMigrator():
         '''
         feature_table = self.configDB.get_table('FEATURE')
         for feature, config in feature_table.items():
-            state = config.pop('status', 'disabled')
-            config['state'] = state
-            self.configDB.set_entry('FEATURE', feature, config)
+            if 'status' in config:
+                state = config.pop('status', 'disabled')
+                config['state'] = state
+                self.configDB.set_entry('FEATURE', feature, config)
 
         container_feature_table = self.configDB.get_table('CONTAINER_FEATURE')
         for feature, config in container_feature_table.items():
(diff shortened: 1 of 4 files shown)


scripts/db_migrator.py
```diff
@@ -181,9 +181,11 @@ def migrate_feature_table(self):
...
         feature_table = self.configDB.get_table('FEATURE')
         for feature, config in feature_table.items():
-            state = config.pop('status', 'disabled')
-            config['state'] = state
-            self.configDB.set_entry('FEATURE', feature, config)
+            state = config.get('status')
+            if state is not None:
+                config['state'] = state
+                config.pop('status')
+                self.configDB.set_entry('FEATURE', feature, config)
```
yxieca, line 188: The behavior is different, the removed code default status to 'disabled'. The added change didn't do that. Is this change intentional? If so, please clarify the intention.

dgsudharsan: > The behavior is different, the removed code default status to 'disabled'. The added change didn't do that. Is this change intentional? If so, please clarify the intention.

Ying, The db_migrator should modify the field only if present. That's why the if check is added. If the field is not present in FEATURE_TABLE the db_migrator shouldn't assume it to be default disabled and leave it to the specific application which handles it to handle the default.

liat-grozovik:
@yxieca FYI the PR was validated internally and the issue seen on 202012 is now fixed. 
Please let us know if you have any additional concern.

yxieca (approved):

qiluo-msft (approved):

PR #1683, opened 2021-06-18
[counterpoll] Adjust port-buffer-drop counter default interval from 60000ms to 10000ms

Junchao-Mellanox:
#### What I did

There is a performance optimization for port buffer drop counter. Adjust port-buffer-drop counter default interval from 60000ms to 10000ms

#### How I did it

1. Change CLI argument range
2. Change CLI show
3. Change unit test

#### How to verify it

Manual test and unit test

### DIFF
--- a/counterpoll/main.py
+++ b/counterpoll/main.py
@@ -90,14 +90,10 @@ def port_buffer_drop():
     """ Port buffer drop  counter commands """
 
 @port_buffer_drop.command()
-@click.argument('poll_interval', type=click.IntRange(30000, 300000))
+@click.argument('poll_interval', type=click.IntRange(1000, 300000))
 def interval(poll_interval):
     """
     Set port_buffer_drop counter query interval
-    This counter group causes high CPU usage when polled,
-    hence the allowed interval is between 30s and 300s.
-    This is a short term solution and
-    should be changed once the performance is enhanced
     """
     configdb = ConfigDBConnector()
     configdb.connect()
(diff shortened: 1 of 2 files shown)


liat-grozovik:
@Junchao-Mellanox  please explain how the system will behave when you have an upgrade from older version to this one. will it use these values? will it overrides user configuration if it is not default? we need to understand the flow before suggesting if to add db migrator or just document the behaviour somewhere.

Junchao-Mellanox:
> @Junchao-Mellanox please explain how the system will behave when you have an upgrade from older version to this one. will it use these values? will it overrides user configuration if it is not default? we need to understand the flow before suggesting if to add db migrator or just document the behaviour somewhere.

When upgrade from older version to this one:

- if there was a config in DB, it still use the interval configured in DB
- if there was no config in DB, it will use the new interval by default

liat-grozovik:
@lguohan and @neethajohn this improvements comes with a performance improvements on Nvidia switch SAI and below. The the idea of improving the counter polling is to provide better resolution of counters for the user. What should be done in order to approve it on other platforms? Can you suggest owners so they can verify and approve it is working for them as well? 
If not, should we consider different counter polling per vendor?

keboliu (approved):

neethajohn:
> @lguohan and @neethajohn this improvements comes with a performance improvements on Nvidia switch SAI and below. The the idea of improving the counter polling is to provide better resolution of counters for the user. What should be done in order to approve it on other platforms? Can you suggest owners so they can verify and approve it is working for them as well?
> If not, should we consider different counter polling per vendor?

@liat-grozovik , I believe this was discussed in one of our meetings and Guohan suggested to add a testcase that could check the cpu usage with this new interval and we could use that to validate if there is any issue on other asics. Based on that, I had provided a comment in https://github.com/Azure/sonic-swss/pull/1792

andriyz-nv:
Hi @neethajohn  and @lguohan can you please provide a path in sonic-mgmt original test that is supposed to check similar behavior? Is it https://github.com/Azure/sonic-mgmt/blob/master/tests/platform_tests/test_cpu_memory_usage.py ?

PR #1689, opened 2021-06-22
[voq] VOQ cli show commands implementation

vganesan-nokia:
#### What I did

Implementation of the following show commands for VOQ chassis.

   - show chassis system-ports [\<system port\>] {-n \<namespace\>}, if multi-asic voq chassis
   - show chassis system-ports [\<system port\>] , if single asic voq chassis
   - show chassis system-neighbors [\<nbr ip address\>] [-n \<asic name\>]
   - show chassis system-lags [\<system lag\>] [-n \<asic name\>] [-l \<linecard or host name\>]

   The existing chassis  commands under "show chassis_modules" and
   "config chassis_modules" have been renamed (moved) under
   "show chassis modules" and "config chassis modules" respectively

#### How I did it

Modified sonic utilities scripts 

#### How to verify it

- In VOQ chassis have system ports and port channel configurations. Have some neighbors attached to some front panel ports.
- Run the CLI commands mentioned above form line card. The commands will show the outputs as mentioned in the design document update PRs https://github.com/Azure/SONiC/pull/801, https://github.com/Azure/SONiC/pull/802 and https://github.com/Azure/SONiC/pull/805

Commands changes:
- The commands under **show chassis_modules** have been moved\/re-named to **show chassis modules**
- The commands under **config chassis_modules** have been moved\/re-named to **config chassis modules**

No change in the output of the existing commands.


anshuv-mfst:
Hi @ysmanman - could you please review , thanks

tests/chassis_modules_test.py
```diff
@@ -39,6 +40,70 @@
 SUPERVISOR0  192.168.1.100            True
 """
 
+show_chassis_system_ports_output_all="""\
+            System Port Name    Port Id    Switch Id    Core    Core Port    Speed
+----------------------------  ---------  -----------  ------  -----------  -------
+   Linecard1|Asic0|Ethernet0          1            0       0            1     100G
+   Linecard1|Asic0|Ethernet0          1            0       0            1     100G
```
arlakshm, line 47: The same output lines are repeated, is this because the same system-port information is present in every asic ?
This output is be confusing.
Can we  either add
 -  a new column `asic_name` to indicate which asic this info is retrived or 
 - make the `-n <asicname>` mandatory for the multi asic linecards , so that the we don't display same system port information multiple times

scripts/voqutil
```diff
@@ -0,0 +1,337 @@
...
+        return "N/A"
+    if status_type == SYSTEM_PORT_SPEED and status != "N/A":
+       status = '{}G'.format(status[:-3])
+    return status
+
+def chassis_app_db_keys_get_system_neigh(chassis_app_db):
+    """
+    Get CHASSIS_APP_DB SYSTEM_NEIGH table Keys
+    """
+    system_neigh_keys = chassis_app_db.keys(chassis_app_db.CHASSIS_APP_DB, "SYSTEM_NEIGH|*")
```
arlakshm, line 88: Use `SYSTEM_NEIGH_TABLE_PREFIX ` iso `SYSTEM_NEIGH|` here ?

scripts/voqutil
```diff
@@ -0,0 +1,337 @@
...
+
+header_system_neigh = ['System Port Interface', 'Neighbor', 'MAC', 'Encap Index']
+
+class SystemNeighStatus(object):
+
+    def __init__(self, namespace_option, display_option):
+        self.db = None
+        self.config_db = None
+        self.table = []
+        self.multi_asic = multi_asic_util.MultiAsic(
```
arlakshm, line 206: Looks `self.multi_asic` like this is not used ?

scripts/voqutil
```diff
@@ -0,0 +1,337 @@
...
+        Class constructor method
+        :param self:
+        :param intf_name: string of system lag name
+        :return:
+        """
+        self.db = None
+        self.config_db = None
+        self.system_lag_name = system_lag_name
+        self.table = []
+        self.multi_asic = multi_asic_util.MultiAsic(
```
arlakshm, line 264: Same as above, looks `self.multi_asic` like this is not used ?

arlakshm (commented):
as comments

tests/chassis_modules_test.py
vganesan-nokia: Yes. The same information will be repeated multiple times - as many times as the number of asics. This is because we have identical system port config in all asics. We display this multiple times so that one can verify for the requirement of indentical system port configuration in all asics.

The option to make the  <asic name> mandatory seems better than displaying all entries with new column since there will be lots of entries if there are many asics.

I'll implement changes to make the asic name mandatory in the command. Thanks for the suggestions.

vganesan-nokia: Fixed. Namespace option is made "required" option.

scripts/voqutil
vganesan-nokia: Fixed, here and in other places also.

scripts/voqutil
vganesan-nokia: Fixed. Removed since this is not  used

scripts/voqutil
vganesan-nokia: Fixed. Removed since this is not used.

scripts/voqutil
```diff
@@ -0,0 +1,369 @@
...
+        else:
+            mock_tables.dbconnector.load_database_config()
+
+except KeyError:
+    pass
+
+
+# ========================== Common VOQ util logic ==========================
+
+SYSTEM_PORT_TABLE_PREFIX = "SYSTEM_PORT_TABLE:"
```
ysmanman, line 34: Is the separator or table name defined in some python lib? If so, it would be great to import them, instead of hardcoding them.

arlakshm: @vganesan-nokia can you address @ysmanman's comment ?

show/chassis_modules.py
```diff
@@ -102,3 +108,55 @@ def midplane_status(chassis_module_name):
...
+@click.option('--verbose', is_flag=True, help="Enable verbose output")
+def system_neighbors(asicname, ipaddress, verbose):
+    """Show VOQ system neighbors information"""
+
+    cmd = "voqutil -c system_neighbors"
+
+    if ipaddress is not None:
+        cmd += " -a {}".format(ipaddress)
+
+    if asicname is not None:
```
arlakshm, line 142: @vganesan-nokia this change will not work on the single asic linecards right ?
I think it will be better to the add the asicname check only on multi asic linecards ?

tests/chassis_modules_test.py
```diff
@@ -39,6 +40,77 @@
...
+Linecard4|Asic2|PortChannel0001         2           22  Linecard4|Asic2|Ethernet29, Linecard4|Asic2|Ethernet30
+"""
+
+show_chassis_system_lags_output_asic1="""\
+                System Lag Name    Lag Id    Switch Id                                     Member System Ports
+-------------------------------  --------  -----------  ------------------------------------------------------
+Linecard2|Asic1|PortChannel0002         1            8  Linecard2|Asic1|Ethernet16, Linecard2|Asic1|Ethernet17
+"""
+
 class TestChassisModules(object):
```
arlakshm, line 114: Do we need to add tests for single asic line cards ?

arlakshm (commented):
as comments

show/chassis_modules.py
```diff
@@ -102,3 +108,55 @@ def midplane_status(chassis_module_name):
...
+        cmd += " -x {}".format(asicname)
+
+    clicommon.run_command(cmd, display_cmd=verbose)
+
+@chassis.command()
+@click.argument('systemlagname', required=False)
+@click.option('--asicname', '-x', 'asicname', default=None, type=str, show_default=False, help='Asic name')
```
arlakshm, line 149: why not use `-n` here ? similar to `show chassis system_ports`

show/chassis_modules.py
```diff
@@ -102,3 +108,55 @@ def midplane_status(chassis_module_name):
...
+        cmd += " -i \"{}\"".format(systemportname)
+    
+    if namespace is not None:
+        cmd += " -n {}".format(namespace)
+
+    clicommon.run_command(cmd, display_cmd=verbose)
+
+@chassis.command()
+@click.argument('ipaddress', required=False)
+@click.option('--asicname', '-x', 'asicname', default=None, type=str, show_default=False, help='Asic name')
```
arlakshm, line 132: why not use `-n` here ? similar to `show chassis system_ports`

scripts/voqutil
vganesan-nokia: Yes. I'll fix this if the table names are defined in some python lib.

show/chassis_modules.py
vganesan-nokia: Please note that for system neighbors and system lags this is not namespace. It is the value of DEVICE_METADATA['localhost']['asic_name'] we filter on in keys .  Therefore, this will/should work on single asic linecards also for the following reasons: (1) Even for the single asic line card, in voq chassis, it is mandatory to configure DEVICE_METADATA['localhost']['asic_name'].  For single asic linecards, it may be always "asic0" or "Asic0". (2) Even if the command is executed from single asic linecard, we should be able to filter the entries for asic name of other line cards that have multiple asics.

tests/chassis_modules_test.py
vganesan-nokia: Yes. Please see my response to comment on chassis_modules.py:139-142

show/chassis_modules.py
vganesan-nokia: For system neighbors and system lags the info are retrieved from supervisor card's chassis app db which is not in any particular namespace. The 'asicname' here is not the namespace. It is the 'asic_name' we look for in the keys of the entries.

show/chassis_modules.py
vganesan-nokia: Please see my response to comments on line# 146-149

scripts/voqutil
vganesan-nokia: Fixed for table names

show/chassis_modules.py
vganesan-nokia: Changed "asicname" option selector to "-n" as suggested here and in sonic chassis subgroup review meeting held on 08/04/2021

show/chassis_modules.py
vganesan-nokia: Changed "asicname" option selector to "-n" as suggested

arlakshm (approved):

PR #1697, opened 2021-06-28
[show] Allow system with no ports in config db run without errors

liorghub:
**What I did**
Allow system with no ports in config db run without errors.
This is needed for modular system which should boot properly without line cards.

**How I did it**
Do not raise warning if ports dictionary is empty.

**How to verify it**
Run show interfaces status

### DIFF
--- a/utilities_common/cli.py
+++ b/utilities_common/cli.py
@@ -125,8 +125,7 @@ class InterfaceAliasConverter(object):
             self.port_dict = self.config_db.get_table('PORT')
         self.alias_max_length = 0
 
-
-        if not self.port_dict:
+        if self.port_dict is None:
             click.echo(message="Warning: failed to retrieve PORT table from ConfigDB!", err=True)
             self.port_dict = {}
 


prsunny:
please provide previous and new output as mentioned in the template

utilities_common/cli.py
```diff
@@ -125,8 +125,7 @@ def __init__(self, db=None):
             self.port_dict = self.config_db.get_table('PORT')
         self.alias_max_length = 0
 
-
-        if not self.port_dict:
+        if self.port_dict is None:
```
dgsudharsan, line 128: If the get_table API doesn't return None, this check may not be needed and we can remove it altogether. Can you please confirm?

liorghub: I agree. Both get_table() and get_port_table() initialize an empty dictionary which is being returned even if 'PORT' table is empty. I will remove this check.

caizhenghui-juniper:
Is this the only change needed for empty port config in the show commands? Or this is just enough to allow system to come up?

liorghub:
@caizhenghui-juniper 
This is the only change needed in sonic utilities (not just to allow clean init).

liorghub:
@prsunny
Can you please approve this change?

PR #1700, opened 2021-06-29
[Config] Update config command of kdump

yozhao101:
Signed-off-by: Yong Zhao <yozhao@microsoft.com>

#### What I did
This PR aims to update the `config` command of Kdump (Linux kernel dump) mechanism.

#### How I did it
Before updating the value of a field in the table of `KDUMP`, we need decide whether the `KDUMP` table and corresponding field does exist in the table.

#### How to verify it
I verifies this on device `str-msn2700-03` and unit test passed.
        
        tests/kdump_test.py::TestKdump::test_config_kdump_disable PASSED                                                                                                 
        tests/kdump_test.py::TestKdump::test_config_kdump_enable PASSED                                                                                                  
        tests/kdump_test.py::TestKdump::test_config_kdump_memory PASSED                                                                                                  
        tests/kdump_test.py::TestKdump::test_config_kdump_num_dumps PASSED

### DIFF
--- a/config/kdump.py
+++ b/config/kdump.py
@@ -1,4 +1,3 @@
-import os
 import sys
 
 import click
--- a/config/kdump.py
+++ b/config/kdump.py
@@ -1,97 +1,90 @@
 import sys
 
 import click
-import utilities_common.cli as clicommon
-from swsscommon.swsscommon import ConfigDBConnector
+from utilities_common.cli import AbbreviationGroup, pass_db
 
 
-@click.group(cls=clicommon.AbbreviationGroup, name="kdump")
+#
+# 'kdump' group ('sudo config kdump ...')
+#
+@click.group(cls=AbbreviationGroup, name="kdump")
 def kdump():
-    """Modify configuration of kdump"""
+    """Configure the KDUMP mechanism"""
     pass
 
+#
+# 'disable' command ('sudo config kdump disable')
+#
+@kdump.command(name="disable", short_help="Disable the KDUMP mechanism")
+@pass_db
+def kdump_disable(db):
+    """Disable the KDUMP mechanism"""
+    kdump_table = db.cfgdb.get_table("KDUMP")
+    if not kdump_table:
+        click.echo("Unable to retrieve 'KDUMP' table from Config DB.")
+        sys.exit(1)
 
-@kdump.command()
-def disable():
-    """Disable kdump feature"""
-    config_db = ConfigDBConnector()
-    if config_db:
-        config_db.connect()
-        kdump_table = config_db.get_table("KDUMP")
-        if not kdump_table:
-            click.echo("Unable to retrieve KDUMP table from CONFIG DB.")
-            sys.exit(1)
+    if "config" not in kdump_table:
+        click.echo("Unable to retrieve key 'config' from KDUMP table.")
+        sys.exit(2)
 
-        if "config" not in kdump_table:
-            click.echo("Unable to retrieve key 'config' from KDUMP table.")
-            sys.exit(2)
+    db.cfgdb.mod_entry("KDUMP", "config", {"enabled": "false"})
 
-        config_db.mod_entry("KDUMP", "config", {"enabled": "false"})
-    else:
-        click.echo("Unable to get an instance of 'ConfigDBConnector'.")
+
+#
+# 'enable' command ('sudo config kdump enable')
+#
+@kdump.command(name="enable", short_help="Enable the KDUMP mechanism")
+@pass_db
+def kdump_enable(db):
+    """Enable the KDUMP mechanism"""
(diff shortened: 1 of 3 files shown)


config/kdump.py
```diff
@@ -1,44 +1,69 @@
-import os
+import sys
+
 import click
 import utilities_common.cli as clicommon
-from swsscommon.swsscommon import ConfigDBConnector
+from swsssdk import ConfigDBConnector
```
qiluo-msft, line 5: >swsssdk [](http://example.com/codeflow?start=5&length=7)

swsssdk is deprecated. Please use libswsscommon instead.

config/kdump.py
```diff
@@ -1,44 +1,69 @@
...
+    pass
+
 
 @kdump.command()
 def disable():
-    """Disable kdump operation"""
+    """Disable kdump feature"""
     config_db = ConfigDBConnector()
-    if config_db is not None:
+    if config_db:
```
qiluo-msft, line 18: >if [](http://example.com/codeflow?start=4&length=2)

If the condition is false, you silently exit. Is it an error?

config/kdump.py
```diff
@@ -1,44 +1,69 @@
...
         config_db.connect()
+        kdump_table = config_db.get_table("KDUMP")
+        if "config" not in kdump_table:
```
qiluo-msft, line 21: >"config" not in kdump_table [](http://example.com/codeflow?start=11&length=27)

In what situation user will meet this error? This `config` command will exit with error code, what should user do to make it working?

config/kdump.py
```diff
@@ -1,44 +1,69 @@
-import os
+import sys
```
qiluo-msft, line 1: >sys [](http://example.com/codeflow?start=7&length=3)

Since you are fixing a bug. Please add a test case which could fail previous code, but pass with new code.

qiluo-msft (commented):
As comment

config/kdump.py
yozhao101: Good catch! If this condition is false, we need show an error message to user instead of exiting silently.

config/kdump.py
yozhao101: Updated!

config/kdump.py
yozhao101: If the configuration of `KDUMP` is not present  in `CONIFG_DB` or some keys are missing , then this error will be shown to user. 

The user can try to run the command `sudo config reload -y` to refresh the `CONFIG_DB` to solve this issue.

config/kdump.py
yozhao101: Added the unit test cases.

config/kdump.py
```diff
@@ -1,44 +1,90 @@
...
+def kdump_memory(db, kdump_memory):
+    """Reserve memory for kdump capture kernel"""
+    kdump_table = db.cfgdb.get_table("KDUMP")
+    if not kdump_table:
+        click.echo("Unable to retrieve KDUMP table from CONFIG DB.")
+        sys.exit(5)
+
+    if "config" not in kdump_table:
+        click.echo("Unable to retrieve key 'config' from KDUMP table.")
+        sys.exit(6)
```
qiluo-msft, line 68: Could you reuse code? I see it 4 times in this file.

yozhao101: Updated!

qiluo-msft (approved):

tests/mock_tables/config_db.json
```diff
@@ -636,6 +636,11 @@
     "DEBUG_COUNTER_DROP_REASON|DEBUG_2|IP_HEADER_ERROR": {},
     "DEBUG_COUNTER_DROP_REASON|DEBUG_2|NO_L3_HEADER": {},
     "DEBUG_COUNTER_DROP_REASON|lowercase_counter|L2_ANY": {},
+    "KDUMP|config": {
+	"enabled": "false",
+	"memory": "256MB",
+	"num_dumps": "3"
```
qiluo-msft, line 642: Please indent like others in this file.

qiluo-msft: I see you mix TAB with SPACE

qiluo-msft (changes requested):
One minor issue added

tests/mock_tables/config_db.json
yozhao101: Good catch! Updated.

qiluo-msft (approved):

qiluo-msft:
This PR could not be cleanly cherry-pick to 202012. Please submit another PR.

yozhao101:
> This PR could not be cleanly cherry-pick to 202012. Please submit another PR.

Please review this PR against 202012: https://github.com/Azure/sonic-utilities/pull/1778.

PR #1707, opened 2021-07-05
[multi-asic][cli][chassis-db] Avoid connecting to chassis db for cli commands executed from linecard

vganesan-nokia:
#### What I did

Currently, for all the cli commands, we connect to all databases mentioned in the database_config.json. The database_config.json also includes the databases from chassis redis server from supervisor card. It is unneccessary to connect to databases from chassis redis server when cli commands are executed form linecard. But we need to allow connection to chassis databases when the cli commands are executed from supervisor card. This problem is fixed by this PR

#### How I did it

The constructor of Db() class which is instantiated for every CLI command execution is changed to skip chassis databases from the list of collected databases if the card is not supervisor card. This PR is dependent on sonic-buildimage PR https://github.com/Azure/sonic-buildimage/pull/8065.

#### How to verify it
In voq chassis execute some CLI commands (show, config, etc.) and make sure that commands are executed without any CLI errors no matter whether there is supervisor card or not.

No command syntax or outputs changed

N/A

### DIFF
--- a/utilities_common/db.py
+++ b/utilities_common/db.py
@@ -1,4 +1,4 @@
-from sonic_py_common import multi_asic
+from sonic_py_common import multi_asic, device_info
 from swsscommon.swsscommon import ConfigDBConnector, SonicV2Connector
 from utilities_common import constants
 from utilities_common.multi_asic import multi_asic_ns_choices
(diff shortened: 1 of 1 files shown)


anshuv-mfst:
Hi @mlorrillere - could you please review, thanks.

mlorrillere (dismissed):

utilities_common/db.py
```diff
@@ -11,7 +11,17 @@ def __init__(self):
         self.cfgdb = ConfigDBConnector()
         self.cfgdb.connect()
         self.db = SonicV2Connector(host="127.0.0.1")
-        for db_id in self.db.get_db_list():
+
+        # Skip connecting to chassis databases in line cards
+        db_list = list(self.db.get_db_list())
+        if not device_info.is_supervisor():
+            try:
+                db_list.remove('CHASSIS_APP_DB')
```
judyjoseph, line 19: Similar comment, we can have a function in py_common/multi_asic.py or py_common/deviceinfo to get the list of chassis db instances and use here. Will help in future when we add more DBs

vganesan-nokia:
@judyjoseph, would you please re-start the azure pipeline again? The build is still not picking up the dependednt changes from Azure/sonic-buildimage#8065. Thanks

PR #1709, opened 2021-07-07
Fixed IP validation in "config interface ip add/remove"

akokhan:
Signed-off-by: Andriy Kokhan <andriyx.kokhan@intel.com>

#### Why I did it
After executing command that should delete ipv6 address from LAG, ipv6 address still remains.
This affects t0 VLAN TCs.

#### What I did
Reworked IP validation for "config interface ip remove" command added by https://github.com/Azure/sonic-utilities/pull/1414

Since the IP address is used as a part of a key in Redis DB, we should not modify IP address format provided by the user (e.g., convert "FC00::71/126" to "fc00::71/126" which causes VLAN TCs failure). At the same time we probably should not tolerate extra zeros in IPv4...

So, reworked IP validation in "config interface ip add/remove" command:
      - Renamed validate_ip_mask() to is_valid_ip_interface() as per code style
      - Updated is_valid_ip_interface() to do not modify the IP address
      - Updated UTs per changes

#### How to verify it
deploy t0
sudo config interface ip remove PortChannel0001 FC00::71/126

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -804,21 +804,26 @@ def validate_mirror_session_config(config_db, session_name, dst_port, src_port, 
 
     return True
 
-def validate_ip_mask(ctx, ip_addr):
+def is_valid_ip_interface(ctx, ip_addr):
     split_ip_mask = ip_addr.split("/")
+    if len(split_ip_mask) < 2:
+        return False
+
     # Check if the IP address is correct or if there are leading zeros.
     ip_obj = ipaddress.ip_address(split_ip_mask[0])
 
+    if isinstance(ip_obj, ipaddress.IPv4Address):
+        # Since the IP address is used as a part of a key in Redis DB,
+        # do not tolerate extra zeros in IPv4.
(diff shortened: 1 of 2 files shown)


akokhan:
@prsunny , @jleveque , @d-dashkov ,
please approve and merge.

akokhan:
The fix must be cherry-picked into 202012 as well

prsunny (dismissed):

akokhan:
In fact, it's wrong approach to convert v6 addresses to upper case as well. Looks like RIF can be loaded from config_db.json into config_db in both lower and upper formats. Ideally, we should not change case of letters.

akokhan:
@lguohan , please review.

prsunny:
@d-dashkov can u pls review?

d-dashkov (approved):

bingwang-ms (approved):

bingwang-ms:
Merged, thanks for the fix

PR #1711, opened 2021-07-13
Coverage uses top level directory as source

qiluo-msft:
#### What I did
Originally, the Azure pipeline "Code Coverage" page will render folder structure wrongly. The first level of folders disappear and all the files in the first level are combined and de-dup. So we lost many main.py, etc.

Fix the coveragerc file so that the rendering will make sense.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,6 @@
 [run]
 branch = True
 source = .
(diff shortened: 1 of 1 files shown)


lguohan (approved):
:shipit:

PR #1718, opened 2021-07-19
Add script null_route_helper

bingwang-ms:
This is a backport of https://github.com/Azure/sonic-utilities/pull/1737

Signed-off-by: bingwang <bingwang@microsoft.com>

#### What I did
This PR introduced a new helper script ```null_route_helper```.

```null_route_helper```  is a utility for blocking and unblocking traffic from given source ip_prefix on ACL tables.
The block operation will insert a DENY rule at the top of the table. The unblock operation will remove an existing DENY rule that has been created by the block operation (i.e. it does NOT insert an ALLOW rule, only removes DENY rules).
Since SONiC supports multi ACL rules share the same priority, all ACL rules created by null_route_helper will use the highest priority(9999).

Example:
```
Block traffic from 10.2.3.4/32:
./null_route_helper acl_table_name block 10.2.3.4/32

Unblock all traffic from 10.2.3.4/32:
./null_route_helper acl_table_name unblock 10.2.3.4/32
```

#### How I did it
The feature is implemented with applying ACL rules.

#### How to verify it
Verified with both unit test and traffic test
```
tests/null_route_helper_test.py::test_ip_validation PASSED                                                                                                                                      [ 11%]
tests/null_route_helper_test.py::test_ip_ver PASSED                                                                                                                                             [ 22%]
tests/null_route_helper_test.py::test_check_table_existence PASSED                                                                                                                              [ 33%]
tests/null_route_helper_test.py::test_build_rule PASSED                                                                                                                                         [ 44%]
tests/null_route_helper_test.py::test_get_rule PASSED                                                                                                                                           [ 55%]
tests/null_route_helper_test.py::test_run_when_table_absent PASSED                                                                                                                              [ 66%]
tests/null_route_helper_test.py::test_run_with_invalid_ip PASSED                                                                                                                                [ 77%]
tests/null_route_helper_test.py::test_block PASSED                                                                                                                                              [ 88%]
tests/null_route_helper_test.py::test_unblock PASSED                                                                                                                                            [100%]
```
```
Name                                          Stmts   Miss Branch BrPart  Cover
-------------------------------------------------------------------------------
scripts/null_route_helper                        96      4     24      1    96%
```
The coverage is not 100 since below line can't be covered
```
if __name__ == "__main__":
    cli()
```

### DIFF
--- a/scripts/null_route_helper
+++ b/scripts/null_route_helper
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 
 """
-Utility for blocking and unblocking traffic from given source ip_prefix on ACL tables.
+Utility for blocking and unblocking traffic from given source ip address on ACL tables.
 
 The block operation will insert a DENY rule at the top of the table. The unblock operation
 will remove an existing DENY rule that has been created by the block operation (i.e. it does
@@ -13,10 +13,13 @@ use the highest priority(9999).
 Example:
 
 Block traffic from 10.2.3.4/32:
-./null_route_helper acl_table_name block 10.2.3.4/32
+./null_route_helper block acl_table_name 10.2.3.4/32
 
 Unblock all traffic from 10.2.3.4/32:
-./null_route_helper acl_table_name unblock 10.2.3.4/32
+./null_route_helper unblock acl_table_name 10.2.3.4/32
+
+List all acl rules added by this script
+./null_route_helper list acl_table_name
 """
 
 
--- a/scripts/null_route_helper
+++ b/scripts/null_route_helper
@@ -26,8 +26,9 @@ import syslog
 import sys
 import click
 import ipaddress
+import tabulate
 
-from swsssdk import ConfigDBConnector
+from swsscommon.swsscommon import ConfigDBConnector
 
 
 CONFIG_DB_ACL_TABLE_TABLE = "ACL_TABLE"
@@ -36,11 +37,15 @@ CONFIG_DB_VLAN_TABLE = "VLAN"
 
 ACTION_ALLOW = "FORWARD"
 ACTION_DENY = "DROP"
+ACTION_LIST = "LIST"
 
 # Since SONiC supports multi ACL rules share the same priority, we use 9999 (the highest) for all rules
 ACL_RULE_PRIORITY = 9999
 # The key of rule will be overridden with BLOCK_RULE_ + ip
 ACL_RULE_PREFIX = 'BLOCK_RULE_'
+
+# Internet Protocol version 4 EtherType
+ETHER_TYPE_IPV4 = 0x0800
 
 def notice(msg):
     """
@@ -63,16 +68,16 @@ def ip_ver(ip_prefix):
     return ipaddress.ip_network(ip_prefix, False).version
 
 
-def check_table_existence(configdb, table_name):
+def confirm_required_table_existence(configdb, sub_table_name):
     """
-    Check the existence of pre-created ACL tables
+    Check the existence of required ACL table, and exit if absent 
     """
-    target_table = configdb.get_entry(CONFIG_DB_ACL_TABLE_TABLE, table_name)
+    target_table = configdb.get_entry(CONFIG_DB_ACL_TABLE_TABLE, sub_table_name)
 
     if not target_table:
-        error("Table {} not found, exiting...".format(table_name))
-
-    return target_table
+        error("Table {} not found, exiting...".format(sub_table_name))
+    
+    return True
 
 
 def get_acl_rule_key(ip_prefix):
@@ -121,7 +126,7 @@ def build_acl_rule(priority, src_ip):
         "PACKET_ACTION": "DROP"
     }
     if ip_ver(src_ip) == 4:
-        rule['ETHER_TYPE'] = str(0x0800)
+        rule['ETHER_TYPE'] = str(ETHER_TYPE_IPV4)
         rule['SRC_IP'] = src_ip
     else:
         rule['IP_TYPE'] = 'IPV6ANY'
@@ -137,8 +142,9 @@ def get_rule(configdb, table_name, ip_prefix):
     key_name = 'SRC_IP' if ip_ver(ip_prefix) == 4 else 'SRC_IPV6'
     all_rules = get_all_acl_rules(configdb, table_name)
     for key, rule in all_rules.items():
-        if ip_prefix == rule[key_name]:
-            return {key: rule}
+        if ip_prefix == rule.get(key_name, None):
+            if ip_prefix:
+                return {key: rule}
 
     return None
 
@@ -149,13 +155,15 @@ def update_acl_table(configdb, acl_table_name, ip_prefix, action):
     For 'DENY', an 'DROP' rule for given ip_prefix will be added if not existed
     For 'ALLOW', we will try to remove the existing 'DENY' rule, and nothing is changed if not existed
     """
-    check_table_existence(configdb, acl_table_name)
+    confirm_required_table_existence(configdb, acl_table_name)
     rule = get_rule(configdb, acl_table_name, ip_prefix)
     rule_key = list(rule.keys())[0] if rule else None
     rule_value = list(rule.values())[0] if rule else None
     if action == ACTION_ALLOW:
         if not rule:
             return
+        # Delete existing BLOCK rule for given ip_prefix
+        # Pass None as data will delete the entry
         configdb.mod_entry(CONFIG_DB_ACL_RULE_TABLE, rule_key, None)
     else:
         if rule:
@@ -172,16 +180,42 @@ def update_acl_table(configdb, acl_table_name, ip_prefix, action):
             configdb.set_entry(CONFIG_DB_ACL_RULE_TABLE, new_rule_key, new_rule_value)
 
 
-def helper(table_name, ip_prefix, action):
+def list_all_null_route_rules(configdb, table_name):
+    """
+    List all rules added by this script
+    """
+    
+    confirm_required_table_existence(configdb, table_name)
+    header = ("Table", "Rule", "Priority", "Action", "Match")
+    all_rules = get_all_acl_rules(configdb, table_name)
+
+    match_keys = ["SRC_IP", "SRC_IPV6"]
+    data = []
+    for (_, rule_id), rule in all_rules.items():
+        priority = rule.get("PRIORITY", "N/A")
+        action = rule.get("PACKET_ACTION", "N/A")
+        match = "N/A"
+        for k in match_keys:
+            if k in rule:
+                match = rule[k]
+                break
+
+        data.append([table_name, rule_id, priority, action, match])
+
+    print(tabulate.tabulate(data, headers=header, tablefmt="simple", missingval=""))
+
+
+def null_route_helper(table_name, action, ip_prefix=None):
     """
     Helper function called by 'click'.
     """
-    ip_prefix = validate_input(ip_prefix)
-
     configdb = ConfigDBConnector()
     configdb.connect()
-
-    update_acl_table(configdb, table_name, ip_prefix, action)
+    if action == ACTION_LIST:
+        list_all_null_route_rules(configdb, table_name)
+    else:
+        ip_prefix = validate_input(ip_prefix)
+        update_acl_table(configdb, table_name, ip_prefix, action)
 
 
 @click.group()
@@ -197,7 +231,7 @@ def block(table_name, ip_prefix):
     """
     Block traffic from given src ip prefix
     """
-    helper(table_name, ip_prefix, ACTION_DENY)
+    null_route_helper(table_name, ACTION_DENY, ip_prefix)
 
 
 # ./null_route_helper unblock table_name 1.2.3.4/21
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: scripts/null_route_helper [hunk 1]
```diff
@@ -0,0 +1,216 @@
...
+    all_rules = configdb.get_table(key)
+    block_rules = {}
+    for k, v in all_rules.items():
+        if k[1].startswith(ACL_RULE_PREFIX):
+            block_rules[k] = v
+
+    return block_rules
+
+
+def validate_input(ip_address):
```
prsunny, line 102: Can you reuse anything from utils?

scripts/null_route_helper
```diff
@@ -0,0 +1,216 @@
...
+    For 'DENY', an 'DROP' rule for given ip_prefix will be added if not existed
+    For 'ALLOW', we will try to remove the existing 'DENY' rule, and nothing is changed if not existed
+    """
+    check_table_existence(configdb, acl_table_name)
+    rule = get_rule(configdb, acl_table_name, ip_prefix)
+    rule_key = list(rule.keys())[0] if rule else None
+    rule_value = list(rule.values())[0] if rule else None
+    if action == ACTION_ALLOW:
+        if not rule:
+            return
```
prsunny, line 158: I think here we should be adding a rule even if it doesn't exist, may be for managing priorities. Why is it skipped here?

prsunny (commented):
Please update command reference guide.

scripts/null_route_helper
```diff
@@ -0,0 +1,216 @@
+#!/usr/bin/env python3
```
qiluo-msft, line 1: >python3 [](http://example.com/codeflow?start=15&length=7)

We need this feature on 201911 branch, so please make sure it could run on python2. Seems we could not cherry-pick directly.

qiluo-msft: If you tested the same script with both python2/3, you can change it to 
```
#!/usr/bin/env python
```

scripts/null_route_helper
```diff
@@ -0,0 +1,216 @@
...
+Since SONiC supports multi ACL rules share the same priority, all ACL rules created by null_route_helper will
+use the highest priority(9999).
+
+Example:
+
+Block traffic from 10.2.3.4/32:
+./null_route_helper acl_table_name block 10.2.3.4/32
+
+Unblock all traffic from 10.2.3.4/32:
+./null_route_helper acl_table_name unblock 10.2.3.4/32
```
qiluo-msft, line 19: >unblock [](http://example.com/codeflow?start=35&length=7)

There is no action to list all the exising rules. So if a user want to `unblock` all, it's inconvenient

scripts/null_route_helper
```diff
@@ -0,0 +1,216 @@
...
+
+
+from __future__ import print_function
+
+import syslog
+import sys
+import click
+import ipaddress
+
+from swsssdk import ConfigDBConnector
```
qiluo-msft, line 30: >swsssdk [](http://example.com/codeflow?start=5&length=7)

Please use swsscommon on master branch.

scripts/null_route_helper
```diff
@@ -0,0 +1,216 @@
...
+
+def ip_ver(ip_prefix):
+    return ipaddress.ip_network(ip_prefix, False).version
+
+
+def check_table_existence(configdb, table_name):
+    """
+    Check the existence of pre-created ACL tables
+    """
+    target_table = configdb.get_entry(CONFIG_DB_ACL_TABLE_TABLE, table_name)
```
qiluo-msft, line 70: >table_name [](http://example.com/codeflow?start=65&length=10)

Both arguments are table names, confusing.

scripts/null_route_helper
```diff
@@ -0,0 +1,216 @@
...
+    syslog.syslog(syslog.LOG_ERR, msg)
+    print(msg, file=sys.stderr)
+    sys.exit(1)
+
+
+def ip_ver(ip_prefix):
+    return ipaddress.ip_network(ip_prefix, False).version
+
+
+def check_table_existence(configdb, table_name):
```
qiluo-msft, line 66: >check_table_existence [](http://example.com/codeflow?start=4&length=21)

-> `get_config_db_acl_table_subtable` or something similiar?

scripts/null_route_helper
```diff
@@ -0,0 +1,216 @@
...
+    """
+    Log an ERR message to the console and syslog, and exit the program with an error code
+    """
+    syslog.syslog(syslog.LOG_ERR, msg)
+    print(msg, file=sys.stderr)
+    sys.exit(1)
+
+
+def ip_ver(ip_prefix):
+    return ipaddress.ip_network(ip_prefix, False).version
```
qiluo-msft, line 63: >ip_network [](http://example.com/codeflow?start=21&length=10)

If you keep this variable, you can reuse it in future, such as check prefix length.

scripts/null_route_helper
```diff
@@ -0,0 +1,216 @@
...
+def build_acl_rule(priority, src_ip):
+    """
+    Bild DROP rule for given src_ip and priority
+    """
+    rule = {
+        "PRIORITY": str(priority),
+        "PACKET_ACTION": "DROP"
+    }
+    if ip_ver(src_ip) == 4:
+        rule['ETHER_TYPE'] = str(0x0800)
```
qiluo-msft, line 124: >0x0800 [](http://example.com/codeflow?start=33&length=6)

Explain magic number?

scripts/null_route_helper
```diff
@@ -0,0 +1,216 @@
...
+    return None
+
+
+def update_acl_table(configdb, acl_table_name, ip_prefix, action):
+    """
+    Update ACL table to apply new rules for given ip_prefix. 'action' is supposed to be in ['DENY', 'ALLOW']
+    For 'DENY', an 'DROP' rule for given ip_prefix will be added if not existed
+    For 'ALLOW', we will try to remove the existing 'DENY' rule, and nothing is changed if not existed
+    """
+    check_table_existence(configdb, acl_table_name)
```
qiluo-msft, line 152: >check_table_existence [](http://example.com/codeflow?start=4&length=21)

If you store the return value, will you optimize below code?

scripts/null_route_helper
qiluo-msft: Because it will be allowed anyway?

scripts/null_route_helper
```diff
@@ -0,0 +1,216 @@
...
+    rule = get_rule(configdb, acl_table_name, ip_prefix)
+    rule_key = list(rule.keys())[0] if rule else None
+    rule_value = list(rule.values())[0] if rule else None
+    if action == ACTION_ALLOW:
+        if not rule:
+            return
+        configdb.mod_entry(CONFIG_DB_ACL_RULE_TABLE, rule_key, None)
```
qiluo-msft, line 159: >mod_entry [](http://example.com/codeflow?start=17&length=9)

Do you actually delete the rule? add some comment to help reading.

scripts/null_route_helper
```diff
@@ -0,0 +1,216 @@
...
+                rule_value['PACKET_ACTION'] = 'DROP'
+                configdb.mod_entry(CONFIG_DB_ACL_RULE_TABLE, rule_key, rule_value)
+        else:
+            priority = ACL_RULE_PRIORITY
+            new_rule_key = (acl_table_name, get_acl_rule_key(ip_prefix))
+            new_rule_value = build_acl_rule(priority, ip_prefix)
+            configdb.set_entry(CONFIG_DB_ACL_RULE_TABLE, new_rule_key, new_rule_value)
+
+
+def helper(table_name, ip_prefix, action):
```
qiluo-msft, line 175: >helper [](http://example.com/codeflow?start=4&length=6)

Use a better name?

scripts/null_route_helper
prsunny: yes

qiluo-msft: So no need to adding a rule.

scripts/null_route_helper
bingwang-ms: The ```validate_input``` in this PR will add prefix len for IP addresses that don't explicitly set prefix. For example, ```1.2.3.4``` is changed to ```1.2.3.4/32```. The common utils may not do this.

scripts/null_route_helper
bingwang-ms: > I think here we should be adding a rule even if it doesn't exist, may be for managing priorities. Why is it skipped here?

I skip here since there will be a default ALLOW rule in the pre-created ACL table. Is it necessary to add an ALLOW rule for a certain prefix?

scripts/null_route_helper
bingwang-ms: ```#!/usr/bin/env python``` doesn't work on ```202012``` image since it will point to python2, and several packages are missing. I think the only way to solve the issue is to create another PR for ```201911``` or earlier image. It's not a clean cherry-pick.

scripts/null_route_helper
bingwang-ms: It's a good suggestion. Thanks.
I was thinking which one is better, a ```list``` command or a ```unblock all``` command? Any suggestions?

scripts/null_route_helper
prsunny: my concern is, lets say user specifies "unblock 1.2.3.4/32" and if there is some rule in between that says 1.2.3.0/24 drop. In this case if we don't explicitly add the "allow" rule, it will get dropped.

scripts/null_route_helper
prsunny: sonic-utilities can no longer be cherry-picked to 201911. We faced it few times and suggest to create a new PR to 201911

scripts/null_route_helper
qiluo-msft: `list` command will be useful in many cases, including
1. unblock all
2. verify existing rules

scripts/null_route_helper
bingwang-ms: That's a good catch. We discuss about it offline, and confirm that all prefix len are 32 (For IPv4) or 128 (IPv6). Based on this acknowledge, it shouldn't be a issue. I will update the ```validate_input``` interface to ensure that. Thanks

scripts/null_route_helper
bingwang-ms: Yes. I checked the code and confirm the entry will be deleted. Also verified in test.
https://github.com/Azure/sonic-swss-common/blob/bf8c832cf1c7a6e72d7b1c843888ffb4a27088c8/common/configdb.cpp#L96

scripts/null_route_helper
bingwang-ms: I added a new interface `list`. Thanks

scripts/null_route_helper
bingwang-ms: I don't like to add a global variable to save the ip version. Any better ideas?

scripts/null_route_helper
bingwang-ms: Updated. Thanks

scripts/null_route_helper
bingwang-ms: Updated. Thanks

scripts/null_route_helper
bingwang-ms: Updated. Thanks

scripts/null_route_helper
bingwang-ms: I don't like to add a global variable to save the return value. Any better ideas?

scripts/null_route_helper
bingwang-ms: Updated. Thanks

scripts/null_route_helper
bingwang-ms: Updated. Thanks

bingwang-ms:
Hi @qiluo-msft @prsunny . All comments were addressed. And the unit test code was also checked in. Please help to review. Thanks

scripts/null_route_helper
qiluo-msft: Add the `list` command in the usage comment?

scripts/null_route_helper
qiluo-msft: I don't like global var either. To clarify, I mean the object of `ip_network` is very useful, and you could deduce many attributes from it such as prefix length, not only the version.

qiluo-msft (approved):

scripts/null_route_helper
```diff
@@ -0,0 +1,267 @@
...
+will remove an existing DENY rule that has been created by the block operation (i.e. it does
+NOT insert an ALLOW rule, only removes DENY rules).
+
+Since SONiC supports multi ACL rules share the same priority, all ACL rules created by null_route_helper will
+use the highest priority(9999).
+
+Example:
+
+Block traffic from 10.2.3.4/32:
+./null_route_helper block acl_table_name 10.2.3.4/32
```
qiluo-msft, line 16: >/32 [](http://example.com/codeflow?start=49&length=3)

we can remove the prefix completely from CLI

prsunny (approved):

scripts/null_route_helper
bingwang-ms: Thanks, I will update.

qiluo-msft (approved):

qiluo-msft (changes requested):
Block. This PR should go master first, and then 201911 and later branches.

bingwang-ms:
> Block. This PR should go master first, and then 201911 and later branches.

OK. I will hold on merging this PR and create a same PR to ```master``` branch.

bingwang-ms:
Same PR created for master branch https://github.com/Azure/sonic-utilities/pull/1737

qiluo-msft (approved):

PR #1723, opened 2021-07-24
[techsupport] Removed interactive option for docker commands and Improved Error Reporting

vivekreddynv:
#### Why I did

Recently, a bug was seen which was related to saisdkdump and particularly showed up when `show techsupport` was invoked. Although, it was fixed, the sonic-mgmt test failed to capture it beforehand. 

This highlighted a few shortcomings of the `generate_dump` script and this PR addresses those and also a few additional issues seen

This PR fixes a few things, I'll explain each of them in the next section.

#### What I did

**1) Remove the "Interactive option (-i) for the docker invocation commands"**
  This was the reason why the bug which was  was not captured previously. When the techsupport was invoked remotely (Eg: using sshpass), the `docker exec -it  <docker> <cmd>` command would fail saying ` ‘the input device is not a TTY'`. Hence the (-i) option was removed.

**2) Change the Return Code**
Currently, the script doesn't return any non-zero error codes for most of the intermediate steps (even though they fail), which makes validation hard.

To handle this, a helper function and trap cmd are used. 
```
handle_error() {
  if [ "$1" != "0" ]; then
    echo "ERR: RC:-$1 observed on line $2" >&2
    RETURN_CODE=1
  fi
}
trap 'handle_error $? $LINENO' ERR # This would trap any executions with non-zero return codes
```
The global variable RETURN_CODE is set when this is called and the same RETURN_CODE is returned when `generate_dump` invocation process exits

You may see this is used in multiple functions instead of placing it once on the top of the script. This is because, every function can itself be considered as a subshell and each of them requires a explicit trap command.

When a command is failed with error, this logic would get append a corresponding log to stderr. 
`ERR: RC:-1 observed on line 729`

**3) Improve Error Reporting for save_cmd function**

Currently any error written to the stderr by the intermediate calls are redirected to the same location as stdout, which is usually the file we see under dump/ dir.  This is perfectly fine, but the sonic-mgmt test only parses the text seen in stdout. 

So, a new option (-r) is added to `generate_dump` script and subsequently to `show techsupport` to redirect any intermediate errors seen to the generate_dump's stderr.

With this option enabled, these sort of errors can be captured on the stderr.

```
root@sonic:/home/admin# show techsupport -r
..........
timeout --foreground 5m show queue counters > /var/dump/sonic_dump_r-tigon-04_20210714_062239/dump/queue.counters_1
  File "/usr/local/bin/queuestat", line 373, in <module>
    main()
  File "/usr/local/bin/queuestat", line 368, in main
    queuestat.get_print_all_stat(json_opt)
  File "/usr/local/bin/queuestat", line 239, in get_print_all_stat
    cnstat_dict = self.get_cnstat(self.port_queues_map[port])
  File "/usr/local/bin/queuestat", line 168, in get_cnstat
    cnstat_dict[queue] = get_counters(queue_map[queue])
  File "/usr/local/bin/queuestat", line 158, in get_counters
    fields[pos] = str(int(counter_data))
ValueError: invalid literal for int() with base 10: ''
handle_error $? $LINENO
ERR: RC:-1 observed on line 199
Command: show queue counters timedout after 5 minutes.
.............

Without that option, this'll be the output seen

root@sonic:/home/admin# show techsupport 
..........
timeout --foreground 5m show queue counters &> /var/dump/sonic_dump_r-tigon-04_20210714_062239/dump/queue.counters_1
handle_error $? $LINENO
ERR: RC:-1 observed on line 199
Command: show queue counters timedout after 5 minutes.
.............
```

**4) Minor Error in sdk-dump collection logic handled**
save_file is only called for the files seen in sdk_dump_path and not for directories
```
cp: -r not specified; omitting directory '/tmp/sdk-dumps'
handle_error $? $LINENO
ERR: RC:-1 observed on line 729
tar: sonic_dump_r-tigon-04_20210714_062239/sai_sdk_dump/sdk-dumps: Cannot stat: No such file or directory
tar: Exiting with failure status due to previous errors
tar append operation failed. Aborting to prevent data loss.
```
The reason being, `find /tmp/sdk-dumps` returns ["/tmp/sdk-dumps"] even if the dir is empty. In the next step, save_file cmd is applied on the folder and thus the error. This can be handled by the change specified above

**5) Minor Error in custom plugins logic handled**
Added a condition to check if the dir exists before proceeding forward.
```
if [[ -d ${PLUGINS_DIR} ]]; then
        local -r dump_plugins="$(find ${PLUGINS_DIR} -type f -executable)"
        for plugin in $dump_plugins; do
            # save stdout output of plugin and gzip it
            save_cmd "$plugin" "$(basename $plugin)" true false
        done
    fi
```
Otherwise,  find command might fail saying
```
root@sonic:/home/admin# find /usr/local/bin/debug-dump -type f -executable
find: ‘/usr/local/bin/debug-dump’: No such file or directory
```

**NOTE: The last two issues were found out because of the error reporting logic added**


dgsudharsan (approved):

liat-grozovik (dismissed):

scripts/generate_dump
```diff
@@ -47,7 +49,15 @@ handle_signal()
 }
 trap 'handle_signal' SIGINT
 
+handle_error() {
+  if [ "$1" != "0" ]; then
+    echo "ERR: RC:-$1 observed on line $2" >&2
+    RETURN_CODE=1
+  fi
```
qiluo-msft, line 56: >fi [](http://example.com/codeflow?start=2&length=2)

In your PR description, this function `return $1`, which is different with the code.

scripts/generate_dump
```diff
@@ -47,7 +49,15 @@ handle_signal()
 }
 trap 'handle_signal' SIGINT
 
+handle_error() {
+  if [ "$1" != "0" ]; then
```
qiluo-msft, line 53: >   [](http://example.com/codeflow?start=0&length=2)

Use 4-space indentation?

scripts/generate_dump
```diff
@@ -144,21 +155,24 @@ save_bcmcmd_all_ns() {
...
 save_cmd() {
+    trap 'handle_error $? $LINENO' ERR
     local start_t=$(date +%s%3N)
     local end_t=0
     local cmd="$1"
     local filename=$2
     local filepath="${LOGDIR}/$filename"
     local do_gzip=${3:-false}
-    local save_stderr=${4:-true}
+    local save_stderr=${4:-$SAVE_STDERR}
```
qiluo-msft, line 165: >save_stderr [](http://example.com/codeflow?start=10&length=11)

The intention of `save_stderr` is different from your new feature `-r`.

Actually I don't see the point of `save_stderr=False`.

Suggest you
1. remove old feature, always save stderr
2. add a new option like `inherit_stderr`, and still keep save stderr.

scripts/generate_dump
```diff
@@ -235,14 +250,15 @@ save_cmd_all_ns() {
...
+    trap 'handle_error $? $LINENO' ERR
     local start_t=$(date +%s%3N)
     local end_t=0
     local docker=$1
     local filename=$2
     local dstpath=$3
     local timeout_cmd="timeout --foreground ${TIMEOUT_MIN}m"
 
-    local touch_cmd="sudo docker exec -i ${docker} touch ${filename}"
+    local touch_cmd="sudo docker exec -t ${docker} touch ${filename}"
```
qiluo-msft, line 261: >-t [](http://example.com/codeflow?start=38&length=2)

No need to add `-t`. In some use case, there is no tty.

scripts/generate_dump
```diff
@@ -235,14 +250,15 @@ save_cmd_all_ns() {
...
 copy_from_docker() {
+    trap 'handle_error $? $LINENO' ERR
     local start_t=$(date +%s%3N)
     local end_t=0
     local docker=$1
     local filename=$2
     local dstpath=$3
     local timeout_cmd="timeout --foreground ${TIMEOUT_MIN}m"
 
-    local touch_cmd="sudo docker exec -i ${docker} touch ${filename}"
```
qiluo-msft, line 245: >-i [](http://example.com/codeflow?start=38&length=2)

Could you detect the environment and add `-i` as the default.

scripts/generate_dump
```diff
@@ -771,14 +808,17 @@ collect_mellanox() {
...
     ${CMD_PREFIX}rm -rf $sai_dump_folder
-    ${CMD_PREFIX}docker exec -it syncd rm -rf $sai_dump_folder
+    ${CMD_PREFIX}docker exec -t syncd rm -rf $sai_dump_folder
     
     # Save SDK error dumps
     local sdk_dump_path=`${CMD_PREFIX}docker exec syncd cat /tmp/sai.profile|grep "SAI_DUMP_STORE_PATH"|cut -d = -f2`
-    if [[ $sdk_dump_path ]]; then
+    if [[ -d $sdk_dump_path ]]; then
         copy_from_docker syncd $sdk_dump_path  /tmp/sdk-dumps
         for file in $(find /tmp/sdk-dumps); do
```
qiluo-msft, line 817: >find [](http://example.com/codeflow?start=22&length=4)

You may fix the issue by option `-type f`

scripts/generate_dump
vivekreddynv: The usecase of stderr=False as i see is only for the sonic-mgmt auto-techsupport test. 

If this is set to false, any intermediate stderr will be redirected to the stderr of the techsupport and the test can be enhanced to capture the stderr. Thereby we can have a single location to view the errors reported by any of the intermediate steps.

scripts/generate_dump
vivekreddynv: Okay, will remove

scripts/generate_dump
qiluo-msft: > The usecase of stderr=False as i see is only for the sonic-mgmt auto-techsupport test.

Could you give a link of the usage in sonic-mgmt repo?

scripts/generate_dump
vivekreddynv: Yeah, Will do that

scripts/generate_dump
vivekreddynv: But why do we require -i for a touch command. Infact i've looked at other docker exec commands used in this script. All of them just write to stdout, so i don't see why -i has to be retained

scripts/generate_dump
vivekreddynv: No, it's not there yet. But we plan to update it once this gets merged.

scripts/generate_dump
vivekreddynv: Handled

scripts/generate_dump
vivekreddynv: Handled

scripts/generate_dump
vivekreddynv: Handled

scripts/generate_dump
vivekreddynv: Let me know what do you think about this

scripts/generate_dump
qiluo-msft, line 165: >SAVE_STDERR [](http://example.com/codeflow?start=28&length=11)

You should add `$SAVE_STDERR` to caller place as an argument, and keep `true` as default parameter value.

vivekreddynv: From what i understand, wouldn't that require adding an extra argument to all the save_cmd invocations across the script? That would be fine if this argument is something like `filename` which might change for different invocations.

But since the scope of this flag is global, i think it makes sense to have it inside the parameter value. Let me know if you think otherwise

dgsudharsan:
@qiluo-msft Can you please review?

scripts/generate_dump
qiluo-msft: I could understand your point. Does it make more sense to only use global variable and remove `$4` ?

vivekreddynv: The stderr arg was first introduced here https://github.com/Azure/sonic-utilities/pull/1335. 

And for some reason, this was set to false in the normal execution itself (probably because he only wanted to collect stdout). I did not want to disturb it and thus retained the argument, but if not provided read from the global variable.

qiluo-msft: Could you discuss with @stepanblyschak in the same company to understand the original intention? To me, use a global variable as default value is weird.

vivekreddynv: @stepanblyschak, confirmed that this local variable $4 can be removed. I'll remove it and will only have the global variable.

vivekreddynv: DONE

qiluo-msft (approved):

qiluo-msft:
This PR could not be cleanly cherry-picked to 202012. Please submit another PR.

vivekreddynv:
> This PR could not be cleanly cherry-picked to 202012. Please submit another PR.

Raised a separate PR: https://github.com/Azure/sonic-utilities/pull/1833

judyjoseph:
This PR could not be cleanly cherry-picked to 202106. Please submit another PR.

qiluo-msft:
@vivekreddynv @dgsudharsan The new behavior is that the script will fail immediately if any command return error. However in reality, this is not convenient. Let's assume there is an image bug, some command fail or this script itself has bug, we still need to collect as much as possible data. Could you improve and let the following commands continue running?

vivekreddynv:
> @vivekreddynv @dgsudharsan The new behavior is that the script will fail immediately if any command return error. However in reality, this is not convenient. Let's assume there is an image bug, some command fail or this script itself has bug, we still need to collect as much as possible data. Could you improve and let the following commands continue running?

Hi @qiluo-msft, 

The script will run in it's entirety (will create the archive with all the dumps) even though if any of the intermediate steps are failed. And in the end, it will exit with an rc=1 . 

But i understand the issue seen here https://github.com/Azure/sonic-utilities/pull/1844.  When a timeout happens, the script returns with a non-zero exit code and bypasses the "Command timedout error log".  i believe that is fixed in the PR. 

Is exiting with non-zero code not okay in the command timedout case?

qiluo-msft:
@yxieca to check

qiluo-msft:

PR #1732, opened 2021-07-29
[intfstat] Fix logic in RIF counters to print right stats

noaOrMlnx:
#### What I did
Fix logic in scripts/intfstat file.
Before the change: 
If a user cleared all counters using `sonic-clear rifcounters` command and then queried for rif counters with a certain interface (e.g. `show interfaces counters rif Ethernet0`), the user would see the older stats. meaning the clear was not performed. 
The reason for this is that when we do sonic-clear rifcounters, the counters are not really being cleared, but current stats are saved into a file, and in the next counters query, it will calculate the diff between current counters and the saved counters. 
If the clear was performed in general (without an interface name), it's saved to a general file (e.g. 0). 
If the clear was performed for a certain interface, it's saved to an interface file (e.g. 0-Ethernet0).
so when we clear in one method and then query for the other, we can receive the info from the wrong file and display different stats. 

#### How I did it
Add the logic: 
when deleting for a certain interface, add the info to the general file. 
when querying for a certain interface, check the general file as well. 

#### How to verify it
* send traffic to interface
* check general rif table (`show interfaces counters rif`)
* clear general rifcounters (`sonic-clear rifcounters`)
* check that the specific interface table is also updated (`show interfaces counters rif Ethernet0`)

* send traffic to interface
* clear interface table (`sonic-clear rifcounters Ethernet0`)
* check that general table is updated as well (`show interfaces counters rif`)
N/A
N/A


BasimShalata:
@liat-grozovik FYI.
@dgsudharsan can you please review.

scripts/intfstat
```diff
@@ -305,6 +305,19 @@ def main():
...
+                        data = pickle.load(open(cnstat_fqn_general_file, 'rb'))
+                        for key, val in cnstat_dict.items():
+                            data[key] = val
+                        pickle.dump(data, open(cnstat_fqn_general_file, 'wb'))
+                    except IOError as e:
+                        sys.exit(e.errno)
+            # Add the information also to interface specific file
             pickle.dump(cnstat_dict, open(cnstat_fqn_file, 'wb'))
         except IOError as e:
             sys.exit(e.errno)
```
dgsudharsan, line 323: If we are making the general file upto date, can you please explain the need for the interface specific file? In my opinion one file should be good enough to save stats.

dgsudharsan (changes requested):

scripts/intfstat
noaOrMlnx: There is another feature we need to take into consideration - tag_name. 
If a user wants to clear counters with a certain tag, we need to keep stats in this tag file. 
so I removed the interface name in file which is indeed unneeded, but changed the logic in the script to not brake tag_name logic.

scripts/intfstat
```diff
@@ -305,6 +305,19 @@ def main():
...
+                cnstat_fqn_general_file = cnstat_fqn_file[:gen_index] + cnstat_fqn_file[gen_index:].split('-')[0]
+                if os.path.isfile(cnstat_fqn_general_file):
+                    try:
+                        data = pickle.load(open(cnstat_fqn_general_file, 'rb'))
+                        for key, val in cnstat_dict.items():
+                            data[key] = val
+                        pickle.dump(data, open(cnstat_fqn_general_file, 'wb'))
+                    except IOError as e:
+                        sys.exit(e.errno)
+            # Add the information also to tag specific file
```
dgsudharsan, line 320: I think this logic will be an issue. Let's assume the user gives clear stats for globally first and then an individual interface (Ethernet0). First logic it will save the stats of all the interfaces in the file since cnstat_dict will contain information on all interfaces. However, second time when interface name is specified, it cnstat_dict will contain the interface stats alone and writing it directly to the file will overwrite previous written information. As done for 'tag' the logic here when clearing per interface should read the contents of the pickle file and then override only the specific interface.

dgsudharsan (changes requested):
Can you please add UT to this file.https://github.com/Azure/sonic-utilities/blob/master/tests/intfstat_test.py ?

scripts/intfstat
noaOrMlnx: Done

dgsudharsan (approved):

dgsudharsan (approved):

PR #1739, opened 2021-08-02
[config][interface][speed] Fixed the config interface speed in multi-asic issue

mlok-nokia:
#### What I did

Fix for issue https://github.com/Azure/sonic-buildimage/issues/8308

#### How I did it

"front_asic_namespaces" is not defined. Changed  function Sonic2VConnector() call with the correct argument "namespace". 

#### How to verify it

- Execute the config command to set the interface speed:
**sudo config interface -n asic0 speed Ethernet0 400000**
- No error shown

- Previously the command showed the following error.
**name 'front_asic_namespaces' is not defined**

- New output does not show the error


zhangyanzhao:
@abdosi  will take a look.

judyjoseph (approved):

PR #1742, opened 2021-08-05
[intfutil]Let speed 100M/2.5G can be correct displayed.

PJHsieh:
Parse the speed string:
The speed string length < len('1000'), display speed by unit 'M'
else display speed by unit 'G'

#### What I did
Let speed 100M/2.5G can be correct displayed

#### How I did it
Parse the speed string:
The speed string length < len('1000'), display speed by unit 'M'
else display speed by 'G'

#### How to verify it
Real device verification.

Verify by the CLI cmd 'show interfaces status' output on the port which is configure with speed 100M
100M shall be shown in the Speed field

Verify by the CLI cmd 'show interfaces status' output on the port which is configured with speed 2.5G
2.5G shall be shown in the Speed field.

### DIFF
--- a/scripts/intfutil
+++ b/scripts/intfutil
@@ -138,12 +138,28 @@ def appl_db_port_status_get(appl_db, intf_name, status_type):
     if status is None:
         return "N/A"
     if status_type == PORT_SPEED and status != "N/A":
-        status = '{}G'.format(status[:-3])
+        if (len(status) < len('1000')):
+            status += 'M'
+        else:
+            if (status.rpartition('000')[1] != '000'):
+                status = status[:-3] + '.' + status[-3:]
+                status = status.rstrip('0') + 'G'
+            else:
+                status = '{}G'.format(status[:-3])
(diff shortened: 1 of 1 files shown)


PJHsieh:
retest

scripts/intfutil
```diff
@@ -138,12 +138,28 @@ def appl_db_port_status_get(appl_db, intf_name, status_type):
     if status is None:
         return "N/A"
     if status_type == PORT_SPEED and status != "N/A":
-        status = '{}G'.format(status[:-3])
+        if (len(status) < len('1000')):
```
Blueve, line 141: I think it would be great if you can extract these lines to a new methods and reuse it
And it might be better to convert to numbers than manipulate text

Blueve (changes requested):

Blueve:
This issue has been addressed in another PR: https://github.com/Azure/sonic-utilities/pull/2110

PR #1747, opened 2021-08-07
disk_check: Script updated to run good in 201811 & 201911

renukamanavalan:
#### What I did
Have independent subdirs for each mounted dir to avoid any collisions of files/dirs by same name.
Adopt for older version of python3

#### How I did it
Changes:
1) Individual subdirs for each dir to be mounted
2) subprocess args made compatible with older version of python3 (tested in version 3.5.3)

#### How to verify it
1) Simulate read-only state
2) Run this script
3) Test ssh via new tacacs user (who had not logged in earlier)

### DIFF
--- a/scripts/disk_check.py
+++ b/scripts/disk_check.py
@@ -64,7 +64,7 @@ def test_writable(dirs):
 
 
 def run_cmd(cmd):
-    proc = subprocess.run(cmd, shell=True, text=True, capture_output=True)
+    proc = subprocess.run(cmd, shell=True, stdout=subprocess.PIPE)
     ret = proc.returncode
     if ret:
         log_err("failed: ret={} cmd={}".format(ret, cmd))
@@ -72,9 +72,9 @@ def run_cmd(cmd):
         log_info("ret={} cmd: {}".format(ret, cmd))
 
     if proc.stdout:
-        log_info("stdout: {}".format(str(proc.stdout)))
+        log_info("stdout: {}".format(proc.stdout.decode("utf-8")))
     if proc.stderr:
-        log_info("stderr: {}".format(str(proc.stderr)))
+        log_info("stderr: {}".format(proc.stderr.decode("utf-8")))
     return ret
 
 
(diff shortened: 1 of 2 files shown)


lguohan:
@qiluo-msft , can you take priority to review this one?

scripts/disk_check.py
```diff
@@ -64,17 +69,17 @@ def test_writable(dirs):
...
+    proc = subprocess.run(cmd, shell=True, stdout=subprocess.PIPE)
     ret = proc.returncode
     if ret:
         log_err("failed: ret={} cmd={}".format(ret, cmd))
     else:
         log_info("ret={} cmd: {}".format(ret, cmd))
 
     if proc.stdout:
-        log_info("stdout: {}".format(str(proc.stdout)))
+        log_info("stdout: {}".format(proc.stdout.decode("utf-8")))
```
qiluo-msft, line 80: >"utf-8" [](http://example.com/codeflow?start=56&length=7)

You can use default parameter value.

tests/disk_check_test.py
```diff
@@ -26,7 +27,7 @@
         "workdir": "/tmp/tmpy",
         "mounts": "overlay_tmpx blahblah",
         "err": "/tmpx is not read-write|READ-ONLY: Mounted ['/tmpx'] to make Read-Write",
-        "cmds": ['mount -t overlay overlay_tmpx -o lowerdir=/tmpx,upperdir=/tmp/tmpx,workdir=/tmp/tmpy /tmpx']
+        "cmds": ['mount -t overlay overlay_tmpx -o lowerdir=/tmpx,upperdir=/tmp/tmpx/tmpx,workdir=/tmp/tmpy/tmpx /tmpx']
```
qiluo-msft, line 30: >upperdir=/tmp/tmpx/tmpx [](http://example.com/codeflow?start=66&length=23)

Could you explain why you need test code change when you change implementation details? Could you make test code not depends on the implementation?

scripts/disk_check.py
renukamanavalan: Prefer to keep it explicit. Any issue with that.

tests/disk_check_test.py
renukamanavalan: This is a base change. We use explicit subdir under upper & work for each dir to mount.

qiluo-msft (approved):
Could you refine the PR title (and the title during squash merge)? It is not clear what you plan to adopt? Audience know nothing about 201811/201911 adoption requirement. I guess you mean adapt code to be compatible with python2?

PR #1754, opened 2021-08-11
[show] Allow system with no ports in config db run without errors

liorghub:
What I did
Allow system with no ports in config db run without errors.
This is needed for modular system which should boot properly without line cards.

How I did it
Do not raise warning if ports dictionary is empty.

How to verify it
Run show interfaces status

root@r-leopard-41:/home/admin# show interfaces status
Warning: failed to retrieve PORT table from ConfigDB!
Warning: failed to retrieve PORT table from ConfigDB!
Interface Lanes Speed MTU FEC Alias Vlan Oper Admin Type Asym PFC
*----------- ------- ------- ----- ----- ------- ------ ------ ------- ------ ----------
root@r-leopard-41:/home/admin#

root@r-leopard-41:/home/admin# show interfaces status
Interface Lanes Speed MTU FEC Alias Vlan Oper Admin Type Asym PFC
*----------- ------- ------- ----- ----- ------- ------ ------ ------- ------ ----------
root@r-leopard-41:/home/admin#

### DIFF
--- a/utilities_common/cli.py
+++ b/utilities_common/cli.py
@@ -127,7 +127,6 @@ class InterfaceAliasConverter(object):
 
 
         if not self.port_dict:
-            click.echo(message="Warning: failed to retrieve PORT table from ConfigDB!", err=True)
             self.port_dict = {}
 
         for port_name in self.port_dict:
--- a/utilities_common/cli.py
+++ b/utilities_common/cli.py
@@ -127,6 +127,7 @@ class InterfaceAliasConverter(object):
 
 
         if not self.port_dict:
+            click.echo(message="Configuration database contains no ports")
             self.port_dict = {}
 
         for port_name in self.port_dict:


dgsudharsan (dismissed):

utilities_common/cli.py
```diff
@@ -127,7 +127,6 @@ def __init__(self, db=None):
 
 
         if not self.port_dict:
-            click.echo(message="Warning: failed to retrieve PORT table from ConfigDB!", err=True)
```
qiluo-msft, line 130: >, err=True [](http://example.com/codeflow?start=86&length=10)

This is a good warning message. Can you keep it and just remove `, err=True` ?

liorghub: @qiluo-msft 
I think that if we are going to leave this message, we should at least change it, something like "Configuration database contains no ports" and of course remove err=True.
What do you think?

qiluo-msft: Agree.

liorghub: Done.

qiluo-msft (approved):

PR #1760, opened 2021-08-16
Improve `config` error handling on version_info

qiluo-msft:
#### What I did
`get_sonic_version_info()` may return `None` if `/etc/sonic/sonic_version.yml` does not exist. Previously the `config` exit without any useful information.

Some `config` subcommands do not depend on sonic version info, we should allow them to run.


arlakshm (approved):
lgtm

PR #1763, opened 2021-08-16
[config][generic-update] Implement ChangeApplier for ConfigDb format

qiluo-msft:
#### What I did
Implement ChangeApplier for ConfigDb format

design doc: https://github.com/Azure/SONiC/blob/master/doc/config-generic-update-rollback/Json_Change_Application_Design.md#312-change-applier

#### How to verify it
TODO: manual test on DUT

### DIFF
--- a/generic_config_updater/generic_updater.py
+++ b/generic_config_updater/generic_updater.py
@@ -1,8 +1,10 @@
 import json
 import os
 from enum import Enum
+from jsonpatch import PatchOperation, AddOperation, RemoveOperation, ReplaceOperation
+from swsscommon.swsscommon import ConfigDBConnector
 from .gu_common import GenericConfigUpdaterError, ConfigWrapper, \
-                       DryRunConfigWrapper, PatchWrapper
+                       DryRunConfigWrapper, PatchWrapper, JsonChange
 from .patch_sorter import PatchSorter
 
 CHECKPOINTS_DIR = "/etc/sonic/checkpoints"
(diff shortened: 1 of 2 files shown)


generic_config_updater/generic_updater.py
```diff
@@ -18,9 +21,94 @@ def release_lock(self):
         pass
 
 class ChangeApplier:
```
ghooo, line 23: maybe move it to a new file?

generic_config_updater/generic_updater.py
```diff
@@ -18,9 +21,94 @@ def release_lock(self):
...
+        services = self.table_service.get(table_name, {})
+        for service in services:
+            commands = self.service_commands.get(service, {})
+            restart_comand = commands.get('restart-command')
+            validate_command = commands.get('validate-commands')
+            if restart_comand:
+                rc = self.exec_command(restart_comand)
+                if rc != 0:
+                    raise GenericConfigUpdaterError(f"Restart command failed: {restart_comand}, rc={rc}")
+                rc = self.exec_command(validate_command)
```
ghooo, line 55: Validate commands do not need to be associated with restart commands, i.e. we should be able to verify the services that absorb the changes automatically.

generic_config_updater/generic_updater.py
```diff
@@ -18,9 +21,94 @@ def release_lock(self):
...
+        parts = op.pointer.parts
+        entry = self.config_db.get_entry(parts[0], parts[1])
+        oldtree = entry
+        for part in parts[1::-1]:
+            oldtree = { part: oldtree }
+        op.apply(oldtree)
+        self.config_db.set_entry(parts[0], parts[1], entry)
+
+    def apply(self, change: JsonChange):
+        patch = change.patch
```
ghooo, line 75: Using the patch within the change is ok, but it is limiting and is a little different than the design document. Using patch will forces specific order of execution on the `ChangeApplier` while in fact there is no ordering. This causes problems like multiple restarts which is solved by the lazy restart.

`JsonChange` on the other hand can be used to just getting the target config, and then the diff. The diff is per table, so we can handle table by table alphabetically.

`JsonChange` only provides one function which is apply (notice the second step in the [diagram](https://github.com/Azure/SONiC/blob/master/doc/config-generic-update-rollback/Json_Change_Application_Design.md#31-overview)) **GenerateTargetConfig**. I think `patch` access should be reduced i.e. to be preceded by `_`

generic_config_updater/generic_updater.py
ghooo, line 55: Validate commands expect ConfigDb before and after the update, from design doc:
>validate-commands - Multiple CLI commands used to validate the service has absorbed the change successfully. The validate commands can expect the config before and after the update to passed to the CLI commands for detailed validation if needed.

This was requested by linkedin team

qiluo-msft: fixed.

generic_config_updater/generic_updater.py
qiluo-msft: will move later.

generic_config_updater/generic_updater.py
ghooo: We discussed this with LinkedIn: Zhenggen Xu <zxu@linkedin.com>; Praveen Chaudhary <pchaudhary@linkedin.com> on 3/15, from the notes:
>
image

Specifically when they delete a port they not only verify it was deleted from ConfigDb, but also from AsicDB. Check code [here](https://github.com/Azure/sonic-utilities/blob/2b12aadeed3a86ca4ede3aa30b451914c4acb00b/config/config_mgmt.py#L380).

So in order to support their request, we are passing to the validation function what commands were deleted/added. I just made it a little bit generic and passed the whole configDb before and after the update. The validate script itself can do diffing to identify these commands that were changed, or they can loop over the PORTs and check which was deleted.

PR #1768, opened 2021-08-17
[Fast-reboot] Set flex counters delay indicator to prevent flex counters enablement after fast-reboot

shlomibitton:
Signed-off-by: Shlomi Bitton <shlomibi@nvidia.com>

#### What I did
Set flex counters delay indicator to prevent flex counters enablement after fast-reboot.

#### How I did it
Modify config DB json file with 'true' status for delay of flex counters indicator.

#### How to verify it
Run fast-reboot and observe counters are created only when enable_counters script is called, even if the tables are present in config DB.

### DIFF
--- a/counterpoll/main.py
+++ b/counterpoll/main.py
@@ -276,7 +276,7 @@ def show():
 
     click.echo(tabulate(data, headers=header, tablefmt="simple", missingval=""))
 
-def _update_config_db(status, filename):
+def _update_config_db_flex_counter_table(status, filename):
     """ Update counter configuration in config_db file """
     with open(filename) as config_db_file:
         config_db = json.load(config_db_file)
@@ -308,16 +308,16 @@ def config_db():
 @click.argument("filename", default="/etc/sonic/config_db.json", type=click.Path(exists=True))
 def enable(filename):
     """ Enable counter configuration in config_db file """
-    _update_config_db("enable", filename)
+    _update_config_db_flex_counter_table("enable", filename)
 
 @config_db.command()
 @click.argument("filename", default="/etc/sonic/config_db.json", type=click.Path(exists=True))
 def disable(filename):
     """ Disable counter configuration in config_db file """
-    _update_config_db("disable", filename)
+    _update_config_db_flex_counter_table("disable", filename)
 
 @config_db.command()
 @click.argument("filename", default="/etc/sonic/config_db.json", type=click.Path(exists=True))
 def delay(filename):
     """ Delay counters in config_db file """
-    _update_config_db("delay", filename)
+    _update_config_db_flex_counter_table("delay", filename)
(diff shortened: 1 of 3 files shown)


yxieca (changes requested):
Please move the change to fast-reboot script itself.

shlomibitton:
> Please move the change to fast-reboot script itself.

@yxieca can you please check now?

yxieca (approved):

counterpoll/main.py
```diff
@@ -283,11 +283,17 @@ def _update_config_db(status, filename):
```
nazariig, line 283: @shlomibitton few comments here:
1. The name of the method `_update_config_db` doesn't reflect the actual implementation
2. No need to read a file if `FLEX_COUNTER_TABLE` is not present

shlomibitton: 1 - Why not? this function is basically modify and save the config_db file.
2 - How can I know if the tables are not present in the file if I would not read it?

nazariig: @shlomibitton you are basically updating the `FLEX_COUNTER_TABLE` but not the entire `CONFIG_DB`. That's why it is not accurate. And for the second - yes, you are right.

shlomibitton: what about _update_config_db_flex_counter_table ?

shlomibitton: Done.

counterpoll/main.py
```diff
@@ -302,11 +308,16 @@ def config_db():
...
 @config_db.command()
 @click.argument("filename", default="/etc/sonic/config_db.json", type=click.Path(exists=True))
 def disable(filename):
     """ Disable counter configuration in config_db file """
-    _update_config_db("disable", filename)
+    _update_config_db_flex_counter_table("disable", filename)
 
+@config_db.command()
+@click.argument("filename", default="/etc/sonic/config_db.json", type=click.Path(exists=True))
+def delay(filename):
```
nazariig, line 321: @shlomibitton do we need a command for the opposite action? Like we have for config-db enable/disable?

shlomibitton: I don't think so, this is basically only for fast-reboot purposes when we want to delay the counters creation, so the only use for this is before fast-reboot.

liat-grozovik:
@shlomibitton is it needed for 202106?
@nazariig could u please refer to the comments?
@vaibhavhd can you please help with review?

shlomibitton:
> @shlomibitton is it needed for 202106?
> @nazariig could u please refer to the comments?
> @vaibhavhd can you please help with review?

@liat-grozovik Yes we should, thanks.

counterpoll/main.py
nazariig: @shlomibitton ok

PR #1773, opened 2021-08-18
[config reload] Call systemctl reset-failed for snmp,telemetry,mgmt-framework services

Junchao-Mellanox:
#### What I did

When issue `config reload -y` or `config load_minigraph -y` command, most of the sonic services will be reset by command `systemctl reset-failed <service_name>`. The purpose is to avoid services reach to its start retry limit and cannot be started. However, `systemctl reset-failed` only resets those services belong to sonic.target, snmp, telemetry and mgmt-framework are not part of them. So if we run `config reload -y` or `config load_minigraph -y` continues, snmp, telemetry and mgmt-framework services might enter into failed state. This PR is to fix the issue.

I would like to cherry-pick this fix to 202012 branch, but this fix also depends on PR https://github.com/Azure/sonic-buildimage/pull/7846. So if we decide to cherry-pick this PR to 202012, we need cherry-pick https://github.com/Azure/sonic-buildimage/pull/7846 first.

#### How I did it

Also call `systemctl reset-failed` for services like snmp, telemetry and mgmt-framework. 

#### How to verify it

Manual test.

No command output changes.

No command output changes.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -681,11 +681,16 @@ def _stop_services():
 
 def _get_sonic_services():
     out = clicommon.run_command("systemctl list-dependencies --plain sonic.target | sed '1d'", return_cmd=True)
-    return [unit.strip() for unit in out.splitlines()]
+    return (unit.strip() for unit in out.splitlines())
+
+
+def _get_delayed_sonic_services():
+    out = clicommon.run_command("systemctl list-dependencies --plain sonic-delayed.target | sed '1d'", return_cmd=True)
+    return (unit.strip().rstrip('.timer') for unit in out.splitlines())
 
 
 def _reset_failed_services():
-    for service in _get_sonic_services():
+    for service in itertools.chain(_get_sonic_services(), _get_delayed_sonic_services()):
         clicommon.run_command("systemctl reset-failed {}".format(service))
 
 
--- a/tests/config_test.py
+++ b/tests/config_test.py
@@ -55,7 +55,7 @@ class TestLoadMinigraph(object):
             traceback.print_tb(result.exc_info[2])
             assert result.exit_code == 0
             assert "\n".join([l.rstrip() for l in result.output.split('\n')]) == load_minigraph_command_output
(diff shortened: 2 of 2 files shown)


config/main.py
```diff
@@ -681,11 +681,16 @@ def _stop_services():
 
 def _get_sonic_services():
     out = clicommon.run_command("systemctl list-dependencies --plain sonic.target | sed '1d'", return_cmd=True)
-    return [unit.strip() for unit in out.splitlines()]
+    return (unit.strip() for unit in out.splitlines())
```
qiluo-msft, line 684: >) [](http://example.com/codeflow?start=53&length=1)

Is there any problem when using list instead of tuple?

tests/config_test.py
```diff
@@ -55,7 +55,7 @@ def test_load_minigraph(self, get_cmd_module, setup_single_broadcom_asic):
             traceback.print_tb(result.exc_info[2])
             assert result.exit_code == 0
             assert "\n".join([l.rstrip() for l in result.output.split('\n')]) == load_minigraph_command_output
-            assert mock_run_command.call_count == 7
+            assert mock_run_command.call_count == 8
```
qiluo-msft, line 58: >call_count [](http://example.com/codeflow?start=36&length=10)

Could you test your feature/fix in this test case?

config/main.py
Junchao-Mellanox: It is a generator. I suppose generator is more efficiency.

```python
a = (str(x) for x in range(10))
print(type(a))
```

Output:
```
<class 'generator'>
```

tests/config_test.py
Junchao-Mellanox: Sure. Added a few checks in this test case, please help review.

qiluo-msft (approved):

qiluo-msft:
This commit could not be cleanly cherry-pick to 202012. Please submit another PR.

PR #1774, opened 2021-08-18
[202012][fast-reboot] Remove FLEX_COUNTER_TABLE from config_db.json b…

vaibhavhd:
Fix https://github.com/Azure/sonic-buildimage/issues/8523

#### What I did
Remove `FLEX_COUNTER_TABLE` from config_db.json before fast-reboot to allow delaying FLEX counter polling after fast-reboot.

Delaying FLEX counter polling is important to keep fastboot dataplane downtime under 30s.

#### How I did it
In the going down path, add a step to modify config_db.json - remove the key:value for `FLEX_COUNTER_TABLE` table.

#### How to verify it

Repro'd the issue in the latest 202012 image.

With the fix, the counter polling is delayed and downtime is back to normal.

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -20,6 +20,7 @@ PLATFORM=$(sonic-cfggen -H -v DEVICE_METADATA.localhost.platform)
 PLATFORM_PLUGIN="${REBOOT_TYPE}_plugin"
 LOG_SSD_HEALTH="/usr/local/bin/log_ssd_health"
 SSD_FW_UPDATE="ssd-fw-upgrade"
+CONFIG_DB_FILE="/etc/sonic/config_db.json"
 TAG_LATEST=yes
 
 # Require 100M available on the hard drive for warm reboot temp files,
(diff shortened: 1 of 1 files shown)


scripts/fast-reboot
```diff
@@ -679,6 +679,14 @@ then
   systemctl stop "$service_name"
 fi
 
+if [[ "$REBOOT_TYPE" = "fast-reboot" ]]; then
+    # Remove FLEX_COUNTER_TABLE from config_db.json
+    # This is done so that in fast-reboot recovery path, FLEX_COUNTER polling is delayed.
+    # Delayed FLEX_COUNTER polling is an attempt keep dataplane downtime below 30s threshold
+    jq --indent 4  'del(.FLEX_COUNTER_TABLE)' ${CONFIG_DB_FILE} > ${CONFIG_DB_FILE}.new
```
prsunny, line 686: How do we restore the config, if use has made some configs?

vaibhavhd: For branches (201811, 201911, 202012):
The user changed configuration to `FLEX_COUNTER_TABLE` will be reset. This is acceptable so far as for the upgrades, we anyway delete the `/host/old_config/config_db.json` file (basically removing all of the old config, and letting new image take the default config).

For newer branches (2021**):
Below PRs are targeted to address the issue of keeping the user configurations after reboot. The idea is to maintain a delay indicator that OA will check and decide to enable/disable flex counter polling.

Adding you to the discussion thread about this topic.

yxieca (approved):

Issue #1753, opened 2021-08-11
[ACL] Incorrect behavior for create ACL table binding to VLAN


ezio-chen:
#### Description

In https://github.com/Azure/sonic-utilities/pull/1475, binding to VLAN will be expanded into VLAN member.

I think the behavior is incorrectly.
If you change the VLAN member ports, the new member will not be restricted by ACL.
Or the member port had join multiple VLANs, other VLAN traffic will also be affected.

I think user shall bind the ACL to member ports and filter VLAN ID in rules if the platforms do not support binding to the VLAN directly.

qiluo-msft:
> Or the member port had join multiple VLANs, other VLAN traffic will also be affected

This is not true. When vlan acl table bind to vlan member port, we adapt the acl rules in the table, ie. add a vlan id rule into the table, so the behavior should be the same.

qiluo-msft:
@ganglyu for awareness.

PR #1775, opened 2021-08-20
[ACL] Rollback the code of #1475 "Expand VLAN into VLAN members..."

ezio-chen:
#### What I did
For the issue of https://github.com/Azure/sonic-utilities/issues/1753, it shall not expand the VLAN to VLAN member ports for all platform.
Also modify the unit test because the behavior is changed.

#### How I did it
Only expand VLAN to VLAN member for asic type is "mellanox"

#### How to verify it
Pass the unit test

```
admin@sonic:~$ show vlan br
+-----------+--------------+-----------+----------------+-----------------------+-------------+
|   VLAN ID | IP Address   | Ports     | Port Tagging   | DHCP Helper Address   | Proxy ARP   |
+===========+==============+===========+================+=======================+=============+
|       100 |              | Ethernet0 | tagged         |                       | disabled    |
+-----------+--------------+-----------+----------------+-----------------------+-------------+
admin@sonic:~$ sudo config acl add table L3_TABLE L3 -p Vlan100
admin@sonic:~$ show acl table
Name      Type    Binding    Description    Stage
--------  ------  ---------  -------------  -------
L3_TABLE  L3      Ethernet0  L3_TABLE       ingress
admin@sonic:~$

```
```
admin@sonic:~$ show vlan br
+-----------+--------------+-----------+----------------+-----------------------+-------------+
|   VLAN ID | IP Address   | Ports     | Port Tagging   | DHCP Helper Address   | Proxy ARP   |
+===========+==============+===========+================+=======================+=============+
|       100 |              | Ethernet0 | tagged         |                       | disabled    |
+-----------+--------------+-----------+----------------+-----------------------+-------------+
admin@sonic:~$ sudo config acl add table L3_TABLE L3 -p Vlan100
admin@sonic:~$ show acl table
Name      Type    Binding    Description    Stage
--------  ------  ---------  -------------  -------
L3_TABLE  L3      Vlan100    L3_TABLE       ingress
admin@sonic:~$

```

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -4569,31 +4569,14 @@ def get_acl_bound_ports():
 
     return list(ports)
 
-
-def expand_vlan_ports(port_name):
-    """
-    Expands a given VLAN interface into its member ports.
-
-    If the provided interface is a VLAN, then this method will return its member ports.
-
-    If the provided interface is not a VLAN, then this method will return a list with only
-    the provided interface in it.
-    """
+def get_acl_valid_vlans():
     config_db = ConfigDBConnector()
     config_db.connect()
-
-    if port_name not in config_db.get_keys("VLAN"):
-        return [port_name]
-
-    vlan_members = config_db.get_keys("VLAN_MEMBER")
-
-    members = [member for vlan, member in vlan_members if port_name == vlan]
-
-    if not members:
-        raise ValueError("Cannot bind empty VLAN {}".format(port_name))
-
-    return members
-
+    vlans = set()
+    vlan_dict = config_db.get_table("VLAN")
+    for key in vlan_dict:
+        vlans.add(key)
+    return list(vlans)
 
 def parse_acl_table_info(table_name, table_type, description, ports, stage):
     table_info = {"type": table_type}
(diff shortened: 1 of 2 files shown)


prsunny (commented):
Why remove the code? Not all platforms may support Vlan bind. So suggest adding an option in CLI to use vlan binding. If the option is not specified, let it expand as it is done currently. @qiluo-msft , @bingwang-ms , what do you think?

bingwang-ms:
> Why remove the code? Not all platforms may support Vlan bind. So suggest adding an option in CLI to use vlan binding. If the option is not specified, let it expand as it is done currently. @qiluo-msft , @bingwang-ms , what do you think?

Agree. The Vlan expland feature is required in some scenarios. Please consider add an option to disable this behavior.

ezio-chen:
I still think this change is unreasonable, if a port join multiple VLAN, the ACL will affect to all VLANs on these ports.

For example, Ethernet0,Ethernet1 is the member of VLAN100 and VLAN200,
and you use the command to create the ACL
 ```sudo config acl add table Vlan100_ACL L3 -p Vlan100```
but the rules in Vlan100_ACL  will affect all traffic in Vlan200.

Why user just use  ```sudo config acl add table Vlan100_ACL L3 -p Ethernet0,Ethernet1 ``` to create ACL table if they really want to bind the ACL on these ports ?

qiluo-msft:
Thanks @ezio-chen for the contribution!
I agree current implementation is not perfect. We introduce this feature to temporarily enable some use case on Mellanox SN2700 platform, which does not support vlan acl binding.

I agree we should remove this feature in long term. To prevent sudden feature broken in short term, could you keep the behavior for Mellanox ASIC platforms and only change the behavior for all others?

qiluo-msft (changes requested):
As comment

ezio-chen:
OK, and for this modification, can you please give some suggestions? 

Method 1: Distinguishing the HWSKU type internal, only expland VLAN on Mellanox.
Method 2: Add CLI options, e.g., config acl add table <table_name> <table_type> [ --expland-vlan ]
or other good ideas  ?

qiluo-msft:
Method 1 looks good to me.

PR #1786, opened 2021-08-27
[config reload][202012] Call systemctl reset-failed for snmp,telemetry,mgmt-f…

Junchao-Mellanox:
Back port PR https://github.com/Azure/sonic-utilities/pull/1773 to 202012 branch.

#### What I did

When issue `config reload -y` or `config load_minigraph -y` command, most of the sonic services will be reset by command `systemctl reset-failed <service_name>`. The purpose is to avoid services reach to its start retry limit and cannot be started. However, `systemctl reset-failed` only resets those services belong to sonic.target, snmp, telemetry and mgmt-framework are not part of them. So if we run `config reload -y` or `config load_minigraph -y` continues, snmp, telemetry and mgmt-framework services might enter into failed state. This PR is to fix the issue.

#### How I did it

Also call `systemctl reset-failed` for services like snmp, telemetry and mgmt-framework. 

#### How to verify it

Manual test.

No command output changes.

No command output changes.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -701,12 +701,16 @@ def _stop_services():
 
 def _get_sonic_services():
     out = clicommon.run_command("systemctl list-dependencies --plain sonic.target | sed '1d'", return_cmd=True)
-    return [unit.strip() for unit in out.splitlines()]
+    return (unit.strip() for unit in out.splitlines())
+
+
+def _get_delayed_sonic_services():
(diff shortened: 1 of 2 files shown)


dprital:
@qiluo-msft  - Can you please review. Thanks.

qiluo-msft (approved):

PR #1791, opened 2021-09-01
Add --skip-platform-check option for sonic_installer when image mismatch

wen587:
#### What I did
Add --skip-platform-check option for sonic_installer; tune up comment for secureboot
#### How I did it
Add --skip-platform-check option for sonic_installer
#### How to verify it
Install a bin file which differs the running platform's ASIC. For example;
install sonic-broadcom.bin to mellanox ASIC platform
install sonic-vs.bin to broadcom ASIC platform
sudo sonic-installer install --help
Usage: sonic-installer install [OPTIONS] URL

  Install image from local binary or URL

Options:
  -y, --yes
  -f, --force_install       Force installation of an image of a type which
                            differs from that of the current running image
  --skip_migration          Do not migrate current configuration to the newly
                            installed image
  --skip-package-migration  Do not migrate current packages to the newly
                            installed image
  --help                    Show this message and exit.

sudo sonic-installer install --help
Usage: sonic-installer install [OPTIONS] URL

  Install image from local binary or URL

Options:
   -y, --yes
   -f, --force, --skip-secure-check
                                  Force installation of an image of a non-
                                  secure type than secure running image
  --skip-platform-check           Force installation of an image of a type
                                  which is not of the same platform
  --skip_migration                Do not migrate current configuration to the
                                  newly installed image
  --skip-package-migration        Do not migrate current packages to the newly
                                  installed image
 --help                    Show this message and exit.

### DIFF
--- a/sonic_installer/bootloader/aboot.py
+++ b/sonic_installer/bootloader/aboot.py
@@ -164,6 +164,9 @@ class AbootBootloader(Bootloader):
         return IMAGE_PREFIX + version.strip()
 
     def verify_binary_image(self, image_path):
+        return os.path.isfile(image_path)
+
+    def verify_secureboot_image(self, image_path):
         try:
             subprocess.check_call(['/usr/bin/unzip', '-tq', image_path])
             return self._verify_secureboot_image(image_path)
--- a/sonic_installer/bootloader/aboot.py
+++ b/sonic_installer/bootloader/aboot.py
@@ -163,7 +163,7 @@ class AbootBootloader(Bootloader):
             return None
         return IMAGE_PREFIX + version.strip()
 
-    def verify_binary_image(self, image_path):
+    def verify_image_platform(self, image_path):
         return os.path.isfile(image_path)
 
     def verify_secureboot_image(self, image_path):
--- a/sonic_installer/bootloader/grub.py
+++ b/sonic_installer/bootloader/grub.py
@@ -98,7 +98,6 @@ class GrubBootloader(OnieInstallerBootloader):
                 asic_type = None
         except (KeyError, TypeError) as e:
             click.echo("Caught an exception: " + str(e))
-        # click.echo('Get running platform ASIC... %s' % asic_type)
 
         # Get installing image's ASIC
         p1 = subprocess.Popen(["sed", "-e", "1,/^exit_marker$/d", image_path], stdout=subprocess.PIPE, preexec_fn=default_sigpipe)
@@ -106,11 +105,9 @@ class GrubBootloader(OnieInstallerBootloader):
         p3 = subprocess.Popen(["sed", "-n", r"s/^machine=\(.*\)/\1/p"], stdin=p2.stdout, stdout=subprocess.PIPE, preexec_fn=default_sigpipe, text=True)
 
         stdout = p3.communicate()[0]
-        p3.wait()
         image_asic = stdout.rstrip('\n')
-        # click.echo('Get installing images ASIC... %s' % image_asic)
 
-        # Backword compatibility if IMAGE_ASIC not set in install.sh
+        # Return true if machine is not found
         if not image_asic:
             return True
 
--- a/sonic_installer/bootloader/grub.py
+++ b/sonic_installer/bootloader/grub.py
@@ -102,6 +102,10 @@ class GrubBootloader(OnieInstallerBootloader):
         image_asic = stdout.rstrip('\n')
         # click.echo('Get installing images ASIC... %s' % image_asic)
 
+        # Backword compatibility if IMAGE_ASIC not set in install.sh
+        if not image_asic:
+            return True
+
         if asic_type == image_asic:
             return True
         return False
--- a/sonic_installer/bootloader/grub.py
+++ b/sonic_installer/bootloader/grub.py
@@ -8,6 +8,7 @@ import subprocess
 
 import click
 
+from sonic_py_common import device_info
 from ..common import (
    HOST_PATH,
    IMAGE_DIR_PREFIX,
@@ -82,14 +83,19 @@ class GrubBootloader(OnieInstallerBootloader):
         run_command('grub-set-default --boot-directory=' + HOST_PATH + ' 0')
         click.echo('Image removed')
 
-    def verify_binary_image(self, image_path):
+    def verify_image_platform(self, image_path):
         if not os.path.isfile(image_path):
             return False
 
         # Get running platform's ASIC
-        sonic_version_yml = open('/etc/sonic/sonic_version.yml', 'r')
-        asic_type = re.search(r"asic_type: (\S+)", sonic_version_yml.read()).group(1)
-        sonic_version_yml.close()
+        try:
+            version_info = device_info.get_sonic_version_info()
+            if version_info:
+                asic_type = version_info['asic_type']
+            else:
+                asic_type = None
+        except (KeyError, TypeError) as e:
+            click.echo("Caught an exception: " + str(e))
         # click.echo('Get running platform ASIC... %s' % asic_type)
 
         # Get installing image's ASIC
--- a/sonic_installer/bootloader/grub.py
+++ b/sonic_installer/bootloader/grub.py
@@ -107,9 +107,9 @@ class GrubBootloader(OnieInstallerBootloader):
         stdout = p3.communicate()[0]
         image_asic = stdout.rstrip('\n')
 
-        # Return true if machine is not found
+        # Return false if machine is not found or unexpected issue occur
         if not image_asic:
-            return True
+            return False
 
         if asic_type == image_asic:
             return True
--- a/sonic_installer/bootloader/grub.py
+++ b/sonic_installer/bootloader/grub.py
@@ -15,6 +15,7 @@ from ..common import (
    run_command,
 )
 from .onie import OnieInstallerBootloader
+from .onie import default_sigpipe
 
 class GrubBootloader(OnieInstallerBootloader):
 
@@ -80,6 +81,30 @@ class GrubBootloader(OnieInstallerBootloader):
 
         run_command('grub-set-default --boot-directory=' + HOST_PATH + ' 0')
         click.echo('Image removed')
+
+    def verify_binary_image(self, image_path):
+        if not os.path.isfile(image_path):
+            return False
+
+        # Get running platform's ASIC
+        sonic_version_yml = open('/etc/sonic/sonic_version.yml', 'r')
+        asic_type = re.search(r"asic_type: (\S+)", sonic_version_yml.read()).group(1)
+        sonic_version_yml.close()
+        # click.echo('Get running platform ASIC... %s' % asic_type)
+
+        # Get installing image's ASIC
+        p1 = subprocess.Popen(["cat", "-v", image_path], stdout=subprocess.PIPE, preexec_fn=default_sigpipe)
+        p2 = subprocess.Popen(["grep", "-m 1", "^image_asic"], stdin=p1.stdout, stdout=subprocess.PIPE, preexec_fn=default_sigpipe)
+        p3 = subprocess.Popen(["sed", "-n", r"s/^image_asic=\"\(.*\)\"$/\1/p"], stdin=p2.stdout, stdout=subprocess.PIPE, preexec_fn=default_sigpipe, text=True)
+
+        stdout = p3.communicate()[0]
+        p3.wait()
+        image_asic = stdout.rstrip('\n')
+        # click.echo('Get installing images ASIC... %s' % image_asic)
+
+        if asic_type == image_asic:
+            return True
+        return False
 
     @classmethod
     def detect(cls):
--- a/sonic_installer/bootloader/grub.py
+++ b/sonic_installer/bootloader/grub.py
@@ -18,6 +18,8 @@ from ..common import (
 from .onie import OnieInstallerBootloader
 from .onie import default_sigpipe
 
+MACHINE_CONF = "installer/machine.conf"
+
 class GrubBootloader(OnieInstallerBootloader):
 
     NAME = 'grub'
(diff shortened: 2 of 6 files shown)


sonic_installer/main.py
```diff
@@ -480,7 +480,9 @@ def sonic_installer():
 @sonic_installer.command('install')
 @click.option('-y', '--yes', is_flag=True, callback=abort_if_false,
               expose_value=False, prompt='New image will be installed, continue?')
-@click.option('-f', '--force', is_flag=True,
+@click.option('--force_secureboot', is_flag=True,
```
qiluo-msft, line 483: >force_secureboot [](http://example.com/codeflow?start=17&length=16)

After double thinking, we should keep old short option and long option, so backward-compatiable. We can modify the help text.

sonic_installer/main.py
```diff
@@ -480,7 +480,9 @@ def sonic_installer():
...
+              help="Force installation of an image of a type which is not secureboot")
+@click.option('-f', '--force_install', is_flag=True,
```
qiluo-msft, line 485: >'-f', '--force_install' [](http://example.com/codeflow?start=14&length=23)

How about '--skip-platform-check'. Since this one is not used much, it's okay to have only long option.

qiluo-msft: Use dash instead of underscore in option name.

sonic_installer/main.py
```diff
@@ -480,7 +480,9 @@ def sonic_installer():
 @sonic_installer.command('install')
 @click.option('-y', '--yes', is_flag=True, callback=abort_if_false,
               expose_value=False, prompt='New image will be installed, continue?')
-@click.option('-f', '--force', is_flag=True,
```
qiluo-msft, line 483: >@click.option('-f', '--force [](http://example.com/codeflow?start=0&length=28)

You can use
```
@click.option('-f', '--force', 'force_secureboot', ...)
```

To keep old option string and have a good python var name.

qiluo-msft: However `force_secureboot` is not accurate. The only use case is the installation path secure->non-secure. So the best var name should be "skip_secure_check"

sonic_installer/main.py
```diff
@@ -530,10 +532,17 @@ def install(url, force, skip_migration=False, skip_package_migration=False,
             echo_and_log('Error: Failed to set image as default', LOG_ERR)
             raise click.Abort()
     else:
+        # Verify that the binary image is secureboot
```
qiluo-msft, line 535: >Verify that the binary image is secureboot [](http://example.com/codeflow?start=10&length=42)

Not accurate.
"Verify not installing non-secure image in a secure running image"

sonic_installer/main.py
```diff
@@ -530,10 +532,17 @@ def install(url, force, skip_migration=False, skip_package_migration=False,
...
+        if not bootloader.verify_secureboot_image(image_path) and not force_secureboot:
+            echo_and_log("Image file '{}' is not a secureboot image.\n".format(url) +
+                "If you are sure you want to install this image, use --force_secureboot.\n" +
+                "Aborting...", LOG_ERR)
+            raise click.Abort()
+
         # Verify that the binary image is of the same type as the running image
-        if not bootloader.verify_binary_image(image_path) and not force:
```
qiluo-msft, line 534: >verify_binary_image [](http://example.com/codeflow?start=26&length=19)

The function name is better named as 'verify_image_platform'

sonic_installer/main.py
```diff
@@ -530,10 +532,17 @@ def install(url, force, skip_migration=False, skip_package_migration=False,
...
             raise click.Abort()
     else:
+        # Verify that the binary image is secureboot
+        if not bootloader.verify_secureboot_image(image_path) and not force_secureboot:
+            echo_and_log("Image file '{}' is not a secureboot image.\n".format(url) +
+                "If you are sure you want to install this image, use --force_secureboot.\n" +
+                "Aborting...", LOG_ERR)
+            raise click.Abort()
+
         # Verify that the binary image is of the same type as the running image
```
qiluo-msft, line 542: >Verify that the binary image is of the same type as the running image [](http://example.com/codeflow?start=10&length=69)

The comment is not accurate. Has nothing to do with "the running image".

sonic_installer/bootloader/grub.py
```diff
@@ -81,6 +82,34 @@ def remove_image(self, image):
...
         click.echo('Image removed')
 
+    def verify_binary_image(self, image_path):
+        if not os.path.isfile(image_path):
+            return False
+
+        # Get running platform's ASIC
+        sonic_version_yml = open('/etc/sonic/sonic_version.yml', 'r')
+        asic_type = re.search(r"asic_type: (\S+)", sonic_version_yml.read()).group(1)
+        sonic_version_yml.close()
```
qiluo-msft, line 92: Could you reuse common library: [https://github.com/Azure/sonic-buildimage/blob/7c9be314e2f390a5f2c522f74128ac5e1e2b6230/src/sonic-py-common/sonic_py_common/device_info.py#L294](https://github.com/Azure/sonic-buildimage/blob/7c9be314e2f390a5f2c522f74128ac5e1e2b6230/src/sonic-py-common/sonic_py_common/device_info.py#L294)

sonic_installer/bootloader/grub.py
```diff
@@ -81,6 +82,34 @@ def remove_image(self, image):
...
+        # click.echo('Get running platform ASIC... %s' % asic_type)
+
+        # Get installing image's ASIC
+        p1 = subprocess.Popen(["cat", "-v", image_path], stdout=subprocess.PIPE, preexec_fn=default_sigpipe)
```
qiluo-msft, line 96: >p1 [](http://example.com/codeflow?start=8&length=2)

Did you handle the return code of each process?

sonic_installer/bootloader/grub.py
```diff
@@ -81,6 +82,34 @@ def remove_image(self, image):
...
+        p2 = subprocess.Popen(["grep", "-m 1", "^image_asic"], stdin=p1.stdout, stdout=subprocess.PIPE, preexec_fn=default_sigpipe)
+        p3 = subprocess.Popen(["sed", "-n", r"s/^image_asic=\"\(.*\)\"$/\1/p"], stdin=p2.stdout, stdout=subprocess.PIPE, preexec_fn=default_sigpipe, text=True)
```
qiluo-msft, line 98: >image_asic [](http://example.com/codeflow?start=49&length=10)

This solution is pending on [https://github.com/Azure/sonic-buildimage/pull/8648/files](https://github.com/Azure/sonic-buildimage/pull/8648/files). Will check after that PR merged.

sonic_installer/bootloader/grub.py
wen587: https://github.com/Azure/sonic-utilities/blob/master/sonic_installer/bootloader/onie.py#L33
I am referring from here. 
If 'image_asic' is missing in binary image, final result string will be empty. It will not impact the comparison.
p1 will throw exception if image_path doesn't exist, but this is checked in previous step.
Is this solve your question?

sonic_installer/main.py
wen587: what about 
Verify the binary image is of the same platform type as the running platform.

sonic_installer/main.py
wen587: Should we leave the echo message as it previous states?
"Image file '{}' is of a different type than running image"

sonic_installer/main.py
```diff
@@ -530,10 +532,17 @@ def install(url, force, skip_migration=False, skip_package_migration=False,
...
-            echo_and_log("Image file '{}' is of a different type than running image.\n".format(url) +
-                "If you are sure you want to install this image, use -f|--force.\n" +
+        # Verify not installing non-secure image in a secure running image
+        if not bootloader.verify_secureboot_image(image_path) and not force:
+            echo_and_log("Image file '{}' is of a non-secure type than secure running image.\n".format(url) +
+                "If you are sure you want to install this image, use -f|--force|--skip-secure-check.\n" +
+                "Aborting...", LOG_ERR)
+            raise click.Abort()
+
+        # Verify that the binary image is of the same platform type than running platform
```
qiluo-msft, line 542: >than [](http://example.com/codeflow?start=68&length=4)

then -> as

sonic_installer/main.py
qiluo-msft: yes, we can keep the ech omessage.

sonic_installer/bootloader/grub.py
qiluo-msft: I see the existing code sample. My question is about general usage of `subprocess.Popen` and pipe. If upstream process exit with a non-zero code, does the whole pipe stop immediately? will the last process `p3` get an non-zero error code? or do we need to check all the return code of p1, p2, p3?

wen587: Checked locally. Downstream could continue to run even upstream exit non-zero code and whole pop will not stop.
I think it is fine.

qiluo-msft:
@Staphylo Could you help review?

sonic_installer/bootloader/aboot.py
```diff
@@ -163,7 +163,10 @@ def get_binary_image_version(self, image_path):
             return None
         return IMAGE_PREFIX + version.strip()
 
-    def verify_binary_image(self, image_path):
+    def verify_image_platform(self, image_path):
+        return os.path.isfile(image_path)
```
Staphylo, line 167: The `verify_image_platform` function will not be doing as much on Arista platforms than it is on ONIE but this is something we can add at a later time.
I think for now it's fine to just check if the file is there.

Staphylo (approved):

sonic_installer/bootloader/grub.py
```diff
@@ -81,6 +85,39 @@ def remove_image(self, image):
...
+            click.echo("Caught an exception: " + str(e))
+        # click.echo('Get running platform ASIC... %s' % asic_type)
+
+        # Get installing image's ASIC
+        p1 = subprocess.Popen(["sed", "-e", "1,/^exit_marker$/d", image_path], stdout=subprocess.PIPE, preexec_fn=default_sigpipe)
+        p2 = subprocess.Popen(["tar", "xf", "-", MACHINE_CONF, "-O"], stdin=p1.stdout, stdout=subprocess.PIPE, preexec_fn=default_sigpipe)
+        p3 = subprocess.Popen(["sed", "-n", r"s/^machine=\(.*\)/\1/p"], stdin=p2.stdout, stdout=subprocess.PIPE, preexec_fn=default_sigpipe, text=True)
+
+        stdout = p3.communicate()[0]
+        p3.wait()
```
lolyu, line 109: No need to call `wait` as `communicate` will ensure the process termination.

sonic_installer/bootloader/grub.py
```diff
@@ -81,6 +85,39 @@ def remove_image(self, image):
...
+        # Get running platform's ASIC
+        try:
+            version_info = device_info.get_sonic_version_info()
+            if version_info:
+                asic_type = version_info['asic_type']
+            else:
+                asic_type = None
+        except (KeyError, TypeError) as e:
+            click.echo("Caught an exception: " + str(e))
+        # click.echo('Get running platform ASIC... %s' % asic_type)
```
lolyu, line 101: Could you pls remove the debug comment?

sonic_installer/bootloader/grub.py
```diff
@@ -81,6 +85,39 @@ def remove_image(self, image):
...
+
+        # Get installing image's ASIC
+        p1 = subprocess.Popen(["sed", "-e", "1,/^exit_marker$/d", image_path], stdout=subprocess.PIPE, preexec_fn=default_sigpipe)
+        p2 = subprocess.Popen(["tar", "xf", "-", MACHINE_CONF, "-O"], stdin=p1.stdout, stdout=subprocess.PIPE, preexec_fn=default_sigpipe)
+        p3 = subprocess.Popen(["sed", "-n", r"s/^machine=\(.*\)/\1/p"], stdin=p2.stdout, stdout=subprocess.PIPE, preexec_fn=default_sigpipe, text=True)
+
+        stdout = p3.communicate()[0]
+        p3.wait()
+        image_asic = stdout.rstrip('\n')
+        # click.echo('Get installing images ASIC... %s' % image_asic)
```
lolyu, line 111: Same as above.

sonic_installer/bootloader/grub.py
```diff
@@ -81,6 +85,39 @@ def remove_image(self, image):
...
+
+        # Backword compatibility if IMAGE_ASIC not set in install.sh
```
lolyu, line 113: `s/backword/backward`

lolyu (approved):

sonic_installer/bootloader/grub.py
qiluo-msft: If you does not check non-zero return code in the pipeline, you will get an empty `image_asic`, and this function will return True. This is very dangerous. Instead, we need to report this as an unexpected failure, or at least return False to be safe.

wen587: Agreed. Change to return False.

qiluo-msft (approved):

lolyu (approved):

sonic_installer/bootloader/grub.py
```diff
@@ -81,6 +85,36 @@ def remove_image(self, image):
...
 
+    def verify_image_platform(self, image_path):
+        if not os.path.isfile(image_path):
+            return False
+
+        # Get running platform's ASIC
+        try:
+            version_info = device_info.get_sonic_version_info()
+            if version_info:
+                asic_type = version_info['asic_type']
```
qiluo-msft, line 96: >version_info['asic_type'] [](http://example.com/codeflow?start=28&length=25)

This not "running platform's ASIC". It is actually the current image's target ASIC.

For example:
1. considering you already force install a broadcom image onto a mellanox platform.
2. and in the running sonic (it is not working well), you want to install a next mellanox image without force
3. version info will tell you current asic is broadcom, and actually the running system ASIC is mellanox.

sonic_installer/main.py
```diff
@@ -530,10 +532,17 @@ def install(url, force, skip_migration=False, skip_package_migration=False,
...
-        if not bootloader.verify_binary_image(image_path) and not force:
+        # Verify not installing non-secure image in a secure running image
+        if not bootloader.verify_secureboot_image(image_path) and not force:
             echo_and_log("Image file '{}' is of a different type than running image.\n".format(url) +
-                "If you are sure you want to install this image, use -f|--force.\n" +
+                "If you are sure you want to install this image, use -f|--force|--skip-secure-check.\n" +
+                "Aborting...", LOG_ERR)
+            raise click.Abort()
+
+        # Verify that the binary image is of the same platform type as running platform
```
qiluo-msft, line 542: >platform [](http://example.com/codeflow?start=54&length=8)

platform -> asic platform

qiluo-msft:
@judyjoseph Could you check if this PR works on broadcom-dnx platforms?

PR #1817, opened 2021-09-13
[202012][sonic-utilities] CLI support for port auto negotiation

vdahiya12:
1. Add CLI support for port auto negotiation feature.
2. Add db_migrator change for auto negotiation feature
2. Add unit test cases for all changes

1. Add new subcommands to "config interface" command group to allow user configuring port auto negotiation
2. Add new subcommands to "show interfaces" command group to allow user show auto negotiation status
3. In db_migrator.py, change auto negotiation related DB field to latest one

Adding this PR because the cherry-pick from original PR #1568 
has conflicts 

#### How to verify it

Ran the changes on Arista7050cx3 testbed 

```
admin@sonic:~$ sudo config interface autoneg Ethernet96 enabled
admin@sonic:~$ echo $?
0
```

```
  Interface    Auto-Neg Mode    Speed    Oper    Admin
-----------  ---------------  -------  ------  -------
  Ethernet0              N/A     100G    down     down
  Ethernet4              N/A     100G    down       up
  Ethernet8              N/A     100G    down       up
 Ethernet12              N/A     100G    down       up
 Ethernet16              N/A     100G    down       up
 Ethernet20              N/A     100G    down     down
 Ethernet24         disabled     100G    down       up
 Ethernet28              N/A     100G    down       up
 Ethernet32              N/A     100G    down       up
 Ethernet36              N/A     100G    down       up
 Ethernet40              N/A     100G    down     down
 Ethernet44              N/A     100G    down       up
 Ethernet48              N/A     100G    down       up
 Ethernet52              N/A     100G    down       up
 Ethernet56              N/A     100G    down       up
 Ethernet60              N/A     100G    down       up
 Ethernet64              N/A     100G    down     down
 Ethernet68              N/A     100G    down     down
 Ethernet72              N/A     100G    down       up
 Ethernet76              N/A     100G    down       up
 Ethernet80              N/A     100G    down       up
 Ethernet84              N/A     100G    down     down
 Ethernet88              N/A     100G    down       up
 Ethernet92              N/A     100G    down       up
 Ethernet96          enabled     100G      up       up
Ethernet100              N/A     100G    down       up
Ethernet104              N/A     100G    down       up
Ethernet108              N/A     100G    down       up
Ethernet112              N/A     100G    down     down
Ethernet116              N/A     100G    down       up
Ethernet120              N/A     100G    down       up
Ethernet124              N/A     100G    down     down
Ethernet128              N/A      10G    down     down
Ethernet132              N/A      10G    down     down
```

[...]

PR #1823, opened 2021-09-16
[sonic-utilities]remove db_migrator logic for autoneg enable/disable

vdahiya12:

#### What I did
This PR removes all the logic for db_migrator which went into #1817 , This is required because this will cause changing the logic to "0/1" to "on/off" in swss and cause a problem in upgrade. 
#### How to verify it
autoneg enable/disable works with the implementation. Removing the db_migration logic so no check required.


scripts/db_migrator.py
```diff
@@ -520,8 +510,6 @@ def version_2_0_0(self):
         Version 2_0_0.
         """
         log.log_info('Handling version_2_0_0')
-        self.migrate_config_db_port_table_for_auto_neg()
-        self.set_version('version_2_0_1')
         return 'version_2_0_1'
 
     def version_2_0_1(self):
```
prsunny, line 515: We can remove this section and keep 2_0_0 as latest

vdahiya12: fixed

prsunny (approved):

PR #1825, opened 2021-09-17
Validate input of ```config mirror_session add```

bingwang-ms:
#### What I did
This PR fixed https://github.com/Azure/sonic-buildimage/issues/7990

Type and value validation is added for CLI ```config mirror_session add```.
The value range is defined below

| Value| Range|
| --- | ----------- |
| src_ip| Invalid IPv4 address |
| dst_ip| Invalid IPv4 address |
|DSCP| [0, 63]|
|TTL|  [0, 255]|
|Queue| [0, 255]|
|GRE_TYPE|[0, 65535]|

#### How I did it
Add a value validation at CLI level.

#### How to verify it
Verified by UT
```
tests/config_mirror_session_test.py::test_mirror_session_add PASSED                                                                                                                                                                                      [ 50%]
tests/config_mirror_session_test.py::test_mirror_session_erspan_add PASSED                                                                                                                                                                               [100%]
```

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -92,6 +92,11 @@ DEFAULT_TPID = "0x8100"
 
 asic_type = None
 
+DSCP_RANGE = click.IntRange(min=0, max=63)
+TTL_RANGE = click.IntRange(min=0, max=255)
+QUEUE_RANGE = click.IntRange(min=0, max=255)
+GRE_TYPE_RANGE = click.IntRange(min=0, max=65535)
+
 #
 # Helper functions
 #
@@ -953,6 +958,19 @@ def cache_arp_entries():
         open(restore_flag_file, 'w').close()
     return success
 
+
+def validate_ipv4_address(ctx, param, ip_addr):
+    """Helper function to validate ipv4 address
+    """
+    try:
+        ip_n = ipaddress.ip_network(ip_addr, False)
+        if ip_n.version != 4:
+            raise click.UsageError("{} is not a valid IPv4 address".format(ip_addr))
+        return ip_addr
+    except ValueError as e:
+        raise click.UsageError(str(e))
+
+
 # This is our main entrypoint - the main 'config' command
 @click.group(cls=clicommon.AbbreviationGroup, context_settings=CONTEXT_SETTINGS)
 @click.pass_context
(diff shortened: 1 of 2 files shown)


raphaelt-nvidia (changes requested):
Since the queue parameter is also used in span, I suggest adding a test for it with the config mirror-session span add command.

bingwang-ms:
> Since the queue parameter is also used in span, I suggest adding a test for it with the config mirror-session span add command.

Thanks. Updated

bingwang-ms:
@raphaelt-nvidia  Hi, could you help to take a look again? Thanks

raphaelt-nvidia:
Question about how the test works: If the test runs only in the mock environment, then LGTM.  If it calls real SAI, then whether trying to configure queue=100 produces errors in the log is platform-dependent, and how it is dependent should change soon when I submit a PR to check the actual maximum queue value via SAI.

bingwang-ms:
> Question about how the test works: If the test runs only in the mock environment, then LGTM. If it calls real SAI, then whether trying to configure queue=100 produces errors in the log is platform-dependent, and how it is dependent should change soon when I submit a PR to check the actual maximum queue value via SAI.

The UT is only running in mocked environment, and there is no orchagent and no syncd, and so there is no SAI. The verification is on CLI level.

raphaelt-nvidia:
SO LGTM.   Should someone else review?

config/main.py
```diff
@@ -953,6 +958,19 @@ def cache_arp_entries():
         open(restore_flag_file, 'w').close()
     return success
 
+
+def validate_ipv4_address(ctx, param, ip_addr):
```
prsunny, line 962: Can you please move this to [common ](https://github.com/Azure/sonic-utilities/blob/master/utilities_common/cli.py#L196)so it can be used by other files in future? -

bingwang-ms: Impmenting a common function to validate IP address is a good suggestion, but this interface is not to be common because this one is implemented as a callback interface of click. I will implement that common interface in another PR.

prsunny (approved):

raphaelt-nvidia:
Our QA has noticed that the new range checking for gre requires the value to be entered in decimal.  The command reference https://github.com/Azure/sonic-utilities/blob/master/doc/Command-Reference.md#mirroring-config-commands doesn't define it explicitly, but the following text suggests that hex format should at least be allowed, if not preferred:

optional - GRE Type in case if user wants to send the packet via GRE tunnel. GRE type could be anything; it could also be left as empty; by default, it is 0x8949 for Mellanox; and 0x88be for the rest of the chips.

@bingwang-ms , what do you think?

PR #1210, opened 2020-11-03
Add `show reboot-cause history` cli

sujinmkang:
- What I did
To make the reboot cause history information available for telemetry service, during processing the reboot-cause on boot, the information is saved in state-db.
The cli command show reboot-cause is changed to support the new file format and file name.
New command show reboot-cause history is added to display the reboot-cause information which is stored in state-db.
- How I did it
Add the 'show reboot-cause history' command to read up to the10 previous reboot-cause information from state-db and display them.
Read the last reboot-cause from the new file "previous-reboot-cause.json" and process the data to display for show reboot-cause

- How to verify it
Unit test added.
Verified also in DUT

- Previous command output (if the output of a command-line utility has changed)

- New command output (if the output of a command-line utility has changed)
```bash
admin@sonic:~$ show reboot-cause history
name                 cause        time                          user    comment
-------------------  -----------  ----------------------------  ------  ---------
2020_10_09_04_53_58  warm-reboot  Fri Oct  9 04:51:47 UTC 2020  admin
2020_10_09_02_33_06  reboot       Fri Oct  9 02:29:44 UTC 2020  admin
2020_10_09_02_00_53  fast-reboot  Fri Oct  9 01:58:04 UTC 2020  admin
2020_10_09_01_56_59  reboot       Fri Oct  9 01:53:49 UTC 2020  admin
```
HLD : Azure/SONiC#669

PR #1826, opened 2021-09-17
[reboot-cause] Fix reboot cause format

Xichen96:
Currently reboot-cause recorded by various reboot scripts is not
recognizable by the show reboot-cause command.

Signed-off-by: Xichen Lin <xichenlin@microsoft.com>

#### What I did
Changed the format of reboot-cause.txt file.

#### How to verify it
Tested on switch

Unknown

User issued 'reboot' command [User: admin, Time: Thu Oct 22 03:11:08 UTC 2020]

### DIFF
--- a/show/reboot_cause.py
+++ b/show/reboot_cause.py
@@ -8,7 +8,7 @@ from swsscommon.swsscommon import SonicV2Connector
 import utilities_common.cli as clicommon
 
 
-PREVIOUS_REBOOT_CAUSE_FILE_PATH = "/host/reboot-cause/previous-reboot-cause.json"
+PREVIOUS_REBOOT_CAUSE_FILE_PATH = "/host/reboot-cause/reboot-cause.txt"
 
 
 def read_reboot_cause_file():
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -710,7 +710,8 @@ fi
 # Update the reboot cause file to reflect that user issued this script
 # Upon next boot, the contents of this file will be used to determine the
 # cause of the previous reboot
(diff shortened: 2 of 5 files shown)


Blueve:
Hi @Xichen96 , could you create an issue and put details in it?

We need more context regarding:
1. Are you testing this with master branch image?
2. Which tests case are failing due to this issue?

Seems like the issue were caused by reboot cause file format is not aligned across different components, some are reading it as a json and some are reading it as plain text.  @yxieca Do you have more information on this?

show/reboot_cause.py
```diff
@@ -8,7 +8,7 @@
 import utilities_common.cli as clicommon
 
 
-PREVIOUS_REBOOT_CAUSE_FILE_PATH = "/host/reboot-cause/previous-reboot-cause.json"
+PREVIOUS_REBOOT_CAUSE_FILE_PATH = "/host/reboot-cause/reboot-cause.txt"
```
Blueve, line 11: Why change this? Who will generated the `previous-reboot-cause.json` file?

I saw the UT were introduced here: https://github.com/Azure/sonic-utilities/pull/1210/files
@sujinmkang Could you please join the review?

Xichen96: I just found a piece of code in sonic-buildimage regarding reboot-cause that I left out. Though reboot-cause still gives "Unknown", I think my fix is only a workaround to passing the test. I will close this pr and create a better solution.

sujinmkang:
@Xichen96 What is the issue you are trying to fix? Can you please open an issue first?

Issue #1830, opened 2021-09-18
[generic-config-updater] Failing to run 'config apply-patch' for a patch to add an ACL


ghooo:
#### Description
Executing the command `config apply-patch` fails for applying a patch to add an ACL

## Steps to reproduce the issue
1. Add the following content in a new file called `acl.json-patch` file
```json
[
 {
  "op": "add",
  "path": "/ACL_TABLE/EVERFLOW",
  "value": {
   "policy_desc": "EVERFLOW",
   "ports": [
    "Ethernet104"
   ],
   "type": "MIRROR"
  }
 }
]
```
2. Run `sudo config apply-patch acl.json-patch`

#### Describe the results you received
```
Failed to apply patch
Usage: config apply-patch [OPTIONS] PATCH_FILE_PATH
Try "config apply-patch -h" for help.

Error: 'BGP_NEIGHBOR'
```

#### Describe the results you expected
```
Failed to apply patch
Usage: config apply-patch [OPTIONS] PATCH_FILE_PATH
Try "config apply-patch -h" for help.

Error: ChangeApplier.apply(change) is not implemented yet
```

#### Additional information you deem important (e.g. issue happens only occasionally)
BGP_NEIGHBOR is a table that does not have YANG model defined

#### Output of `show version`

```
SONiC Software Version: SONiC.master.37690-1a2e85248
Distribution: Debian 10.10
Kernel: 4.19.0-12-2-amd64
Build commit: 1a2e85248
Build date: Fri Sep 17 14:17:43 UTC 2021
Built by: AzDevOps@sonic-build-workers-000PC0

Platform: x86_64-kvm_x86_64-r0
HwSKU: Force10-S6000
ASIC: vs
ASIC Count: 1
Serial Number: N/A
Model Number: N/A
Hardware Revision: N/A
Uptime: 05:10:04 up  5:39,  1 user,  load average: 0.53, 0.59, 0.64

```

ghooo:
Reason for failure is the sonic-yang-mgmt pkg we consume modifies an input parameter in-place, I don't think this should happen.

Here is how:
https://github.com/Azure/sonic-buildimage/blob/e6699a0132c141b000af99521eab1fabf6942921/src/sonic-yang-mgmt/sonic_yang_ext.py#L702
```python
    """
    load_data: load Config DB, crop, xlate and create data tree from it. (Public)
    input:    data
              debug Flag
    returns:  True - success   False - failed
    """
[...]

PR #1831, opened 2021-09-18
[generic-config-updater] Cropping current/target config before sorting

ghooo:
#### What I did
Fixing issue #1830

#### How I did it
Problem we consume the function `loadData` from sonic-yang-mgmt pkg that always crops tables without YANG models from configdb json object. It does it as a side-effect and is not an expected outcome of the function.

The fix here is to crop the current/target tables before doing any sorting, this way gurantee we avoid this bug. I think the better soln is to fix sonic-yang-mgmt pkg. Will look into this more next week.

#### How to verify it
Run the command in the issue, it will result in the expected outcome

**TODO: add unit-tests**

### DIFF
--- a/generic_config_updater/patch_sorter.py
+++ b/generic_config_updater/patch_sorter.py
@@ -997,7 +997,10 @@ class PatchSorter:
         current_config = self.config_wrapper.get_config_db_as_json()
         target_config = self.patch_wrapper.simulate_patch(patch, current_config)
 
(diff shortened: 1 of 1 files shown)


qiluo-msft (approved):

qiluo-msft:
You mentioned "TODO: add unit-tests". Will it happen in this PR?

PR #1836, opened 2021-09-22
Use devicelist to check instead of machine.conf

wen587:
#### What I did
Check by devices list instead of machine.conf
#### How I did it
Add a devices list for the same ASIC while build the image
#### How to verify it
Install a bin file which differs the running platform's ASIC. For example;
install sonic-broadcom.bin to mellanox ASIC platform
install sonic-vs.bin to broadcom ASIC platform

### DIFF
--- a/sonic_installer/bootloader/grub.py
+++ b/sonic_installer/bootloader/grub.py
@@ -18,7 +18,7 @@ from ..common import (
 from .onie import OnieInstallerBootloader
 from .onie import default_sigpipe
 
-MACHINE_CONF = "installer/machine.conf"
+DEVICES_DIR = "installer/devices"
 
 class GrubBootloader(OnieInstallerBootloader):
 
@@ -85,35 +85,29 @@ class GrubBootloader(OnieInstallerBootloader):
         run_command('grub-set-default --boot-directory=' + HOST_PATH + ' 0')
         click.echo('Image removed')
 
+    def platform_in_devices_list(self, platform, image_path):
+        """
+        For those images that don't have devices list builtin, check_output will raise an exception.
+        In this case, we simply return True to make it worked compatible as before.
+        Otherwise, we need to check if platform is inside the supported target devices list.
+        """
+        try:
+            with open(os.devnull, 'w') as fnull:
+                p = subprocess.Popen(["sed", "-e", "1,/^exit_marker$/d", image_path], stdout=subprocess.PIPE, preexec_fn=default_sigpipe)
+                output = subprocess.check_output(["tar", "tf", "-", DEVICES_DIR], stdin=p.stdout, stderr=fnull, preexec_fn=default_sigpipe, text=True)
+                return platform in output
+        except subprocess.CalledProcessError:
+            return True
+
     def verify_image_platform(self, image_path):
         if not os.path.isfile(image_path):
             return False
 
-        # Get running platform's ASIC
-        try:
-            version_info = device_info.get_sonic_version_info()
-            if version_info:
-                asic_type = version_info['asic_type']
-            else:
-                asic_type = None
-        except (KeyError, TypeError) as e:
-            click.echo("Caught an exception: " + str(e))
+        # Get running platform
+        platform = device_info.get_platform()
 
-        # Get installing image's ASIC
-        p1 = subprocess.Popen(["sed", "-e", "1,/^exit_marker$/d", image_path], stdout=subprocess.PIPE, preexec_fn=default_sigpipe)
-        p2 = subprocess.Popen(["tar", "xf", "-", MACHINE_CONF, "-O"], stdin=p1.stdout, stdout=subprocess.PIPE, preexec_fn=default_sigpipe)
-        p3 = subprocess.Popen(["sed", "-n", r"s/^machine=\(.*\)/\1/p"], stdin=p2.stdout, stdout=subprocess.PIPE, preexec_fn=default_sigpipe, text=True)
-
-        stdout = p3.communicate()[0]
-        image_asic = stdout.rstrip('\n')
-
-        # Return false if machine is not found or unexpected issue occur
-        if not image_asic:
-            return False
-
-        if asic_type == image_asic:
-            return True
-        return False
+        # Check if platform is inside image's target platforms
+        return self.platform_in_devices_list(platform, image_path)
 
     @classmethod
     def detect(cls):
--- a/sonic_installer/main.py
+++ b/sonic_installer/main.py
@@ -533,14 +533,14 @@ def install(url, force, skip_platform_check=False, skip_migration=False, skip_pa
             raise click.Abort()
     else:
         # Verify not installing non-secure image in a secure running image
-        if not bootloader.verify_secureboot_image(image_path) and not force:
+        if not force and not bootloader.verify_secureboot_image(image_path):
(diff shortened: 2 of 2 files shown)


wen587:
To make this PR functional, it requires https://github.com/Azure/sonic-buildimage/pull/8814 to work.
But it can also be merged alone. Because it works backward compatible.

sonic_installer/main.py
```diff
@@ -533,14 +533,14 @@ def install(url, force, skip_platform_check=False, skip_migration=False, skip_pa
...
     else:
         # Verify not installing non-secure image in a secure running image
-        if not bootloader.verify_secureboot_image(image_path) and not force:
+        if not force and not bootloader.verify_secureboot_image(image_path):
             echo_and_log("Image file '{}' is of a different type than running image.\n".format(url) +
                 "If you are sure you want to install this image, use -f|--force|--skip-secure-check.\n" +
                 "Aborting...", LOG_ERR)
             raise click.Abort()
 
         # Verify that the binary image is of the same platform type as running platform
```
qiluo-msft, line 542: >platform [](http://example.com/codeflow?start=54&length=8)

Platform has multiple meanings in SONiC context. Suggest platform -> asic
running platform -> running platform's asic type

sonic_installer/bootloader/grub.py
```diff
@@ -85,35 +85,31 @@ def remove_image(self, image):
...
-            version_info = device_info.get_sonic_version_info()
-            if version_info:
-                asic_type = version_info['asic_type']
-            else:
-                asic_type = None
-        except (KeyError, TypeError) as e:
-            click.echo("Caught an exception: " + str(e))
+            with open(os.devnull, 'w') as fnull:
+                p = subprocess.Popen(["sed", "-e", "1,/^exit_marker$/d", image_path], stdout=subprocess.PIPE, preexec_fn=default_sigpipe)
+                output = subprocess.check_output(["tar", "xf", "-", PLATFORMS_ASIC, -O], stdin=p.stdout, stderr=fnull, preexec_fn=default_sigpipe, text=True)
```
qiluo-msft, line 97: >-O [](http://example.com/codeflow?start=84&length=2)

Did you test this line?

qiluo-msft, line 97: >stderr=fnull [](http://example.com/codeflow?start=105&length=12)

Is it the same as the default value `None`?

sonic_installer/bootloader/grub.py
```diff
@@ -85,35 +85,31 @@ def remove_image(self, image):
...
+                return platform in output
```
qiluo-msft, line 98: >output [](http://example.com/codeflow?start=35&length=6)

`output` could be a long list. You can reduce the memory consuption by continuing pipe and grep, or use line-by-line file manipulation.

sonic_installer/bootloader/aboot.py
```diff
@@ -164,7 +165,22 @@ def get_binary_image_version(self, image_path):
...
+        # Get running platform
+        platform = device_info.get_platform()
+
+        # If .platforms_asic is not existed, we simply return True for backward
+        # compatibility. Otherwise, we check if current platform is inside the
+        # supported target platforms list.
+        try:
+            output = subprocess.check_output(['/usr/bin/unzip', '-qop', image_path, '.platforms_asic'], text=True)
+            return platform in output
+        except subprocess.CalledProcessError:
```
qiluo-msft, line 180: >CalledProcessError [](http://example.com/codeflow?start=26&length=18)

This section is good consideration for downgrade use case.

However, `CalledProcessError` is larger than the the case ".platforms_asic is not existed". Could you check it in straighforward way?

sonic_installer/bootloader/aboot.py
```diff
@@ -164,7 +165,22 @@ def get_binary_image_version(self, image_path):
...
+            pass
```
qiluo-msft, line 181: >pass [](http://example.com/codeflow?start=12&length=4)

Your comment said "simply return True"

sonic_installer/bootloader/grub.py
```diff
@@ -85,35 +85,31 @@ def remove_image(self, image):
...
+        except subprocess.CalledProcessError:
+            pass
```
qiluo-msft, line 100: >pass [](http://example.com/codeflow?start=12&length=4)

simply return True

qiluo-msft:
Regarding the PR description, you can move "Check by devices list instead of machine.conf" into the section "Check by devices list instead of machine.conf".

Please give more explanation on the question "What I did". This PR actually changes behavior.

sonic_installer/bootloader/aboot.py
wen587: I am refering here. Seems the image_path is guaranteed to exist. So it only raise exeption if platform_asic is not existed. Maybe it is enough to just for checking existence?
[image_version](https://github.com/Azure/sonic-utilities/blob/master/sonic_installer/bootloader/aboot.py#L159)

sonic_installer/bootloader/aboot.py
wen587: will change

sonic_installer/bootloader/grub.py
wen587: Yes. It works as expected.

wen587: No. I tried locally. `None` still prompt error msg.

sonic_installer/bootloader/grub.py
wen587: will change

sonic_installer/main.py
wen587: will change

sonic_installer/bootloader/grub.py
wen587: I was thinking using pipe and grep. But seems Popen dont have better ways to handle the case of unexisted file. 
Because 'platform_asic nonexistence' and 'platform unmatch' will both produce the empty string if using pipe and grep.
I am thinking of  Popen(['sed'...) first then check_call(['tar'...) first to see if there is any CalledProcessError to confirm platforms_asic exist. Then do the above sed & tar then pipe and grep.
But it will cost more time as you can see. Do you prefer the pipe and grep over check_output?

sonic_installer/bootloader/grub.py
wen587: My bad. Will change later.

qiluo-msft: Then the `with` scope is too small, you need to cover `p2.wait()`, to be safer, cover until function exit

sonic_installer/common.py
```diff
@@ -41,3 +42,9 @@ def run_command_or_raise(argv, raise_exception=True):
         raise SonicRuntimeException("Failed to run command '{0}'".format(argv))
 
     return out.rstrip("\n")
+
+# Needed to prevent "broken pipe" error messages when piping
+# output of multiple commands using subprocess.Popen()
+def default_sigpipe():
```
qiluo-msft, line 48: >default_sigpipe [](http://example.com/codeflow?start=4&length=15)

what is the impact of ""broken pipe" error messages when piping" ?

qiluo-msft (approved):

qiluo-msft (approved):

PR #1842, opened 2021-09-27
[db_migrator] Fix the target db version of the port-channel key test

stephenxs:
#### What I did
The port-channel key migrator was introduced in version `2_0_2` so the expected database version of the test case should be `2_0_2`. 
It was modified to `2_0_3` when the new version was introduced by mistake. This won't fail the test but disable the require its database version to be updated every time a new version is introduced. (Refer #1566 and #1614 for details)
This is to correct it by changing it back to `2_0_2`.


keboliu (approved):

liat-grozovik:
@yxieca could you please help to review or suggest assignee?

yxieca (approved):

PR #1843, opened 2021-09-27
[techsupport] [202106] Removed -i option for docker commands and Improved Error Reporting

vivekreddynv:
Signed-off-by: Vivek Reddy Karri <vkarri@nvidia.com>

#### What I did

PR https://github.com/Azure/sonic-utilities/pull/1723 cannot be cherry-picked directly to 202106. Thus raised a separate PR

PR #1833, opened 2021-09-20
[techsupport] [202012] Removed -i option for docker commands and Improved Error Reporting

vivekreddynv:
Signed-off-by: Vivek Reddy Karri <vkarri@nvidia.com>

#### What I did

PR https://github.com/Azure/sonic-utilities/pull/1723 cannot be cherry-picked directly to 202012. Thus raised a separate PR

PR #1844, opened 2021-09-28
[show techsupport] fix bash errors in generate_dump script

yxieca:
#### What I did
Fix: https://github.com/Azure/sonic-buildimage/issues/8850

Issue was introduced by #1723, #1833, and #1843 (pending merge)

The error_handler is a great idea but the bash script needs to be error free first.

#### How I did it
Fix bash script errors.

#### How to verify it
run show techsupport test..

### DIFF
--- a/scripts/generate_dump
+++ b/scripts/generate_dump
@@ -195,8 +195,9 @@ save_cmd() {
         if $NOOP; then
             echo "${timeout_cmd} $cmd $redirect '$filepath'"
         else
-            eval "${timeout_cmd} $cmd" "$redirect" "$filepath"
-            if [ $? -ne 0 ]; then
+            RC=0
+            eval "${timeout_cmd} $cmd" "$redirect" "$filepath" || RC=$?
+            if [ $RC -ne 0 ]; then
                 echo "Command: $cmd timedout after ${TIMEOUT_MIN} minutes."
             fi
         fi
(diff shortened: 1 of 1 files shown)


scripts/generate_dump
```diff
@@ -195,8 +195,9 @@ save_cmd() {
         if $NOOP; then
             echo "${timeout_cmd} $cmd $redirect '$filepath'"
         else
-            eval "${timeout_cmd} $cmd" "$redirect" "$filepath"
-            if [ $? -ne 0 ]; then
```
smaheshm, line 199: This code looks good to me. Wondering where does it fail.

smaheshm: oh, I see, it can fail when there's timeout.

smaheshm (approved):
:shipit:

scripts/generate_dump
yxieca: Didn't look deep into it. One possibility is that command timed out?

yxieca: If you were asking what the issue with this old structure. if line 198 returns non-zero code, then it would trigger the error handling and bypass the handling at 199.

smaheshm: got it!

qiluo-msft:
>             eval "${timeout_cmd} bash -c \"${cmds}\""

Does this line have the same issue?

---
Refers to: scripts/generate_dump:189 in 1bb4365. [](commit_id = 1bb4365cd5daf8deaf5221c9eb145963a539c0a1, deletion_comment = False)

yxieca:
> > ```
> >         eval "${timeout_cmd} bash -c \"${cmds}\""
> > ```
> 
> Does this line have the same issue?
> 
> Refers to: scripts/generate_dump:189 in [1bb4365](https://github.com/Azure/sonic-utilities/commit/1bb4365cd5daf8deaf5221c9eb145963a539c0a1). [](commit_id = [1bb4365](https://github.com/Azure/sonic-utilities/commit/1bb4365cd5daf8deaf5221c9eb145963a539c0a1), deletion_comment = False)

This one didn't cause trouble in my test.

yxieca:
@qiluo-msft  you are right, there are more issues in the script that didn't get caught by test but code structure also needs to be updated. This PR by itself fixes the test issue. I'll add another PR to address more code structural issues.

PR #1848, opened 2021-09-29
Disable pfcwd forward action config for cisco-8000.

alpeshspatel:
1. Disable pfcwd forward action config for cisco-8000.
2. Do not configure pfcwd if pfc is not configured on a port

This PR is in association with: https://github.com/Azure/sonic-swss/pull/1748 where PFC-WD forward action is disabled for cisco-8000 platform. 

Signed-off-by: Alpesh S Patel <alpesh@cisco.com>

#### What I did

There are two changes of interest:
1. disables pfcwd forward action configuration from pfcwd command
2. in the pfcwd script, adds a check to enable pfcwd configuration on a port only if it has pfc enabled. If config_db.json has an empty entry like: "pfc_enable" : "",
and subsequently pfcwd is configured, swss/orchagent throws an error logs

#### How to verify it
1. can be verified on cisco-8000 platform. It is a no-op for other platforms

```
# pfcwd start -a forward 400 -r 400
SKIPPED: PFC WD 'forward' action not supported on asic 'cisco-8000'
```

2. for a single port, in /etc/sonic/config_db.json, set no pfc enable bits by setting: 

```
"pfc_enable" :  ""
```
for atleast 1 port.

without this change, see the logs in syslog. With this change, there will just be a log on console.

### DIFF
--- a/pfcwd/main.py
+++ b/pfcwd/main.py
@@ -11,7 +11,7 @@ from sonic_py_common.multi_asic import get_external_ports
 from tabulate import tabulate
 from utilities_common import multi_asic as multi_asic_util
 from utilities_common import constants
-from sonic_py_common import logger
+from sonic_py_common import device_info, logger
 
 SYSLOG_IDENTIFIER = "config"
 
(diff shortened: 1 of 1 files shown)


neethajohn (approved):

alpeshspatel:
@neethajohn Can you please double commit/cherry-pick to the master branch as well.

alpeshspatel:
@neethajohn Once you merge this in master, what is the process/cadence of putting that in azure sonic-buildimage/master.
Can you do the needful there or let me know the next step.

PR #1854, opened 2021-10-03
[202012] [techsupport] Techsupport Error Reporting pending fixes

vivekreddynv:
#### What I did

This PR include some fixes which were missed while manually porting the error reporting PR onto 202012 https://github.com/Azure/sonic-utilities/pull/1833. 

i.e. removing -it option from the docker exec commands. to understand why the -it option was removed, refer https://github.com/Azure/sonic-utilities/pull/1723 

This also include another fix which removes -d from the show ip interface command, which fails otherwise

#### How to verify it

- Run the show tech-support and check the return status. It should be zero. (Atleast, it was on mellanox platform. I couldn't check the functions which were platform specific)
- Run the "show techsupport" test. one case will still fail complaining about this. This is expected as -it option was removed. and the test has to be updated. 

`cmd_not_found = defaultdict(<type 'list'>, {'docker_cmds': ['docker exec -it syncd saidump', 'docker exec -it lldp lldpcli show statistics']})`

### DIFF
--- a/scripts/generate_dump
+++ b/scripts/generate_dump
@@ -86,8 +86,9 @@ save_bcmcmd() {
             if [ $ret -eq 124 ]; then
                 echo "Command: $cmd timedout after ${TIMEOUT_MIN} minutes."
             else
-                grep "polling socket timeout: Success" ${filepath} &>/dev/null
-                if [ $? -eq 0 ]; then
+                RC=0
+                grep "polling socket timeout: Success" ${filepath} &>/dev/null || RC=$?
+                if [ $RC -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


yxieca (changes requested):
Please make matching sonic-mgmt test_techsupport_commands() change.

vivekreddynv:
> Please make matching sonic-mgmt test_techsupport_commands() change.
> 
> Again, have you tested your change with the test mentioned above?

Raised the changes to sonic-mgmt https://github.com/Azure/sonic-mgmt/pull/4504

yxieca (approved):

liat-grozovik:
@qiluo-msft could you please help to merge?

PR #1856, opened 2021-10-04
Generic_upater: Apply JSON change

renukamanavalan:
#### What I did
Apply JSON change

#### How I did it
Get running config, apply json change and set the updates onto running redis.

### DIFF
--- a/generic_config_updater/change_applier.py
+++ b/generic_config_updater/change_applier.py
@@ -6,17 +6,6 @@ from swsscommon.swsscommon import ConfigDBConnector
 from .gu_common import JsonChange
 
 
-def get_running_config():
-    (_, fname) = tempfile.mkstemp(suffix="_changeApplier")
-    os.system("sonic-cfggen -d --print-data > {}".format(fname))
-    run_data = {}
-    with open(fname, "r") as s:
-        run_data = json.load(s)
-    if os.path.isfile(fname):
-        os.remove(fname)
-    return run_data
-
-
 def get_config_db():
     config_db = ConfigDBConnector()
     config_db.connect()
@@ -41,11 +30,23 @@ class ChangeApplier:
 
 
     def apply(self, change):
-        run_data = get_running_config()
+        run_data = self._get_running_config()
         upd_data = change.apply(copy.deepcopy(run_data))
 
         for tbl in set(run_data.keys()).union(set(upd_data.keys())):
             self._upd_data(tbl, run_data.get(tbl, {}), upd_data.get(tbl, {}))
+
+
+    def _get_running_config(self):
+        (_, fname) = tempfile.mkstemp(suffix="_changeApplier")
+        os.system("sonic-cfggen -d --print-data > {}".format(fname))
+        run_data = {}
+        with open(fname, "r") as s:
+            run_data = json.load(s)
+        if os.path.isfile(fname):
+            os.remove(fname)
+        return run_data
+
 
             
 
--- a/generic_config_updater/change_applier.py
+++ b/generic_config_updater/change_applier.py
@@ -1,0 +1,52 @@
+import copy
+import json
+import os
+import tempfile
+from swsscommon.swsscommon import ConfigDBConnector
+from .gu_common import JsonChange
+
+
+def get_running_config():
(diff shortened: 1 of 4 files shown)


generic_config_updater/change_applier.py
```diff
@@ -0,0 +1,103 @@
...
+def set_config(config_db, tbl, key, data):
+    config_db.set_entry(tbl, key, data)
+
+
+UPDATER_CONF_FILE = "/etc/sonic/generic_config_updater.conf"
+updater_data = None
+
+class ChangeApplier:
+    def __init__(self):
+        global updater_data, log_level
```
ghooo, line 27: why do we need `log_level`?

ghooo: maybe revert logging logic since it is added in another PR: #1864

generic_config_updater/change_applier.py
```diff
@@ -0,0 +1,103 @@
...
+    def apply(self, change):
+        run_data = self._get_running_config()
+        upd_data = change.apply(copy.deepcopy(run_data))
+        upd_keys = defaultdict(dict)
+
+        for tbl in set(run_data.keys()).union(set(upd_data.keys())):
+            self._upd_data(tbl, run_data.get(tbl, {}),
+                    upd_data.get(tbl, {}), upd_keys)
+
+        ret = _services_validate(run_data, upd_data, upd_keys)
```
ghooo, line 81: In the [design doc](https://github.com/Azure/SONiC/blob/master/doc/config-generic-update-rollback/Json_Change_Application_Design.md#31-overview) we loop over the tables and update them one-by-one, after updating config db for each table we run the validation checks.
```
for table in tables:
  if table has diff:
    update_config_db(table)
    restart_and_validate_services(table)
```

Here we update all the configdb tables first, then run all the validations together. 
```
upd_tables_keys
for table in tables:
  if table has diff:
    update_config_db(table, keys)
    upd_tables_keys.add(keys)

cmds = get_validation_commands(upd_tables_keys)
for cmd in cmds:
  run(cmd, old_config, upd_config, keys)
```

I think both look OK to me, I think the second approach reduces the number of calls to validate/restart services. Is there other advantages? i think we might need to update design doc

ghooo: Also can the tables be processed in `alphabetical` order? makes it easier to read the logs i think where we have expectations of the order of the tables.

qiluo-msft: The first approach may be safer since we focus one table each time. We don't need to worry about the combination of modifying the different table at the same time, which may need througly test.

generic_config_updater/change_applier.py
```diff
@@ -0,0 +1,103 @@
...
+        with open(fname, "r") as s:
+            run_data = json.load(s)
+        if os.path.isfile(fname):
+            os.remove(fname)
+        return run_data
+
+
+            
+
+
```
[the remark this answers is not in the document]
renukamanavalan: sure

generic_config_updater/change_applier.py
```diff
@@ -0,0 +1,103 @@
...
+UPDATER_CONF_FILE = "/etc/sonic/generic_config_updater.conf"
+updater_data = None
+
+class ChangeApplier:
+    def __init__(self):
+        global updater_data, log_level
+
+        self.config_db = get_config_db()
+        if updater_data == None:
```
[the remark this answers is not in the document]
renukamanavalan: I was handling the possibility of reading empty dict from .conf. But that is not going to be the case. So make lgtm happy.

generic_config_updater/change_applier.py
```diff
@@ -0,0 +1,141 @@
...
+import importlib
+import os
+import tempfile
+from collections import defaultdict
+from swsscommon.swsscommon import ConfigDBConnector
+from .gu_common import genericUpdaterLogging
+
+
+UPDATER_CONF_FILE = "/etc/sonic/generic_config_updater.conf"
+updater_data = None
```
qiluo-msft, line 13: global vars are only used in one class, better to move as class member.

Add some comment, especiall what is `updater_data`?

PR #1861, opened 2021-10-06
[TH3] Skipp Control Plane Assist on WARM Reboot for TH3 HWSKUs

gechiang:
#### What I did
Added Code to Skip over Control Plane Assist (CPA) executed during Warm reboot operation for those HW that is not capable of handling CPA such as TH3 based HW SKUs.

#### How to verify it
Execute testcase "arp/test_wr_arp.py" to ensure the operation to program VxLAN Tunnel and Mirror ACLs are all skipped.
This is skipped during setup as well as finalizer time.  
Syslog captured that shows the skipping is made when HW SKU of TH3 is detected:

```
```

Also added new test cases to specifically test out the new helper function I added.

### DIFF
--- a/scripts/neighbor_advertiser
+++ b/scripts/neighbor_advertiser
@@ -561,6 +561,20 @@ def reset_vxlan_tunnel():
 
 
 #
+# Helper function to determine of HWSKU is capable of handling CPA
+#
+CPA_skip_hwsku = ['DellEMC-Z9332f-M-O16C64', 'DellEMC-Z9332f-M-O16C64-lab']
+
+def check_hwsku_not_capable_handle_CPA():
+    hwsku = get_switch_hwsku()
+    if hwsku in CPA_skip_hwsku:
+        log.log_warning('HWSKU {} not able to handle CPA programming. Skipping it gracefully.'.format(hwsku))
+        return True
+    else:
+        return False
+
+
+#
 # Main function
 #
 
@@ -579,6 +593,11 @@ def main():
 
     connect_config_db()
     connect_app_db()
+
+    # Some HWSKU not able to support CPA and should be skipped gracefully
+    if check_hwsku_not_capable_handle_CPA():
+        sys.exit(0)
+
     if operation_mode == 'set':
         set_success = False
 
--- a/tests/neighbor_advertiser_test.py
+++ b/tests/neighbor_advertiser_test.py
@@ -66,3 +66,15 @@ class TestNeighborAdvertiser(object):
         for key in expected_mapping.keys():
             assert(key in tunnel_mapping.keys())
             assert(expected_mapping[key] == tunnel_mapping[key])
+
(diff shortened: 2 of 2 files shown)


prsunny (commented):
Can the neighbor_advertiser be skipped while invoking warmboot instead of modifying the script?

gechiang:
> Can the neighbor_advertiser be skipped while invoking warmboot instead of modifying the script?

Reason I am doing it inside the script is because the setup for warmboot and finalizer are two seperate files and I think it is easier to  have this inside the script which is called by both cases instead of spread out to multiple files.

gechiang:
> > > Can the neighbor_advertiser be skipped while invoking warmboot instead of modifying the script?
> > 
> > 
> > Reason I am doing it inside the script is because the setup for warmboot and finalizer are two seperate files and I think it is easier to have this inside the script which is called by both cases instead of spread out to multiple files.
> 
> If you fix this in setup for warmboot, isn't finalizer a no-op? As in, even if it try to remove the entries in finalizer, it will not trigger anything since its non-existent, right?

Interesting idea.  Let me give it a try just on the setup for warmboot to see if that works.  If so, I will create a new PULL request.
Thanks!

gechiang:
@prsunny I have modified the code per your comment.  Please help review the new changes.
Thanks!

scripts/fast-reboot
```diff
@@ -277,8 +277,13 @@ function backup_database()
 function setup_control_plane_assistant()
 {
     if [[ -n "${ASSISTANT_IP_LIST}" && -x ${ASSISTANT_SCRIPT} ]]; then
-        debug "Setting up control plane assistant: ${ASSISTANT_IP_LIST} ..."
-        ${ASSISTANT_SCRIPT} -s ${ASSISTANT_IP_LIST} -m set
+        # TH3 HW is not capable of VxLAN programming thus skipping TH3 platforms
+        if [[ "$HWSKU" != "DellEMC-Z9332f-M-O16C64" && "$HWSKU" != "DellEMC-Z9332f-M-O16C64-lab" ]]; then
```
vaibhavhd, line 281: `$HWSKU` can be replaced by `${HWSKU}` for consistency and probably bash safety too.

gechiang:
> The failures seem to be coming due to a sonic-swss dvs switch creation error. Yesterday this was supposedly fixed on sonic-swss - [Azure/sonic-swss#1949](https://github.com/Azure/sonic-swss/pull/1949)
> 
> Buildimage update is in progress - [Azure/sonic-buildimage#8915](https://github.com/Azure/sonic-buildimage/pull/8915)
> 
> I suggest to retry after buildimage PR is merged.

Thanks!  I was about to ask how to address the tests failure which are unrelated to my changes...
Will wait for that swss PR pulled into sonic-buildimage before attempting restart azp run.

PR #1868, opened 2021-10-11
Add trap flow counter support

Junchao-Mellanox:
#### What I did

Add trap flow counter CLI support. See HLD: https://github.com/Azure/SONiC/pull/858

#### How I did it

Add new command:

- counterpoll flowcnt-trap enable/disable
- counterpoll flowcnt-trap interval
- show flowcnt-trap stats
- sonic-clear flowcnt-trap
- Extend command: counterpoll show

#### How to verify it

Unit test/Manual test

### DIFF
--- a/scripts/flow_counters_stat
+++ b/scripts/flow_counters_stat
@@ -1,0 +1,256 @@
+#!/usr/bin/env python3
+
+import argparse
+import os
+import _pickle as pickle
+import sys
+
+from natsort import natsorted
+from tabulate import tabulate
+
+# mock the redis for unit test purposes #
+try:
+    if os.environ["UTILITIES_UNIT_TESTING"] == "2":
+        modules_path = os.path.join(os.path.dirname(__file__), "..")
+        tests_path = os.path.join(modules_path, "tests")
+        sys.path.insert(0, modules_path)
+        sys.path.insert(0, tests_path)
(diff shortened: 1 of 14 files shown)


Junchao-Mellanox:
Hi @qiluo-msft , could you please suggest a reviewer from MSFT side? Thanks!

prsunny:
@dgsudharsan, can you review this as well?

scripts/flow_counters_stat
```diff
@@ -0,0 +1,256 @@
...
+            if ns not in old_data:
+                continue
+            old_stats = old_data[ns]
+            for name, values in stats.items():
+                if name not in old_stats:
+                    continue
+
+                old_values = old_stats[name]
+                if values[-1] != old_values[-1]:
+                    # Counter OID not equal means the trap was removed and added again. Removing a trap would cause
```
dgsudharsan, line 222: This is a good logic to handle. However, what if OID returned is the same as before delete? Should we rather not check if the diff is negative and consider it as a case where trap got deleted?

dgsudharsan (changes requested):

dgsudharsan (approved):

keboliu (approved):

prsunny (approved):

PR #1886, opened 2021-10-19
[generate_dump] remove secrets from dump files

liuh-80:
Remove secrets from dump files.

#### What I did
    Add bash functions to remove secrets from dump files.

#### How I did it
    For tacacs key, radius key, snmp community srring, use sed command with regex to remove user secrets from dump files.
    For certs, update tar command exclude list to remove those certs from dump file.

#### How to verify it
    Run 'show techsupport' command and check secrets removed from dump files.


lguohan:
need to back port this to 202012, 202006, 201911

qiluo-msft:

PR #1822, opened 2021-09-15
[Reclaim buffer] [Mellanox] Db migrator support reclaiming reserved buffer for unused ports

stephenxs:
#### What I did
Db migrator support reclaiming reserved buffer for unused ports

Signed-off-by: Stephen Sun stephens@nvidia.com
#### How I did it

For admin down ports, if the buffer objects configuration aligns with default configuration, set the buffer objects configuration as:

- Dynamic model: all normal buffer objects are configured on admin down ports. Buffer manager will apply zero profiles on admin down ports.
- Static model: zero buffer objects are configured on admin down ports.

#### How to verify it
1. Unit test.
2. Manually test.

PR #1897, opened 2021-10-26
[Reclaiming buffer][202106] Database migrator for reclaiming buffer

stephenxs:
This is to cherry-pick community PR #1822 to 202106.

#### What I did
Db migrator support reclaiming reserved buffer for unused ports

As there is no empty slot for database version between 202106 and master (202111), this migration will be done regardless of whether database version is changed. The DB migrator should be idempotent.

Signed-off-by: Stephen Sun stephens@nvidia.com
#### How I did it

For admin down ports, if the buffer objects configuration aligns with default configuration, set the buffer objects configuration as:

- Dynamic model: all normal buffer objects are configured on admin down ports. Buffer manager will apply zero profiles on admin down ports.
- Static model: zero buffer objects are configured on admin down ports.

#### How to verify it
1. Unit test.
2. Manually test.


scripts/db_migrator.py
```diff
@@ -582,6 +582,16 @@ def common_migration_ops(self):
...
 
+        # To migrate buffer on Mellanox platforms
+        # For legacy branches, this is the only place it can be called because
+        #  - Putting it in version 2_0_1 causes db_migrator not able to run
+        #    when the switch is migrated from 202106-no-reclaiming-buffer whose db version is 2_0_2
+        #    to 202106-reclaiming-buffer, which causes reclaiming buffer not take effect
+        #  - Putting it in version 2_0_2 is not a solution either because
+        #    version 2_0_3 has been occupied by 202111
+        if self.asic_type == "mellanox":
+            self.mellanox_buffer_migrator.mlnx_reclaiming_unused_buffer()
```
neethajohn, line 593: As discussed, please ensure that we have a testcase for migration within the same branch since this logic will get called all the time and we need to ensure there is no corruption

stephenxs: Actually, there is no difference in the logic of migrating `CONFIG_DB`. The difference is in the logic for `APPL_DB`.
For config db migration, the logic is like is: if the data contains zero buffer stuff, the configuration of each buffer object on admin down ports will not be the default, so no action for them. this scenario can be covered by all test cases with non-default port configuration.
For warm reboot, in 202012 and 202106, there is a condition like [this](https://github.com/Azure/sonic-utilities/pull/1898/files#diff-14f9672a21874da22d9e80f19da24ce853d68e7e331f3800da5fd29f150ebb5aR1061) when testing whether a profile can be removed from `APPL_DB`.
```
            keys = self.appDB.keys(self.appDB.APPL_DB, "BUFFER_PROFILE_TABLE:*")
            for buffer_profile_key in keys:
                profile = buffer_profile_key.split(':')[1]
                if profile[:12] == 'pg_lossless_' and profile not in referenced_profiles and profile not in buffer_profile_table.keys():
                    self.appDB.delete(self.appDB.APPL_DB, buffer_profile_key)
```
while in master the if condition is like this `if profile not in referenced_profiles and profile not in buffer_profile_table.keys():`
which means on legacy branches we do need to check whether the profile name starts with `pg_lossless_`, otherwise the zero profiles just pushed to APPL_DB in the previous round will be removed.
We have a test case to verify this case. You can check the test cases starting with `warm-reboot-same-version`.

stephenxs:
Also cherry-picked a missing logic to test whether it's a double pool module or single pool module.

scripts/db_migrator.py
neethajohn: @stephenxs, why is this logic diff in master? Shouldn't the same case apply there as well? We remove only lossless profiles and should not remove the zero profiles that were inserted?

neethajohn (approved):

PR #1901, opened 2021-10-27
generic_updater: Fix service validator related issues

renukamanavalan:
#### What I did

    1) Copy generic_updater_config.conf.json as part of install
    2) Read conf file from install dir
    3) Drop empty keys & tables upon jsonpatch.JsonPatch.apply to be in sync with
       redis update
    4) Prefix service_validator module path with "generic_updater"

#### How to verify it
For updates on tables configured with services, the service code is invoked, as expected.


qiluo-msft (dismissed):

ghooo (approved):

generic_config_updater/change_applier.py
```diff
@@ -8,8 +8,8 @@
 from swsscommon.swsscommon import ConfigDBConnector
 from .gu_common import genericUpdaterLogging
 
-
-UPDATER_CONF_FILE = "/etc/sonic/generic_config_updater.conf"
```
qiluo-msft, line 12: >generic_config_updater.conf [](http://example.com/codeflow?start=32&length=27)

Regarding the filename, `generic_config_updater.conf.json` seems make more sense. It is confusing to see `..updater.conf..`

PR #1906, opened 2021-11-01
[port2alias]: Fix to get right number of return values

SuvarnaMeenakshi:
from get_port_config. Modified to get ports from all
namespaces and added unit test cases.

Signed-off-by: SuvarnaMeenakshi <sumeenak@microsoft.com>

#### What I did
get_port_config was modified to return different set number of arguments in PR: https://github.com/Azure/sonic-buildimage/pull/4222. Changes in this PR is:
1. To address the different set of return values
2.  To get the ports from all namespaces
3. To add unit-test
#### How I did it
Modify port2alias to read ports from all namespaces and also add unit-test
#### How to verify it
Test on single and multi-asic platforms.
unit-test passed.


abdosi (approved):

qiluo-msft:

PR #1907, opened 2021-11-01
[201911][port2alias]: Fix to get right number of return values

SuvarnaMeenakshi:
#### What I did
master branch PR: https://github.com/Azure/sonic-utilities/pull/1906
get_port_config was modified to return different set number of arguments in PR: Azure/sonic-buildimage#4222. Changes in this PR is:

To address the different set of return values
To get the ports from all namespaces
How I did it
Modify port2alias to read ports from all namespaces and also add unit-test

How to verify it

### DIFF
--- a/scripts/port2alias
+++ b/scripts/port2alias
@@ -50,7 +50,7 @@ def translate_line(line, ports):
 
 def main():
     (platform, hwsku) = get_platform_hwsku()
-    (ports, _) = get_port_config(hwsku, platform)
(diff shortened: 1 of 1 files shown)


abdosi:
can we add ut ?

abdosi:
@SuvarnaMeenakshi please add UT as part of another PR. Merging this as of now based on manual validation.

abdosi (commented):
LGTM

PR #1919, opened 2021-11-10
generic_config_updater: Filename changed & VLAN validator added

renukamanavalan:
#### What I did
Change filename per review comments from last PR.
Removed print statement from change_applier
Added vlan validator & test code for the same


generic_config_updater/services_validator.py
```diff
@@ -3,15 +3,41 @@
...
+
+
+def vlan_validator(old_config, upd_config, keys):
+    old_vlan = old_config.get("VLAN", {})
+    upd_vlan = upd_config.get("VLAN", {})
+
+    for key in set(old_vlan.keys()).union(set(upd_vlan.keys())):
+        if (old_vlan.get(key, {}).get("dhcp_servers", []) != 
+                upd_vlan.get(key, {}).get("dhcp_servers", [])):
+            return _service_restart("dhcp_relay")
```
ghooo, line 40: NICE! We just restart dhcp at the first occurrence of mismatch between dhcp_servers in old vs updated.

ghooo (approved):

setup.py
```diff
@@ -58,7 +58,7 @@
         'watchdogutil',
     ],
     package_data={
-        'generic_config_updater': ['generic_updater_config.conf.json'],
+        'generic_config_updater': ['generic_config_updater.conf.json'],
```
qiluo-msft, line 61: >generic_config_updater [](http://example.com/codeflow?start=36&length=22)

Thanks!

generic_config_updater/change_applier.py
```diff
@@ -9,29 +9,27 @@
...
+        logger.set_min_log_priority_debug()
+    else:
+        logger.set_min_log_priority_notice()
 
 
-def log_debug(m):
-    logger.log_debug(m, print_to_console)
-    if print_to_stdout:
-        print(m)
+def _log_debug(m):
```
qiluo-msft, line 27: >_log_debug [](http://example.com/codeflow?start=4&length=10)

Minor issue: the original name without starting underscore looks nicer.

generic_config_updater/services_validator.py
```diff
@@ -3,15 +3,41 @@
...
 def _service_restart(svc_name):
-    os.system(f"systemctl restart {svc_name}")
-    logger.log_notice(f"Restarted {svc_name}")
+    rc = os.system(f"systemctl restart {svc_name}")
+    logger.log(logger.LOG_PRIORITY_NOTICE,
+            f"Restarted {svc_name}", print_to_console)
+    return rc == 0
 
 
 def ryslog_validator(old_config, upd_config, keys):
```
qiluo-msft, line 25: >ryslog_validator [](http://example.com/codeflow?start=4&length=16)

ryslog_validator -> rsyslog_validator ?

generic_config_updater/generic_config_updater.conf.json
```diff
@@ -39,6 +39,9 @@
         },
         "DHCP_SERVER": {
             "services_to_validate": [ "dhcp-relay" ]
+        },
+        "VLAN": {
+            "services_to_validate": [ "vlan-service" ]
```
qiluo-msft, line 44: >services_to_validate [](http://example.com/codeflow?start=13&length=20)

According to the design doc [https://github.com/Azure/SONiC/blob/master/doc/config-generic-update-rollback/Json_Change_Application_Design.md#3212-metadata](https://github.com/Azure/SONiC/blob/master/doc/config-generic-update-rollback/Json_Change_Application_Design.md#3212-metadata). It fits better as `restart-command` than `services_to_validate`

renukamanavalan: This is not blindly restart. 
For example, in case of VLAN table update, we call vlan_validator, which *may* or may not restart dhcp_sevice.
Again a validator, could potentially restart more than one and possibly none too, as in vlan case.

So I prefer this name validator, which is table specific.

qiluo-msft: I see the intelligence of restarting dhcp_service for vlan-service. However, the validator is design as read-only operations for the purpose of validation, including validating services are completely restarted and healthy.

renukamanavalan: I don't get your last comments. But Design doc updated.

qiluo-msft (approved):

Issue #1908, opened 2021-11-02
JSON change: remove op for empty table is invalid


renukamanavalan:
#### Description

```
Patch Applier: The patch was sorted into 2 changes:
Patch Applier: * [{"op": "remove", "path": "/TACPLUS/global"}]
Patch Applier: * [{"op": "remove", "path": "/TACPLUS"}]
Patch Applier: Applying changes in order.
```

Upon applying first change, the redis removes `"TACPLUS|global"`.
Now no more "TACPLUS"
The second change to remove `"/TACPLUS"` throws in JsonChange.apply, which does "jsonpatch.JsonPatch.apply" as there is no TACPLUS key.

Error message:

Error: can't remove a non-existent object 'TACPLUS
## Steps to reproduce the issue
1. Ensure no TACPLUS key in running config  --(load minigraph can simulate)
2. Apply patch 
`[ { "op": "add", "path": "/TACPLUS", "value": { "global": { "auth_type": "login" } } } ]`
3. Apply patch
`[ { "op": "remove", "path": "/TACPLUS" } ]`

This should fail as below.
```
Failed to apply patch
Usage: config apply-patch [OPTIONS] PATCH_FILE_PATH
Try "config apply-patch -h" for help.

Error: can't remove a non-existent object 'TACPLUS'

```
#### Output of `show version`

```
(paste your output here)
```

ghooo:
addressed in #1923

PR #1923, opened 2021-11-11
[generic-config-updater] Handling empty tables while sorting a patch

ghooo:
#### What I did
Fixing issue #1908 

#### How I did it
- When the given patch produces empty-table, reject it since ConfigDb cannot show empty tables. Achieved that by:
  - Adding a validation step before patch-sorting
- The patch-sorter should not produce steps which result in empty-tables because it is not possible in ConfigDb, we should instead delete the whole table. Achieved that by:
  - Adding a new validator to reject moves producing empty tables
  - No need to add a generator for deleting the whole table, current generators take care of that. They do by the following:
    - The move to empty a table is rejected by `NoEmptyTableMoveValidator`
    - The previous rejected move is used to generate moves using `UpperLevelMoveExtender` which produces either
      - Remove move for parent -- This is good, we are removing the table
      - Replace move for parent -- This later on will be extended to a Delete move using `DeleteInsteadOfReplaceMoveExtender`

The generators/validators are documented in the `PatchSorter.py`, check the documentation out.

#### How to verify it
unit-test

```sh
admin@vlab-01:~$ sudo config apply-patch empty-a-table.json-patch -v
Patch Applier: Patch: [{"op": "replace", "path": "/DEVICE_METADATA", "value": {}}]
Patch Applier: Validating patch is not making changes to tables without YANG models.
Patch Applier: Validating target config according to YANG models.
... sonig-yang-mgmt logs
... sonig-yang-mgmt logs
Patch Applier: The patch was sorted into 11 changes:
Patch Applier:   * [{"op": "remove", "path": "/DEVICE_METADATA/localhost/bgp_asn"}]
Patch Applier:   * [{"op": "remove", "path": "/DEVICE_METADATA/localhost/buffer_model"}]
Patch Applier:   * [{"op": "remove", "path": "/DEVICE_METADATA/localhost/default_bgp_status"}]
Patch Applier:   * [{"op": "remove", "path": "/DEVICE_METADATA/localhost/default_pfcwd_status"}]
Patch Applier:   * [{"op": "remove", "path": "/DEVICE_METADATA/localhost/deployment_id"}]
Patch Applier:   * [{"op": "remove", "path": "/DEVICE_METADATA/localhost/docker_routing_config_mode"}]
Patch Applier:   * [{"op": "remove", "path": "/DEVICE_METADATA/localhost/hostname"}]
Patch Applier:   * [{"op": "remove", "path": "/DEVICE_METADATA/localhost/hwsku"}]
Patch Applier:   * [{"op": "remove", "path": "/DEVICE_METADATA/localhost/mac"}]
Patch Applier:   * [{"op": "remove", "path": "/DEVICE_METADATA/localhost/platform"}]
Patch Applier:   * [{"op": "remove", "path": "/DEVICE_METADATA/localhost"}]
Patch Applier: Applying changes in order.
Patch Applier: Verifying patch updates are reflected on ConfigDB.
Failed to apply patch

Error: After applying patch to config, there are still some parts not updated
admin@vlab-01:~$ 
```

The above error occurs because post the update, the whole `DEVICE_METADATA` table is deleted, not just showing as empty i.e. `"DEVICE_METADATA": {}`

```
admin@vlab-01:~$ sudo config apply-patch empty-a-table.json-patch -v
Patch Applier: Patch: [{"op": "replace", "path": "/DEVICE_METADATA", "value": {}}]
Patch Applier: Validating patch is not making changes to tables without YANG models.
Failed to apply patch

Error: Given patch is not valid because it will result in empty tables which is not allowed in ConfigDb. Table: DEVICE_METADATA
admin@vlab-01:~$ 
```

### DIFF
--- a/generic_config_updater/generic_updater.py
+++ b/generic_config_updater/generic_updater.py
@@ -55,8 +55,8 @@ class PatchApplier:
         # Validate target config does not have empty tables since they do not show up in ConfigDb
         self.logger.log_notice("Validating target config does not have empty tables, " \
                                "since they do not show up in ConfigDb.")
-        has_empty_tables, empty_tables = self.config_wrapper.has_empty_tables(target_config)
-        if has_empty_tables:
+        empty_tables = self.config_wrapper.get_empty_tables(target_config)
+        if empty_tables: # if there are empty tables
             empty_tables_txt = ", ".join(empty_tables)
             raise ValueError("Given patch is not valid because it will result in empty tables " \
                              "which is not allowed in ConfigDb. " \
--- a/generic_config_updater/generic_updater.py
+++ b/generic_config_updater/generic_updater.py
@@ -52,7 +52,17 @@ class PatchApplier:
         self.logger.log_notice("Simulating the target full config after applying the patch.")
         target_config = self.patch_wrapper.simulate_patch(patch, old_config)
 
-        # Validate target config
+        # Validate target config does not have empty tables since they do not show up in ConfigDb
+        self.logger.log_notice("Validating target config does not have empty tables, " \
+                               "since they do not show up in ConfigDb.")
+        has_empty_tables, empty_tables = self.config_wrapper.has_empty_tables(target_config)
+        if has_empty_tables:
+            empty_tables_txt = ", ".join(empty_tables)
+            raise ValueError("Given patch is not valid because it will result in empty tables " \
+                             "which is not allowed in ConfigDb. " \
+                            f"Table{'s' if len(empty_tables) != 1 else ''}: {empty_tables_txt}")
+
+        # Validate target config according to YANG models
         self.logger.log_notice("Validating target config according to YANG models.")
         if not(self.config_wrapper.validate_config_db_config(target_config)):
             raise ValueError(f"Given patch is not valid because it will result in an invalid config")
--- a/generic_config_updater/patch_sorter.py
+++ b/generic_config_updater/patch_sorter.py
@@ -599,8 +599,8 @@ class NoEmptyTableMoveValidator:
         return True
 
     def _validate_table(self, table, config):
-        is_empty = table in config and not(config[table])
-        return not(is_empty)
+        # the only invalid case is if table exists and is empty
+        return table not in config or config[table]
 
 class LowLevelMoveGenerator:
     """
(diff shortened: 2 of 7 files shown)


generic_config_updater/generic_updater.py
```diff
@@ -52,7 +52,17 @@ def apply(self, patch):
         self.logger.log_notice("Simulating the target full config after applying the patch.")
         target_config = self.patch_wrapper.simulate_patch(patch, old_config)
 
-        # Validate target config
+        # Validate target config does not have empty tables since they do not show up in ConfigDb
+        self.logger.log_notice("Validating target config does not have empty tables, " \
+                               "since they do not show up in ConfigDb.")
+        has_empty_tables, empty_tables = self.config_wrapper.has_empty_tables(target_config)
```
qiluo-msft, line 58: >empty_tables [](http://example.com/codeflow?start=26&length=12)

Not need to return bool, you can check `len(empty_tables)`.

generic_config_updater/patch_sorter.py
```diff
@@ -573,6 +573,35 @@ def _find_ref_paths(self, paths, config):
...
+
+    def _validate_tables(self, tables, config):
+        for table in tables:
+            if not(self._validate_table(table, config)):
+                return False
+        return True
+
+    def _validate_table(self, table, config):
+        is_empty = table in config and not(config[table])
+        return not(is_empty)
```
qiluo-msft, line 603: >not [](http://example.com/codeflow?start=15&length=3)

too many `not`. How about `table not in config or config[table]`

qiluo-msft (commented):
As comments

renukamanavalan (dismissed):

generic_config_updater/patch_sorter.py
ghooo: It will look like this:
```python
    def _validate_table(self, table, config):
        return table not in config or config[table]
```

This means the table is either in not in config, and table has content. Why does this mean the table is not empty? It will take future developer a while to understand it. I think it is easier to have a flag `is_empty` and revert the flag on return.

generic_config_updater/generic_updater.py
ghooo: updated

generic_config_updater/patch_sorter.py
qiluo-msft: Either way is okay to me.
Could you add function level comment "the only invalid case is if table exist and is empty"?

ghooo: [offline discussion] will update and add a small comment

ghooo: updated

qiluo-msft (approved):

PR #1925, opened 2021-11-15
[flex counter] Flex counter threads consume too much CPU resources. #9202

StormLiangMS:
#### What I did
In SWSS module, the default interval of watermark counters is changed from 10 seconds to 60 seconds. To increase the interval CLI upper limit from 30 seconds to 60 seconds.

### DIFF
--- a/counterpoll/main.py
+++ b/counterpoll/main.py
@@ -22,7 +22,7 @@ def queue():
     """ Queue counter commands """
 
 @queue.command()
-@click.argument('poll_interval', type=click.IntRange(100, 30000))
+@click.argument('poll_interval', type=click.IntRange(100, 60000))
 def interval(poll_interval):
     """ Set queue counter query interval """
     configdb = ConfigDBConnector()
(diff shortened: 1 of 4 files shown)


counterpoll/main.py
```diff
@@ -203,7 +203,7 @@ def watermark():
     """ Watermark counter commands """
 
 @watermark.command()
-@click.argument('poll_interval', type=click.IntRange(1000, 30000))
+@click.argument('poll_interval', type=click.IntRange(1000, 60000))
```
neethajohn, line 206: this is the only interval that needs to be updated. Please remove the rest of the changes. They are not related to watermark counters

counterpoll/main.py
```diff
@@ -203,9 +203,15 @@ def watermark():
...
 def interval(poll_interval):
-    """ Set watermark counter query interval for both queue and PG watermarks """
+    """ 
+    Set watermark counter query interval for both queue and PG watermarks
```
neethajohn, line 209: I would suggest to remove this verbose content regarding the interval change since this will show up in the command usage. We can capture this in the PR description

neethajohn (approved):

PR #1931, opened 2021-11-16
[ci] allowPartiallySucceededBuilds when copying sonic-swss-common artifact

qiluo-msft:
#### What I did
Otherwise, a partial failure like test failure will block the build job.


yxieca (approved):

qiluo-msft:
sonic-swss-common 202012 branch build finally pass, and this workaround is not needed any more. It is actually harmful if sonic-swss-common fails and has no artifacts. Please stop using it.

PR #1933, opened 2021-11-17
Fix invalid output of syslog IPv6 servers

wen587:
#### What I did
Modify the filter function of `show runningconfiguration syslog`
#### How I did it
Filter by ending "]" rather than split by ":"
#### How to verify it
add a syslog v6 server `sudo config syslog add f587::1:1`
run command `show runningconfiguration syslog`
```
admin@vlab-01:~$ show runningconfiguration syslog
Syslog Servers
----------------
[10.0.0.5]
[10.0.0.6]
[f587
```
```
admin@vlab-01:~$ show runningconfiguration syslog
Syslog Servers
----------------
[10.0.0.5]
[10.0.0.6]
[f587::1:1]
```

### DIFF
--- a/show/main.py
+++ b/show/main.py
@@ -1357,16 +1357,19 @@ def show_run_snmp(db, ctx):
 @click.option('--verbose', is_flag=True, help="Enable verbose output")
 def syslog(verbose):
     """Show Syslog running configuration
-    To match below cases:
+    To match below cases(port is optional):
     *.* @IPv4:port
+    *.* @@IPv4:port
     *.* @[IPv4]:port
+    *.* @@[IPv4]:port
     *.* @[IPv6]:port
+    *.* @@[IPv6]:port
     """
     syslog_servers = []
     syslog_dict = {}
-    re_ipv4_1 = re.compile(r'^\*\.\* @(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}):\d+')
-    re_ipv4_2 = re.compile(r'^\*\.\* @\[(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})\]:\d+')
-    re_ipv6 = re.compile(r'^\*\.\* @\[([0-9a-fA-F:]+)\]:\d+')
+    re_ipv4_1 = re.compile(r'^\*\.\* @{1,2}(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})(:\d+)?')
+    re_ipv4_2 = re.compile(r'^\*\.\* @{1,2}\[(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})\](:\d+)?')
+    re_ipv6 = re.compile(r'^\*\.\* @{1,2}\[([0-9a-fA-F:]+)\](:\d+)?')
     with open("/etc/rsyslog.conf") as syslog_file:
         data = syslog_file.readlines()
     for line in data:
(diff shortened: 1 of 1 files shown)


show/main.py
```diff
@@ -1363,8 +1363,10 @@ def syslog(verbose):
         data = syslog_file.readlines()
     for line in data:
         if line.startswith("*.* @"):
-            line = line.split(":")
-            server = line[0][5:]
+            end = line.find("]")
```
qiluo-msft, line 1366: >find [](http://example.com/codeflow?start=23&length=4)

Better to use regex or something more smarter, in order to make parsing flexible to all the valid config:
1. @ipv4:port
2. @[ipv4]:port
3. @[ipv6]:port

show/main.py
```diff
@@ -1356,16 +1356,29 @@ def show_run_snmp(db, ctx):
...
-            server = line[0][5:]
-            syslog_servers.append(server)
+        if re_ipv4_1.match(line):
+            server =  re_ipv4_1.match(line).group(1)
+        elif re_ipv4_2.match(line):
+            server =  re_ipv4_2.match(line).group(1)
+        elif re_ipv6.match(line):
+            server =  re_ipv6.match(line).group(1)
+        else:
+            continue
```
qiluo-msft, line 1380: >continue [](http://example.com/codeflow?start=12&length=8)

Should we report error message on stderr instead of being silent?

show/main.py
```diff
@@ -1356,16 +1356,29 @@ def show_run_snmp(db, ctx):
 @runningconfiguration.command()
 @click.option('--verbose', is_flag=True, help="Enable verbose output")
 def syslog(verbose):
-    """Show Syslog running configuration"""
+    """Show Syslog running configuration
+    To match below cases:
+    *.* @IPv4:port
+    *.* @[IPv4]:port
+    *.* @[IPv6]:port
```
qiluo-msft, line 1363: >port [](http://example.com/codeflow?start=16&length=4)

The port is optional and the default is 514. ref: [https://linux.die.net/man/5/rsyslog.conf](https://linux.die.net/man/5/rsyslog.conf)

show/main.py
```diff
@@ -1356,16 +1356,29 @@ def show_run_snmp(db, ctx):
...
+    """
     syslog_servers = []
     syslog_dict = {}
+    re_ipv4_1 = re.compile(r'^\*\.\* @(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}):\d+')
```
qiluo-msft, line 1367: >^\*\.\* @(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}):\d+ [](http://example.com/codeflow?start=29&length=49)

Thanks for being strict when writing a regex!
However since it is a `show` command, and the data is from a file in filesystem, should we be more relax and do not check for a valid ipv4/v6 address?

we can delegate the check to `rsyslogd -N1`. If anything wrong, we report an empty syslog running config.

wen587: Hi Qi, according to [rsyslog IPv6 support](https://github.com/Azure/sonic-buildimage/pull/3754/files), I think we can just consider the cases of    ` *.* @[IPv4]:port`  and `*.* @[IPv6]:port` which are all included in square bracket.
This way, I think the initial commit in this PR which checks ending ']' is enough.

show/main.py
wen587: I don't think so. I am checking all lines in rsyslogd.conf.
Even if only check lines start with `*.* @`, it is possible we have tcp format server starts with `*.* @@`. It is not enabled in the rsyslogd.conf.
Seems current `show run syslog` command only care about UDP server

show/main.py
wen587: I can make the port optional.

qiluo-msft (approved):

show/main.py
```diff
@@ -1356,16 +1356,32 @@ def show_run_snmp(db, ctx):
 @runningconfiguration.command()
 @click.option('--verbose', is_flag=True, help="Enable verbose output")
 def syslog(verbose):
```
qiluo-msft, line 1358: Could you add some unit tests?

PR #1948, opened 2021-11-24
[Auto-Techsupport] Issues related to Multiple Cores crashing handled

vivekreddynv:
Signed-off-by: Vivek Reddy Karri <vkarri@nvidia.com>

#### What I did

**Issues seen when multiple cores are crashed in very quick succession:**
1) The **rate_limit_interval** is not honored. Because, i previously was finding out the last created tech-support using the glob pattern `sonic_dump_*tar*`, which  will not include the dumps which are being currently run. These existing dump will not have .tar.gz extension. Thus, modified the `get_ts_dumps` to search based on the TS_ROOT i.e `sonic_dump_*`
2) **show auto-tech support history** is not showing all the created dumps. I've previously used to take the diff of tech support dumps before and after running the invocation and used to assign the diff as the corresponding techsupport for this core. This approach is prone to race condition as we can have multiple dumps in the diff found in the interval. 
Avoided this by parsing the stdout returned by `show techsupport` invocation

#### How to verify it

1) Unit Tests
2) Generate core-dumps in very quick succession. Use the default rate limit interval. Should only see one entry in tech-support history
3) Set global rate limit interval to 0. Generate cores in quick succession. Should see a few entries in the history.


vivekreddynv:
@ganglyu @qiluo-msft Please help review

scripts/coredump_gen_handler.py
```diff
@@ -111,26 +111,34 @@ def get_since_arg(self):
             return since_cfg
         return SINCE_DEFAULT
 
+    def parse_ts_dump_name(self, ts_stdout):
+        """ Figure out the ts_dump name from the techsupport stdout """
+        matches = re.findall(TS_PTRN, ts_stdout)
+        if matches:
+            return matches[-1]
+        else:
```
ganglyu, line 119: else is not necessary here.

tests/coredump_gen_handler_test.py
```diff
@@ -315,10 +328,13 @@ def test_invalid_since_argument(self):
         populate_state_db(redis_mock)
         with Patcher() as patcher:
             def mock_cmd(cmd, env):
+                ts_dump = "/var/dump/sonic_dump_random3.tar.gz"
                 cmd_str = " ".join(cmd)
-                if "show techsupport --since '2 days ago'" in cmd_str:
-                    patcher.fs.create_file("/var/dump/sonic_dump_random3.tar.gz")
-                    return 0, "", ""
+                print(cmd_str)
```
ganglyu, line 333: Is print used for debug?

vivekreddynv: Yes, Will remove

ganglyu (approved):
lgtm

PR #1952, opened 2021-11-30
[port] Fix port speed set

mykolaxgerasymenko:
Signed-off-by: Mykola Gerasymenko <mykolax.gerasymenko@intel.com>

#### What I did
Wrong port speed is set if included as a substring in a speed literal.
For example if we have a valid supported speeds for a port with 4 lanes: 40000 and 100000. Then if we try to set 4000.
`sudo config interface speed Ethernet0 4000`
As a result, we will get a wrong set speed 4G:
```
dut:~$ show interfaces status
Interface  Lanes  Speed  MTU  FEC   Alias     Vlan   Oper  Admin   Type      Asym PFC
--------- ------ ------ ----  ---- --------   -----  ----- ------ -----------   -----
Ethernet0  0,1,2,3  4G  9100  rs  Ethernet1/1 routed  up    up   QSFP28 or later     off
```
#### How I did it
Fixed finding the target (set) speed in the supported speeds list.

#### How to verify it
Set the wrong speed, which included as a substring in a speed literal.
For example, for a speed 40000 it is 4000.
If the wrong speed is set, an error message should appear:
```
admin@sonic:~$ sudo config interface speed Ethernet0 4000
Invalid speed specified: 4000
Valid speeds:40000,100000
```
Only speeds from the list of "Valid speeds" have been successfully set.
`sudo config interface speed Ethernet0 40000`

### DIFF
--- a/scripts/portconfig
+++ b/scripts/portconfig
@@ -109,12 +109,12 @@ class portconfig(object):
     def set_speed(self, port, speed):
         if self.verbose:
             print("Setting speed %s on port %s" % (speed, port))
-        supported_speeds_str = self.get_supported_speeds(port)
-        if supported_speeds_str:
-            if supported_speeds_str.find(str(speed)) == -1:
-                print('Invalid speed specified: {}'.format(speed))
-                print('Valid speeds:{}'.format(supported_speeds_str))
-                exit(1)
+        supported_speeds_str = self.get_supported_speeds(port) or ''
+        supported_speeds = [int(s) for s in supported_speeds_str.split(',')]
+        if int(speed) not in supported_speeds:
+            print('Invalid speed specified: {​​​​​​​}​​​​​​​'.format(speed))
+            print('Valid speeds:{​​​​​​​}​​​​​​​'.format(supported_speeds_str))
+            exit(1)
         self.db.mod_entry(PORT_TABLE_NAME, port, {PORT_SPEED_CONFIG_FIELD_NAME: speed})
 
     def set_fec(self, port, fec):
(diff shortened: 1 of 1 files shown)


mykolaxgerasymenko:
@andriymoroz-mlnx @yxieca Please review.

novikauanton (approved):

scripts/portconfig
```diff
@@ -109,12 +109,12 @@ class portconfig(object):
...
         if self.verbose:
             print("Setting speed %s on port %s" % (speed, port))
-        supported_speeds_str = self.get_supported_speeds(port)
-        if supported_speeds_str:
-            if supported_speeds_str.find(str(speed)) == -1:
-                print('Invalid speed specified: {}'.format(speed))
-                print('Valid speeds:{}'.format(supported_speeds_str))
-                exit(1)
+        supported_speeds_str = self.get_supported_speeds(port) or ''
+        supported_speeds = [int(s) for s in supported_speeds_str.split(',')]
```
novikauanton, line 113: in case `supported_speeds_str` is blank string (``),  next line will faile, because blank string can't be converted to int, 
```
>>> [int(s) for s in  ''.split(',')]
  File "<stdin>", line 1, in <listcomp>
ValueError: invalid literal for int() with base 10: ''
```

You need to check that value you are casting to int is not blank, 
```
 [int(s) for s in  ''.split(',') if s]
```

novikauanton (changes requested):

scripts/portconfig
mykolaxgerasymenko: Thanks for the review. Fixed.

novikauanton (approved):

mykolaxgerasymenko:
@andriymoroz-mlnx @yxieca please review.

mykolaxgerasymenko:
@jleveque @Junchao-Mellanox please review.

mykolaxgerasymenko:
@jleveque @Junchao-Mellanox Could you please review?

yxieca (approved):

PR #1956, opened 2021-12-01
[fast-reboot]Let PMON be graceful stopped

PJHsieh:
#### What I did
Let PMON service be graceful stopped.
#### How I did it
Append pmon to the end of SERVICES_TO_STOP to lead to safe exit.
#### How to verify it
The following strings shall be in the output of "sudo fast-reboot -v"

Stopping pmon service ...
Stopped pmon service ...


PJHsieh:
retest

PR #1957, opened 2021-12-01
[ssdutil] Inform if the disk is not SSD

sujinmkang:
https://github.com/Azure/sonic-buildimage/issues/9407

#### What I did
Inform that the disk is not SSD for the platform with non-SSD disk

#### How I did it
Check disk type and exit the ssdutil command 

#### How to verify it
Check `show platform ssdhealth` command on the platform with non SSD disk

### DIFF
--- a/ssdutil/main.py
+++ b/ssdutil/main.py
@@ -97,22 +97,18 @@ def ssdutil():
         sys.exit(1)
 
     parser = argparse.ArgumentParser()
-    parser.add_argument("-d", "--device", help="Device name to show health info", default=None)
+    parser.add_argument("-d", "--device", help="Device name to show health info", default=get_default_disk())
     parser.add_argument("-v", "--verbose", action="store_true", default=False, help="Show verbose output (some additional parameters)")
     parser.add_argument("-e", "--vendor", action="store_true", default=False, help="Show vendor output (extended output if provided by platform vendor)")
     args = parser.parse_args()
 
-    if args.device:
-        disk_device = args.device
-    else:
-        disk_device = get_default_disk()
 
-    disk_type = get_disk_type(disk_device)
-    if disk_type != DISK_TYPE_SSD:
+    disk_type = get_disk_type(args.device)
+    if DISK_TYPE_SSD not in disk_type:
         print("Disk type is not SSD")
-        sys.exit(1)
 
-    ssd = import_ssd_api(disk_device)
+    ssd = import_ssd_api(args.device)
+
 
     print("Device Model : {}".format(ssd.get_model()))
     if args.verbose:
--- a/ssdutil/main.py
+++ b/ssdutil/main.py
@@ -30,7 +30,7 @@ def get_disk_type(diskdev):
     proc = subprocess.Popen(cmd, shell=True, text=True, stdout=subprocess.PIPE)
     outs = proc.stdout.readlines()
     for out in outs:
-        if out.split()[0] in diskdev_name:
+        if out.split()[0] is diskdev_name:
               cmd = "cat /sys/block/{}/queue/rotational".format(diskdev_name)
               proc = subprocess.Popen(cmd, shell=True, text=True, stdout=subprocess.PIPE)
               out = proc.stdout.readline()
--- a/ssdutil/main.py
+++ b/ssdutil/main.py
@@ -18,26 +18,44 @@ except ImportError as e:
 DEFAULT_DEVICE="/dev/sda"
 SYSLOG_IDENTIFIER = "ssdutil"
 DISK_TYPE_SSD = "0"
+DISK_INVALID = "-1"
 
 # Global logger instance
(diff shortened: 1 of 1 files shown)


ssdutil/main.py
```diff
@@ -65,6 +75,11 @@ def ssdutil():
     parser.add_argument("-e", "--vendor", action="store_true", default=False, help="Show vendor output (extended output if provided by platform vendor)")
     args = parser.parse_args()
 
+    disk_type = get_disk_type(args.device)
+    if disk_type != DISK_TYPE_SSD:
+        print("Disk is not SSD")
+        sys.exit(1)
```
neethajohn, line 81: @sujinmkang , can you also ensure that ssd related testcases in sonic-mgmt are skipped for such platforms after this change gets merged?

prgeor (dismissed):

ssdutil/main.py
neethajohn: Azure/sonic-mgmt#4807

neethajohn (dismissed):

ssdutil/main.py
```diff
@@ -16,11 +17,20 @@
 
 DEFAULT_DEVICE="/dev/sda"
 SYSLOG_IDENTIFIER = "ssdutil"
+DISK_TYPE_SSD = "0"
 
 # Global logger instance
 log = logger.Logger(SYSLOG_IDENTIFIER)
 
 
+def get_disk_type(diskdev):
```
prgeor, line 26: better to validate if the listed diskdev is valid as listed by "lsblk" otherwise the utility will crash. you can use 
```
lsblk -l -n | grep 'disk' 
```
to get the list of all disk names available in the host

ssdutil/main.py
```diff
@@ -16,11 +17,29 @@
...
+    """Check disk type"""
+    diskdev_name = diskdev.replace('/dev/','')
+    cmd = "lsblk -l -n |grep disk"
+    proc = subprocess.Popen(cmd, shell=True, text=True, stdout=subprocess.PIPE)
+    outs = proc.stdout.readlines()
+    for out in outs:
+        if out.split()[0] is diskdev_name:
```
prgeor, line 33: what if diskdev_name is one of  ['sda', 'sdb', 'sdc'], but you seem to be checking only the first element

neethajohn:
@prgeor, can you look at the new changes?

prgeor (dismissed):

ssdutil/main.py
```diff
@@ -16,10 +17,46 @@
...
+def get_disk_type(diskdev):
+    """Check disk type"""
+    diskdev_name = diskdev.replace('/dev/','')
+    cmd = "lsblk -l -n |grep {}".format(diskdev_name)
+    proc = subprocess.Popen(cmd, shell=True, text=True, stdout=subprocess.PIPE)
+    out = proc.stdout.readline()
+    if diskdev_name not in out:
+        return DISK_INVAILD
+    cmd = "cat /sys/block/{}/queue/rotational".format(diskdev_name)
+    proc = subprocess.Popen(cmd, shell=True, text=True, stdout=subprocess.PIPE)
```
alexrallen, line 55: Forking to `cat` here seems unnecessary. Is there a reason you chose not to use native python io here?

alexrallen: It is also more pythonic to do this and add error handling in order to handle the `DISK_INVALID` case rather than forking `lsblk` as well.

ssdutil/main.py
```diff
@@ -16,10 +17,46 @@
...
 log = logger.Logger(SYSLOG_IDENTIFIER)
 
+def get_default_disk():
+    """Check default disk"""
+    default_device = DEFAULT_DEVICE
+    host_mnt = '/host'
+    cmd = "lsblk -l -n |grep {}".format(host_mnt)
+    proc = subprocess.Popen(cmd, shell=True, text=True, stdout=subprocess.PIPE)
+    out = proc.stdout.readline()
+    if host_mnt in out:
```
alexrallen, line 33: Is this check necessary? `grep` should take care of this. Perhaps check if the result is not empty (or grep return code)?

ssdutil/main.py
```diff
@@ -16,10 +17,46 @@
...
+DISK_INVALID = "-1"
 
 # Global logger instance
 log = logger.Logger(SYSLOG_IDENTIFIER)
 
+def get_default_disk():
+    """Check default disk"""
+    default_device = DEFAULT_DEVICE
+    host_mnt = '/host'
+    cmd = "lsblk -l -n |grep {}".format(host_mnt)
```
alexrallen, line 30: You may be able to do this cleaner with `df`

```
admin@r-lionfish-16:~$ df --output=source /host
Filesystem
/dev/sda3
```

alexrallen (changes requested):
Hey @sujinmkang while looking this over I had a few comments / questions. Please take a look, thanks.

PR #1958, opened 2021-12-03
Refine doc on show loopback/mgmt interfaces

qiluo-msft:
#### What I did
Add explicit sections for showing management/loopback interfaces

### DIFF
--- a/doc/Command-Reference.md
+++ b/doc/Command-Reference.md
@@ -6,6 +6,7 @@
 * [Introduction](#introduction)
 * [Basic Tasks](#basic-tasks)
   * [SSH Login](#ssh-login)
+  * [Show Management Interface](#show-management-interface)
   * [Configuring Management Interface](#configuring-management-interface)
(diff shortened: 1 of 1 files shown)


doc/Command-Reference.md
```diff
@@ -254,6 +256,10 @@ By default, login takes the user to the default prompt from which all the show c
 
 Go Back To [Beginning of the document](#) or [Beginning of this section](#basic-tasks)
 
+### Show Management Interface
+
+Please check [show ip interfaces](#show-ip-interfaces)
```
shi-su, line 261: Does this mean there is no specific command only to show management and loopback interfaces? When these interfaces need to be shown, just call `show ip interfaces` and find the ones of interest?

qiluo-msft: Yes. Do you have any concern?

shi-su: Nope, just to clarify

shi-su (approved):

PR #1963, opened 2021-12-08
[soft-reboot] Add support for platforms based on Device Tree

dflynn-Nokia:
The soft-reboot command currently expects a platform that uses GRUB.
Information about the current kernel boot arguments is extracted from the
GRUB config file. This commit adds support for platforms that do not use GRUB
but rather uses Device Tree to define the current kernel boot arguments.
Example platform architectures using Device Tree include armhf and arm64.

#### What I did
Add soft-reboot command support for the armhf and arm64 platforms

#### How I did it
Add logic to retrieve kernel boot arguments from the Device Tree rather than from the GRUB config for these platforms

#### How to verify it
Execute the soft-reboot command on an armhf or arm64 platform

#### Previous command output (if the output of a command-line utility has changed)
sudo soft-reboot
Error response from daemon: Container 68b38ffde1c8a95922eed7706a2f92c46e5b6582539cecc4aaa7263803db56e3 is not running
cat: /host/grub/grub.cfg: No such file or directory
requested COLD shutdown
syscall kexec_file_load not available.
Read on /host failed: Is a directory
Cannot read /hostFailed to arm Watchdog for 180 seconds
Nothing has been loaded!

#### New command output (if the output of a command-line utility has changed)
Successful reboot via kexec to Sonic login prompt

PR #1966, opened 2021-12-10
[warm/fast-reboot] Fix kexec portion to support platforms based on De…

dflynn-Nokia:
…vice Tree

The warm-reboot and fast-reboot commands currently expect a platform that uses
GRUB.  Information about the current kernel boot arguments is extracted from
the GRUB config file. This commit adds support for platforms that do not use
GRUB but rather use Device Tree to define the current kernel boot arguments.
Example platform architectures using Device Tree include armhf and arm64.

This commit also includes a minor improvement to a similar change made to the
soft-reboot command in PR# [1963](https://github.com/Azure/sonic-utilities/pull/1963)

#### What I did
Fix kexec portion of warm-reboot and fast-reboot commands for the armhf and arm64 platforms

#### How I did it
Add logic to retrieve kernel boot arguments from the Device Tree rather than from the GRUB config for these platforms

#### How to verify it
Execute the warm-reboot and fast-reboot commands on an armhf or arm64 platform and confirm that kexec reboots the switch  as expected

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -27,6 +27,7 @@ SSD_FW_UPDATE="ssd-fw-upgrade"
 TAG_LATEST=yes
 DETACH=no
 LOG_PATH="/var/log/${REBOOT_TYPE}.txt"
+UIMAGE_HDR_SIZE=64
 
 # Require 100M available on the hard drive for warm reboot temp files,
 # Size is in 1K blocks:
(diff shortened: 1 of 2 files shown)


sujinmkang (approved):

qiluo-msft:

dflynn-Nokia:
I submitted PR# [1980](https://github.com/Azure/sonic-utilities/pull/1980) to deliver the fix the 202012 branch.

PR #1967, opened 2021-12-10
[config reload] Fixing config reload when timer based delayed services are disabled

dgsudharsan:
Signed-off-by: Sudharsan Dhamal Gopalarathnam sudharsand@nvidia.com

#### What I did
When timer based delayed services like mgmt-framework, telemetry and snmp are disabled and config reload is execute it fails Failed to reset failed state of unit mgmt-framework.service: Unit mgmt-framework.service not loaded.
The reason is these services don't get masked like regular services and these are derived from timers. So when reset-failed is tried on these services it leads to exception.

#### How I did it
When the feature related to these services are disabled their timers would be masked and wouldn't be "enabled". So when deriving the services from timers the services which are not enabled will be skipped.

#### How to verify it
Disable services like mgmt-framework, snmp and telemetry and execute config reload. The config reload should execute without failure


tests/config_test.py
```diff
@@ -92,7 +94,7 @@ def test_load_minigraph(self, get_cmd_module, setup_single_broadcom_asic):
             mock_run_command.assert_any_call('systemctl reset-failed swss')
             # Verify "systemctl reset-failed" is called for services under sonic-delayed.target 
             mock_run_command.assert_any_call('systemctl reset-failed snmp')
-            assert mock_run_command.call_count == 10
+            assert mock_run_command.call_count == 11
```
liat-grozovik, line 97: @dgsudharsan what this hard coded number stands for? should we count it in a better way than just a number?

Junchao-Mellanox (approved):

config/main.py
```diff
@@ -733,7 +733,9 @@ def _get_sonic_services():
 
 def _get_delayed_sonic_services():
     out = clicommon.run_command("systemctl list-dependencies --plain sonic-delayed.target | sed '1d'", return_cmd=True)
-    return (unit.strip().rstrip('.timer') for unit in out.splitlines())
+    timers = [unit for unit in out.splitlines() if(clicommon.run_command("systemctl is-enabled {}".format(unit),
```
qiluo-msft, line 736: >systemctl [](http://example.com/codeflow?start=74&length=9)

you can use one command to check all the units. You may use `xargs` to bypass the command line too long issue.

qiluo-msft, line 736: >unit [](http://example.com/codeflow?start=14&length=4)

This statement is too complex. Could you rewrite in traditional control flow, like for-loop or if-block?

config/main.py
```diff
@@ -733,7 +733,12 @@ def _get_sonic_services():
 
 def _get_delayed_sonic_services():
     out = clicommon.run_command("systemctl list-dependencies --plain sonic-delayed.target | sed '1d'", return_cmd=True)
-    return (unit.strip().rstrip('.timer') for unit in out.splitlines())
+    timers = []
+    for unit in out.splitlines():
+        state = clicommon.run_command("systemctl is-enabled {}".format(unit), return_cmd=True)
+        if(state.strip() == "enabled"):
```
qiluo-msft, line 739: >( [](http://example.com/codeflow?start=10&length=1)

Add a blank after `if`, you can remove the outer `()`.

qiluo-msft (approved):

tests/config_test.py
dgsudharsan: I believe the logic is testing for number of calls to clicommon.run_command which is internal implementation. There is no way of knowing it in test script (even in the code). If tomorrow an additional command gets added this needs to be changed again. I agree this is not a deterministic method but its an existing code and I am not sure of the motivation.
For e.g below are the calls internally invoked.
sudo systemctl stop sonic.target --job-mode replace-irreversibly
/usr/local/bin/sonic-cfggen -H -m --write-to-db
config qos reload --no-dynamic-buffer
pfcwd start_default
systemctl list-dependencies --plain sonic.target | sed '1d'
systemctl list-dependencies --plain sonic-delayed.target | sed '1d'
systemctl is-enabled snmp.timer
systemctl reset-failed swss
systemctl reset-failed snmp
sudo systemctl restart sonic.target
sudo monit reload

liat-grozovik:
@dgsudharsan is this also required for 202111? if so, i will add the label

liat-grozovik (approved):

PR #1975, opened 2021-12-17
[GCU] Using test yang models instead of real yang models for unit-tests

ghooo:
#### What I did
While running tests, do not rely on yang models on the box. Instead use the yang models in the `files/test-yang-models` directory.

This changes makes sure whenever the real YANG models change, our unit-test don't break.

Nightly builds should be used to capture issues with the real YANG models.

#### How I did it
Copied yang models that are needed for testing to the test folder

#### How to verify it
unit-tests

### DIFF
--- a/tests/generic_config_updater/generic_updater_test.py
+++ b/tests/generic_config_updater/generic_updater_test.py
@@ -3,7 +3,7 @@ import os
 import shutil
 import unittest
 from unittest.mock import MagicMock, Mock, call
-from .gutest_helpers import create_side_effect_dict, Files
+from .gutest_helpers import create_side_effect_dict, Files, init_tests
 
 import generic_config_updater.generic_updater as gu
 import generic_config_updater.patch_sorter as ps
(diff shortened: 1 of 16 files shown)


qiluo-msft:
> Nightly builds should be used to capture issues with the real YANG models.

Do you have any plan on this? Originally the unit test could expose some issues in sonic-buildimage's/src/sonic-yang-models folder. If we lose the capability, how can we capture them in nightly test?

tests/generic_config_updater/generic_updater_test.py
```diff
@@ -12,6 +12,8 @@
 # sys.path.insert(0,'../../generic_config_updater')
 # import generic_updater as gu
 
+init_tests()
```
qiluo-msft, line 15: >init_tests [](http://example.com/codeflow?start=0&length=10)

Normally we don't write code outside the test class/functions in file scope. Do you want to use a fixture [https://docs.pytest.org/en/latest/how-to/fixtures.html](https://docs.pytest.org/en/latest/how-to/fixtures.html)?

I guess it fits into a session fixture.

ghooo:
@qiluo-msft I am moving the patch-sorting unit-tests to a JSON file in preparation for using them for the nightly test check PR #1977

ghooo:
Actually thinking about this for a bit more, I think it is OK to test using the real YANG models. These unit-tests can help capture in YANG models.

Issue #1976, opened 2021-12-17
[GCU] PatchSorter.sort can generate different correct sorted steps for the same input -- always generate same steps


ghooo:
#### Description
When running the unit-test `REMOVING_CREATE_ONLY_FIELD__SUCCESS` it generates different results with each run.

Note: The flag `skip_exact_change_list_match` is getting adding in this PR: #1977

## Steps to reproduce the issue
1. in ` tests/generic_config_updater/patch_sorter_test.py` set `skip_exact_change_list_match=False`
2. Run the `TestPatchSorter` tests

#### Describe the results you received
unit-test fails, expected and actual changes are not matched

#### Describe the results you expected
unit-test succeed, expected and actual changes match

#### Output of `show version`

```
(paste your output here)
```

PR #1977, opened 2021-12-17
[GCU] Moving PatchSorter unit-test to json file to make it easier to read/maintain

ghooo:
#### What I did
Moved PatchSorter unit-test to json file to make it easier to read/maintain

In the future this test json file can be used for adding nightly tests.

While testing I had to set `skip_exact_change_list_match=True` because different runs of the same input produce correct but different changes. It is better to always generate the same steps for easier debugging. Created issue: #1976

#### How I did it
Copied to Json file the following:
- current_config i.e. current running config
- patch 
- expected change list

#### How to verify it
unit-tests

### DIFF
--- a/tests/generic_config_updater/files/patch_sorter_test_failure.json
+++ b/tests/generic_config_updater/files/patch_sorter_test_failure.json
@@ -1,6 +1,6 @@
 {
     "EMPTYING_A_CONFIGDB_TABLE__FAILURE": {
-        "desc": "Emptying a configdb table fails because empty tables are not allowed in configdb.",
+        "desc": "Emptying a configdb table fails because empty tables are not allowed in configdb. User should remove whole table instead e.g. remove /ACL_TABLE in this case.",
         "current_config": {
             "ACL_TABLE": {
                 "EVERFLOW": {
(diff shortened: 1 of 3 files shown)


tests/generic_config_updater/files/patch_sorter_test_failure.json
```diff
@@ -0,0 +1,71 @@
+{
+    "EMPTYING_A_CONFIGDB_TABLE__FAILURE": {
+        "desc": "Emptying a configdb table fails because empty tables are not allowed in configdb.",
```
qiluo-msft, line 3: >allowed [](http://example.com/codeflow?start=78&length=7)

I understand this is not allowed. Do you expect user to remove `/ACL_TABLE/`? If yes, add it into `desc` ?

ghooo: This test is not about emptying `ACL_TABLE` it is about emptying any configdb table. I think adding `ACL_TABLE` to the name will be confusing.

ghooo: User can delete any configdb table, as long as it passes YANG validation.

ghooo: Offline discussion:
[15:00] Mohamed Ghoneim
Ah, you mean what should the user do in case they encounter this error? good point

[15:00] Qi Luo
yes

Updated

qiluo-msft (approved):

qiluo-msft (approved):

PR #1983, opened 2021-12-23
[GCU] Copying config_db before calling sonic_yang.loadData

ghooo:
#### What I did
Adding unit-test for cases where `PatchSorter.sort` is called to update config that has tables without YANG. This would sometimes throw the error
```
KeyError: 'NameOfTableWithoutYang'
```
This error happened because `find_ref_paths` would remove tables without yang as a side-effect.

Earlier this was fixed by cropping tables without yang models from current and target configs. That fix had 2 issues:
- It is not addressing the real problem which is cropping the config as a side effect to `find_ref_paths`
- If the move generated by the inner logic of `PatchSorter` was to replace whole config, it would result in replacing whole config with the config with cropped non-yang tables

#### How I did it
- Fixing `find_ref_paths` to not crop the given config as a side-effect
  - Added a unit-test for that
- Revert the change to crop current config and target config in `PatchSorter.sort`
  - Added a unit-test to verify replace whole config move does not result unintentionally in removing tables without yang
- Added unit-test to verify calling `PatchSorter.sort` on a config with non-yang tables does not result in the exception: 
  ```
  KeyError: 'NameOfTableWithoutYang'
  ```

#### How to verify it
unit-tests

### DIFF
--- a/generic_config_updater/patch_sorter.py
+++ b/generic_config_updater/patch_sorter.py
@@ -36,6 +36,13 @@ class Diff:
     def has_no_diff(self):
         return self.current_config == self.target_config
 
+    def __str__(self):
+        return f"""current_config: {self.current_config}
+target_config: {self.target_config}"""
+
+    def __repr__(self):
+        return str(self)
+
 class JsonMove:
     """
     A class similar to JsonPatch operation, but it allows the path to refer to non-existing middle elements.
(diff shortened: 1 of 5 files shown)


qiluo-msft:
Could you explain your motivation in PR description? Is it a bug fix or refactoring?

generic_config_updater/patch_sorter.py
```diff
@@ -36,6 +36,13 @@ def apply_move(self, move):
     def has_no_diff(self):
         return self.current_config == self.target_config
 
+    def __str__(self):
```
ghooo, line 39: Adding `__str__` and `__repr__` are not related to this PR, but they are small improvements to the `Diff` class.

qiluo-msft (approved):

PR #1985, opened 2021-12-23
[Auto Techsupport] Added Event Driven TS to Command Reference 

vivekreddynv:
#### What I did

Added the Event Driven Tech Support related information to the Command Reference Guide.  [HLD](https://github.com/Azure/SONiC/blob/master/doc/auto_techsupport_and_coredump_mgmt.md)


doc/Command-Reference.md
```diff
@@ -9946,8 +9947,55 @@ In SONiC, there usually exists a set of tables related/relevant to a particular
...
+  config auto-techsupport global since <string>
   
+  config auto-techsupport-feature add <feature_name> --state <enabled/disabled> --rate-limit-interval <uint16>
+  config auto-techsupport-feature delete <feature_name>
+  config auto-techsupport-feature update <feature_name> --state <enabled/disabled> 
+  config auto-techsupport-feature update <feature_name> --rate-limit-interval <uint16>
 		
+  2) Show CLI:
+  show auto-techsupport global
+  show auto-techsupport-feature
```
liat-grozovik, line 9971: is - is correct for feature?

doc/Command-Reference.md
```diff
@@ -9946,8 +9947,55 @@ In SONiC, there usually exists a set of tables related/relevant to a particular
...
   ```
+
+### Event Driven Techsupport Invocation
+
+This feature/capability makes the techsupport invocation event-driven based on core dump generation. This feature is only applicable for the processes running in the containers. More detailed explanation can be found in the HLD https://github.com/Azure/SONiC/blob/master/doc/auto_techsupport_and_coredump_mgmt.md
+
+- Usage:
+  ```
+  1) Configuration:
+  config auto-techsupport global state <enabled/disabled>
```
liat-grozovik, line 9958: a question in general, should we have more info and examples for the config part? I believe user should not go to the HLD for CLI commands only if he would like to understand the internals. So my questions is what is the minimum info we can provide to users for the CLI documentation to avoid going to an HLD.

doc/Command-Reference.md
vivekreddynv: yes it is. Eg: https://github.com/Azure/SONiC/blob/master/doc/auto_techsupport_and_coredump_mgmt.md#show-cli

doc/Command-Reference.md
vivekreddynv: For the users to tune the configuration options, they have to understand them, and for that best place is to visit HLD. I believe any more description about those parameters here would be a duplicate information and not necessary. 

Let me know if you think otherwise

liat-grozovik (approved):

PR #1988, opened 2021-12-25
[GCU] Different apply-patch runs should produce same sorted steps

### LINKED ISSUE #1976: [GCU] PatchSorter.sort can generate different correct sorted steps for the same input -- always generate same steps

ghooo:
#### What I did
Fixes #1976

Different runs of `apply-patch` would produce different sorted steps. It is better to be consistent while producing the sorted steps to make it easier to debug issues in the future.

The issue is with using `set(list)` to remove duplicates from the list. Looping over the set elements does not guarantee same order every time we iterate. To reproduce this do the following:
1. Create a file named `test.py` with the content
```python
x = ["XYZ", "ABC", "DEF"]
y = set(x)
print(y)
```
2. Run `python3 test.py`
3. Again `python3 test.py`
4. Again `python3 test.py`
The output of these runs will be different each time.

#### How I did it
Instead of returning a `set`, return a list of `ref_paths` which does not have duplicated elements inserted to begin with.

#### How to verify it
unit-tests

### DIFF
--- a/tests/generic_config_updater/patch_sorter_test.py
+++ b/tests/generic_config_updater/patch_sorter_test.py
@@ -1747,6 +1747,9 @@ class TestSortAlgorithmFactory(unittest.TestCase):
         self.assertCountEqual(expected_validator, actual_validators)
 
 class TestPatchSorter(unittest.TestCase):
+    def setUp(self):
+        self.config_wrapper = ConfigWrapper()
+
     def test_patch_sorter_success(self):
         # Format of the JSON file containing the test-cases:
         #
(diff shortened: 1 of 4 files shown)


tests/generic_config_updater/patch_sorter_test.py
```diff
@@ -1787,7 +1788,7 @@ def run_single_success_case(self, data, skip_exact_change_list_match):
         simulated_config = current_config
         for change in actual_changes:
             simulated_config = change.apply(simulated_config)
-            self.assertTrue(ConfigWrapper().validate_config_db_config(simulated_config))
+            self.assertTrue(self.config_wrapper.validate_config_db_config(simulated_config))
```
ghooo, line 1791: Using a pre-created `config_wrapper` so we don't have to create one each time and was time loading yang models from disk each time.

qiluo-msft (approved):

PR #1989, opened 2021-12-28
[config] Fix snmpagentaddress doesn't support the uppercase letters for ipv6 address

### LINKED ISSUE #4852: [GCU] Need enhancement to support deleting keys one by one without empty-table restriction
### Description

Need enhancement to support delete keys one by one, without bothering about SONiC database restriction of empty table.

`config apply-patch` callers (automation scripts, incremental config push) typically emit per-key removes for the objects they want gone, for example every VLAN they manage:

```
{"op": "remove", "path": "/VLAN/Vlan10"}
{"op": "remove", "path": "/VLAN/Vlan20"}
```

They do not inspect running ConfigDB to learn whether those keys are the last remaining entries in the table. ConfigDB cannot store an empty table, so when the patch happens to delete the last keys, GCU simulates a target of `"VLAN": {}` and rejects the patch. The same intent succeeds if the caller instead sends a table-level remove:

```
{"op": "remove", "path": "/VLAN"}
```

That requires the caller to know device state up front. apply-patch should accept the per-key form and handle the empty-table restriction internally.

This is only requested for key-level `remove` operations (`/TABLE/KEY`). Other patch shapes (add/replace, field-level remove, table-level remove, mixed replace on the same table) should keep today's behavior. Field-operation checks (for example the illegal remove of `/LOOPBACK_INTERFACE/Loopback0`) must still run on the original transition.

### Steps to reproduce the issue

1. Configure a device whose `VLAN` table contains only `Vlan10` and `Vlan20`.
2. Apply a JSON patch that removes those keys one by one, with no table-level `/VLAN` remove:

```
[
  {"op": "remove", "path": "/VLAN/Vlan10"},
  {"op": "remove", "path": "/VLAN/Vlan20"}
]
```

3. `sudo config apply-patch <patch.json>`

### Describe the results you received

```
given patch is not valid because it will result in empty tables which is not allowed in ConfigDb. Table: VLAN
```

Nothing is applied. The caller has to re-read running config and resubmit a table-level remove.

### Describe the results you expected

apply-patch succeeds. GCU treats the per-key deletes as "remove these keys, and if that empties the table, remove the table" so automation does not have to special-case the ConfigDB empty-table restriction.

A mixed patch that also updates an unrelated table (for example a `PORT` replace next to the VLAN key deletes) should still apply; only the emptied table is rewritten.

### Additional information

Empty-table rejection itself is still the right ConfigDB rule (#1908 / #1923). This request is to rewrite only the key-level remove case so callers can delete keys one by one. It is the table-level counterpart of emptying a leaf-list (#4731).

xwjiang2021:
Fixes #4852 (https://github.com/Azure/sonic-mgmt/issues/4852)

#### What I did
          Snmpagentaddress now support upper letters input for ipv6 address
          Fix some grammer mistakes
#### How I did it
          Compare the lowercase of input and netifaces addresses
          delete ; after break, convert addfress to address
#### How to verify it
          'sudo config snmpagentadress add FC00:1::32 -v mgmt'
          The ipv6 address can be obtained by command 'sudo ifconfig'
          'IP addfress is not available'
          we can see the entry by echoing 'show snmpagentaddress'


qiluo-msft:
Do you intend to fix https://github.com/Azure/sonic-mgmt/issues/4852? It is not clear from your PR description.

config/main.py
```diff
@@ -2490,11 +2492,11 @@ def add_snmp_agent_address(ctx, agentip, port, vrf):
             for ipaddr in ipaddresses[ip_family[ip.version]]:
                 if agentip == ipaddr['addr']:
```
qiluo-msft, line 2493: >ipaddr['addr'] [](http://example.com/codeflow?start=30&length=14)

Currently the ConfigDB has no way to enforce the lowercase for SNMP_AGENT_ADDRESS_CONFIG lowercase key. So there may be already lowercase entries there.

Suggest you compare ipv6 address case insensitive. This library may help [https://docs.python.org/3/library/ipaddress.html](https://docs.python.org/3/library/ipaddress.html)

qiluo-msft:
@sumanbrcm Could you also help review?

xwjiang2021:
> Do you intend to fix [Azure/sonic-mgmt#4852](https://github.com/Azure/sonic-mgmt/issues/4852)? It is not clear from your PR description.

Yes I want to fix this issue, and I have edited the description.

config/main.py
xwjiang2021: I see, it's better to compare the lowercase of IPv6 addresses, and I have modified my PR, thanks.

yxieca (approved):

PR #2001, opened 2022-01-07
[flow counter] Fix issue: should not compare str with int

Junchao-Mellanox:
#### What I did

When clearing stats, current flow counter stats is serialized to a file for further compare. All stats value is saved as string value. When doing diff, the loaded stats value shall be convert into int and do the compare. The PR is to fix this issue.

#### How I did it

1. Convert value to int when doing compare
2. Save value as string to the dump file

#### How to verify it

1. Adjust unit test case to cover the change

### DIFF
--- a/scripts/flow_counters_stat
+++ b/scripts/flow_counters_stat
@@ -230,7 +230,7 @@ class FlowCounterStats(object):
                     # counter OID is changed.
                     old_values[-1] = values[-1]
                     for i in diff_column_positions:
-                        old_values[i] = 0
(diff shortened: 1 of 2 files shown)


keboliu (approved):

dgsudharsan (approved):

Junchao-Mellanox:
Hi @prsunny, could you please help cherry-pick to 202111?

Junchao-Mellanox:
Hi @judyjoseph , could you please help cherry-pick to 202111?

PR #2003, opened 2022-01-11
[GCU] Show default option for '--format'

ghooo:
#### What I did
Show the default enum for the option `--format`

#### How I did it
Added `show_default=True`

#### How to verify it
Checkout output below

```bash
admin@vlab-01:~$ sudo config apply-patch -h

  Apply given patch of updates to Config. A patch is a JsonPatch which
  follows rfc6902. This command can be used do partial updates to the config
  with minimum disruption to running processes. It allows addition as well
  as deletion of configs. The patch file represents a diff of ConfigDb(ABNF)
  format or SonicYang format.

  <patch-file-path>: Path to the patch file on the file-system.

Options:
  -f, --format [CONFIGDB|SONICYANG]
                                  format of config of the patch is either
                                  ConfigDb(ABNF) or SonicYang
  -d, --dry-run                   test out the command without affecting
                                  config state
  -v, --verbose                   print additional details of what the
                                  operation is doing
  -h, -?, --help                  Show this message and exit.
admin@vlab-01:~$ 
```

```sh
admin@vlab-01:~$ sudo config apply-patch -h

  Apply given patch of updates to Config. A patch is a JsonPatch which
  follows rfc6902. This command can be used do partial updates to the config
  with minimum disruption to running processes. It allows addition as well
  as deletion of configs. The patch file represents a diff of ConfigDb(ABNF)
  format or SonicYang format.

  <patch-file-path>: Path to the patch file on the file-system.

Options:
  -f, --format [CONFIGDB|SONICYANG]
                                  format of config of the patch is either
                                  ConfigDb(ABNF) or SonicYang  [default:
                                  CONFIGDB]
  -d, --dry-run                   test out the command without affecting
                                  config state
  -v, --verbose                   print additional details of what the
                                  operation is doing
  -h, -?, --help                  Show this message and exit.
admin@vlab-01:~$ 
```

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1188,7 +1188,8 @@ def print_dry_run_message(dry_run):
 @click.argument('patch-file-path', type=str, required=True)
 @click.option('-f', '--format', type=click.Choice([e.name for e in ConfigFormat]),
                default=ConfigFormat.CONFIGDB.name,
-               help='format of config of the patch is either ConfigDb(ABNF) or SonicYang')
+               help='format of config of the patch is either ConfigDb(ABNF) or SonicYang',
+               show_default=True)
 @click.option('-d', '--dry-run', is_flag=True, default=False, help='test out the command without affecting config state')
 @click.option('-n', '--ignore-non-yang-tables', is_flag=True, default=False, help='ignore validation for tables without YANG models', hidden=True)
 @click.option('-i', '--ignore-path', multiple=True, help='ignore validation for config specified by given path which is a JsonPointer', hidden=True)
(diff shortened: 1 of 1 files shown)


config/main.py
```diff
@@ -1188,7 +1188,8 @@ def print_dry_run_message(dry_run):
 @click.argument('patch-file-path', type=str, required=True)
 @click.option('-f', '--format', type=click.Choice([e.name for e in ConfigFormat]),
                default=ConfigFormat.CONFIGDB.name,
-               help='format of config of the patch is either ConfigDb(ABNF) or SonicYang')
+               help='format of config of the patch is either ConfigDb(ABNF) or SonicYang',
```
qiluo-msft, line 1191: >SonicYang [](http://example.com/codeflow?start=79&length=9)

Regarding the format string, we prefer ConfigYang or config_yang (ref: [https://github.com/Azure/sonic-utilities/blob/055ed4fafeaece1f00e4dcb20631805a5e621fb6/config/main.py#L1335](https://github.com/Azure/sonic-utilities/blob/055ed4fafeaece1f00e4dcb20631805a5e621fb6/config/main.py#L1335)).

The reason is to align with the naming of config_db. And in future we may have state_db/state_yang (just a conceptual example).

If you agree and would like to use another PR, we can merge this PR first.

qiluo-msft (approved):

config/main.py
ghooo: let me address this in a different PR

PR #2006, opened 2022-01-12
[GCU] Disallowing DeleteInsteadOfReplaceMoveExtender from generating delete whole config move

ghooo:
#### What I did
Not generating delete whole config move because it is not allowed by JsonPatch library and it is not possible for ConfigDb to be equal to NULL.

#### How I did it
- Verified UpperLevelMoveExtender does not generate delete whole config move
- Modified DeleteInsteadOfReplaceMoveExtender to not generate delete whole config move
- DeleteRefsMoveExtender cannot generate delete whole config move, because it deletes the referrers

#### How to verify it
unit-test


tests/generic_config_updater/files/patch_sorter_test_success.json
```diff
@@ -2811,5 +2811,88 @@
                 }
             ]
         ]
+    },
+    "ADDING_LOOPBACK0_VRF_NAME__DELETES_LOOPBACK0_AND_IPS_DOES_NOT_AFFECT_OTHER_TABLES": {
```
ghooo, line 2815: This test case was failing because at some point the patch-sorter
- tries to delete the whole config move which is rejected using which was generated using: DeleteInsteadOfReplaceMoveExtender 
- tries to delete the referrers to the whole config : DeleteRefsMoveExtender 
- why trying to remove referrers of whole config, it actually tries to delete referrers to PORT table i.e. CABLE_LENGTH.

renukamanavalan (approved):

wen587:
Verified locally. Vrf change can pass now.

tests/generic_config_updater/files/patch_sorter_test_success.json
wen587: resolve the issue

wen587 (approved):

PR #2015, opened 2022-01-15
[build-failure-fix][GCU] Sort referrer paths alphabetically directly on 202111

ghooo:
#### What I did
Make sure the same order is always returned while getting the referrer paths i.e. data dependencies.

#### How I did it
Sort the returned list.

#### How to verify it
unit-tests


judyjoseph:
Local build is good, merging in.

PR #2017, opened 2022-01-16
Revert "[config reload] Fixing config reload when timer based delayed…

qiluo-msft:
… services are disabled (#1967)"

This reverts commit 055ed4fafeaece1f00e4dcb20631805a5e621fb6.


dgsudharsan:
@qiluo-msft May I know the reason why it is reverted?

PR #2044, opened 2022-01-27
[GCU] Supporting Groupings during path-xpath translation

### LINKED ISSUE #2041: [GCU] apply-patch failed to get token from 'using' group
#### Description
When `apply-patch` a json file which yang has grouping member, the method cannot grab token.

1. bgp-monitor.yang has member that defined in sonic-bgp-common.yang
https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-bgp-monitor.yang#L43
2. When apply-patch, it failed to grab token.
```
admin@vlab-01:~$ sudo config apply-patch -i /FEATURE -i /QUEUE -i /SCHEDULER bgpl/add.json
Patch Applier: Patch: [{"op": "add", "path": "/BGP_MONITORS", "value": {"11.0.0.1": {"admin_status": "up", "asn": "65100", "holdtime": "10", "keepalive": "3", "local_addr": "10.1.0.32", "name": "BGPMonitor", "nhopself": "0", "rrclient": "0"}}}]
...
Failed to apply patch

Error: Token not found 
```
3. some debug info when error occur in https://github.com/Azure/sonic-utilities/blob/master/generic_config_updater/gu_common.py#L551
`raise ValueError("Token not found model {} token_index {} path_tokens {}".format(model, token_index, path_tokens))`
The token is only got from `leaf` not from `using`
```
model OrderedDict([('@name', 'BGP_MONITORS_LIST'), 
('key', OrderedDict([('@value', 'addr')])), 
('description', OrderedDict([('text', 'This list is to support BGP monitor configuration handled by bgpcfgd')])), 
('leaf', OrderedDict([('@name', 'addr'), ('type', OrderedDict([('@name', 'inet:ip-address')])), 
('description', OrderedDict([('text', 'BGP monitor address')])), ('__isleafList', False)])), 

('uses', OrderedDict([('@name', 'bgpcmn:sonic-bgp-cmn-neigh'),    <===================================================

('refine', OrderedDict([('@target-node', 'name'), 
('must', OrderedDict([('@condition', "current() = 'BGPMonitor'"), 
('error-message', OrderedDict([('value', 'Invalid BGP monitor name')]))]))]))]))]) 
token_index 2 path_tokens ['BGP_MONITORS', '11.0.0.1', 'asn']
```
See above.

Fix in two ways:
1. We don't using grouping in Yang such that bgp-monitor has all member defined in itself.
2. Handle `using` keyword in yang model.

#### Output of `show version`

```
(paste your output here)
```

ghooo:
#### What I did
Fixes #2041

Supporting groupings during `config-db path` <-> `config-yang xpath` translation

This mimics the `config-db` <-> `config-yang`  translation as added by https://github.com/Azure/sonic-buildimage/pull/8318

#### How I did it
Handled groupings of leaf commands only as that's what is supported in sonic-yang-mgmt

#### How to verify it
unit-test

### DIFF
--- a/generic_config_updater/gu_common.py
+++ b/generic_config_updater/gu_common.py
@@ -772,7 +772,7 @@ class PathAddressing:
             return None
 
         # a model can be a single dict or a list of dictionaries, unify to a list of dictionaries
-        if isinstance(uses_s, dict):
+        if not isinstance(uses_s, list):
             uses_s = [uses_s]
 
         sy = self._create_sonic_yang_with_loaded_models()
@@ -780,14 +780,21 @@ class PathAddressing:
         table_module = sy.confDbYangMap[table]['yangModule']
         # uses Example: "@name": "bgpcmn:sonic-bgp-cmn"
         for uses in uses_s:
+            if not isinstance(uses, dict):
+                raise GenericConfigUpdaterError(f"'uses' is expected to be a dictionary found '{type(uses)}'.\n" \
+                                                f"  uses: {uses}\n  model: {model}\n  table: {table}\n  token: {token}")
+
             # Assume ':'  means reference to another module
             if ':' in uses['@name']:
-                prefix = uses['@name'].split(':')[0].strip()
-                uses_module = sy._findYangModuleFromPrefix(prefix, table_module)
+                name_parts = uses['@name'].split(':')
+                prefix = name_parts[0].strip()
+                uses_module_name = sy._findYangModuleFromPrefix(prefix, table_module)
+                grouping = name_parts[-1].strip()
             else:
-                uses_module = table_module
-            grouping = uses['@name'].split(':')[-1].strip()
-            leafs = sy.preProcessedYang['grouping'][uses_module][grouping]
+                uses_module_name = table_module['@name']
+                grouping = uses['@name']
+
+            leafs = sy.preProcessedYang['grouping'][uses_module_name][grouping]
 
             leaf_model = self._get_model(leafs, token)
             if leaf_model:
--- a/generic_config_updater/gu_common.py
+++ b/generic_config_updater/gu_common.py
@@ -360,6 +360,9 @@ class PathAddressing:
 
         return f"{PathAddressing.XPATH_SEPARATOR}{PathAddressing.XPATH_SEPARATOR.join(str(t) for t in tokens)}"
 
+    def _create_sonic_yang_with_loaded_models(self):
+        return self.config_wrapper.create_sonic_yang_with_loaded_models()
+
     def find_ref_paths(self, path, config):
         """
         Finds the paths referencing any line under the given 'path' within the given 'config'.
@@ -401,7 +404,7 @@ class PathAddressing:
         return self._find_leafref_paths(path, config)
 
     def _find_leafref_paths(self, path, config):
-        sy = self.config_wrapper.create_sonic_yang_with_loaded_models()
+        sy = self._create_sonic_yang_with_loaded_models()
 
         tmp_config = copy.deepcopy(config)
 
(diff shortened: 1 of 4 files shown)


generic_config_updater/gu_common.py
```diff
@@ -746,6 +763,38 @@ def _get_model(self, model, name):
...
+    def _get_uses_leaf_model(self, model, table, token):
+        """
+          Getting leaf model in uses model matching the given token.
+        """
+        uses_s = model.get('uses')
+        if not uses_s:
+            return None
+
+        # a model can be a single dict or a list of dictionaries, unify to a list of dictionaries
+        if isinstance(uses_s, dict):
```
qiluo-msft, line 775: >isinstance [](http://example.com/codeflow?start=11&length=10)

Better to check "is not a list"

generic_config_updater/gu_common.py
```diff
@@ -746,6 +763,38 @@ def _get_model(self, model, name):
...
+            uses_s = [uses_s]
+
+        sy = self._create_sonic_yang_with_loaded_models()
+        # find yang module for current table
+        table_module = sy.confDbYangMap[table]['yangModule']
+        # uses Example: "@name": "bgpcmn:sonic-bgp-cmn"
+        for uses in uses_s:
```
qiluo-msft, line 782: >uses [](http://example.com/codeflow?start=12&length=4)

check each "uses" is a dict

generic_config_updater/gu_common.py
```diff
@@ -746,6 +763,38 @@ def _get_model(self, model, name):
...
+            # Assume ':'  means reference to another module
+            if ':' in uses['@name']:
+                prefix = uses['@name'].split(':')[0].strip()
```
qiluo-msft, line 785: >split [](http://example.com/codeflow?start=39&length=5)

split twice. Could you do once?

tests/generic_config_updater/files/config_db_with_bgp_neighbor.json
```diff
@@ -0,0 +1,20 @@
+{
+    "BGP_NEIGHBOR": {
```
qiluo-msft, line 2: >BGP_NEIGHBOR [](http://example.com/codeflow?start=5&length=12)

Test BGP_MONITORS?

generic_config_updater/gu_common.py
ghooo: good idea

generic_config_updater/gu_common.py
ghooo: updated

generic_config_updater/gu_common.py
ghooo: sure, done

tests/generic_config_updater/files/config_db_with_bgp_neighbor.json
ghooo: added

tests/generic_config_updater/gu_common_test.py
```diff
@@ -735,6 +735,18 @@ def check(path, xpath, config=None):
...
+        check(path="/BGP_NEIGHBOR/1.2.3.4/holdtime",
+              xpath="/sonic-bgp-neighbor:sonic-bgp-neighbor/BGP_NEIGHBOR/BGP_NEIGHBOR_TEMPLATE_LIST[neighbor='1.2.3.4']/holdtime",
+              config=Files.CONFIG_DB_WITH_BGP_NEIGHBOR)
+        check(path="/BGP_NEIGHBOR/default|1.2.3.4/asn",
+              xpath="/sonic-bgp-neighbor:sonic-bgp-neighbor/BGP_NEIGHBOR/BGP_NEIGHBOR_LIST[vrf_name='default'][neighbor='1.2.3.4']/asn",
+              config=Files.CONFIG_DB_WITH_BGP_NEIGHBOR)
+        check(path="/BGP_MONITORS/5.6.7.8/name",
+              xpath="/sonic-bgp-monitor:sonic-bgp-monitor/BGP_MONITORS/BGP_MONITORS_LIST[addr='5.6.7.8']/name",
+              config=Files.CONFIG_DB_WITH_BGP_NEIGHBOR)
+        check(path="/LLDP/GLOBAL/mode",
```
ghooo, line 747: This test is added because LLDP has a `grouping` defined in the model itself, although commented out I have added a test for it anyway https://github.com/Azure/sonic-buildimage/blob/bcf5388f250d99fbd626dead98be511a0306ec6d/src/sonic-yang-models/yang-models/sonic-lldp.yang#L119

wen587 (approved):
Test locally and BGP_MONITOR tests works fine.

qiluo-msft (approved):

qiluo-msft (approved):

PR #2047, opened 2022-01-27
[GCU] Adding unit-test where path and ref paths are under the same YANG container

ghooo:
#### What I did
Did a debug session with Ping Mao and found this case is failing in her PR: https://github.com/Azure/sonic-buildimage/pull/9545 I think it is an interesting case and I added it explicitly to GCU.

Adding unit-test where path and ref paths are under the same YANG container

[sonic-loopback-interface.yang](https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-loopback-interface.yang) has:
```yang
    container sonic-loopback-interface {
        container LOOPBACK_INTERFACE {
            list LOOPBACK_INTERFACE_LIST {
                ...
                leaf name{
                    type string;
                }
                ...
            list LOOPBACK_INTERFACE_IPPREFIX_LIST {
                leaf name{
                    ...
                    type leafref {
                        path "../../LOOPBACK_INTERFACE_LIST/name";
                    }
                }
               ...
```

#### How I did it
Unit-test added

#### How to verify it
unit-test

### DIFF
--- a/tests/generic_config_updater/gu_common_test.py
+++ b/tests/generic_config_updater/gu_common_test.py
@@ -675,6 +675,20 @@ class TestPathAddressing(unittest.TestCase):
         # Assert
         self.assertEqual(expected, actual)
 
+    def test_find_ref_paths__path_and_ref_paths_are_under_same_yang_container__returns_ref_paths(self):
+        # Arrange
+        path = "/LOOPBACK_INTERFACE/Loopback0"
+        expected = [
+            "/LOOPBACK_INTERFACE/Loopback0|10.1.0.32~132",
+            "/LOOPBACK_INTERFACE/Loopback0|1100:1::32~1128",
+        ]
+
+        # Act
+        actual = self.path_addressing.find_ref_paths(path, Files.CONFIG_DB_WITH_LOOPBACK_INTERFACES)
+
+        # Assert
+        self.assertEqual(expected, actual)
+
     def test_find_ref_paths__does_not_remove_tables_without_yang(self):
         # Arrange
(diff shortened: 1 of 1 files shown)


renukamanavalan (dismissed):

tests/generic_config_updater/gu_common_test.py
```diff
@@ -675,6 +675,20 @@ def test_find_ref_paths__whole_config_path__returns_all_refs(self):
         # Assert
         self.assertEqual(expected, actual)
 
+    def test_find_ref_paths__path_and_ref_paths_are_under_same_yang_container__returns_ref_paths(self):
+        # Arrange
+        path = "/LOOPBACK_INTERFACE/Loopback0"
+        expected = [
+            "/LOOPBACK_INTERFACE/Loopback0|10.1.0.32~132",
```
qiluo-msft, line 682: >~1 [](http://example.com/codeflow?start=52&length=2)

To improve readability regarding the `~1` escape, we con construct the paths from components.

eg: [https://github.com/Azure/sonic-mgmt/blob/d23d893a40e898fb3aeb70abd3bd60d08492b484/tests/generic_config_updater/gu_utils.py#L244](https://github.com/Azure/sonic-mgmt/blob/d23d893a40e898fb3aeb70abd3bd60d08492b484/tests/generic_config_updater/gu_utils.py#L244)

ghooo: updated

qiluo-msft (dismissed):

qiluo-msft (approved):

PR #2053, opened 2022-02-02
Validate destination port is not LAG

rupesh-k:
#### What I did
LAG as mirror destionation is not supported in SPAN. Add validation in CLI.
This fixes issue https://github.com/Azure/sonic-buildimage/issues/7989#issue-930956772
Also fixes https://github.com/Azure/sonic-buildimage/issues/7943
#### How I did it
Add extra validation to check if the destion port is LAG or not.
#### How to verify it
Add mirror config with LAG as destination port.
Allow mirror config and orchagent logs error
Show error message that Destination port is not supported


raphaelt-nvidia:
In the bug, I mentioned two possible directions of solution.  The other one was "The HLD hints that a portchannel might be valid as a mirror session destination: Existing CLI 'config mirror_session add/remove' to be extended to include destination port/portchannel".

Did you determine that 

- the HLD didn't really mean it, and the current fix is permanent, or 
- the HLD did mean it, and the current fix is temporary?
-

rupesh-k:
Hi @raphaelt-nvidia ,

This is the final fix. 
There was no plan to support mirroring to port-channel. 
The orchagent code explicitly handles this condition already and swss schema also captures that. 
Here we are blocking at CLI level.

HLD also captures this in config db update. There was typo as u mentioned , will take care of that in HLD.

Thanks

raphaelt-nvidia:
LGTM

bingwang-ms (dismissed):

bingwang-ms:
LGTM.  Thanks for the fix.

bingwang-ms (approved):

liat-grozovik:
@rupesh-k could you please check failures and if needed restart checkers?

bingwang-ms (approved):

PR #2064, opened 2022-02-12
[warmboot] Migrate 10G ports during warm-reboot on s6100

tjchadaga:
#### What I did
Changes include db-migration logic to add back 10G ports on 6100 during warm-reboot from earlier image to ensure config and app DB are in sync with new port_config.ini

#### How I did it
Set configdb manually and update corresponding new entries in appdb

#### How to verify it
This PR fixes https://github.com/Azure/sonic-buildimage/issues/9976


scripts/db_migrator.py
```diff
@@ -125,6 +126,50 @@ def migrate_interface_table(self):
...
+            log.log_notice("Port count not found - {}, skip migration".format(portCount))
+            return True
+
+        self.portCount = int(portCount)
+        if self.portCount > 64:
+            log.log_notice("Port count is {}, no migration needed".format(self.portCount))
+            return True        
+        
+        log.log_notice("Migrating management ports Ethernet64 & Ethernet65 to configDB for warm-reboot on {}".format(self.hwsku))
+        self.configDB.set_entry('PORT', 'Ethernet64', {'alias': 'tenGige1/1', 'description': 'tenGige1/1', 'index': '64', 'lanes': '129', 'mtu': '9100', 'pfc_asym': 'off', 'speed': '10000'})
```
gechiang, line 149: Question: Is there any chance that Ethernet64 and/or Ethernet65 already exists on configDB already?  Also, are we assuming that these two ports MUST be the two mgmt ports?  The previous portCount check seems to suggest that as long as the port count is 64 or less then this will always be the case... Not sure if there are any other S6100 variants currently or maybe in the future that this is not the case?

tjchadaga: Updated code based on discussion. Please review

scripts/db_migrator.py
```diff
@@ -125,6 +126,43 @@ def migrate_interface_table(self):
...
+        '''
+        During warm-reboot, add back two 10G management ports which got removed from 6100
+        to ensure no change in bcm.config from older image
+        '''
+        if device_info.is_warm_restart_enabled('swss') == False:
+            log.log_notice("Skip migration on {}, warm-reboot flag not set".format(self.hwsku))
+            return True
+
+        entries = {}
+        entries['Ethernet64'] = {'alias': 'tenGige1/1', 'description': 'tenGige1/1', 'index': '64', 'lanes': '129', 'mtu': '9100', 'pfc_asym': 'off', 'speed': '10000'}
```
aravindmani-1, line 139: Could you change the alias name to tenGigE instead of tenGige?.

scripts/db_migrator.py
```diff
@@ -125,6 +126,43 @@ def migrate_interface_table(self):
...
+        entries['Ethernet65'] = {'alias': 'tenGige1/2', 'description': 'tenGige1/2', 'index': '65', 'lanes': '131', 'mtu': '9100', 'pfc_asym': 'off', 'speed': '10000'}
```
aravindmani-1, line 140: Could you change the alias name to tenGigE instead of tenGige?.

aravindmani-1 (changes requested):

scripts/db_migrator.py
tjchadaga: Done

yxieca (approved):

aravindmani-1:
@tjchadaga Is there any reason why builds are failing here?.

gechiang (approved):

PR #2079, opened 2022-02-22
[ci] Use official build debian pkg instead and parameterize source branch.

liushilongbuaa:
#### What I did
1. Use official build debian pkg.
2. parameterize download source branch for future release branch.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -29,14 +29,22 @@ stages:
       image: sonicdev-microsoft.azurecr.io:443/sonic-slave-buster:latest
 
     steps:
+    - script: |
+        sourceBranch=$(Build.SourceBranchName)
+        if [[ "$(Build.Reason)" == "PullRequest" ]];then
+          sourceBranch=$(System.PullRequest.TargetBranch)
+        fi
+        echo "Download artifact branch: $sourceBranch"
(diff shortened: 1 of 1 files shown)


xumia (approved):

qiluo-msft (approved):

qiluo-msft:
Will override vstest failure to fix master build.

ghooo:
> Will override vstest failure to fix master build.

@qiluo-msft For my understanding, I think this PR is not impacting builds that are on top of master branch. It is affecting builds that are on top on other branches like 202111. Looking at the logs of this PR I see the following 
```
Download artifact branch: master
```
So it is not changing.

The builds in master branch in sonic-utils will pass `Test Python 3` now because we have master branch branch successful build.
https://dev.azure.com/mssonic/build/_build/results?buildId=74810&view=results

liushilongbuaa:
@ghooo yes, this PR will not change the download branch for sonic-utilities master branch. It will change future release branches.
Former two builds on pipeline 1 failed for unknown reason, so it's better to use pipeline 142. There are two main reason: 1. pipeline 142 don't use cache, so it is more clean. 2. pipeline 1 run a build for each commit, any breaking PR merging will impact sonic-utilities job.

ghooo:
@ liushilongbuaa got it, thanks for the detailed explanation.

PR #2081, opened 2022-02-24
[config/main.py]: Enable config validation using YANG with config reload.

praveen-li:
Changes:
-- Perform config validation before stopping any service.
-- Perform config validation before and after db migration.
-- TODO: Make config reload aware of migration so that config
   validation is done either before or after db migration.
-- white space changes.

Signed-off-by: Praveen Chaudhary <pchaudhary@linkedin.com>

#### What I did

Enable config validation using YANG with config reload.

#### How I did it

Changes:
-- Perform config validation before stopping any service.
-- Perform config validation before and after db migration.
-- TODO: Make config reload aware of migration so that config
   validation is done either before or after db migration.
-- white space changes.

Considerations: 

**----- config in YANG format: [Current Implementation]** 

Today if config is provided in yang format, we validate it before updating config_db inside sonic-cfgen using SonicYangCfgDbGenerator class. In case of invalid configuration it does no update to config_db. But It seems to be validating config, little too late, because we have already stopped services. In this case, we restart services but config is flushed already. 

https://github.com/Azure/sonic-buildimage/blob/3bbe562bcaeb3e16fec86f27c185d30b5b44cf6e/src/sonic-config-engine/sonic-cfggen#L322 

Flushed the DB here: 
https://github.com/Azure/sonic-utilities/blob/a9bbaf2273169ed9f4c764cc39e97708995e0fa3/config/main.py#L1455 

***No change will be done with above in this PR.***

**----- Migration Scenarios:** 

**1.) Avoid validation before db-migration step.** 

When we migrate using sonic-installer, old config file will be copied to /etc/sonic folder. We need to avoid Validation at start of config-reload i.e. before db migration step is done. Reason to avoid config validation at start is, YANG model may have changed b/w releases. And we may be doing validation of old config against new YANG models. 

How to avoid config validation in this scenario? Should we pass an option, when migration is done, from config_setup file. 

***[This Part will be handled in future PR. With this PR, validation will be always performed at start i.e. before stopping any services.]*** 

**2. Validation must be done after db-migration step only when upgrade is performed using sonic-installer.**

 Validation will be done after DB migration step during config reload always. 
***[This PR will cover it, though in future the config validation should be done after db-migration step only when upgrade is performed using sonic-installer].***

 In this case, If validation fails after DB migration, we return immediately, i.e. no services starts is done.

***Doubts:* Is above a right option for warm-reboot cases. [We may be fine here, since It implies, that there is a bug in db_migrator.py] .**

----- yang file format validation at start, i.e. before stopping services: 
***Not covered in this PR.***

----- config load scenario: 
***Not covered in this PR.***

#### How to verify it
No new test is added as of now, but I expect many previous tests to fail due to config not matching with yang models.
So the goal of this PR is to find the config for which yang models are not written yet. Once we reached there, I will add test if still no coverage for new code.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -42,7 +42,7 @@ from . import nat
 from . import vlan
 from . import vxlan
 from . import plugins
-from .config_mgmt import ConfigMgmtDPB
+from .config_mgmt import ConfigMgmt, ConfigMgmtDPB
 from . import mclag
 
 # mock masic APIs for unit test
@@ -789,7 +789,7 @@ def _per_namespace_swss_ready(service_name):
         return False
 
 def _swss_ready():
-    list_of_swss = [] 
+    list_of_swss = []
     num_asics = multi_asic.get_num_asics()
     if num_asics == 1:
         list_of_swss.append("swss.service")
@@ -802,7 +802,7 @@ def _swss_ready():
         if _per_namespace_swss_ready(service_name) == False:
             return False
 
-    return True 
+    return True
 
 def _is_system_starting():
     out = clicommon.run_command("sudo systemctl is-system-running", return_cmd=True)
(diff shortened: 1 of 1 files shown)


zhangyanzhao:
will be covered by CLI with YANG validation effort. No need on this function and close

PR #2082, opened 2022-02-24
[counterpoll] Display the correct default poll interval for watermark counters

neethajohn:

Azure/sonic-swss#2031 updated the flex counter polling interval to 60s for watermark counters but the show command wasn't updated to reflect the correct default value

#### What I did
Display the correct poll interval for watermark related counters in the 'counterpoll show' command

#### How I did it
Update the default interval to be the same as the one updated by the Orchs

#### How to verify it
Issue "counterpoll show", the queue, pg and buffer pool watermark should show default 60s
Updated the counterpoll unit tests to reflect the same

### DIFF
--- a/tests/counterpoll_test.py
+++ b/tests/counterpoll_test.py
@@ -17,16 +17,16 @@ from .mock_tables import dbconnector
 
 import counterpoll.main as counterpoll
 
-expected_counterpoll_show = """Type                    Interval (in ms)  Status
+expected_counterpoll_show = """Type                  Interval (in ms)    Status
 --------------------  ------------------  --------
-QUEUE_STAT                         10000  enable
-PORT_STAT                           1000  enable
-PORT_BUFFER_DROP                   60000  enable
-QUEUE_WATERMARK_STAT               60000  enable
-PG_WATERMARK_STAT                  60000  enable
-PG_DROP_STAT                       10000  enable
-ACL                                10000  enable
-FLOW_CNT_TRAP_STAT                 10000  enable
+QUEUE_STAT            10000               enable
+PORT_STAT             1000                enable
+PORT_BUFFER_DROP      60000               enable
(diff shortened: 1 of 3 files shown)


StormLiangMS (approved):
LGTM

tests/counterpoll_test.py
```diff
@@ -17,16 +17,16 @@
...
 --------------------  ------------------  --------
-QUEUE_STAT                         10000  enable
-PORT_STAT                           1000  enable
-PORT_BUFFER_DROP                   60000  enable
-QUEUE_WATERMARK_STAT               10000  enable
-PG_WATERMARK_STAT                  10000  enable
-PG_DROP_STAT                       10000  enable
-ACL                                10000  enable
-FLOW_CNT_TRAP_STAT                 10000  enable
+QUEUE_STAT            10000               enable
```
StormLiangMS, line 22: This format is different with the output of counterpoll show on tb as bellowing, wondering if this test case can pass?

admin@str2-7260cx3-acs-12:~$ counterpoll show
Type                        Interval (in ms)    Status
--------------------------  ------------------  --------
QUEUE_STAT                  default (10000)     enable
PORT_STAT                   default (1000)      enable
PORT_BUFFER_DROP            default (60000)     enable
RIF_STAT                    default (1000)      enable
QUEUE_WATERMARK_STAT        default (10000)     enable
PG_WATERMARK_STAT           default (10000)     enable
PG_DROP_STAT                default (10000)     enable
BUFFER_POOL_WATERMARK_STAT  default (10000)     enable
admin@str2-7260cx3-acs-12:~$

neethajohn: Yes. that's because in the testbed all our default values but here some are configured and some are default. python wheel was built successfully in the pr run

yxieca (approved):

qiluo-msft:

PR #2103, opened 2022-03-15
[generic-config-updater] Add caclrule validator

wen587:
After the json patch could be gathered in some way, it should fix nightly random failure issue.
#### What I did
When gcu make change to control plane ACL_RULE, the iptables doen't change immediately. There is a delay because caclmgr will update in 0.5 sec if no more update being made to config. So I add a caclrule validator to make sure the iptable is updated.
#### How I did it
When caclrule is being changed, add a 1sec sleep to make sure caclmgr does update.
#### How to verify it
Run sonic-utilities unit test

### DIFF
--- a/generic_config_updater/generic_config_updater.conf.json
+++ b/generic_config_updater/generic_config_updater.conf.json
@@ -44,7 +44,7 @@
             "services_to_validate": [ "vlan-service" ]
         },
         "ACL_RULE": {
-            "services_to_validate": [ "caclrule-service" ]
+            "services_to_validate": [ "caclmgrd-service" ]
         }
     },
     "services": {
@@ -63,8 +63,8 @@
         "vlan-service": {
             "validate_commands": [ "generic_config_updater.services_validator.vlan_validator" ]
         },
-        "caclrule-service": {
-            "validate_commands": [ "generic_config_updater.services_validator.caclrule_validator" ]
+        "caclmgrd-service": {
+            "validate_commands": [ "generic_config_updater.services_validator.caclmgrd_validator" ]
         }
     }
 }
--- a/generic_config_updater/generic_config_updater.conf.json
+++ b/generic_config_updater/generic_config_updater.conf.json
@@ -42,6 +42,9 @@
         },
         "VLAN": {
             "services_to_validate": [ "vlan-service" ]
+        },
+        "ACL_RULE": {
+            "services_to_validate": [ "caclrule-service" ]
         }
     },
     "services": {
@@ -59,6 +62,9 @@
         },
         "vlan-service": {
             "validate_commands": [ "generic_config_updater.services_validator.vlan_validator" ]
+        },
+        "caclrule-service": {
+            "validate_commands": [ "generic_config_updater.services_validator.caclrule_validator" ]
         }
     }
 }
--- a/generic_config_updater/services_validator.py
+++ b/generic_config_updater/services_validator.py
@@ -93,8 +93,8 @@ def caclmgrd_validator(old_config, upd_config, keys):
         if (old_aclrule.get(key, {}) != upd_aclrule.get(key, {})):
             # caclmgrd will update in 0.5 sec when configuration stops,
             # we sleep 1 sec to make sure it does update.
-            rc = os.system("sleep 1s")
-            return rc == 0
+            time.sleep(1)
+            return True
     # No update to ACL_RULE.
     return True
 
--- a/generic_config_updater/services_validator.py
+++ b/generic_config_updater/services_validator.py
@@ -71,3 +71,16 @@ def vlan_validator(old_config, upd_config, keys):
     # No update to DHCP servers.
     return True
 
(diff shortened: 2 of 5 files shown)


generic_config_updater/generic_config_updater.conf.json
```diff
@@ -42,6 +42,9 @@
         },
         "VLAN": {
             "services_to_validate": [ "vlan-service" ]
+        },
+        "ACL_RULE": {
+            "services_to_validate": [ "caclrule-service" ]
```
qiluo-msft, line 47: >caclrule-service [](http://example.com/codeflow?start=39&length=16)

This is uncommon name. How about `caclmgrd`?

generic_config_updater/services_validator.py
```diff
@@ -71,3 +71,35 @@ def vlan_validator(old_config, upd_config, keys):
     # No update to DHCP servers.
     return True
 
+def caclrule_validator(old_config, upd_config, keys):
+    old_acltable = old_config.get("ACL_TABLE", {})
+    upd_acltable = upd_config.get("ACL_TABLE", {})
+
+    try:
+        old_cacltable = [table for table in old_acltable
```
qiluo-msft, line 79: >old_acltable [](http://example.com/codeflow?start=44&length=12)

You can iterate `old_acltable.items()` and get key and value at the same time. It will simplify the code.

generic_config_updater/services_validator.py
```diff
@@ -71,3 +71,35 @@ def vlan_validator(old_config, upd_config, keys):
...
+    old_caclrule = [rule for rule in old_aclrule
+                    if rule.split("|")[0] in old_cacltable]
+    upd_caclrule = [rule for rule in upd_aclrule
+                    if rule.split("|")[0] in upd_cacltable]
+
+    for key in set(old_caclrule).union(set(upd_caclrule)):
+        if (old_aclrule.get(key, {}) != upd_aclrule.get(key, {})):
+            # caclmgrd will update in 0.5 sec when configuration stops,
+            # we sleep 1 sec to make sure it does update.
+            rc = os.system("sleep 1s")
```
qiluo-msft, line 101: >os.system("sleep 1s") [](http://example.com/codeflow?start=17&length=21)

python has `time.sleep()`.

generic_config_updater/services_validator.py
```diff
@@ -71,3 +71,35 @@ def vlan_validator(old_config, upd_config, keys):
...
+    old_aclrule = old_config.get("ACL_RULE", {})
+    upd_aclrule = upd_config.get("ACL_RULE", {})
+
+    old_caclrule = [rule for rule in old_aclrule
+                    if rule.split("|")[0] in old_cacltable]
+    upd_caclrule = [rule for rule in upd_aclrule
+                    if rule.split("|")[0] in upd_cacltable]
+
+    for key in set(old_caclrule).union(set(upd_caclrule)):
+        if (old_aclrule.get(key, {}) != upd_aclrule.get(key, {})):
```
qiluo-msft, line 98: Could you add code comment about `for` loop? Why not just sleep once during one call?

generic_config_updater/services_validator.py
wen587: This is to check the test cmds https://github.com/Azure/sonic-utilities/blob/13f88c1aad4c6a5293fef978b7affb95335d1f04/tests/generic_config_updater/service_validator_test.py#L106 also avoid conflict use with 
https://github.com/Azure/sonic-utilities/blob/13f88c1aad4c6a5293fef978b7affb95335d1f04/generic_config_updater/services_validator.py#L37

generic_config_updater/services_validator.py
wen587: added inline

qiluo-msft (approved):

wen587:
@renukamanavalan could you help review?

renukamanavalan:
> @renukamanavalan could you help review?

@prsunny, can you please help?

prsunny (approved):

generic_config_updater/services_validator.py
```diff
@@ -71,3 +71,30 @@ def vlan_validator(old_config, upd_config, keys):
...
+                    if rule.split("|")[0] in old_cacltable]
+    upd_caclrule = [rule for rule in upd_aclrule
+                    if rule.split("|")[0] in upd_cacltable]
+
+    # Only sleep when cacl rule is changed as this will update iptable.
+    for key in set(old_caclrule).union(set(upd_caclrule)):
+        if (old_aclrule.get(key, {}) != upd_aclrule.get(key, {})):
+            # caclmgrd will update in 0.5 sec when configuration stops,
+            # we sleep 1 sec to make sure it does update.
+            time.sleep(1)
```
prsunny, line 96: IMO, this is a blind sleep. If `caclmgrd` is changed in future to wait 1sec, we may have to revisit here. I don't have a better suggestion though.

PR #2107, opened 2022-03-17
[Mellanox] [reboot] [asan] stop asan-enabled containers on reboot

Yakiv-Huryk:
During a cold reboot on an image built with ENABLE_ASAN=y, swss and syncd (only on Mellanox platform) are explicitly stopped.
This will make the relevant processes receive SIGTERM and generate the ASAN reports.
The behavior for fast/warm reboots is not affected.
The behavior of regular (ENABLE_ASAN=n) images is not affected.

#### What I did
Modified the reboot script so the asan logs are generated on reboot.
#### How I did it
Added stop_services_asan() to reboot script.
#### How to verify it
Added "asan: 'yes'" to the /etc/sonic/sonic_version.yml
Checked that relevant containers are stopped and logs are generated

### DIFF
--- a/scripts/reboot
+++ b/scripts/reboot
@@ -25,6 +25,7 @@ REBOOT_USER=$(logname)
 PLATFORM=$(sonic-cfggen -H -v DEVICE_METADATA.localhost.platform)
 ASIC_TYPE=$(sonic-cfggen -y /etc/sonic/sonic_version.yml -v asic_type)
 SUBTYPE=$(sonic-cfggen -d -v DEVICE_METADATA.localhost.subtype)
+ASAN=$(sonic-cfggen -y /etc/sonic/sonic_version.yml -v asan)
 VERBOSE=no
 EXIT_NEXT_IMAGE_NOT_EXISTS=4
 EXIT_SONIC_INSTALLER_VERIFY_REBOOT=21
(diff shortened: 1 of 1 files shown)


liat-grozovik:
@Yakiv-Huryk  could you please add to the PR description the behavior expected? what will be done in case of reboot/warm/fast? beside of that LGTM

liat-grozovik (approved):

scripts/reboot
```diff
@@ -78,6 +79,17 @@ function stop_sonic_services()
     stop_pmon_service
 }
 
+function stop_services_asan()
+{
+    if [[ x"$ASIC_TYPE" == x"mellanox" ]]; then
+        debug "Stopping syncd for ASAN"
+        systemctl stop syncd
```
theasianpianist, line 86: Stopping SWSS should also stop syncd automatically, is it necessary to explicitly stop syncd here?

Yakiv-Huryk: Good catch, thanks. I've updated the PR: https://github.com/Azure/sonic-utilities/pull/2107/commits/5484b43d83a50de6217c63cef9290fe9830c2068

liat-grozovik (approved):

liat-grozovik:
@theasianpianist can you please signoff?

scripts/reboot
```diff
@@ -78,6 +79,12 @@ function stop_sonic_services()
     stop_pmon_service
 }
 
+function stop_services_asan()
+{
+    debug "Stopping swss for ASAN"
+    systemctl stop swss
```
theasianpianist, line 85: Want to confirm here that SWSS needs to be stopped on all platforms, not just on Mellanox

Yakiv-Huryk: It's for all platforms since asan for swss is platform-independent.

liat-grozovik:
@theasianpianist kindly reminder to signoff

theasianpianist (approved):

PR #2118, opened 2022-03-27
[yang] extend ConfigMgmt constructor to pass YANG options

ayurkiv-nvda:
Should be merged after  https://github.com/Azure/sonic-buildimage/pull/10359 Used  **SonicYang** with extended list of params to init object.

#### What I did
Pass YANG attribute that allow do not automatically search for schemas in current working directory, which is by default searched automatically (despite not recursively).

#### How I did it
extend ConfigMgmt constructor

#### How to verify it
1) Create some invalid link on switch :
**ln -s /usr/abc xxx**
2) run **spm list**
--> There should not be these messages:
```
libyang[1]: Unable to get information about "xxx" file in "/tmp" when searching for (sub)modules (No such file or directory)
libyang[1]: Unable to get information about "xxx" file in "/tmp" when searching for (sub)modules (No such file or directory)
libyang[1]: Unable to get information about "xxx" file in "/tmp" when searching for (sub)modules (No such file or directory)
libyang[1]: Unable to get information about "xxx" file in "/tmp" when searching for (sub)modules (No such file or directory)
```

### DIFF
--- a/sonic_package_manager/manager.py
+++ b/sonic_package_manager/manager.py
@@ -5,6 +5,7 @@ import functools
 import os
 import pkgutil
 import tempfile
+import yang as ly
 from inspect import signature
 from typing import Any, Iterable, List, Callable, Dict, Optional
 
@@ -1002,7 +1003,7 @@ class PackageManager:
         docker_api = DockerApi(docker.from_env(), ProgressManager())
         registry_resolver = RegistryResolver()
         metadata_resolver = MetadataResolver(docker_api, registry_resolver)
-        cfg_mgmt = config_mgmt.ConfigMgmt(source=INIT_CFG_JSON)
+        cfg_mgmt = config_mgmt.ConfigMgmt(source=INIT_CFG_JSON, sonicYangOptions=ly.LY_CTX_DISABLE_SEARCHDIR_CWD)
         cli_generator = CliGenerator(log)
         feature_registry = FeatureRegistry(SonicDB)
         service_creator = ServiceCreator(feature_registry,
--- a/config/config_mgmt.py
+++ b/config/config_mgmt.py
@@ -35,7 +35,7 @@ class ConfigMgmt():
     to verify config for the commands which are capable of change in config DB.
     '''
 
-    def __init__(self, source="configDB", debug=False, allowTablesWithoutYang=True):
+    def __init__(self, source="configDB", debug=False, allowTablesWithoutYang=True, sonicYangOptions=0):
         '''
         Initialise the class, --read the config, --load in data tree.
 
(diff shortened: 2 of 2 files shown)


sonic_package_manager/manager.py
```diff
@@ -1002,7 +1003,7 @@ def get_manager() -> 'PackageManager':
         docker_api = DockerApi(docker.from_env(), ProgressManager())
         registry_resolver = RegistryResolver()
         metadata_resolver = MetadataResolver(docker_api, registry_resolver)
-        cfg_mgmt = config_mgmt.ConfigMgmt(source=INIT_CFG_JSON)
```
praveen-li, line 1005: I did not understand this use. 
So ConfigMgmt searches for yang models at a fixed location 'YANG_DIR = "/usr/local/yang-models"', then in what case it is an error ?.

Kindly elaborate, thx.

ayurkiv-nvda: Not exactly 
ConfigMgmt searches for yang models in current working directory AND "/usr/local/yang-models"

dprital:
@praveen-li  - Now that the comment was addressed, can you please approve this PR ?

ayurkiv-nvda:
It looks like PR checks failed because of sonic-buildimage pointer should be updated after https://github.com/Azure/sonic-buildimage/pull/10359  merging

zhangyanzhao:
@praveen-li would you please help to review and approve if you are ok with the latest update? Thanks.

qiluo-msft:
@ghooo Could you help review this PR while @praveen-li is not available?

ayurkiv-nvda:
@ghooo Could you please help to review changes?

sonic_package_manager/manager.py
ghooo: Since the expectation is to only read schema from `/user/local/yang-models` let's disable reading schema from CWD (current working dir) in sonic-mgmt directly, no need to create a flags object and pass it.

libyang ref: https://netopeer.liberouter.org/doc/libyang/master/html/group__contextoptions.html
PR to add SonicYangOptions needs to be reverted: https://github.com/Azure/sonic-buildimage/pull/10359

ayurkiv-nvda: @ghooo  I have considered this option earlier, but after the discussion I chose the current approach. It looks more flexible to me and allows to pass different yang options to sonic-yang-mgmt

dprital:
@ghooo - As the comments were addressed, can you please approve this PR ?

dprital:
@ghooo - Can you please approve this PR ?

ghooo (approved):

PR #2122, opened 2022-03-30
[QoS] Fix issues in clear_qos

stephenxs:
#### What I did

Fix issues in `clear_qos`

- Remove tables BUFFER_PORT_INGRESS_PROFILE_LIST and BUFFER_PORT_EGRESS_PROFILE_LIST
  When `clear_qos` is executed, all QoS- and buffer-related tables should be removed. Currently, both tables are missed. This is to fix it.
- Adjust the order in which the buffer and QoS tables are removed
  Remove the tables whose entries depend on other tables first and then the tables whose entries are referenced by other tables.
  Any object can be removed from SAI only if there is no object referencing it. There are retrying and dependency-tracking mechanisms to guarantee it in `orchagent`.
  If CLI operates the table in an order where the referencing tables removed first before referenced tables, it is possible to reduce the probability to retry theoretically.
  The order does not guarantee it but adjusting the order is almost zero-cost so we would like to do it.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -635,6 +635,8 @@ def _clear_cbf():
 
 def _clear_qos():
     QOS_TABLE_NAMES = [
+            'PORT_QOS_MAP',
+            'QUEUE',
             'TC_TO_PRIORITY_GROUP_MAP',
             'MAP_PFC_PRIORITY_TO_QUEUE',
             'TC_TO_QUEUE_MAP',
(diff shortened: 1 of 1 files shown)


keboliu (approved):

liat-grozovik:
@stephenxs is this also an issue for 202111? 202012?

liat-grozovik (approved):

neethajohn (approved):

PR #2123, opened 2022-03-30
Fixing get port speed when oper status is down

dgsudharsan:
Signed-off-by: Sudharsan Dhamal Gopalarathnam <sudharsand@nvidia.com>

#### What I did
Recently enhancements update port oper speed in state_db and query it do display in show commands. However the oper speed gets updated only when admin is up. When admin goes down and if the user configures new speed it doesn't get updated in state_db. So a query to show interface status displays the last set port speed and not newly updated speed.

#### How I did it
To overcome the issue, the port speed is read from state_db only when oper status is up. Else it is obtained from config_db.

#### How to verify it
Added test configuration to verify the scenario.

### DIFF
--- a/scripts/intfutil
+++ b/scripts/intfutil
@@ -154,7 +154,8 @@ def port_oper_speed_get(db, intf_name):
     Get port oper speed
     """
     oper_speed = db.get(db.STATE_DB, PORT_STATE_TABLE_PREFIX + intf_name, PORT_SPEED)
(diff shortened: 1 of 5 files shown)


Junchao-Mellanox (approved):

prgeor (approved):

dgsudharsan:
@prgeor  Can you please help to merge the fix?

PR #2140, opened 2022-04-26
[minigraph]  Consume golden_config_db.json while loading minigraph

wen587:
#### What I did
This PR is for supporting consume `golden_config_db.json` while loading minigraph. User can put the `golden_config_db.json` with minigraph file to override configDB after reload minigraph.
The `golden_config_db.json` looks just like a config_db.json and it will be placed on `/etc/sonic/golden_config_db.json`.
#### How I did it
Add code `config override-config-table` command to let it consume `golden_config_db.json`
#### How to verify it
Add UT tests and run.
```
admin@vlab-01:~$ sudo config override-config-table -h
Usage: config override-config-table [OPTIONS] INPUT_CONFIG_DB

  Override current configDB with input config.

Options:
  --dry_run TEXT  Dry run, writes config to the given file
  -h, -?, --help  Show this message and exit.
```

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1678,10 +1678,8 @@ def load_port_config(config_db, port_config_path):
 
 def override_config_by(golden_config_path):
     # Override configDB with golden config
-    cmd = "config override-config-table {}".format(golden_config_path)
-    click.echo(click.style("Running command: ", fg='cyan') +
-               click.style(cmd, fg='green'))
-    os.system(cmd)
+    clicommon.run_command('config override-config-table {}'.format(
+        golden_config_path), display_cmd=True)
     return
 
 
--- a/config/main.py
+++ b/config/main.py
@@ -15,7 +15,7 @@ import itertools
 
 from collections import OrderedDict
 from generic_config_updater.generic_updater import GenericUpdater, ConfigFormat
-from minigraph import parse_device_desc_xml
+from minigraph import parse_device_desc_xml, minigraph_encoder
 from natsort import natsorted
 from portconfig import get_child_ports
 from socket import AF_INET, AF_INET6
@@ -27,7 +27,7 @@ from utilities_common.db import Db
 from utilities_common.intf_filter import parse_interface_in_filter
 from utilities_common import bgp_util
 import utilities_common.cli as clicommon
-from utilities_common.general import load_db_config
+from utilities_common.general import load_db_config, load_module_from_source
 
 from .utils import log
 
@@ -98,6 +98,9 @@ DSCP_RANGE = click.IntRange(min=0, max=63)
 TTL_RANGE = click.IntRange(min=0, max=255)
 QUEUE_RANGE = click.IntRange(min=0, max=255)
 GRE_TYPE_RANGE = click.IntRange(min=0, max=65535)
+
+# Load sonic-cfggen from source since /usr/local/bin/sonic-cfggen does not have .py extension.
+sonic_cfggen = load_module_from_source('sonic_cfggen', '/usr/local/bin/sonic-cfggen')
 
 #
 # Helper functions
@@ -1619,6 +1622,9 @@ def load_minigraph(db, no_service_restart):
                 cfggen_namespace_option = " -n {}".format(namespace)
             clicommon.run_command(db_migrator + ' -o set_version' + cfggen_namespace_option)
 
+    # Load golden_config_db.json
+    load_golden_config('/etc/sonic/golden_config_db.json')
+
     # We first run "systemctl reset-failed" to remove the "failed"
     # status from all services before we attempt to restart them
     if not no_service_restart:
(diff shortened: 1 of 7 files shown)


config/main.py
```diff
@@ -99,6 +99,9 @@
 QUEUE_RANGE = click.IntRange(min=0, max=255)
 GRE_TYPE_RANGE = click.IntRange(min=0, max=65535)
 
+# Load sonic-cfggen from source since /usr/local/bin/sonic-cfggen does not have .py extension.
+sonic_cfggen = load_module_from_source('sonic_cfggen', '/usr/local/bin/sonic-cfggen')
```
qiluo-msft, line 103: >load_module_from_source [](http://example.com/codeflow?start=15&length=23)

We can move  some useful functions to an official python module (ref: src/sonic-config-engine/setup.py) so it will be easy to reuse. This ask is in low priority, you can check this after all others resolved.

config/main.py
```diff
@@ -1619,6 +1622,9 @@ def load_minigraph(db, no_service_restart):
                 cfggen_namespace_option = " -n {}".format(namespace)
             clicommon.run_command(db_migrator + ' -o set_version' + cfggen_namespace_option)
 
+    # Load golden_config_db.json
+    load_golden_config('/etc/sonic/golden_config_db.json')
```
qiluo-msft, line 1626: >/etc/sonic/golden_config_db.json [](http://example.com/codeflow?start=24&length=32)

Make this path configurable, with a default value of this string.

qiluo-msft, line 1626: >load_golden_config [](http://example.com/codeflow?start=4&length=18)

Better name: override_config_from

config/main.py
```diff
@@ -1667,6 +1673,65 @@ def load_port_config(config_db, port_config_path):
                 port_name), display_cmd=True)
     return
 
+
+def load_golden_config(golden_config_path):
+    if not os.path.isfile(golden_config_path):
+        return
```
qiluo-msft, line 1679: This check should be outside the function. It is caller's responsible to treat missing file as failure or silently skipping.

config/main.py
```diff
@@ -1667,6 +1673,65 @@ def load_port_config(config_db, port_config_path):
...
+
+    # Override configDB with golden config
+    clicommon.run_command('config override-config-table {}'.format(
```
qiluo-msft, line 1682: >config [](http://example.com/codeflow?start=27&length=6)

This function is already inside `config`, so no need to start another process, you can call a python function.

config/main.py
```diff
@@ -1667,6 +1673,65 @@ def load_port_config(config_db, port_config_path):
...
+    return
+
+
+#
+# 'override-config-table' command ('config override-config-table ...')
+#
+@config.command('override-config-table')
+@click.argument('golden-config-db', required=True)
+@click.option(
+    '--dry_run', type=click.STRING,
```
qiluo-msft, line 1693: >STRING [](http://example.com/codeflow?start=28&length=6)

string means requiring a string after `--dry_run`.

qiluo-msft, line 1693: >_ [](http://example.com/codeflow?start=10&length=1)

Prefer `-` than `_`.

config/main.py
```diff
@@ -1667,6 +1673,65 @@ def load_port_config(config_db, port_config_path):
...
+    help="Dry run, writes config to the given file"
+)
+@clicommon.pass_db
+def override_config_table(db, golden_config_db, dry_run):
+    """Override current configDB with golden config."""
+
+    try:
+        # Load golden config json
+        golden_config_input = read_json_file(golden_config_db)
+    except Exception:
```
qiluo-msft, line 1703: >Exception [](http://example.com/codeflow?start=11&length=9)

Use specific exception types, general one will hide bugs.

config/main.py
```diff
@@ -1667,6 +1673,65 @@ def load_port_config(config_db, port_config_path):
...
+        with open(dry_run, "w+") as f:
+            json.dump(current_config, f, sort_keys=True, indent=4, cls=minigraph_encoder)
+    else:
+        # Deserialized golden config to DB recognized format
+        sonic_cfggen.FormatConverter.to_deserialized(golden_config_input)
+        # Delete table from DB then mod_config to apply golden config
+        click.echo("Removing configDB overriden table first ...")
+        for table in golden_config_input:
+            config_db.delete_table(table)
+        click.echo("Writing golden config to configDB ...")
```
qiluo-msft, line 1731: >Writing [](http://example.com/codeflow?start=20&length=7)

Overriding

qiluo-msft, line 1731: >golden [](http://example.com/codeflow?start=28&length=6)

I am considering remove `golden` advertisement from code/comment/usage text. User has the flexibility to use this CLI to achieve whatever overriding goal, and the input file may be not so "golden"

config/main.py
```diff
@@ -1667,6 +1673,65 @@ def load_port_config(config_db, port_config_path):
...
+        data = sonic_cfggen.FormatConverter.output_to_db(golden_config_input)
+        config_db.mod_config(data)
```
qiluo-msft, line 1733: >mod_config [](http://example.com/codeflow?start=18&length=10)

If user invoke this CLI directly, is it possible to warn user that no services are restarted.

config/main.py
wen587: Add print for exception
```
click.secho("Bad format: json file broken. {}".format(str(e))
```

qiluo-msft (approved):
LGTM. Let's wait a while for more eyes.

prsunny:
Can you please provide more details in the description?

For e.g, what is overridden from the golden_config? Is any config taken from the minigraph in this case?

wen587:
> Can you please provide more details in the description?
> 
> For e.g, what is overridden from the golden_config? Is any config taken from the minigraph in this case?

Updated in description. 
In short, configDB will be overriden by all golden_config_db.json 's configurations. The other Table configurations will keep the same as minigraph generated config if not listed in Golden Config,

config/main.py
```diff
@@ -1619,6 +1623,10 @@ def load_minigraph(db, no_service_restart):
                 cfggen_namespace_option = " -n {}".format(namespace)
             clicommon.run_command(db_migrator + ' -o set_version' + cfggen_namespace_option)
 
+    # Load golden_config_db.json
+    if os.path.isfile(DEFAULT_GOLDEN_CONFIG_DB_FILE):
```
judyjoseph, line 1627: @wen587, @qiluo-msft, to support multi-asic devices, to start with could we also call this API override_config_by() per namespace as well?  This will help with the pilot of this feature MACSEC_PROFILE table which is needed in the redis_db per namespace as well -- macsec/swss dockers run per asic/namespace.

Later when we design the table contents in the golden_config_db.json file, we (NDM) could name the table based on the asic/namespace name  eg: ASIC0_PORT_TABLE, ASIC1_PORT_TABLE etc - so that the respective tables will be pushed to the right redis DB in namespaces.

PR #2157, opened 2022-05-10
[config reload] Fix invalid rstrip.

ganglyu:
Signed-off-by: Gang Lv ganglv@microsoft.com

#### What I did
'rstrip' behavior is wrong, if unit is 'gnmi.timer', the result is 'gn', and 'gnmi' is expected.

#### How I did it
Use 're.sub' to replace 'rstrip'.

#### How to verify it
Run unit test for sonic-utilities.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -746,7 +746,7 @@ def _get_delayed_sonic_services():
     services = []
     for unit in timer:
         if state[timer.index(unit)] == "enabled":
-            services.append(re.sub('.timer$', '', unit, 1))
+            services.append(re.sub('\.timer$', '', unit, 1))
     return services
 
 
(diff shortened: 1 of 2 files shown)


config/main.py
```diff
@@ -746,7 +746,7 @@ def _get_delayed_sonic_services():
     services = []
     for unit in timer:
         if state[timer.index(unit)] == "enabled":
-            services.append(unit.rstrip(".timer"))
+            services.append(unit.replace(".timer", ""))
```
qiluo-msft, line 749: >replace [](http://example.com/codeflow?start=33&length=7)

Even better, use `re.sub`?

tests/config_test.py
```diff
@@ -81,10 +81,10 @@ def mock_run_command_side_effect(*args, **kwargs):
 
     if kwargs.get('return_cmd'):
         if command == "systemctl list-dependencies --plain sonic-delayed.target | sed '1d'":
-            return 'snmp.timer'
```
qiluo-msft, line 84: Keep original test case and add a new one?

qiluo-msft (approved):

PR #2173, opened 2022-05-18
[CLI] Move hostname, mgmt interface/vrf config to hostcfgd

fastiuk:
#### Why I did it
To be able to configure the management interface and hostname standalone by changing database config at runtime.
From the CLI perspective fo view, the following behavior is the same. But now you have two ways of configuring it: CLI, directly through the database.

#### How I did it
Moved configuration part of the interface and hostname to "hostcfgd".

#### How to verify it
* Built an image
* Flash it to the switch
* Run CLI commands
```
# Set IP address: verify address is set on the iface
sudo config interface ip add eth0 10.210.25.127/22 10.210.24.1
ip address show eth0
# 2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
#     link/ether 98:03:9b:a2:be:80 brd ff:ff:ff:ff:ff:ff
#     inet 10.210.25.127/22 brd 10.210.27.255 scope global eth0
#        valid_lft forever preferred_lft forever
#     inet6 fe80::9a03:9bff:fea2:be80/64 scope link
#       valid_lft forever preferred_lft forever

# Remove IP address: verify you received address form DHCP
sudo config interface ip remove eth0 10.210.25.127/22
ip address show eth0
# 2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
#     link/ether 98:03:9b:a2:be:80 brd ff:ff:ff:ff:ff:ff
#     inet 10.210.25.127/22 brd 10.210.27.255 scope global eth0
#        valid_lft forever preferred_lft forever
#     inet6 fe80::9a03:9bff:fea2:be80/64 scope link
#       valid_lft forever preferred_lft forever

# Enable/disable mgmt VRF
ip address show mgmt
# Device "mgmt" does not exist.

sudo config vrf add mgmt
ip address show mgmt
# 72: mgmt: <NOARP,MASTER,UP,LOWER_UP> mtu 65575 qdisc noqueue state UP group default qlen 1000
#     link/ether fa:9b:ad:7b:1e:83 brd ff:ff:ff:ff:ff:ff

sudo config vrf del mgmt
ip address show mgmt
# Device "mgmt" does not exist.

# Setting the hostname
admin@r-anaconda-27:~$ sudo config hostname bla
# Login / Logout
admin@bla:~$
```

Description for the changelog
* Moved management interface configuration to hostcfgd.
* Moved management VRF configuration to hostcfgd.
* Moved hostname configuration to hostcfgd.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -22,6 +22,7 @@ from socket import AF_INET, AF_INET6
 from sonic_py_common import device_info, multi_asic
 from sonic_py_common.interface import get_interface_table_name, get_port_table_name, get_intf_longname
 from utilities_common import util_base
+from swsscommon import swsscommon
 from swsscommon.swsscommon import SonicV2Connector, ConfigDBConnector
 from utilities_common.db import Db
 from utilities_common.intf_filter import parse_interface_in_filter
@@ -1889,19 +1890,11 @@ def hostname(new_hostname):
 
     config_db = ConfigDBConnector()
     config_db.connect()
-    config_db.mod_entry('DEVICE_METADATA' , 'localhost', {"hostname" : new_hostname})
-    try:
-        command = "service hostname-config restart"
-        clicommon.run_command(command, display_cmd=True)
-    except SystemExit as e:
-        click.echo("Restarting hostname-config  service failed with error {}".format(e))
-        raise
+    config_db.mod_entry(swsscommon.CFG_DEVICE_METADATA_TABLE_NAME, 'localhost',
+                        {'hostname': new_hostname})
 
-    # Reload Monit configuration to pick up new hostname in case it changed
-    click.echo("Reloading Monit configuration ...")
-    clicommon.run_command("sudo monit reload")
-
-    click.echo("Please note loaded setting will be lost after system reboot. To preserve setting, run `config save`.")
+    click.echo('Please note loaded setting will be lost after system reboot. To'
+               ' preserve setting, run `config save`.')
 
 #
 # 'synchronous_mode' command ('config synchronous_mode ...')
--- a/config/main.py
+++ b/config/main.py
@@ -2820,22 +2820,6 @@ def warm_restart_bgp_eoiu(ctx, enable):
     db = ctx.obj['db']
     db.mod_entry('WARM_RESTART', 'bgp', {'bgp_eoiu': enable})
 
-def mvrf_restart_services():
-    """Restart interfaces-config service and NTP service when mvrf is changed"""
-    """
-    When mvrf is enabled, eth0 should be moved to mvrf; when it is disabled,
-    move it back to default vrf. Restarting the "interfaces-config" service
-    will recreate the /etc/network/interfaces file and restart the
-    "networking" service that takes care of the eth0 movement.
-    NTP service should also be restarted to rerun the NTP service with or
-    without "cgexec" accordingly.
-    """
-    cmd="service ntp stop"
-    os.system (cmd)
-    cmd="systemctl restart interfaces-config"
-    os.system (cmd)
-    cmd="service ntp start"
-    os.system (cmd)
 
 def vrf_add_management_vrf(config_db):
     """Enable management vrf in config DB"""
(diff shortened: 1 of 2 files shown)


dgsudharsan (changes requested):
Please add UT for the hostname add. Else the coverage will fail.

liat-grozovik:
@fastiuk when do you think UT will be avaialble so we can review and signoff?

fastiuk:
> @fastiuk when do you think UT will be avaialble so we can review and signoff?

Done

dgsudharsan (dismissed):

fastiuk:
@dgsudharsan Please take a look at this one as well. It is part of the same changes: https://github.com/Azure/sonic-buildimage/pull/10868

config/main.py
```diff
@@ -1755,19 +1756,11 @@ def hostname(new_hostname):
...
-        raise
+    config_db.mod_entry(swsscommon.CFG_DEVICE_METADATA_TABLE_NAME, 'localhost',
+                        {'hostname': new_hostname})
 
-    # Reload Monit configuration to pick up new hostname in case it changed
-    click.echo("Reloading Monit configuration ...")
-    clicommon.run_command("sudo monit reload")
-
-    click.echo("Please note loaded setting will be lost after system reboot. To preserve setting, run `config save`.")
+    click.echo('Please note loaded setting will be lost after system reboot. To'
```
qiluo-msft, line 1762: >( [](http://example.com/codeflow?start=14&length=1)

Could you use `click.wrap_text()` to wrap long line instead of manually wrap?

config/main.py
```diff
@@ -2671,22 +2664,6 @@ def warm_restart_bgp_eoiu(ctx, enable):
     db = ctx.obj['db']
     db.mod_entry('WARM_RESTART', 'bgp', {'bgp_eoiu': enable})
 
-def mvrf_restart_services():
```
prsunny, line 2674: I see that the code to restart services are moved to hostcfgd. What is the benefit that we are getting by this approach? What is missing in the current approach?

prsunny (changes requested):

config/main.py
fastiuk: I am wrapping just the code string here. The log line in runtime will be the same as it was before this change.

config/main.py
prsunny: I don't think that is expected to be supported. We use CLI to configure the DB and there are multiple validations done at the CLI level (For e.g, validate IP address corrrectness etc). So why should not we use the CLI here?

fastiuk: I disagree.
Example: SONiC has the ability to configure syslog servers. It has a dedicated table for it but doesn't have cli to do such configuration. That means that the only way to configure it is through the DB.
Another example: What if you don't need CLI at all? You have SNMP server and it configures your switch. It won't be possible if commands which are required to make configuration reside here in cli.

prsunny: afaik, there is CLI for syslog, you can also refer this. https://github.com/Azure/sonic-utilities/pull/2191. The point is, if there is a need for configuring after the initial config, then it should ideally have a CLI. We don't need CLI for those configs that can be applied at startup (via config_db.json or equivalent..) and not to be changed later.

fastiuk: > The point is, if there is a need for configuring after the initial config, then it should ideally have a CLI.

Agree. But we can do it without CLI as well, right? It is not prohibited, SONiC is an open Linux system, so why a user can't just use it the way it wants?

> We don't need CLI for those configs that can be applied at startup (via config_db.json or equivalent..)

Agree, so that is why I proposing those changes. It allows users to have yet another way to configure its switch. The CLI is still here, so you can still configure your switch via CLI (And most users will do so).

prsunny (approved):
Discussed offline on the use-case. Signing-off from my side.

qiluo-msft (approved):

PR #2174, opened 2022-05-18
[GCU] Handling non-compliant leaf-list with string values

### LINKED ISSUE #2170: [GCU] Handle non-compliant leaf-list with string values
#### Description
sonic_yang_mgmt follows these [guidelines](https://github.com/Azure/SONiC/blob/master/doc/mgmt/SONiC_YANG_Model_Guidelines.md) in order to convert from `config_db` to `config_yang` and vice versa.
in GCU we rely on the same [guidelines](https://github.com/Azure/SONiC/blob/master/doc/mgmt/SONiC_YANG_Model_Guidelines.md) to convert from `config_db path` to `config_yang xpath` and vice versa.

There were modifications to how sonic_yang_mgmt converts data recently, and these changes are not supported by gcu, the changes are the following:
- Supporting `Non compliant leaf list in config_db schema` https://github.com/Azure/sonic-buildimage/pull/10291

1.
2.
3.

#### Output of `show version`

```
(paste your output here)
```

ghooo:
#### What I did
Fixes #2170

Converting xpath to path:
- There is only 1 case where the path is referring to the list itself. Example:
```
path="/BUFFER_PORT_EGRESS_PROFILE_LIST/Ethernet9/profile_list",
xpath="/sonic-buffer-port-egress-profile-list:sonic-buffer-port-egress-profile-list/BUFFER_PORT_EGRESS_PROFILE_LIST/BUFFER_PORT_EGRESS_PROFILE_LIST_LIST[port='Ethernet9']/profile_list",
```
- There is no other case as we the list is just a single string, and we cannot refer to individual elements in string in ConfigDb.

Converting path to xpath, there are 2 cases:
- Xpath is pointing to leaf-list as a whole, return path of whole list
```
xpath="/sonic-buffer-port-egress-profile-list:sonic-buffer-port-egress-profile-list/BUFFER_PORT_EGRESS_PROFILE_LIST/BUFFER_PORT_EGRESS_PROFILE_LIST_LIST[port='Ethernet9']/profile_list",
path="/BUFFER_PORT_EGRESS_PROFILE_LIST/Ethernet9/profile_list",
```
- Xpath is pointing to an element of the leaf-list, return path of whole list as well since it is not possible to point to specific element -- here we are pointing to the element `egress_lossy_profile`
```
xpath="/sonic-buffer-port-egress-profile-list:sonic-buffer-port-egress-profile-list/BUFFER_PORT_EGRESS_PROFILE_LIST/BUFFER_PORT_EGRESS_PROFILE_LIST_LIST[port='Ethernet9']/profile_list[.='egress_lossy_profile']",
path="/BUFFER_PORT_EGRESS_PROFILE_LIST/Ethernet9/profile_list",
```

#### How I did it
Only a single change if the xpath is pointing to a specific element, just return whole list token from ConfigDb format.

Please note this solution is future proof, if ConfigDb changes how a list is saved as a string or json list, we don't care.

#### How to verify it
unit-test

### DIFF
--- a/generic_config_updater/gu_common.py
+++ b/generic_config_updater/gu_common.py
@@ -757,6 +757,12 @@ class PathAddressing:
             #   path: /BUFFER_PORT_EGRESS_PROFILE_LIST/Ethernet9/profile_list
             if isinstance(list_config, str):
                 return [leaf_list_name]
+
+            if not isinstance(list_config, list):
+                raise ValueError(f"list_config is expected to be of type list or string. Found {type(list_config)}.\n  " + \
+                                 f"model: {model}\n  token_index: {token_index}\n  " + \
+                                 f"xpath_tokens: {xpath_tokens}\n  config: {config}"))
+
             list_idx = list_config.index(leaf_list_value)
             return [leaf_list_name, list_idx]
 
--- a/generic_config_updater/gu_common.py
+++ b/generic_config_updater/gu_common.py
@@ -761,7 +761,7 @@ class PathAddressing:
             if not isinstance(list_config, list):
                 raise ValueError(f"list_config is expected to be of type list or string. Found {type(list_config)}.\n  " + \
                                  f"model: {model}\n  token_index: {token_index}\n  " + \
-                                 f"xpath_tokens: {xpath_tokens}\n  config: {config}"))
+                                 f"xpath_tokens: {xpath_tokens}\n  config: {config}")
 
             list_idx = list_config.index(leaf_list_value)
             return [leaf_list_name, list_idx]
--- a/generic_config_updater/gu_common.py
+++ b/generic_config_updater/gu_common.py
@@ -750,6 +750,13 @@ class PathAddressing:
             # leaf_list_name = match.group(1)
             leaf_list_value = match.group(1)
             list_config = config[leaf_list_name]
+            # Workaround for those fields who is defined as leaf-list in YANG model but have string value in config DB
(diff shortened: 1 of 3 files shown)


generic_config_updater/gu_common.py
```diff
@@ -750,6 +750,13 @@ def _get_path_tokens_from_leaf(self, model, token_index, xpath_tokens, config):
...
             leaf_list_value = match.group(1)
             list_config = config[leaf_list_name]
+            # Workaround for those fields who is defined as leaf-list in YANG model but have string value in config DB
+            # No need to lookup the item index in ConfigDb since the list is represented as a string, return path to string immediately
+            # Example:
+            #   xpath: /sonic-buffer-port-egress-profile-list:sonic-buffer-port-egress-profile-list/BUFFER_PORT_EGRESS_PROFILE_LIST/BUFFER_PORT_EGRESS_PROFILE_LIST_LIST[port='Ethernet9']/profile_list[.='egress_lossy_profile']
+            #   path: /BUFFER_PORT_EGRESS_PROFILE_LIST/Ethernet9/profile_list
+            if isinstance(list_config, str):
+                return [leaf_list_name]
             list_idx = list_config.index(leaf_list_value)
```
qiluo-msft, line 760: >list_config [](http://example.com/codeflow?start=23&length=11)

Do you want to add `else if` block? What is the expected type?

ghooo: The return type is a list of tokens (strings). I don't like adding else to an if that has `return`. I prefer to avoid indentation when possible.

PR #2175, opened 2022-05-19
[dump] Optimized dump state cli and modified tests to not use common data 

vivekreddynv:
#### Why I did

**Optimize Dump State Cli:**
With the optimizations made, an entire run of "dump state" will only create 1 `Sonicv2Connector `Object and only call `connect` method once per db. i.e one `DBConnector `Object per DB

**Test Stability:** 
Since the dump state tests are based on data (key and field-value pairs) any changes to those tables will lead to failure in dump state tests. Thus migrated the data_sources for dump_state_test & match_engine_tests from mock_tables to dump_input/ thus making them more stable.

#### How to verify it

```
vkarri@230fae5c3278:/sonic/src/sonic-utilities$ pytest-3 tests/dump_tests/
========================================= test session starts ==========================================
platform linux -- Python 3.9.2, pytest-6.0.2, py-1.10.0, pluggy-0.13.0
rootdir: /sonic/src/sonic-utilities/tests, configfile: pytest.ini
plugins: pyfakefs-4.5.6, cov-2.10.1
collected 140 items                                                                                    

tests/dump_tests/dump_state_test.py .............                                                [  9%]
tests/dump_tests/match_engine_test.py .................................                          [ 32%]
tests/dump_tests/module_tests/copp_test.py ..........                                            [ 40%]
tests/dump_tests/module_tests/evpn_test.py ......                                                [ 44%]
tests/dump_tests/module_tests/fdb_test.py ............                                           [ 52%]
tests/dump_tests/module_tests/interface_test.py ..........                                       [ 60%]
tests/dump_tests/module_tests/port_test.py ......                                                [ 64%]
tests/dump_tests/module_tests/portchannel_member_test.py ...                                     [ 66%]
tests/dump_tests/module_tests/portchannel_test.py ....                                           [ 69%]
tests/dump_tests/module_tests/route_test.py ..........                                           [ 76%]
tests/dump_tests/module_tests/vlan_test.py .....................                                 [ 91%]
tests/dump_tests/module_tests/vxlan_tunnel_map_test.py .....                                     [ 95%]
tests/dump_tests/module_tests/vxlan_tunnel_test.py .......                                       [100%]

========================================= 140 passed in 1.04s ==========================================
```


dgsudharsan (approved):

dump/main.py
```diff
@@ -146,19 +145,20 @@ def filter_out_dbs(db_list, collected_info):
...
     db_cfg_file = JsonSource()
-    db_conn = ConnectionPool().initialize_connector(namespace)
     for db_name in all_dbs:
         if db_name == "CONFIG_FILE":
             db_cfg_file.connect(plugins.dump_modules[module].CONFIG_FILE, namespace)
         else:
-            db_conn.connect(db_name)
+            conn_pool.get(db_name, namespace)
+    
+    db_conn = conn_pool.cache.get(namespace, {}).get('conn', None)
```
SuvarnaMeenakshi, line 161: Instead of using 'conn' string, will be good to define a string and use it.
The same for the keys in ConnectionPool like "connected_to" and "conn", as it is used in multiple locations and will be hard to keep track.

tests/dump_tests/match_engine_test.py
```diff
@@ -182,75 +217,60 @@ def test_just_keys_false(self):
...
-        from ..mock_tables import dbconnector
-        dbconnector.load_namespace_config()
-
-    def teardown_class(cls):
-        print("TEARDOWN")
-        os.environ["UTILITIES_UNIT_TESTING"] = "0"
-        os.environ["UTILITIES_UNIT_TESTING_TOPOLOGY"] = ""
-
-    def test_namespace_asic0(self):
+    def test_namespace_asic0(self, match_engine):
```
SuvarnaMeenakshi, line 257: How is the config_db data from namespace fetched if  setup is removed to load namespace config?

SuvarnaMeenakshi: How is match_engine toggling between single namespace and mulitple namespace scenario?

vivekreddynv: I'm pre-populating ConnectionPool based on namespace as the key ahead of time and using it to construct MatchEngine. 

Once the Sonicv2Connector Objects are cached in the ConnectionPool, MatchEngine doesn't re-initialize them but will rather use the existing ones and thus there is no different b/w both the scenarios.

SuvarnaMeenakshi: so this testing will be checking data in DEFAULT_NAMESPACE and a specific namespace like namepsace asic0, which will be like a multi-asic DUT.
Earlier, the unit-test was testing on DEFAULT_NAMESPACE of single asic DUT and asic0 of multi-asic DUT. correct?

vivekreddynv: Yeah, based on the ns argument provided to MatchRequest. The differentiation is actually made/required in dump_state_test.py (These tests invoke the command using the )

There you could see i've used different fixtures for masic and single asic

dump/main.py
vivekreddynv: Updated

SuvarnaMeenakshi (approved):
lgtm

PR #2181, opened 2022-05-23
[subinterface]Avoid removing the subinterface when last configured ip is removed

dgsudharsan:
Signed-off-by: Sudharsan Dhamal Gopalarathnam <sudharsand@nvidia.com>

#### What I did
Avoid removing the subinterface when last configured ip is removed

#### How I did it
In case of other interface types like INTERFACE, VLAN_INTERFACE, PORTCHANNEL_INTERFACE, when the last ip address is removed the CLI implicitly delete the interface table entry of type INTERFACE|intf_name. However this logic doesn't apply to VLAN_SUB_INTERFACE since the creation and removal will be controlled by config subinterface command

Without this fix the subinterface gets removed and show subinterface status will not list the subinterface.

#### How to verify it
Added UT to verify it.


keboliu (approved):

liat-grozovik:
@prsunny could you please review and signoff?

prsunny (approved):

PR #2182, opened 2022-05-23
Warm/Fast reboot: Remove duplicate kernel options

vaibhavhd:
#### What I did

Continuous warm upgrades fail at one point on Arista platforms.

The failure is seen during the step "Load kernel into the memory", where below error is seen:

```
May 19 10:16:33.530520 str-7260cx3-acs-2 NOTICE admin: warm-reboot failure (255) cleanup ...
May 19 10:16:33.745967 str-7260cx3-acs-2 NOTICE admin: Cancel warm-reboot: code (0)
'Cannot load /host/image-20201231.66/boot/vmlinuz-4.19.0-12-2-amd64\\n'
```

Failure is because due to continuous upgrades (cleanup prev image included after every boot), `kernel-params-base` file keeps growing. This is because this file is stored at /host/ and is retained after every boot up. Subsequent sonic-installer adds to the list without checking the already existing options.

#### How I did it

In the warmboot path, make sure that the `BOOT_OPTIONS` used for kernel load are unique and hence remove duplicate options, if present.

There should also be a corresponding fix in sonic-installer or Arista platform scripts which generate these duplicate options into `kernel-params-base` file.
#### How to verify it
Tested on physical testbed. Continuous warm upgrades don't fail with this error anymore.

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -365,6 +365,9 @@ function setup_reboot_variables()
         local fstype=$(blkid -o value -s TYPE ${sonic_dev})
         BOOT_OPTIONS="${BOOT_OPTIONS} ssd-upgrader-part=${sonic_dev},${fstype}"
     fi
+    BOOT_OPTIONS_ORIG=(${BOOT_OPTIONS})
+    BOOT_OPTIONS_UNIQ=($(printf "%s\n" "${BOOT_OPTIONS_ORIG[@]}" | awk '!seen[$0]++'))
+    BOOT_OPTIONS="${BOOT_OPTIONS_UNIQ[*]}"
 }
 
 function check_docker_exec()


yxieca (approved):

Staphylo:
@vaibhavhd  This is possibly a bug in the `boot0` script.
Could you share which release you are experiencing this issue on?
I believe I merged a fix on master at some point but it might have never gone into older releases.

vaibhavhd:
> @vaibhavhd This is possibly a bug in the `boot0` script. Could you share which release you are experiencing this issue on? I believe I merged a fix on master at some point but it might have never gone into older releases.

Hi @Staphylo,  I tried 201811 and 202012 and experienced issues on both. I also agree this is unrelated to warmboot script, but has impacted warmboot nevertheless.

Staphylo:
@vaibhavhd I opened https://github.com/Azure/sonic-buildimage/pull/11161 to handle this duplicate parameter situation.
I believe this PR will no longer be necessary.

vaibhavhd:
Closing in favor of a more generic fix at https://github.com/Azure/sonic-buildimage/pull/11161

PR #2183, opened 2022-05-24
Fix sonic-installer and 'show version' command crash when database docker not running issue.

liuh-80:
Fix sonic-installer and 'show version' command crash when database docker not running issue.

#### Description
    Global variable utilies_common.cli.iface_alias_converter will connect to config DB when initialize itself. If database docker not running, the initialize code will crash. 
    To fix this issue, change the variable to lazy initialize variable with lazy_object_proxy.

#### Motivation and Context

Fix this issue: https://github.com/Azure/sonic-buildimage/issues/10434

#### How Has This Been Tested?
    Pass all UT and sonic-buildimage E2E test.


liuh-80:
Code change in this PR will validate and pass E2E test with this: https://github.com/Azure/sonic-buildimage/pull/10909

utilities_common/cli.py
```diff
@@ -183,8 +184,11 @@ def alias_to_name(self, interface_alias):
         # interface_alias not in port_dict. Just return interface_alias
         return interface_alias if sub_intf_sep_idx == -1 else interface_alias + VLAN_SUB_INTERFACE_SEPARATOR + vlan_id
 
-# Global class instance for SONiC interface name to alias conversion
-iface_alias_converter = InterfaceAliasConverter()
+def GetInterfaceAliasConverter():
```
qiluo-msft, line 187: >GetInterfaceAliasConverter [](http://example.com/codeflow?start=4&length=26)

`Get` is not proper verb for your intention. If user keep `Get`, it actually like creating many converters.

For this common usage, suggest use lambda instead of a new function.

liuh-80: Fixed, use lambda and remove this method.

liuh-80:
Test with build image failed, not caused by change in this PR, seems fixed in master branch, will merge latest code and test again.

qiluo-msft (approved):

qiluo-msft:
Could you update PR description?
1. "Remove global variable iface_alias_converter" this is not true
2. Please share more insight what is wrong with old code, since this is a fix.

stepanblyschak:
@qiluo-msft @liuh-80 Is there going to be a UT covering this fix ?

qiluo-msft:
@liuh-80 Please add unit test for this fix.

liuh-80:
> @liuh-80 Please add unit test for this fix.

Here is UT PR: https://github.com/Azure/sonic-mgmt/pull/5722

stepanblyschak:
@liuh-80 @qiluo-msft Could you please cherry pick this PR to 2205?

PR #2190, opened 2022-05-31
YANG validation for ConfigDB Updates: portchannel add/remove, loopback interface, VLAN YANG validation Using GCU

isabelmsft:
#### What I did
Added YANG validation to add/remove portchannel using GCU apply patch

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -14,6 +14,7 @@ import time
 import itertools
 import copy
 
+from jsonpatch import JsonPatchConflict
 from collections import OrderedDict
 from generic_config_updater.generic_updater import GenericUpdater, ConfigFormat
 from minigraph import parse_device_desc_xml, minigraph_encoder
@@ -31,6 +32,7 @@ from utilities_common import bgp_util
 import utilities_common.cli as clicommon
 from utilities_common.helper import get_port_pbh_binding, get_port_acl_binding
 from utilities_common.general import load_db_config, load_module_from_source
+from .validated_config_db_connector import ValidatedConfigDBConnector
 import utilities_common.multi_asic as multi_asic_util
 
 from .utils import log
@@ -104,6 +106,7 @@ DSCP_RANGE = click.IntRange(min=0, max=63)
 TTL_RANGE = click.IntRange(min=0, max=255)
 QUEUE_RANGE = click.IntRange(min=0, max=255)
 GRE_TYPE_RANGE = click.IntRange(min=0, max=65535)
+ADHOC_VALIDATION = True
 
 # Load sonic-cfggen from source since /usr/local/bin/sonic-cfggen does not have .py extension.
 sonic_cfggen = load_module_from_source('sonic_cfggen', '/usr/local/bin/sonic-cfggen')
@@ -2031,51 +2034,64 @@ def portchannel(db, ctx, namespace):
 @click.pass_context
 def add_portchannel(ctx, portchannel_name, min_links, fallback, fast_rate):
     """Add port channel"""
-    if is_portchannel_name_valid(portchannel_name) != True:
-        ctx.fail("{} is invalid!, name should have prefix '{}' and suffix '{}'"
-                 .format(portchannel_name, CFG_PORTCHANNEL_PREFIX, CFG_PORTCHANNEL_NO))
-
-    db = ctx.obj['db']
-
-    if is_portchannel_present_in_db(db, portchannel_name):
-        ctx.fail("{} already exists!".format(portchannel_name))
-
+    
     fvs = {
         'admin_status': 'up',
         'mtu': '9100',
         'lacp_key': 'auto',
         'fast_rate': fast_rate.lower(),
     }
+
     if min_links != 0:
         fvs['min_links'] = str(min_links)
     if fallback != 'false':
         fvs['fallback'] = 'true'
-    db.set_entry('PORTCHANNEL', portchannel_name, fvs)
-
+    
+    if ADHOC_VALIDATION:
+        db = ctx.obj['db']
+        if is_portchannel_name_valid(portchannel_name) != True:
+            ctx.fail("{} is invalid!, name should have prefix '{}' and suffix '{}'"
+                    .format(portchannel_name, CFG_PORTCHANNEL_PREFIX, CFG_PORTCHANNEL_NO))
+        if is_portchannel_present_in_db(db, portchannel_name):
+            ctx.fail("{} already exists!".format(portchannel_name)) # TODO: MISSING CONSTRAINT IN YANG MODEL
+    else:
+        db = ValidatedConfigDBConnector(ctx.obj['db']) 
+    
+    try:
+        db.set_entry('PORTCHANNEL', portchannel_name, fvs)
+    except ValueError:
+        ctx.fail("{} is invalid!, name should have prefix '{}' and suffix '{}'".format(portchannel_name, CFG_PORTCHANNEL_PREFIX, CFG_PORTCHANNEL_NO))
+ 
 @portchannel.command('del')
 @click.argument('portchannel_name', metavar='<portchannel_name>', required=True)
 @click.pass_context
 def remove_portchannel(ctx, portchannel_name):
     """Remove port channel"""
-    if is_portchannel_name_valid(portchannel_name) != True:
-        ctx.fail("{} is invalid!, name should have prefix '{}' and suffix '{}'"
-                 .format(portchannel_name, CFG_PORTCHANNEL_PREFIX, CFG_PORTCHANNEL_NO))
+    
+    if ADHOC_VALIDATION:
+        db = ctx.obj['db']
+        if is_portchannel_name_valid(portchannel_name) != True:
+            ctx.fail("{} is invalid!, name should have prefix '{}' and suffix '{}'"
+                    .format(portchannel_name, CFG_PORTCHANNEL_PREFIX, CFG_PORTCHANNEL_NO))
 
-    db = ctx.obj['db']
+        # Don't proceed if the port channel does not exist
+        if is_portchannel_present_in_db(db, portchannel_name) is False:
+            ctx.fail("{} is not present.".format(portchannel_name))
 
-    # Dont proceed if the port channel does not exist
-    if is_portchannel_present_in_db(db, portchannel_name) is False:
-        ctx.fail("{} is not present.".format(portchannel_name))
+        # Dont let to remove port channel if vlan membership exists
+        for k,v in db.get_table('VLAN_MEMBER'): # TODO: MISSING CONSTRAINT IN YANG MODEL
+            if v == portchannel_name:
+                ctx.fail("{} has vlan {} configured, remove vlan membership to proceed".format(portchannel_name, str(k)))
 
-    # Dont let to remove port channel if vlan membership exists
-    for k,v in db.get_table('VLAN_MEMBER'):
-        if v == portchannel_name:
-            ctx.fail("{} has vlan {} configured, remove vlan membership to proceed".format(portchannel_name, str(k)))
-
-    if len([(k, v) for k, v in db.get_table('PORTCHANNEL_MEMBER') if k == portchannel_name]) != 0:
-        click.echo("Error: Portchannel {} contains members. Remove members before deleting Portchannel!".format(portchannel_name))
+        if len([(k, v) for k, v in db.get_table('PORTCHANNEL_MEMBER') if k == portchannel_name]) != 0: # TODO: MISSING CONSTRAINT IN YANG MODEL
+            ctx.fail("Error: Portchannel {} contains members. Remove members before deleting Portchannel!".format(portchannel_name))
     else:
+        db = ValidatedConfigDBConnector(ctx.obj['db'])
+    
+    try:
         db.set_entry('PORTCHANNEL', portchannel_name, None)
+    except JsonPatchConflict:
+        ctx.fail("{} is not present.".format(portchannel_name))
 
 @portchannel.group(cls=clicommon.AbbreviationGroup, name='member')
 @click.pass_context
@@ -2104,8 +2120,8 @@ def add_portchannel_member(ctx, portchannel_name, port_name):
     # Dont proceed if the port channel does not exist
     if is_portchannel_present_in_db(db, portchannel_name) is False:
         ctx.fail("{} is not present.".format(portchannel_name))
-
-    # Dont allow a port to be member of port channel if it is configured with an IP address
+ 
+    # Don't allow a port to be member of port channel if it is configured with an IP address
     for key,value in db.get_table('INTERFACE').items():
         if type(key) == tuple:
             continue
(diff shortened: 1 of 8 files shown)


config/main.py
```diff
@@ -1893,43 +1893,49 @@ def portchannel(db, ctx, namespace):
...
+
+    gcu_json = {"op": "add",
+                "path": "/PORTCHANNEL/{}".format(portchannel_name)}
+    gcu_json["value"] = fvs
+    gcu_json_input.append(gcu_json)
+    gcu_patch = jsonpatch.JsonPatch(gcu_json_input)
+    format = ConfigFormat.CONFIGDB.name
+    config_format = ConfigFormat[format.upper()]
+    GenericUpdater().apply_patch(patch=gcu_patch, config_format=config_format, verbose=False, dry_run=False, ignore_non_yang_tables=False, ignore_paths=None)
+
```
qiluo-msft, line 1918: Encapsulate GCU implementation into a ConfigDB decorator class. So user could still use `db.set_entry('PORTCHANNEL', portchannel_name, fvs)`

config/main.py
```diff
@@ -1893,38 +1892,30 @@ def portchannel(db, ctx, namespace):
...
     if is_portchannel_present_in_db(db, portchannel_name):
         ctx.fail("{} already exists!".format(portchannel_name))
+    
+    db = ValidatedConfigDBConnector()
+    db.connect()
 
-    fvs = {'admin_status': 'up',
-           'mtu': '9100',
-           'lacp_key': 'auto'}
+    fvs = {"admin_status": "up",
```
qiluo-msft, line 1902: >fvs [](http://example.com/codeflow?start=4&length=3)

Seems like most changes are `'` -> `"`. Not necessary anymore.

config/main.py
```diff
@@ -1893,38 +1892,30 @@ def portchannel(db, ctx, namespace):
...
 @portchannel.command('del')
 @click.argument('portchannel_name', metavar='<portchannel_name>', required=True)
 @click.pass_context
 def remove_portchannel(ctx, portchannel_name):
     """Remove port channel"""
-    if is_portchannel_name_valid(portchannel_name) != True:
-        ctx.fail("{} is invalid!, name should have prefix '{}' and suffix '{}'"
-                 .format(portchannel_name, CFG_PORTCHANNEL_PREFIX, CFG_PORTCHANNEL_NO))
-
-    db = ctx.obj['db']
```
qiluo-msft, line 1923: >ctx.obj['db'] [](http://example.com/codeflow?start=9&length=13)

We need to use `ctx.obj['db']` inside ValidatedConfigDBConnector implementation, and the vision is move the decorating to ctx itself, so every place referring `ctx.obj['db']` will benefit with zero code change.

config/validated_config_db_connector.py
```diff
@@ -0,0 +1,24 @@
+import jsonpatch
+
+from swsscommon.swsscommon import SonicV2Connector, ConfigDBConnector
+from generic_config_updater.generic_updater import GenericUpdater, ConfigFormat
+
+class ValidatedConfigDBConnector(ConfigDBConnector):
```
qiluo-msft, line 6: >ConfigDBConnector [](http://example.com/codeflow?start=33&length=17)

The base class is not used in the implementation. Decorator pattern makes more sense.

isabelmsft: Changed to decorator pattern

config/main.py
```diff
@@ -1893,43 +1892,35 @@ def portchannel(db, ctx, namespace):
...
 def add_portchannel(ctx, portchannel_name, min_links, fallback):
     """Add port channel"""
-    if is_portchannel_name_valid(portchannel_name) != True:
-        ctx.fail("{} is invalid!, name should have prefix '{}' and suffix '{}'"
-                 .format(portchannel_name, CFG_PORTCHANNEL_PREFIX, CFG_PORTCHANNEL_NO))
-
+    
     db = ctx.obj['db']
-
+    validated_db = ValidatedConfigDBConnector(db)
```
qiluo-msft, line 1897: >validated_db = ValidatedConfigDBConnector(db) [](http://example.com/codeflow?start=4&length=45)

To be even cleaner, you can combine 2 lines into 1.
```
db = ValidatedConfigDBConnector(ctx.obj['db'])
```

config/main.py
```diff
@@ -1893,43 +1892,35 @@ def portchannel(db, ctx, namespace):
...
 
     db = ctx.obj['db']
-
-    # Dont proceed if the port channel does not exist
-    if is_portchannel_present_in_db(db, portchannel_name) is False:
-        ctx.fail("{} is not present.".format(portchannel_name))
+    validated_db = ValidatedConfigDBConnector(db)
 
     if len([(k, v) for k, v in db.get_table('PORTCHANNEL_MEMBER') if k == portchannel_name]) != 0:
         click.echo("Error: Portchannel {} contains members. Remove members before deleting Portchannel!".format(portchannel_name))
```
qiluo-msft, line 1921: >Portchannel {} contains members [](http://example.com/codeflow?start=27&length=31)

Can we depend yang model to validate this? If capable, you can remove this if branch.

isabelmsft: The check exists, though the error message is not intuitive: "Data Loading failed, points to a non-existing leaf." 
I can remove this if branch for now and clarify the error messages at a later point

config/main.py
```diff
@@ -1893,43 +1892,33 @@ def portchannel(db, ctx, namespace):
...
-    # Dont proceed if the port channel does not exist
-    if is_portchannel_present_in_db(db, portchannel_name) is False:
-        ctx.fail("{} is not present.".format(portchannel_name))
+    db = ValidatedConfigDBConnector(ctx.obj['db'])
 
     if len([(k, v) for k, v in db.get_table('PORTCHANNEL_MEMBER') if k == portchannel_name]) != 0:
         click.echo("Error: Portchannel {} contains members. Remove members before deleting Portchannel!".format(portchannel_name))
     else:
-        db.set_entry('PORTCHANNEL', portchannel_name, None)
+        validated_db.set_entry('PORTCHANNEL', portchannel_name, None)
```
qiluo-msft, line 1921: >validated_db [](http://example.com/codeflow?start=8&length=12)

validated_db -> db

qiluo-msft:
Please fix LGTM alert.

config/main.py
```diff
@@ -1893,38 +1894,35 @@ def portchannel(db, ctx, namespace):
...
     """Add port channel"""
-    if is_portchannel_name_valid(portchannel_name) != True:
-        ctx.fail("{} is invalid!, name should have prefix '{}' and suffix '{}'"
-                 .format(portchannel_name, CFG_PORTCHANNEL_PREFIX, CFG_PORTCHANNEL_NO))
+    
+    error_map = {}
+    error_map[ValueError] = "{} is invalid!, name should have prefix '{}' and suffix '{}'".format(portchannel_name, CFG_PORTCHANNEL_PREFIX, CFG_PORTCHANNEL_NO)
 
-    db = ctx.obj['db']
+    db = validate(ctx.obj['db'], error_map, ctx)
```
qiluo-msft, line 1901: >validate [](http://example.com/codeflow?start=9&length=8)

revert this part.

config/validated_config_db_connector.py
```diff
@@ -0,0 +1,32 @@
+import jsonpatch
+
+from generic_config_updater.generic_updater import GenericUpdater, ConfigFormat
+
+def validate(config_db_connector, error_map, ctx=None):
```
qiluo-msft, line 5: >validate [](http://example.com/codeflow?start=4&length=8)

`ValidatedConfigDBConnector` name is better for a function name.

qiluo-msft: This is still applicable. Just for function name.

qiluo-msft:
> def is_loopback_name_valid(loopback_name):

Remove unused function.

---
Refers to: config/main.py:5898 in 8829d96. [](commit_id = 8829d96e4569920d19eec6259aaa119a8ec413c2, deletion_comment = False)

config/validated_config_db_connector.py
```diff
@@ -0,0 +1,37 @@
...
+    if value is not None:
+        op = "add"
+    else:
+        op = "remove"
+    path, value = make_path_value_jsonpatch_compatible(table, key, value) 
+
+    gcu_json_input = []
+    gcu_json = {"op": "{}".format(op),
+                "path": "{}".format(path)}
+    if op is "add":
```
qiluo-msft, line 29: >is [](http://example.com/codeflow?start=10&length=2)

Use `==` for equality testing.

config/validated_config_db_connector.py
```diff
@@ -0,0 +1,37 @@
...
+from generic_config_updater.generic_updater import GenericUpdater, ConfigFormat
+
+def ValidatedConfigDBConnector(config_db_connector):
+    config_db_connector.set_entry = validated_set_entry
+    return config_db_connector
+
+def make_path_value_jsonpatch_compatible(table, key, value):
+    if type(key) == tuple:
+        path = JsonPointer.from_parts([table, '|'.join(key)]).path
+    else:
```
qiluo-msft, line 13: >else [](http://example.com/codeflow?start=4&length=4)

In this branch, you can also leverage `JsonPointer.from_parts`, and actually compare with above block, you can do fine granularity if/else.

config/validated_config_db_connector.py
```diff
@@ -0,0 +1,37 @@
+import jsonpatch
+from jsonpointer import JsonPointer
+
+from generic_config_updater.generic_updater import GenericUpdater, ConfigFormat
+
+def ValidatedConfigDBConnector(config_db_connector):
+    config_db_connector.set_entry = validated_set_entry
+    return config_db_connector
+
+def make_path_value_jsonpatch_compatible(table, key, value):
```
qiluo-msft, line 10: >make_path_value_jsonpatch_compatible [](http://example.com/codeflow?start=4&length=36)

This will be a common function. Suggest add enough unit test cases.

config/validated_config_db_connector.py
```diff
@@ -0,0 +1,37 @@
+import jsonpatch
+from jsonpointer import JsonPointer
+
+from generic_config_updater.generic_updater import GenericUpdater, ConfigFormat
+
+def ValidatedConfigDBConnector(config_db_connector):
```
qiluo-msft, line 6: As we discussed on design, we need to check global disable/enable option. So suggest add the check in this function.

qiluo-msft: Sorry I did not mean `fallback_adhoc_validation`. I mean `yang_config_validation` you added in ConfigDB. You may check @wen587 PR [https://github.com/sonic-net/sonic-buildimage/pull/11715/files](https://github.com/sonic-net/sonic-buildimage/pull/11715/files)

config/main.py
```diff
@@ -1893,43 +1895,53 @@ def portchannel(db, ctx, namespace):
...
-    if is_portchannel_name_valid(portchannel_name) != True:
-        ctx.fail("{} is invalid!, name should have prefix '{}' and suffix '{}'"
-                 .format(portchannel_name, CFG_PORTCHANNEL_PREFIX, CFG_PORTCHANNEL_NO))
-
-    db = ctx.obj['db']
-
+    
+    if ADHOC_VALIDATION:
+        if is_portchannel_name_valid(portchannel_name) != True:
+            ctx.fail("{} is invalid!, name should have prefix '{}' and sufix '{}'"
```
qiluo-msft, line 1901: >sufix [](http://example.com/codeflow?start=71&length=5)

typo

config/main.py
```diff
@@ -101,6 +102,7 @@
 TTL_RANGE = click.IntRange(min=0, max=255)
 QUEUE_RANGE = click.IntRange(min=0, max=255)
 GRE_TYPE_RANGE = click.IntRange(min=0, max=65535)
+ADHOC_VALIDATION = False
```
qiluo-msft, line 105: >ADHOC_VALIDATION [](http://example.com/codeflow?start=0&length=16)

This is good change, let's default to `True`.

config/validated_config_db_connector.py
qiluo-msft: `fallback_adhoc_validation` should be independent with `ValidatedConfigDBConnector`. We have use cases like
1. True, True
2. True, False
3. False, True
4. False, False
So just remove this parameter in this class.

config/main.py
```diff
@@ -5930,36 +5942,46 @@ def loopback(ctx, redis_unix_socket_path):
...
 @click.pass_context
 def del_loopback(ctx, loopback_name):
-    config_db = ctx.obj['db']
-    if is_loopback_name_valid(loopback_name) is False:
-        ctx.fail("{} is invalid, name should have prefix '{}' and suffix '{}' "
-                .format(loopback_name, CFG_LOOPBACK_PREFIX, CFG_LOOPBACK_NO))
-
+    config_db = ValidatedConfigDBConnector(ctx.obj['db'], ADHOC_VALIDATION)
+    
     lo_config_db = config_db.get_table('LOOPBACK_INTERFACE')
```
qiluo-msft, line 5967: >lo_config_db = config_db.get_table('LOOPBACK_INTERFACE') [](http://example.com/codeflow?start=4&length=56)

This line is moved. However you can keep it at original place, right?

config/vlan.py
```diff
@@ -20,18 +25,24 @@ def add_vlan(db, vid):
...
     
     if vid == 1:
         ctx.fail("{} is default VLAN".format(vlan))
     
-    if clicommon.check_if_vlanid_exist(db.cfgdb, vlan):
-        ctx.fail("{} already exists".format(vlan))
-
-    db.cfgdb.set_entry('VLAN', vlan, {'vlanid': vid})
+    if ADHOC_VALIDATION:
+        if clicommon.check_if_vlanid_exist(db.cfgdb, vlan): # MISSING CONSTRAINT IN YANG MODEL
```
qiluo-msft, line 38: >MISSING CONSTRAINT IN YANG MODEL [](http://example.com/codeflow?start=62&length=32)

Add prefix "TODO: "

config/vlan.py
```diff
@@ -42,26 +53,35 @@ def del_vlan(db, vid):
...
     
-    if keys:
-        ctx.fail("VLAN ID {} can not be removed. First remove all members assigned to this VLAN.".format(vid))
+        if keys: # MISSING CONSTRAINT IN YANG MODEL
+            ctx.fail("VLAN ID {} can not be removed. First remove all members assigned to this VLAN.".format(vid))
         
-    db.cfgdb.set_entry('VLAN', 'Vlan{}'.format(vid), None)
+    config_db = ValidatedConfigDBConnector(db.cfgdb, ADHOC_VALIDATION)
+    try: 
+        config_db.set_entry('VLAN', 'Vlan{}'.format(vid), None)
```
qiluo-msft, line 78: >None [](http://example.com/codeflow?start=58&length=4)

You should handle `set_entry` `None` as special case in decorator.

isabelmsft: Added special case to decorator

config/vlan.py
```diff
@@ -118,46 +135,54 @@ def add_vlan_member(db, vid, port, untagged):
...
+            ctx.fail("{} is already a member of {}".format(port, vlan))
+
+        if clicommon.is_valid_port(db.cfgdb, port):
+            is_port = True
+        elif clicommon.is_valid_portchannel(db.cfgdb, port):
+            is_port = False
+        else:
+            ctx.fail("{} does not exist".format(port))
+
+        if (is_port and clicommon.is_port_router_interface(db.cfgdb, port)) or \ # TODO: MISSING CONSTRAINT IN YANG MODEL
```
qiluo-msft, line 167: >\ # [](http://example.com/codeflow?start=79&length=3)

lgtm caught a true bug. You could not add comment after `\<CR>`. Suggest move the comment to the above line.

config/validated_config_db_connector.py
isabelmsft: Changed to use `JsonPointer.from_parts` in both branches

config/main.py
isabelmsft: `lo_config_db` is used with and without ADHOC_VALIDATION, so I moved the definition to before ADHOC_VALIDATION

config/validated_config_db_connector.py
isabelmsft: Added unit tests

config/validated_config_db_connector.py
qiluo-msft: Still applicable.

generic_config_updater/gu_common.py
```diff
@@ -16,6 +16,9 @@
 class GenericConfigUpdaterError(Exception):
     pass
 
+class EmptyTableError(Exception):
```
qiluo-msft, line 19: >Exception [](http://example.com/codeflow?start=22&length=9)

Inherit from `ValueError`?

config/validated_config_db_connector.py
```diff
@@ -0,0 +1,47 @@
...
+    gcu_patch = jsonpatch.JsonPatch(gcu_json_input)
+    format = ConfigFormat.CONFIGDB.name
+    config_format = ConfigFormat[format.upper()]
+
+    config_db = ConfigDBConnector()
+    config_db.connect()
+    try:
+        GenericUpdater().apply_patch(patch=gcu_patch, config_format=config_format, verbose=False, dry_run=False, ignore_non_yang_tables=False, ignore_paths=None)
+    except EmptyTableError:
+        print(config_db)
```
qiluo-msft, line 45: >print(config_db) [](http://example.com/codeflow?start=8&length=16)

Remove debug code?

config/validated_config_db_connector.py
```diff
@@ -0,0 +1,47 @@
...
+    gcu_json_input.append(gcu_json)
+    gcu_patch = jsonpatch.JsonPatch(gcu_json_input)
+    format = ConfigFormat.CONFIGDB.name
+    config_format = ConfigFormat[format.upper()]
+
+    config_db = ConfigDBConnector()
+    config_db.connect()
+    try:
+        GenericUpdater().apply_patch(patch=gcu_patch, config_format=config_format, verbose=False, dry_run=False, ignore_non_yang_tables=False, ignore_paths=None)
+    except EmptyTableError:
```
qiluo-msft, line 44: >EmptyTableError [](http://example.com/codeflow?start=11&length=15)

I understand GCU could not delete the last vlan entry in vlan table, because it will lead to an empty vlan table. And GCU is okay to delete the whole vlan table in this case.

Suggest catch the exception, if the original patch is delete one entry, we change the patch to delete the whole table and retry `apply_patch`.

qiluo-msft: Or you can decorate `delete_table`, and call it instead.

qiluo-msft: Or
```
gcu_json_input.append(gcu_json)
gcu_json_input.append({
    "op": "remove",
    "path": table
})
```

isabelmsft: I decorated `delete_table` and changed exception handling to call `validated_delete_table`

config/validated_config_db_connector.py
isabelmsft: Removed

generic_config_updater/gu_common.py
isabelmsft: Changed to inherit from `ValueError`

config/validated_config_db_connector.py
isabelmsft: I changed to make `fallback_adhoc_validation` independent of `ValidatedConfigDBConnector`. In `ValidatedConfigDBConnector`, we use `yang_config_enabled`

qiluo-msft (approved):

PR #2191, opened 2022-05-31
[ssip]: Add CLI

nazariig:
**HLD:** https://github.com/sonic-net/SONiC/pull/1002

#### What I did
* Implemented CLI for Syslog Source IP

#### How I did it
* N/A

#### How to verify it
* N/A

* N/A

```
root@sonic:/home/admin# show syslog
SERVER IP    SOURCE IP    PORT    VRF
-----------  -----------  ------  --------
2.2.2.2      1.1.1.1      514     default
3.3.3.3      1.1.1.1      514     mgmt
2222::2222   1111::1111   514     Vrf-Data
```

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -6015,58 +6015,6 @@ def enable(enable):
     clicommon.run_command(command, display_cmd=True)
 
 #
-# 'syslog' group ('config syslog ...')
-#
-@config.group(cls=clicommon.AbbreviationGroup, name='syslog')
-@click.pass_context
-def syslog_group(ctx):
-    """Syslog server configuration tasks"""
-    config_db = ConfigDBConnector()
-    config_db.connect()
-    ctx.obj = {'db': config_db}
-
-@syslog_group.command('add')
-@click.argument('syslog_ip_address', metavar='<syslog_ip_address>', required=True)
-@click.pass_context
-def add_syslog_server(ctx, syslog_ip_address):
-    """ Add syslog server IP """
-    if not clicommon.is_ipaddress(syslog_ip_address):
-        ctx.fail('Invalid ip address')
-    db = ctx.obj['db']
-    syslog_servers = db.get_table("SYSLOG_SERVER")
-    if syslog_ip_address in syslog_servers:
-        click.echo("Syslog server {} is already configured".format(syslog_ip_address))
-        return
-    else:
-        db.set_entry('SYSLOG_SERVER', syslog_ip_address, {'NULL': 'NULL'})
-        click.echo("Syslog server {} added to configuration".format(syslog_ip_address))
-        try:
-            click.echo("Restarting rsyslog-config service...")
-            clicommon.run_command("systemctl restart rsyslog-config", display_cmd=False)
-        except SystemExit as e:
-            ctx.fail("Restart service rsyslog-config failed with error {}".format(e))
-
-@syslog_group.command('del')
-@click.argument('syslog_ip_address', metavar='<syslog_ip_address>', required=True)
-@click.pass_context
-def del_syslog_server(ctx, syslog_ip_address):
-    """ Delete syslog server IP """
-    if not clicommon.is_ipaddress(syslog_ip_address):
-        ctx.fail('Invalid IP address')
-    db = ctx.obj['db']
-    syslog_servers = db.get_table("SYSLOG_SERVER")
-    if syslog_ip_address in syslog_servers:
-        db.set_entry('SYSLOG_SERVER', '{}'.format(syslog_ip_address), None)
-        click.echo("Syslog server {} removed from configuration".format(syslog_ip_address))
-    else:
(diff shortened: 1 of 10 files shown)


config/main.py
```diff
@@ -4970,6 +4974,12 @@ def del_vrf(ctx, vrf_name):
         ctx.fail("'vrf_name' is not start with Vrf, mgmt or management!")
     if len(vrf_name) > 15:
         ctx.fail("'vrf_name' is too long!")
+    syslog_table = config_db.get_table("SYSLOG_SERVER")
+    syslog_vrf_dev = "mgmt" if vrf_name == "management" else vrf_name
```
qiluo-msft, line 4978: >vrf_name [](http://example.com/codeflow?start=31&length=8)

@prsunny, Could you check VRF related part?

show/main.py
```diff
@@ -1550,35 +1554,14 @@ def show_run_snmp(db, ctx):
...
-        elif re_ipv4_2.match(line):
-            server =  re_ipv4_2.match(line).group(1)
-        elif re_ipv6.match(line):
-            server =  re_ipv6.match(line).group(1)
-        else:
-            continue
-        syslog_servers.append("[{}]".format(server))
-    syslog_dict['Syslog Servers'] = syslog_servers
-    print(tabulate(syslog_dict, headers=list(syslog_dict.keys()), tablefmt="simple", stralign='left', missingval=""))
+@clicommon.pass_db
```
wen587, line 1557: Hi, thanks for the modification. I would suggest to read from `/etc/rsyslog.conf`
I have seen scenarios where the SYSLOG_SERVER is different between configDB and `rsyslog.conf`. 
For example, SYSLOG_SERVER have configured four servers while `rsyslogd` only contains a subset(2 or 3) of those server.
That could be caused by the fact that rsyslogd wasn't able to consume the configDB very well.

nazariig: @wen587 IMHO, getting syslog server info by directly grepping `/etc/rsyslog.conf` is a kind of overhead and simply looks like a bad design (configuration should be read from Config DB - that's a SONiC philosophy). Syslog configuration is by design to be rendered from a template file. If something is missing or a divergence is present - then it's completely user's responsibility.

config/main.py
prsunny: lgtm, minor comment: Later part of the syslog implementation checks for the presence of 'source_ip' in linux during add operation. But this is not checked during an `ip address` delete. Is this intentional or am i missing something?

nazariig: @prsunny you are right, this is intentionally. The reason for that is because we don't reference IP in Config DB, so there is no point to check

show/main.py
wen587: Hi @nazariig, Syslog server updating needs to restart rsyslog each time. It could be the restarting is too frequent and SONiC missed some updates. 
However, the `rsyslog.conf` if I understand it right, is the REAL number of syslog server starts.
We can treat ConfigDB as the ground truth if and only if SONiC consumes all the config correctly.
Hi @qiluo-msft , could you correct me if I am wrong?

nazariig: > Hi @nazariig, Syslog server updating needs to restart rsyslog each time. It could be the restarting is too frequent and SONiC missed some updates.

@wen587 i understand your concerns and you are totally right. This PR resolves the issue you mentioned

nazariig: @qiluo-msft can you please advise here?

qiluo-msft: I checked many subcommands under `show running` and the implementation is not consistent. Reading `/etc/rsyslog.conf` has its own benefit, which acts more accurate as "running config". On the other side, reading ConfigDB acts more like "show user config".

@nazrrig Could you explain on the "overhead" part?

nazariig: @qiluo-msft IMHO, parsing generated files with regex for "running" config is an overhead. Still, if you believe that this is more reliable way - please let me know and i will align.

*P.S:* if we want to follow such an approach (instead of Config DB), would be great to define some kind of doc to explain where and how it supposed to be used.

liat-grozovik:
Are there any such examples in current sonic arch? If not I don't think we should go there unless we decide to move to a different way in sonic and for that I don't think we should use this specific implementation to discuss it

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

show/main.py
qiluo-msft: Please help explore parsing `/etc/rsyslog.conf` since this is the original solution. Let me know if you find it is unnecessary too complex.

nazariig: @qiluo-msft done

config/main.py
nazariig: @prsunny can you please approve if no more questions

wen587:
LGTM

qiluo-msft (approved):

PR #2193, opened 2022-06-03
[yang] remove mistakenly added parameter for 'get_module_name'

ayurkiv-nvda:
Signed-off-by: Andriy Yurkiv <ayurkiv@nvidia.com>

#### What I did
Revert parameter from function that was mistakenly added in https://github.com/Azure/sonic-utilities/pull/2118

We don't need to pass 'sonic_yang_options' to static function get_module_name()

#### How I did it
remove 'sonic_yang_options' from SonicYang object creation in get_module_name()

### DIFF
--- a/config/config_mgmt.py
+++ b/config/config_mgmt.py
@@ -281,7 +281,7 @@ class ConfigMgmt():
         """
 
         # Instantiate new context since parse_module_mem() loads the module into context.
-        sy = sonic_yang.SonicYang(YANG_DIR, sonic_yang_options=self.sonicYangOptions)
+        sy = sonic_yang.SonicYang(YANG_DIR)
         module = sy.ctx.parse_module_mem(yang_module_str, ly.LYS_IN_YANG)
         return module.name()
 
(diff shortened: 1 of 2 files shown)


volodymyrsamotiy:
@qiluo-msft, could you please help merging this PR?

Coverage is failing because it is 0%. But in this PR we just revert 1 line of code that was changed recently by another PR. And before that coverage was 0% as well. So I think we can be safely merge with coverage 0% (because in fact we just move to previous state).

We need to revert this 1 line because it caused degradation in application extension installation flow (BTW we even didn't need this change in that another PR, it was changed by mistake).

Please check previous comment for some more details.

qiluo-msft:
Could you add a unit test to cover the flow of `get_module_name()`? The old code will fail on this unit test because it has python syntax error, but new code will pass.

ayurkiv-nvda:
> Could you add a unit test to cover the flow of `get_module_name()`? The old code will fail on this unit test because it has python syntax error, but new code will pass.

@qiluo-msft I have updated PR, please take a look at new UT added to check flow of  "get_module_name"

qiluo-msft (approved):

PR #2194, opened 2022-06-06
Fix issue where generate_dump will abort if docker is not running lea…

jeffhtgt:
…ving

an incomplete support file.

Signed-off-by: Jeff Henning <jeffhtgt@gmail.com>

#### What I did

Fix generate_dump to not abort if a docker is not running.

#### How I did it

Add check for running docker in copy_from_docker() and return if the docker is not running (ported
from a 202012 branch).

#### How to verify it

Got a successful dump.tgz after running "show techsupport".  Previously the file would be
incomplete due to command abort.

### DIFF
--- a/scripts/generate_dump
+++ b/scripts/generate_dump
@@ -284,6 +284,13 @@ copy_from_docker() {
     local dstpath=$3
     local timeout_cmd="timeout --foreground ${TIMEOUT_MIN}m"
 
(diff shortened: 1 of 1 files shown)


jeffhtgt:
@lguohan @prsunny Could you please help with review and moving this forward?  Thank you!

PR #2196, opened 2022-06-07
Potential risk of acl_loader missing ctx in main function

inspurSDN:
Issue:
Potential risk about cli() function without ctx in main function of acl_loader

Root cause:
SONiC will call cli() directly in acl-loader.
But it still exist potential risk when run main in acl-loader

Solution:
Remover main funcion which missing ctx object and remain cli() function for acl-loader call directly

### DIFF
--- a/acl_loader/main.py
+++ b/acl_loader/main.py
@@ -1078,11 +1078,3 @@ def delete(ctx, table, rule):
 
     acl_loader.delete(table, rule)
 
-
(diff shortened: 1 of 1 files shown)


inspurSDN:
@lguohan could you help to review it?

lguohan:
@qiluo-msft , can you check this pr?

qiluo-msft (approved):

PR #2200, opened 2022-06-08
[config reload] Fixing config reload when timer based services are disabled

dgsudharsan:
Signed-off-by: Sudharsan Dhamal Gopalarathnam <sudharsand@nvidia.com>

DO NOT MERGE UNTIL https://github.com/Azure/sonic-mgmt/pull/5783 IS MERGED
#### What I did
Fixed config reload when timer based delayed services are disabled. When they are disabled, the property property=LastTriggerUSecMonotonic returns "0". This will cause config reload to fail even though all enabled services are up.

#### How I did it
Fixed the delayed services logic to check if the services are enabled before getting the property LastTriggerUSecMonotonic . Additionally fixed the return codes when config reload fails due to system checks

#### How to verify it
Added UT to verify it. Modified sonic-mgmt tests to verify it additionally.

PR #2201, opened 2022-06-08
[config reload][202111] Fixing config reload when timer based services are disabled

dgsudharsan:
Porting of https://github.com/Azure/sonic-utilities/pull/2200 into 202111
#### What I did
Fixed config reload when timer based delayed services are disabled. When they are disabled, the property property=LastTriggerUSecMonotonic returns "0". This will cause config reload to fail even though all enabled services are up.

#### How I did it
Fixed the delayed services logic to check if the services are enabled before getting the property LastTriggerUSecMonotonic . Additionally fixed the return codes when config reload fails due to system checks

#### How to verify it
Added UT to verify it. Modified sonic-mgmt tests to verify it additionally.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -734,7 +734,7 @@ def _get_sonic_services():
     return (unit.strip() for unit in out.splitlines())
 
 
-def _get_delayed_sonic_services():
+def _get_delayed_sonic_units(get_timers=False):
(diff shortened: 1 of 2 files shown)


stepanblyschak (approved):

dgsudharsan:
@qiluo-msft @judyjoseph  Can we get this approved? The main PR is approved by @qiluo-msft

qiluo-msft (approved):

dgsudharsan:
@qiluo-msft  Can we merge this?

PR #2205, opened 2022-06-09
Update db_migrator to support `PORT_QOS_MAP|global`

bingwang-ms:
Signed-off-by: bingwang <wang.bing@microsoft.com>

#### What I did
This PR is to update the `db_migrator` to support generation of `PORT_QOS_MAP|global`.

After PR https://github.com/Azure/sonic-buildimage/pull/10565, there will be two DSCP_TO_TC_MAP

- DSCP_TO_TC_MAP|AZURE is the default map, which is used at port level and switch level
- DSCP_TO_TC_MAP|AZURE_TUNNEL is used to remap the priority of tunnel traffic in dualtor deployment

To address the issue, an entry `PORT_QOS_MAP|global` will be added into `config_db`
```
"PORT_QOS_MAP": {
        "global": {
            "dscp_to_tc_map": "AZURE"
        }
}
```
To handle the upgrade from older image, `db_migrator` is updated to generate the entry.
There is no way to check if `PORT_QOS_MAP|global` is supported, so the migration is ran on all platforms. There is check in `sonic-swss` to check the switch capability [code](https://github.com/Azure/sonic-swss/blob/ad8f5e43951d6ab8cb2e777f45f4da5a9245a982/orchagent/qosorch.cpp#L245-250
), so it will not cause issue.

#### How I did it
Update `db_migrator` to add a common migration step.

#### How to verify it
Verified by vstest
```
collected 78 items                                                                                                                                                                                    

tests/db_migrator_test.py ..............................................................................                                                                                        [100%]

===================================================================================== 78 passed in 26.55 seconds ======================================================================================
```


scripts/db_migrator.py
```diff
@@ -485,6 +485,20 @@ def migrate_qos_fieldval_reference_format(self):
...
+    def migrate_port_qos_map_global(self):
+        """
+        Generate dscp_to_tc_map for switch.
+        """
+        dscp_to_tc_map_table_names = self.configDB.get_keys('DSCP_TO_TC_MAP')
+        if len(dscp_to_tc_map_table_names) == 0:
+            return True
+        
+        qos_maps = self.configDB.get_table('PORT_QOS_MAP')
+        if 'global' not in qos_maps.keys():
```
stephenxs, line 497: I suggest skipping Mellanox devices here because the global `PORT_QOS_MAP` is not required and will not be applied on Mellanox devices.
```suggestion
        if 'global' not in qos_maps.keys() and self.asic_type != "mellanox":
```

stephenxs (changes requested):

scripts/db_migrator.py
```diff
@@ -485,6 +485,23 @@ def migrate_qos_fieldval_reference_format(self):
         self.migrate_qos_db_fieldval_reference_remove(qos_table_list, self.configDB, self.configDB.CONFIG_DB, '|')
         return True
 
+    def migrate_port_qos_map_global(self):
+        """
+        Generate dscp_to_tc_map for switch.
+        """
+        asics_require_global_dscp_to_tc_map = ["broadcom"]
+        if self.asic_type not in asics_require_global_dscp_to_tc_map:
+            return True
```
neethajohn, line 494: "return" should do since we are not capturing the return value of this function

scripts/db_migrator.py
```diff
@@ -485,6 +485,23 @@ def migrate_qos_fieldval_reference_format(self):
...
+        dscp_to_tc_map_table_names = self.configDB.get_keys('DSCP_TO_TC_MAP')
+        if len(dscp_to_tc_map_table_names) == 0:
+            return True
```
neethajohn, line 497: same comment as above

tests/db_migrator_test.py
```diff
@@ -373,3 +373,30 @@ def test_pfc_enable_migrator(self):
...
+    @classmethod
+    def teardown_class(cls):
+        os.environ['UTILITIES_UNIT_TESTING'] = "0"
+        dbconnector.dedicated_dbs['CONFIG_DB'] = None
+
+    def test_global_dscp_to_tc_map_migrator(self):
+        dbconnector.dedicated_dbs['CONFIG_DB'] = os.path.join(mock_db_path, 'config_db', 'qos_map_table_global_input')
+        import db_migrator
+        dbmgtr = db_migrator.DBMigrator(None)
+        dbmgtr.asic_type = "broadcom"
```
neethajohn, line 391: please add a testcase for asic_type == 'mellanox' as well

scripts/db_migrator.py
bingwang-ms: Thanks. I added a check for ASIC type. The table entry is only added for a platform now.

scripts/db_migrator.py
bingwang-ms: Fixed. Thanks

scripts/db_migrator.py
bingwang-ms: Fixed.  Thanks

tests/db_migrator_test.py
bingwang-ms: Added.

neethajohn (approved):

neethajohn:
@stephenxs, please take a final look

scripts/db_migrator.py
```diff
@@ -725,6 +742,8 @@ def common_migration_ops(self):
             self.migrate_mgmt_ports_on_s6100()
         else:
             log.log_notice("Asic Type: {}, Hwsku: {}".format(self.asic_type, self.hwsku))
+        
+        self.migrate_port_qos_map_global()
```
prsunny, line 746: why is this in the common migration path? I thought it should be based on a version.

stephenxs (dismissed):

stephenxs:
> @stephenxs, please take a final look

LGTM. Approved.

scripts/db_migrator.py
bingwang-ms: Updated. Please take a look. Thanks

prsunny (approved):

neethajohn (approved):

PR #2210, opened 2022-06-11
[config][muxcable] fix minor config DB logic issue

vdahiya12:
For active-standby type cable there will not be any soc_ipv4 entry in the "MUX_CABLE" table, hence when updating the config DB entry with update 
for 
```sudo config mux mode <state> <port>```
we should just check the cable_type and update the table entry
if no soc_ipv4 the value written in none in that case

#### How to verify it
Unit-tests cover the change

### DIFF
--- a/config/muxcable.py
+++ b/config/muxcable.py
@@ -254,7 +254,7 @@ def lookup_statedb_and_update_configdb(db, per_npu_statedb, config_db, port, sta
     if str(state_cfg_val) == str(configdb_state):
         port_status_dict[port_name] = 'OK'
     else:
-        if cable_type is not None and soc_ipv4_value is not None:
+        if cable_type is not None or soc_ipv4_value is not None:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: config/muxcable.py [hunk 1]
```diff
@@ -254,7 +254,7 @@ def lookup_statedb_and_update_configdb(db, per_npu_statedb, config_db, port, sta
     if str(state_cfg_val) == str(configdb_state):
         port_status_dict[port_name] = 'OK'
     else:
-        if cable_type is not None and soc_ipv4_value is not None:
+        if cable_type is not None or soc_ipv4_value is not None:
```
zjswhhh, line 257: Are we assuming `cable_type` field only presents for `active-active` ports? It doesn't harm to check `if cable_type == "active-active"`?

vdahiya12: so basically, for "active-standby" type cable there is no soc_ipv4 but we still need to write the value of "cable_type" back

zjswhhh (approved):

PR #2212, opened 2022-06-13
[generic-config-updater] Add NTP validator 

wen587:
#### What I did
When GCU config NTP_SERVER table, its change is not actually taken info effect. Because NTP service is not restarted. So I add NTP service validator and restart service when NTP_SERVER is changed.
#### How I did it
When NTP_SERVER table being configured through GCU, the ntp service will restart.
#### How to verify it
Run GCU E2E test for NTP_SERVER table.

### DIFF
--- a/generic_config_updater/generic_config_updater.conf.json
+++ b/generic_config_updater/generic_config_updater.conf.json
@@ -47,7 +47,7 @@
             "services_to_validate": [ "caclmgrd-service" ]
         },
         "NTP_SERVER": {
-            "services_to_validate": [ "ntp" ]
+            "services_to_validate": [ "ntp-service" ]
         }
     },
     "services": {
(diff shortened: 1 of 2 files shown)


generic_config_updater/services_validator.py
```diff
@@ -98,3 +98,7 @@ def caclmgrd_validator(old_config, upd_config, keys):
     # No update to ACL_RULE.
     return True
 
+
+def ntp_validator(old_config, upd_config, keys):
+    return _service_restart("ntp-config")
+
```
qiluo-msft, line 104: Remove unnecessary blank lines at the file end.

generic_config_updater/generic_config_updater.conf.json
```diff
@@ -45,6 +45,9 @@
         },
         "ACL_RULE": {
             "services_to_validate": [ "caclmgrd-service" ]
+        },
+        "NTP_SERVER": {
+            "services_to_validate": [ "ntp" ]
```
qiluo-msft, line 50: >ntp [](http://example.com/codeflow?start=39&length=3)

ntp-service or ntp-config-service

qiluo-msft (approved):

PR #2215, opened 2022-06-14
[202012] [cherry-pick] Update db_migrator to support `PORT_QOS_MAP|global`

bingwang-ms:
#### What I did
This PR is to cherry-pick https://github.com/Azure/sonic-utilities/pull/2205 into `202012` branch after resolving conflicts.

This PR is to update the `db_migrator` to support generation of `PORT_QOS_MAP|global`.

After PR https://github.com/Azure/sonic-buildimage/pull/10565, there will be two DSCP_TO_TC_MAP

- DSCP_TO_TC_MAP|AZURE is the default map, which is used at port level and switch level
- DSCP_TO_TC_MAP|AZURE_TUNNEL is used to remap the priority of tunnel traffic in dualtor deployment

To address the issue, an entry `PORT_QOS_MAP|global` will be added into `config_db`
```
"PORT_QOS_MAP": {
        "global": {
            "dscp_to_tc_map": "AZURE"
        }
}
```
To handle the upgrade from older image, `db_migrator` is updated to generate the entry.
There is no way to check if `PORT_QOS_MAP|global` is supported, so the migration is ran on all platforms. There is check in `sonic-swss` to check the switch capability [code](https://github.com/Azure/sonic-swss/blob/ad8f5e43951d6ab8cb2e777f45f4da5a9245a982/orchagent/qosorch.cpp#L245-250
), so it will not cause issue.

#### How I did it
Update `db_migrator` to add a common migration step.

#### How to verify it
Verified by vstest
```
collected 76 items                                                                                                                                                                                    

tests/db_migrator_test.py ............................................................................                                                                                          [100%]

===================================================================================== 76 passed in 18.26 seconds ======================================================================================
```


stephenxs:
Comment left on the original PR #2205 is also applicable here.

scripts/db_migrator.py
```diff
@@ -422,6 +422,23 @@ def migrate_pfcwd_sw_enable_table(self):
                 v['pfcwd_sw_enable'] = v['pfc_enable']
                 self.configDB.set_entry('PORT_QOS_MAP', k, v)
 
+    def migrate_port_qos_map_global(self):
+        """
+        Generate dscp_to_tc_map for switch.
+        """
+        asics_require_global_dscp_to_tc_map = ["broadcom"]
+        if self.asic_type not in asics_require_global_dscp_to_tc_map:
+            return True
```
neethajohn, line 431: "return" should do since we are not capturing the return value of this function

scripts/db_migrator.py
```diff
@@ -422,6 +422,23 @@ def migrate_pfcwd_sw_enable_table(self):
...
+        dscp_to_tc_map_table_names = self.configDB.get_keys('DSCP_TO_TC_MAP')
+        if len(dscp_to_tc_map_table_names) == 0:
+            return True
```
neethajohn, line 434: same comment as above

tests/db_migrator_test.py
```diff
@@ -274,3 +274,29 @@ def test_pfc_enable_migrator(self):
...
+    @classmethod
+    def teardown_class(cls):
+        os.environ['UTILITIES_UNIT_TESTING'] = "0"
+        dbconnector.dedicated_dbs['CONFIG_DB'] = None
+
+    def test_global_dscp_to_tc_map_migrator(self):
+        dbconnector.dedicated_dbs['CONFIG_DB'] = os.path.join(mock_db_path, 'config_db', 'qos_map_table_global_input')
+        import db_migrator
+        dbmgtr = db_migrator.DBMigrator(None)
+        dbmgtr.asic_type = "broadcom"
```
neethajohn, line 291: please add a testcase for asic_type == 'mellanox' as well

bingwang-ms:
> Comment left on the original PR #2205 is also applicable here.

Thanks. I address the issue by checking the ASIC type.  The table is only generated on one platform now. Also a test scenario is added to cover that.

scripts/db_migrator.py
bingwang-ms: Fixed. Thanks

scripts/db_migrator.py
bingwang-ms: Fixed. Thanks

tests/db_migrator_test.py
bingwang-ms: Good suggestion. A new test scenario is added.

neethajohn (approved):

stephenxs (approved):

scripts/db_migrator.py
```diff
@@ -558,9 +575,17 @@ def version_1_0_6(self):
 
     def version_2_0_0(self):
         """
-        Current latest version. Nothing to do here.
+        Version 2_0_0.
         """
         log.log_info('Handling version_2_0_0')
+        self.migrate_port_qos_map_global()
```
neethajohn, line 581: We cannot add this here because all the subsequent versions have already been taken by other branches. for eg, 202111 is at version 2_0_4. There are many migration tasks that need to be done when moving from 202012 to 202111. refer to this.
https://github.com/Azure/sonic-utilities/blob/202111/scripts/db_migrator.py

bingwang-ms: Yes, that is also my concern. 
But, do we have to add all upcoming migration for `202012` branch to `common_migration` ? Otherwise we will run into the version issue.

prsunny: I think this is fine, because this path run for all versions below it which is what we need, correct?

bingwang-ms: > I think this is fine, because this path run for all versions below it which is what we need, correct?

Yes, Prince, for this migration, the code is fine. I think what Neetha was concerning is how we handle the migration from `202012` image to `202111` image.
For example, we already have the migrator in branch `202111`, which version is `2_0_0`. It's for migrating `migrate_config_db_port_table_for_auto_neg()`. Now, we are going to define migrator for `202012` branch, which is also at `2_0_0`. So, when we are upgrading from `202012` to `202111`, the migrator at `v_2_0_0` will not be executed since the version in db is already `2_0_0`.
One possible solution may be https://github.com/Azure/sonic-utilities/pull/2272

prsunny (approved):

PR #2234, opened 2022-06-27
[GCU] Moving UniqueLanes from only validating moves, to be a supplemental YANG validator

ghooo:
#### What I did
- Added a new supplemental YANG validator to validate UniqueLanes in `validate_config_db_config`
- Removed UniqueLanesMoveValidator as the lanes validation will be taken care of by FullConfigMoveValidator which uses `validate_config_db_config`

The benefit of this is at the beginning of `apply-patch` we make a call to `validate_config_db_config` to check if the given patch is valid or not ([code](https://github.com/Azure/sonic-utilities/blob/e6e4f8ceb9a59fb7b3767a65ffc4f017d0807832/generic_config_updater/patch_sorter.py#L1522)). Now we will fail early, instead of going for the move generation and not being able to generate a moves.

#### How I did it
Check code.

#### How to verify it
Added unit-tests.

### DIFF
--- a/generic_config_updater/gu_common.py
+++ b/generic_config_updater/gu_common.py
@@ -115,7 +115,7 @@ class ConfigWrapper:
 
         # TODO: Move these validators to YANG models
         supplemental_yang_validators = [self.validate_bgp_peer_group,
-                                        self.validate_unique_lanes]
+                                        self.validate_lanes]
 
         try:
             tmp_config_db_as_json = copy.deepcopy(config_db_as_json)
@@ -133,21 +133,35 @@ class ConfigWrapper:
 
         return True, None
 
-    def validate_unique_lanes(self, config_db):
+    def validate_lanes(self, config_db):
         if "PORT" not in config_db:
             return True, None
 
         ports = config_db["PORT"]
-        existing = {}
+
+        # Validate each lane separately, make sure it is not empty, and is a number
+        port_to_lanes_map = {}
         for port in ports:
             attrs = ports[port]
             if "lanes" in attrs:
                 lanes_str = attrs["lanes"]
-                lanes = lanes_str.split(", ")
+                lanes_with_whitespaces = lanes_str.split(",")
+                lanes = [lane.strip() for lane in lanes_with_whitespaces]
                 for lane in lanes:
-                    if lane in existing:
-                        return False, f"'{lane}' lane is used multiple times in PORT: {set([port, existing[lane]])}"
-                    existing[lane] = port
+                    if not lane:
+                        return False, f"PORT '{port}' has an empty lane"
+                    if not lane.isdigit():
+                        return False, f"PORT '{port}' has an invalid lane '{lane}'"
+                port_to_lanes_map[port] = lanes
+
+        # Validate lanes are unique
+        existing = {}
+        for port in port_to_lanes_map:
+            lanes = port_to_lanes_map[port]
+            for lane in lanes:
+                if lane in existing:
+                    return False, f"'{lane}' lane is used multiple times in PORT: {set([port, existing[lane]])}"
+                existing[lane] = port
         return True, None
 
     def validate_bgp_peer_group(self, config_db):
--- a/generic_config_updater/gu_common.py
+++ b/generic_config_updater/gu_common.py
@@ -113,6 +113,10 @@ class ConfigWrapper:
     def validate_config_db_config(self, config_db_as_json):
         sy = self.create_sonic_yang_with_loaded_models()
 
+        # TODO: Move these validators to YANG models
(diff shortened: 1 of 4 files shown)


generic_config_updater/gu_common.py
```diff
@@ -113,18 +113,43 @@ def validate_sonic_yang_config(self, sonic_yang_as_json):
...
+        if "PORT" not in config_db:
+            return True, None
+
+        ports = config_db["PORT"]
+        existing = {}
+        for port in ports:
+            attrs = ports[port]
+            if "lanes" in attrs:
+                lanes_str = attrs["lanes"]
+                lanes = lanes_str.split(", ")
```
qiluo-msft, line 146: >,  [](http://example.com/codeflow?start=41&length=2)

This is too strict "comma, followed by a space, as a separator". More often, it is just one comma.

generic_config_updater/gu_common.py
```diff
@@ -113,18 +113,43 @@ def validate_sonic_yang_config(self, sonic_yang_as_json):
...
+                for lane in lanes:
+                    if lane in existing:
```
qiluo-msft, line 148: >lane [](http://example.com/codeflow?start=23&length=4)

check `lane` is a string of number?

generic_config_updater/gu_common.py
```diff
@@ -113,18 +113,43 @@ def validate_sonic_yang_config(self, sonic_yang_as_json):
...
+                        return False, f"'{lane}' lane is used multiple times in PORT: {set([port, existing[lane]])}"
+                    existing[lane] = port
+        return True, None
```
qiluo-msft, line 151: >None [](http://example.com/codeflow?start=21&length=4)

return `""` instead of `None`.

ghooo: In C#, we use the pattern `TryParse(...)` the return is a boolean and the output is passed as an out paramter.

Here I think I am doing the same, but instead `out param`, returning 2 outputs

In C#, we return `null` as the result of parsing if parsing fails. This is how the built in class `IPAddress` handles `TryParse`

```csharp
IPAddress.TryParse("invalid", out IPAddress invalid_ip);
Console.WriteLine(invalid_ip);
```
Output
```
null
```

Is that not good in python? or you want this improvement for reduce the four-letter word `None` to `""`

generic_config_updater/gu_common.py
ghooo: Sure, we can double check here since that's already confirmed by YANG validation done earlier.

generic_config_updater/gu_common.py
ghooo: Ah good point, let me fix this.

ghooo: Updated

generic_config_updater/gu_common.py
ghooo: Updated

generic_config_updater/gu_common.py
```diff
@@ -113,18 +113,57 @@ def validate_sonic_yang_config(self, sonic_yang_as_json):
...
 
+        return True, None
+
+    def validate_lanes(self, config_db):
+        if "PORT" not in config_db:
+            return True, None
+
+        ports = config_db["PORT"]
+
+        # Validate each lane separately, make sure it is not empty, and is a number
```
ghooo, line 142: I don't like validating single lane and unique lanes in 1 function, I think it would be better if we can split the lanes in ConfigDb to YANG conversion, then validate unique lanes later. Just writing this comment here for future reference.

generic_config_updater/gu_common.py
qiluo-msft: It's a minor issue, not critical.

qiluo-msft (approved):

generic_config_updater/gu_common.py
ghooo: Ah, did not know we can do that with assert. I think for the current case if the return is `False` the error should neither be `None` or `Empty`. I think this is interesting let me think about it separately.

generic_config_updater/gu_common.py
```diff
@@ -113,18 +113,57 @@ def validate_sonic_yang_config(self, sonic_yang_as_json):
...
+        port_to_lanes_map = {}
+        for port in ports:
+            attrs = ports[port]
+            if "lanes" in attrs:
+                lanes_str = attrs["lanes"]
+                lanes_with_whitespaces = lanes_str.split(",")
```
isabelmsft, line 148: Can we assume that there will always be a comma-separated list? Is it possible that the user may push just one lane?

ghooo: lanes can have a single element, check test case `test_validate_lanes__single_valid_lane__success`

PR #2240, opened 2022-06-29
Add 'traffic_shift_away' option to config load_minigraph

tjchadaga:
#### What I did
Add a CLI option to 'config load_minigraph' to keep the device in maintenance

#### How I did it
Added changes to execute TSA after minigraph is loaded to ensure the device is kept in maintenance. 

#### How to verify it
1. Execute 'sudo load_minigraph -t'
2.  After the new config is fully applied, execute 'TSC' to ensure the device mode is "Maintenance"

sudo config load_minigraph -h
Usage: config load_minigraph [OPTIONS]

  Reconfigure based on minigraph.

Options:
  -y, --yes
  -n, --no_service_restart  Do not restart docker services
  -?, -h, --help            Show this message and exit.

sudo config load_minigraph -h
Usage: config load_minigraph [OPTIONS]

  Reconfigure based on minigraph.

Options:
  -y, --yes
  -n, --no_service_restart  Do not restart docker services
  **-t, --traffic_shift_away  Keep device in maintenance with TSA**
  -?, -h, --help            Show this message and exit.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1641,8 +1641,9 @@ def load_mgmt_config(filename):
 @click.option('-y', '--yes', is_flag=True, callback=_abort_if_false,
                 expose_value=False, prompt='Reload config from minigraph?')
 @click.option('-n', '--no_service_restart', default=False, is_flag=True, help='Do not restart docker services')
+@click.option('-t', '--traffic_shift_away', default=False, is_flag=True, help='Keep device in maintenance with TSA')
 @clicommon.pass_db
-def load_minigraph(db, no_service_restart):
+def load_minigraph(db, no_service_restart, traffic_shift_away):
     """Reconfigure based on minigraph."""
     log.log_info("'load_minigraph' executing...")
 
@@ -1711,6 +1712,10 @@ def load_minigraph(db, no_service_restart):
     # Load golden_config_db.json
     if os.path.isfile(DEFAULT_GOLDEN_CONFIG_DB_FILE):
         override_config_by(DEFAULT_GOLDEN_CONFIG_DB_FILE)
+
+    # Keep device in maintenance with TSA 
+    if traffic_shift_away:
+        clicommon.run_command("TSA", display_cmd=True)
 
     # We first run "systemctl reset-failed" to remove the "failed"
     # status from all services before we attempt to restart them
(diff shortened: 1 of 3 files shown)


config/main.py
```diff
@@ -1712,6 +1713,10 @@ def load_minigraph(db, no_service_restart):
     if os.path.isfile(DEFAULT_GOLDEN_CONFIG_DB_FILE):
         override_config_by(DEFAULT_GOLDEN_CONFIG_DB_FILE)
 
+    # Keep device in maintenance with TSA 
+    if traffic_shift_away:
+        clicommon.run_command("TSA", display_cmd=True)
```
qiluo-msft, line 1718: >run_command [](http://example.com/codeflow?start=18&length=11)

This may override above golden config. Actually you can achieve your goal with golden config, if golden config is used.

qiluo-msft:
@wen587 Please help review.

config/main.py
tjchadaga: This will only apply when the command is triggered with "--traffic_shift_away" option, in which case, I think the user intention is to override golden config and keep the device in maintenance.

wen587:
LGTM

config/main.py
qiluo-msft: Your implementation on a special case that DEFAULT_GOLDEN_CONFIG_DB_FILE exists and user provided "--traffic_shift_away" option:
1. first override by golden config
2. TSA applied

So TSA will win the override because it is applied later. However in this special case, I think golden config should be the winner since user could use golden config to achieve the effect of TSA. Suggest ignore TSA option if DEFAULT_GOLDEN_CONFIG_DB_FILE exists and user provided "--traffic_shift_away" option

wen587: > Your implementation on a special case that DEFAULT_GOLDEN_CONFIG_DB_FILE exists and user provided "--traffic_shift_away" option:
> 
> 1. first override by golden config
> 2. TSA applied
> 
> So TSA will win the override because it is applied later. However in this special case, I think golden config should be the winner since user could use golden config to achieve the effect of TSA. Suggest ~ignore TSA option if DEFAULT_GOLDEN_CONFIG_DB_FILE exists and user provided "--traffic_shift_away" option~ swap the order of Step 1 and Step 2

That's right. Customers can achieve TSA status by modifying the golden config.

tjchadaga: Thanks, updated and added a warning since the CLI option may not work in this case anymore

qiluo-msft (approved):

PR #2242, opened 2022-06-30
Added cisco config platform commands

yucgu:
Signed-off-by: Yucai Gu <yucgu@cisco.com>

#### What I did
Add cisco sub-command option under 'config platform' command

#### How I did it
In config/main.py, check the platform type and import the cisco.py file under cisco platform code when it's cisco-8000.

#### How to verify it
Run config platform -h to see all commands. We will be able to see config platform cisco. This is only available on cisco devices.


kevinskwang (dismissed):

yxieca (approved):

yxieca:
@yucgu can you add a unit test to cover this simple change?

yucgu:
> @yucgu can you add a unit test to cover this simple change?
Hi Xing,
thanks for looking at it. I tried various changes, includes adding the test cases or calling the function directly, but still not able to make the UT coverage pass. 
    Probably the UT infra doesn't support this kind of coverage by now, there are two reasons:
    1) the parent function where I added those 3 lines of new codes is a main entry point of a command chain, it's not called/covered, I checked other main entry points of other command chains, all are not called/covered, looks like the UT infra doesn't support that coverage.
    2) even there is some way to make it called, and we use mock to make asic_type as cisco-8000, the code will still not be tested because the cisco.py is in the platform code repo, the UT of sonic-utilities doesn't have that file.

PR #2255, opened 2022-07-06
[show] add new CLI to show tunnel route objects

zjswhhh:
#### What I did
Add new CLI support to show tunnel route objects in ASIC DB.

sign-off: Jing Zhang zhangjing@microsoft.com

#### How I did it
* Check if tunnel route object exists for **server_ipv4, server_ipv6, soc_ipv4**. If existing, print it out. 
* If not specifying port name, print all tunnel route objects. 
  
#### How to verify it
* Added unit tests. 
* Tested on dual testbeds: 
  *  show mux tunnel-route Ethernet140
     ```
      PORT         DEST_TYPE    DEST_ADDRESS    
      -----------  -----------  ---------------  
      Ethernet140  server_ipv4  192.168.0.29/32  
     ```
  *  show mux tunnel-route --json Ethernet140
     ```
      {
          "TUNNEL_ROUTE": {
              "Ethernet140": {
                  "server_ipv4": {
                      "DEST": "192.168.0.29/32"
                  }
              }
          }
      }
     ```

### DIFF
--- a/show/muxcable.py
+++ b/show/muxcable.py
@@ -505,20 +505,20 @@ def create_json_dump_per_port_config(db, port_status_dict, per_npu_configdb, asi
 
 def get_tunnel_route_per_port(db, port_tunnel_route, per_npu_configdb, per_npu_asic_db, asic_id, port):
 
-    mux_cfg_dict[asic_id] = per_npu_configdb[asic_id].get_all(
+    mux_cfg_dict = per_npu_configdb[asic_id].get_all(
     per_npu_configdb[asic_id].CONFIG_DB, 'MUX_CABLE|{}'.format(port))
     dest_names = ["server_ipv4", "server_ipv6", "soc_ipv4"]
 
     for name in dest_names:
-        dest_address = mux_cfg_dict[asic_id].get(name, None)
+        dest_address = mux_cfg_dict.get(name, None)
 
         if dest_address is not None:
             route_keys = per_npu_asic_db[asic_id].keys(
                 per_npu_asic_db[asic_id].ASIC_DB, 'ASIC_STATE:SAI_OBJECT_TYPE_ROUTE_ENTRY:*{}*'.format(dest_address))
             
-            if route_keys is not None:
+            if route_keys is not None and len(route_keys):
                 route_dict = per_npu_asic_db[asic_id].get_all(
-                    per_npu_asic_db[asic_id].ASIC_DB, route_keys[0])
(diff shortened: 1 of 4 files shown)


zjswhhh:
> I tent to disagree with this change. First of all, `Nexthop ID` doesn't convey any meaningful information. Secondly, just "ip route show 192.168.0.29/32" will display if its tunnel or non-tunnel route.

Hi Prince - Yes `ip route show` will have that information. It's just recently we have been doing static active-active (forcing both ToR to be active) on active-standby dualToRs, and to verify the tunnel route is removed, I first need to get server ip and then check the route.  So I wanted to add a CLI to check in batches. But it's not a must-have.

yxieca (dismissed):

show/muxcable.py
```diff
@@ -503,6 +503,42 @@ def create_json_dump_per_port_config(db, port_status_dict, per_npu_configdb, asi
     if soc_ipv4_value is not None:
         port_status_dict["MUX_CABLE"]["PORTS"][port_name]["SERVER"]["soc_ipv4"] = soc_ipv4_value
 
+def get_tunnel_route_per_port(db, port_tunnel_route, per_npu_configdb, per_npu_asic_db, asic_id, port):
```
prsunny, line 506: If the intention is to only display the list of Tunnel route, please fetch data from the APP_DB "TUNNEL_ROUTE_TABLE" and display and avoid this complexity of looping through ASIC DB.

prsunny (changes requested):

show/muxcable.py
zjswhhh: Will there be any chance of inconsistency between APP_DB and ASIC_DB? Are there any concerns of using ASIC_DB?

prsunny: There should not be any inconsistency. If it is there, its a bug scenario. You don't have to use the translation logic  for ASIC_DB objects if using APP_DB. Moreover, its straightforward. Tunnel routes is what you are looking for and it is part of the TUNNEL_ROUTE_TABLE.

zjswhhh: Thanks for clarifying. Will update accordingly.

prsunny (approved):
lgtm

yxieca (approved):

qiluo-msft:

liuh-80:
Cherry-pick PR merged: https://github.com/sonic-net/sonic-utilities/pull/2659

PR #2257, opened 2022-07-07
Add interface link-training command into the CLI doc

ds952811:
- What I did
Add interface link-training command into the CLI doc

- How I did it
Update Command-Reference.md

- How to verify it
N/A

Signed-off-by: Dante Su <dante.su@broadcom.com>

### DIFF
--- a/doc/Command-Reference.md
+++ b/doc/Command-Reference.md
@@ -441,14 +441,15 @@ The same syntax applies to all subgroups of `show` which themselves contain subc
     -?, -h, --help  Show this message and exit.
 
   Commands:
-    counters     Show interface counters
-    description  Show interface status, protocol and...
-    naming_mode  Show interface naming_mode status
-    neighbor     Show neighbor related information
-    portchannel  Show PortChannel information
-    status       Show Interface status information
-    tpid         Show Interface tpid information
-    transceiver  Show SFP Transceiver information
+    counters       Show interface counters
+    description    Show interface status, protocol and...
+    link-training  Show interface link-training information
+    naming_mode    Show interface naming_mode status
+    neighbor       Show neighbor related information
+    portchannel    Show PortChannel information
+    status         Show Interface status information
+    tpid           Show Interface tpid information
+    transceiver    Show SFP Transceiver information
   ```
 
 Go Back To [Beginning of the document](#) or [Beginning of this section](#getting-help)
@@ -3642,6 +3643,35 @@ This command displays the key fields of the interfaces such as Operational Statu
   Ethernet4    down       up  hundredGigE1/2  T0-2:hundredGigE1/30
   ```
 
+**show interfaces link-training (Versions >= 202211)**
+
+This command is to display the link-training status of the selected interfaces. If **interface_name** is not specicied, this command shows the link-training status of all interfaces.
+
+- Usage:
+  ```
+  show interfaces link-training status [<interface_name>]
+  ```
+
+- Example:
+  ```
+  admin@sonic:~$ show interfaces link-training status
+    Interface      LT Oper    LT Admin    Oper    Admin
+  -----------  -----------  ----------  ------  -------
+    Ethernet0      trained          on      up       up
+    Ethernet8      trained          on      up       up
+   Ethernet16          off         off    down       up
+   Ethernet24  not trained          on    down       up
+   Ethernet32          off         off    down       up
+  ```
+
+- Example (to only display the link-training status of interface Ethernet8):
+  ```
+  admin@sonic:~$ show interfaces link-training status Ethernet8
+    Interface      LT Oper    LT Admin    Oper    Admin
+  -----------  -----------  ----------  ------  -------
+    Ethernet8      trained          on      up       up
+  ```
+
 **show interfaces mpls**
 
 This command is used to display the configured MPLS state for the list of configured interfaces.
@@ -3840,6 +3870,7 @@ This sub-section explains the following list of configuration on the interfaces.
 10) type - to set interface type
 11) mpls - To add or remove MPLS operation for the interface
 12) loopback-action - to set action for packet that ingress and gets routed on the same IP interface
+13) link-training - to set interface link-training mode
 
 From 201904 release onwards, the “config interface” command syntax is changed and the format is as follows:
 
@@ -4410,6 +4441,29 @@ Loopback action can be drop or forward.
   admin@sonic:~$ config interface ip loopback-action Ethernet0 forward
 
   ```
+
+**config interface link-training <interface_name> (Versions >= 202211)**
+
+This command is used for setting link-training mode of a interface.
+
+- Usage:
+  ```
+  sudo config interface link-training --help
+  Usage: config interface link-training [OPTIONS] <interface_name> <mode>
+
+    Set interface link-training mode
+
+  Options:
+    -v, --verbose   Enable verbose output
+    -h, -?, --help  Show this message and exit.
+  ```
+
+- Example:
+  ```
+  admin@sonic:~$ sudo config interface link-training Ethernet0 on
+  admin@sonic:~$ sudo config interface link-training Ethernet0 off
+  ```
+
 Go Back To [Beginning of the document](#) or [Beginning of this section](#interfaces)
 
 ## Interface Naming Mode
--- a/scripts/intfutil
+++ b/scripts/intfutil
@@ -799,8 +799,6 @@ class IntfLinkTrainingStatus(object):
                 if lt_admin not in ['on', 'off']:
                     lt_admin = 'N/A'
                 lt_status = state_db_port_status_get(self.db, key, PORT_LINK_TRAINING_STATUS)
-                if lt_status in ['', None]:
-                    lt_status = 'N/A'
                 table.append((key,
                               lt_status.replace('_', ' '),
                               lt_admin,
--- a/scripts/intfutil
+++ b/scripts/intfutil
@@ -797,10 +797,10 @@ class IntfLinkTrainingStatus(object):
                     continue
                 lt_admin = appl_db_port_status_get(self.db, key, PORT_LINK_TRAINING)
                 if lt_admin not in ['on', 'off']:
-                    lt_admin = '-'
+                    lt_admin = 'N/A'
                 lt_status = state_db_port_status_get(self.db, key, PORT_LINK_TRAINING_STATUS)
-                if lt_status in ['N/A', '', None]:
-                    lt_status = 'off'
+                if lt_status in ['', None]:
+                    lt_status = 'N/A'
                 table.append((key,
                               lt_status.replace('_', ' '),
                               lt_admin,
--- a/tests/intfutil_test.py
+++ b/tests/intfutil_test.py
@@ -101,15 +101,15 @@ show_interface_link_training_status_output = """\
   Interface      LT Oper    LT Admin    Oper    Admin
 -----------  -----------  ----------  ------  -------
   Ethernet0  not trained          on    down       up
- Ethernet16          off           -      up       up
- Ethernet24          off           -      up       up
- Ethernet28          off           -      up       up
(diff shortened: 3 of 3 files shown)


adyeung:
@prgeor with all the link training code enhancements merged, pls review and merge the cmd ref

prgeor:
@dgsudharsan can you review?

Junchao-Mellanox (approved):

dgsudharsan (approved):

doc/Command-Reference.md
```diff
@@ -3631,6 +3632,35 @@ This command displays the key fields of the interfaces such as Operational Statu
...
+This command is to display the link-training status of the selected interfaces. If **interface_name** is not specicied, this command shows the link-training status of all interfaces.
+
+- Usage:
+  ```
+  show interfaces link-training status [<interface_name>]
+  ```
+
+- Example:
+  ```
+  admin@sonic:~$ show interfaces link-training status
```
prgeor, line 3646: @ds952811 what will be the output if a platform does not implement these LT configuration and status query?

ds952811: 1. If LT config command is issued and enabled on a platform without LT support

    1-1. Due to the missing SAI_PORT_ATTR_SUPPORTED_LINK_TRAINING_MODE in the SAI commit associated with sonic-sairedis, **LT is now assumed to be available on all platforms right now**
https://github.com/opencomputeproject/SAI/blob/master/inc/saiport.h#L2105

    1-2. If LT is administratively enabled (i.e. Setting 'link_training=on' to PORT_TABLE of APPL_DB), the PortsOrch will issue the request to SAI/syncd and error out
https://github.com/sonic-net/sonic-swss/blob/master/orchagent/portsorch.cpp#L2368

    **NOTE: Gearbox is not supported and is outside the scope of this PR, it's clearly mentioned in the HLD**
    https://github.com/sonic-net/SONiC/blob/master/doc/port_link_training/port-link-training-design.md#scope

2. If LT show command is issued on a platform without LT support

    2-1. For the ports with state info in STATE_DB

```
admin@sonic:~$ sonic-db-dump -n "STATE_DB" -k "PORT*Ethernet0" -y
{
  "PORT_INDEX_TABLE|Ethernet0": {
    "expireat": 1659577967.5626447,
    "ttl": -0.001,
    "type": "hash",
    "value": {
      "ifindex": "149",
      "index": "1"
    }
  },
  "PORT_TABLE|Ethernet0": {
    "expireat": 1659577967.562652,
    "ttl": -0.001,
    "type": "hash",
    "value": {
      "admin_status": "down",
      "mtu": "9100",
      "netdev_oper_status": "down",
      "state": "ok",
      "supported_speeds": "400000,10000,40000,25000,50000,100000,200000"
    }
  }
}admin@sonic:~$show interfaces link-training status Ethernet0
  Interface    LT Oper    LT Admin    Oper    Admin
-----------  ---------  ----------  ------  -------
  Ethernet0        off           -    down     down
```

    2-2. For the ports without state info in STATE_DB

```
admin@sonic:~$ sonic-db-dump -n "STATE_DB" -k "PORT*Ethernet256" -y
{}admin@sonic:~$ show interfaces link-training status Ethernet256
  Interface    LT Oper    LT Admin    Oper    Admin
-----------  ---------  ----------  ------  -------
Ethernet256        off           -     N/A     down

```

prgeor: @ds952811 can we have "LT Oper" to be "N/A" if platform does not support LT so that we know clearly its not supported by the platform?

ds952811: It's not a good idea to use "N/A" for 'LT Oper', the user will get confused, because it's not a valid state when LT is available on the chip. 
e.g.
On a port with LT support, the LT Oper is "on" only when it's administratively enabled, otherwise, it's "off", which means it's disabled. Hence having 'LT Oper=N/A', it actually means using the driver/hardware defaults, and the user will never know what's the accurate hardware state of that particular port.

To resolve this issue, the port initial sequence will need to be revised to retrieve LT status during OA startup, because of the lack of SAI attribute support, the non-broadcom chips is likely to fail and cause issues even when the LT is not administratively enabled. This is exactly something that's outside our expectations.
Hence this is not a good idea.

dgsudharsan: @ds952811 2-2. For the ports without state info in STATE_DB
I believe the output should be N/A for all the fields instead of having a '-'. We cannot blindly assume that LT is on or off in the hardware by default N/A means not applicable which is what we use when fields cannot be determined. If this needs to be updated in code please update

dgsudharsan (changes requested):

ds952811:
The '-' is observed only in 'LT Admin', not in 'LT Oper'
LT Admin: The user admin config, it could be '-' or 'on' or 'off', where '-' means 'not admin configured'
LT Oper: The hardware operational state, it could be either 'on' or 'off'
Hence we don't need 'N/A'.

Please note LT operational state is either ON or OFF, from a hardware perspective, tri-state is not applicable, and the link down is expected if either local or remote on a link with a different LT operational state.

i.e.
LT off + LT off -->  This is the default/generic behavior for platforms with or without LT support.
LT on + LT off -->  Link down is expected, as it gets stuck in not_trained state, and most likely failed with frame_lockup error
LT off + LT on -->  Link down is expected, as it gets stuck in not_trained state, and most likely failed with frame_lockup error
LT on + LT on -->  Link comes up when the link is trained, otherwise, it's down and further status could be available via SAI_PORT_ATTR_LINK_TRAINING_FAILURE_STATUS

On the other hand, on a platform with LT enabled without SONiC acknowledgment or even without SAI get attribute support, the possible solution is as follows:

1. Update PortsOrch::initializePort() to publish the actual hardware operational state into STATE_DB via SAI_PORT_ATTR_LINK_TRAINING_RX_STATUS and SAI_PORT_ATTR_LINK_TRAINING_ENABLE
2. Update PortsOrch::initializePort() to publish the **platform-specific hard coded** LT operational state into STATE_DB
3. Update the vendor-specific SDK config (e.g. config.bcm in the Broadcom chips) to have LT disabled by default

dgsudharsan:
> The '-' is observed only in 'LT Admin', not in 'LT Oper' LT Admin: The user admin config, it could be '-' or 'on' or 'off', where '-' means 'not admin configured' LT Oper: The hardware operational state, it could be either 'on' or 'off' Hence we don't need 'N/A'.
> 
> Please note LT operational state is either ON or OFF, from a hardware perspective, tri-state is not applicable, and the link down is expected if either local or remote on a link with a different LT operational state.
> 
> i.e. LT off + LT off --> This is the default/generic behavior for platforms with or without LT support. LT on + LT off --> Link down is expected, as it gets stuck in not_trained state, and most likely failed with frame_lockup error LT off + LT on --> Link down is expected, as it gets stuck in not_trained state, and most likely failed with frame_lockup error LT on + LT on --> Link comes up when the link is trained, otherwise, it's down and further status could be available via SAI_PORT_ATTR_LINK_TRAINING_FAILURE_STATUS
> 
> On the other hand, on a platform with LT enabled without SONiC acknowledgment or even without SAI get attribute support, the possible solution is as follows:
> 
> 1. Update PortsOrch::initializePort() to publish the actual hardware operational state into STATE_DB via SAI_PORT_ATTR_LINK_TRAINING_RX_STATUS and SAI_PORT_ATTR_LINK_TRAINING_ENABLE
> 2. Update PortsOrch::initializePort() to publish the **platform-specific hard coded** LT operational state into STATE_DB
> 3. Update the vendor-specific SDK config (e.g. config.bcm in the Broadcom chips) to have LT disabled by default

I don't necessarily agree that '-' needs to be displayed for LT Admin platforms without support. I also don't agree that when support is not present you cannot display LT oper as off.
If a specific vendor SAI doesn't support an attribute what we do today is display show as 'N/A' which means not available. We should display LT admin and LT oper as 'N/A' which means those status are not available to be queried from the SAI and hardware. N/A doesn't imply it is enabled or disabled. It simply means OS (SONiC) couldn't query the current state because SAI doesn't support.

ds952811:
> > The '-' is observed only in 'LT Admin', not in 'LT Oper' LT Admin: The user admin config, it could be '-' or 'on' or 'off', where '-' means 'not admin configured' LT Oper: The hardware operational state, it could be either 'on' or 'off' Hence we don't need 'N/A'.
> > Please note LT operational state is either ON or OFF, from a hardware perspective, tri-state is not applicable, and the link down is expected if either local or remote on a link with a different LT operational state.
> > i.e. LT off + LT off --> This is the default/generic behavior for platforms with or without LT support. LT on + LT off --> Link down is expected, as it gets stuck in not_trained state, and most likely failed with frame_lockup error LT off + LT on --> Link down is expected, as it gets stuck in not_trained state, and most likely failed with frame_lockup error LT on + LT on --> Link comes up when the link is trained, otherwise, it's down and further status could be available via SAI_PORT_ATTR_LINK_TRAINING_FAILURE_STATUS
> > On the other hand, on a platform with LT enabled without SONiC acknowledgment or even without SAI get attribute support, the possible solution is as follows:
> > 
> > 1. Update PortsOrch::initializePort() to publish the actual hardware operational state into STATE_DB via SAI_PORT_ATTR_LINK_TRAINING_RX_STATUS and SAI_PORT_ATTR_LINK_TRAINING_ENABLE
> > 2. Update PortsOrch::initializePort() to publish the **platform-specific hard coded** LT operational state into STATE_DB
> > 3. Update the vendor-specific SDK config (e.g. config.bcm in the Broadcom chips) to have LT disabled by default
> 
> I don't necessarily agree that '-' needs to be displayed for LT Admin platforms without support. I also don't agree that when support is not present you cannot display LT oper as off. If a specific vendor SAI doesn't support an attribute what we do today is display show as 'N/A' which means not available. We should display LT admin and LT oper as 'N/A' which means those status are not available to be queried from the SAI and hardware. N/A doesn't imply it is enabled or disabled. It simply means OS (SONiC) couldn't query the current state because SAI doesn't support.

In the case of **'LT Admin'**, it's okay to replace '-' with 'N/A' to indicate a missing user admin config, but it should not be messed up with LT support availability in the SAI library. A dedicated show command for the port features is a better choice, and it's better to create a dedicated PR for this.

In the case of 'LT Oper', while it's not a good idea to use 'N/A' as it's less useful when debugging the link failure, but yes, it's also a bad idea to assume 'LT Oper' is OFF when the state is unavailable in STATE_DB due to the missing SAI support, it's not accurate for chips with the feature support in the SAI .

Honestly, the same issue is applicable to the AutoNeg as well, while it could be enabled by default in the vendor-specific SDK config, the SONiC will never know the actual hardware state and assume both AN and LT are disabled by default.

Again, it's better to review these out-of-bound configurations for SONiC in a dedicated PR for features that are optionally available and could be enabled outside SONiC configuration system.

ds952811:
I double-checked the changes in this PR against sonic-utilities, it does not mention the possible values for 'LT Oper'.
On the other hand, such information should be introduced into the HLD below

https://github.com/sonic-net/SONiC/pull/925/files#diff-116bb65f69b25d13cb802e3e6e41d8260fb79aa67c4f5b4cf69230a7b814aa1bR185

Hence could we have this CLI command manual merged, and then move on the enhancement for the chips without SAI attribute support or out-of-bound SDK configurtation? Thanks.

ds952811:
I think again, and you're right, we could have 'N/A' for both LT Oper/Admin before moving on the enhacnement, as it's in the same repo. with this doc.

ds952811:
@prgeor 
Excuse me, but do you have a chance to review the current changeset? Thanks you

dgsudharsan (dismissed):

dgsudharsan:
@ds952811 Can you please resolve conflicts?

ds952811:
> @ds952811 Can you please resolve conflicts?

Sorry for the late update, it's ready now

dgsudharsan (approved):

prgeor (approved):

PR #2266, opened 2022-07-11
Remove disabled and not loaded services before calling reset-failed and restart services

zbud-msft:
#### What I did

Added logic to remove disabled and not loaded services before calling reset-failed/restart services. Certain services like telemetry can go down and become disabled, which would cause load_minigraph to fail when resetting failed services. Services that are not loaded or disabled should not impact reset or start of other services.

#### How I did it

Added logic to remove services that are disabled or not loaded from the group of listed services for that specific operation. such as resetting failed or restart.

#### How to verify it

Manual testing. Bring down a service such as telemetry via mask or config feature state telemetry disabled, and it should not impact load_minigraph

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -559,12 +559,11 @@ def _get_disabled_services_list():
 
 def _get_not_loaded_services_list(services_list):
     not_loaded_services_list = []
-    with open(os.devnull, 'w') as devnull:
-        for service in services_list:
-            command = "sudo systemctl show -p LoadState --value %s" % service
-            status = subprocess.check_output(command, shell=True).decode()
-            if status != "loaded\n":
-                not_loaded_services_list.append(service)
+    for service in services_list:
+        command = "sudo systemctl show -p LoadState --value %s" % service
+        status = subprocess.check_output(command, shell=True).decode()
+        if status != "loaded\n":
+            not_loaded_services_list.append(service)
 
     return not_loaded_services_list
 


config/main.py
```diff
@@ -601,6 +617,7 @@ def _reset_failed_services():
         'telemetry'
     ]
 
+    _remove_invalid_services(services_to_reset)
```
qiluo-msft, line 620: >_remove_invalid_services [](http://example.com/codeflow?start=4&length=24)

If a service is disabled in FEATURE table, we can still "reset failed", right?

config/main.py
```diff
@@ -623,12 +640,7 @@ def _restart_services():
         'telemetry'
     ]
 
-    disable_services = _get_disabled_services_list()
-
-    for service in disable_services:
-        if service in services_to_restart:
-            services_to_restart.remove(service)
-
+    _remove_invalid_services(services_to_restart)
```
qiluo-msft, line 643: >_remove_invalid_services [](http://example.com/codeflow?start=4&length=24)

Please a testcase which fails old code and pass new code. It could be unit test or sonic-mgmt testcase.

config/main.py
zbud-msft: No, is there a case in which we would want to?

qiluo-msft: I am thinking of below timeline:
1. a service became unhealthy and trapped in a failed status
2. user decided to disabled it in FEATURE table
3. user `config reload`

The expectation should be the "failed" status be "reset", like a fresh new service.

You state "No". Do you mean technically impossible?

zbud-msft: If user disables service in feature table and reloads config.db, hostcfgd will mask and disable the feature. Meaning the service is no longer loaded or active.

If we try to reset-failed the service, reset-failed command will exit with error "service not loaded". 

The fix is that if service is masked we do not want to run reset-failed or restart commands as those will return errors.

qiluo-msft (approved):

qiluo-msft:
What about older or newer branches?

PR #2267, opened 2022-07-14
[sflow] Fix sflow polling_interval.

LGH-12:
Signed-off-by: liuteng <liuteng@asterfusion.com>

#### What I did
If polling interval is not between 5-300, echo "Polling interval must be between 5-300 (0 to disable)". It should return instead of writing to  db.
#### How I did it
Replace click.echo with ctx.fail in "config sflow polling-interval" command.
#### How to verify it
Added unit test.


LGH-12:
@prsunny could you please help to review or assign someone?

PR #2273, opened 2022-07-21
fixing required value validator, and adding replacelane generator

### LINKED ISSUE #2263: [GCU] Ethernet lane update fails, hung on patch sorting step
#### Description
GCU patch application to replace/update lanes for Ethernet interface never terminates, hangs on on `Sorting patch updates` step

1. Apply patch to replace lanes for Ethernet interface

Patch application never terminates

Successful patch application, ethernet interface lane updates go through 

#### Output of `show version`

```
SONiC Software Version: SONiC.internal-202205.57377412-84a9a7f11b
Distribution: Debian 11.3
Kernel: 5.10.0-12-2-amd64
Build commit: 84a9a7f11b
Build date: Wed Jul  6 12:59:01 UTC 2022
Built by: cloudtest@d7f67934c00000B



Platform: x86_64-mlnx_msn2700-r0
HwSKU: Mellanox-SN2700
ASIC: mellanox
ASIC Count: 1
Serial Number: MT2028X35367
Model Number: MSN2700-CS2FO
Hardware Revision: A2
Uptime: 00:27:50 up 2 days,  3:27,  3 users,  load average: 1.85, 1.39, 1.21
Date: Sat 09 Jul 2022 00:27:50
```
See compressed folder for sample patches and DUT config that result in this error 
[test_files.zip](https://github.com/Azure/sonic-utilities/files/9075565/test_files.zip)

ghooo:
#### What I did
Fixes #2263

What is the problem?
* Lanes is create-only field, so the parent `/PORT/Ethernet124` for example needs to be deleted then added.
* Deletion of the `/PORT/Ethernet124` requires deletion of all of its dependencies.
* In the DUT there are around 18 different dependencies
* The search algorithm to find the moves to apply is not very efficient in this case, what it does is find the first dependency delete it. Then it will try to add the dependency back because it thinks this will bring us closer to the final goal config. But then it realizes adding the config back, will bring us back to a visited state. So we move on and try to delete another dependency, after that the search algorithm will try to add the first dependency, but this time the config state was not visited before because that's a new state.

Check the following to better understand:
initial state:
```
port ethernet124
dep1
dep2
dep3
```
Deletion dep1
```
port ethernet124
dep2
dep3
```
Adding dep1 back will rejected, because it will move us back to the initial state
Try and delete dep2
```
port ethernet124
dep3
```
Try and add dep1
```
port ethernet124
dep1
dep3
```
Now the above is a valid state, but it is useless. If you can expand that to 18 dependencies we end up with a lot of tries that are useless.

Solution is to add a custom validator that understand that once a dependency is deleted, do not add it back until we delete the `/PORT/Ethernet124`

Also add a generator for the deletion to make it faster to pick the dependencies to remove.

#### How I did it
Added new validator/generator for handling the lane replacement case. The validator/generator understand that for lane replacement the port and its dependencies need to be removed.

#### How to verify it
UTs TBA

### DIFF
--- a/generic_config_updater/patch_sorter.py
+++ b/generic_config_updater/patch_sorter.py
@@ -399,10 +399,10 @@ class JsonPointerFilter:
         if token == "*":
             matching_keys = config.keys()
         elif token.startswith("*|"):
-            suffix = token[2:]
+            suffix = token[1:] # the suffix will be `|...`
             matching_keys = [key for key in config.keys() if key.endswith(suffix)]
         elif token.endswith("|*"):
-            prefix = token[:-2]
+            prefix = token[:-1] # the prefix will be `...|`
             matching_keys = [key for key in config.keys() if key.startswith(prefix)]
         elif token in config:
             matching_keys = [token]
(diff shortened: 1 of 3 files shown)


generic_config_updater/patch_sorter.py
```diff
@@ -399,10 +399,10 @@ def _get_paths_recursive(self, config, pattern_tokens, matching_tokens, idx, com
         if token == "*":
             matching_keys = config.keys()
         elif token.startswith("*|"):
-            suffix = token[2:]
+            suffix = token[1:] # the suffix will be `|...`
```
ghooo, line 402: Without this change the pattern `Ethernet12|*` would match `Ethernet124|0` and `Ethernet12|0`

generic_config_updater/patch_sorter.py
```diff
@@ -399,10 +399,10 @@ def _get_paths_recursive(self, config, pattern_tokens, matching_tokens, idx, com
...
             matching_keys = [key for key in config.keys() if key.endswith(suffix)]
         elif token.endswith("|*"):
-            prefix = token[:-2]
+            prefix = token[:-1] # the prefix will be `...|`
```
ghooo, line 405: Without this change the pattern `*|10` would match `Ethernet1|10` and `Ethernet1|11110`

generic_config_updater/patch_sorter.py
```diff
@@ -921,6 +981,8 @@ def validate(self, move, diff):
                 for required_path, required_value in data[path]:
                     current_value = self.identifier.get_value_or_default(current_config, required_path)
                     simulated_value = self.identifier.get_value_or_default(simulated_config, required_path)
+                    if simulated_value == None: # Simulated config does not have this value at all.
```
ghooo, line 984: If simulated config does not have the value, it means the move is deleting its parent. in such case skip and assume the mvoe is valid. Let other validators decide.

PR #2277, opened 2022-07-27
[202012] Migrate missed config on cross branch warm upgrade to 202012

vaibhavhd:
#### What I did
Warm upgrade from 201811 (or 201911) to 202012 branch generates config which does not contain some new tables, config attributes introduced in 202012 branch.

1. Missing tables are `RESTAPI`, `TELEMETRY` and `CONSOLE_SWITCH`.
2. Missing attr is `synchronous_mode` in `METADATA`
3. Missing VXLAN entries in APPDB. This one cause warmboot failure. Fix warmboot issue from 201811 to 202012. Cause of issue:

201811 image does not have vxlanmgrd. Thus, during warmboot from 201811 to 202012, vxlanmgrd does not do warmstart and reconciliation as restore_count for vxlanmgrd was not set in 201811 image. Since nothing is reconciled to appdb, vxlanorch has nothing to process. Hence, orchagent bake process misses VXLAN_TUNNEL* entries.
Consequently, syncd reports differences between CURRENT and TEMP view and attempts to remove VXLAN entries in new image. This remove operation fail and swss, syncd both crash.

#### How I did it

Advanced db_migrator version on 202012 branch to handle the new migration logic.

#### How to verify it

Tested on physical device and executed unit tests locally.


yxieca (approved):

stephenxs:
Hi @yxieca @vaibhavhd 
`db_migrator` runs not only during upgrading but also during the configuration initialization.
On a switch installed from ONIE, the `TELEMETRY` table will also be inserted to the `CONFIG_DB` but no keys generated by default. Just want to clarify, is it the user's responsibility to generate the keys in this case?
Thanks.

PR #2278, opened 2022-07-27
Fix issues for sonic_installer upgrade-docker and sonic_installer rollback-docker

Junchao-Mellanox:
#### What I did

Fix issues:
1. upgrade-docker shall not check STATE DB for those container which do not support warm mode
2. rollback-docker sometimes cannot really rollback the container because it cannot get the correct docker image tag

#### How I did it

1. for containers do not support warm mode, ignore STATE_DB check
2. use another command to get docker image tag

#### How to verify it

Manual test.

UT is not created for this change because there are many docker command running for these two CLIs, thus:
1. Mock so many docker command causes the test case very tricky. The case would be even more complicated than the CLIs.
2. UT cannot really test the command in case of so many docker command

Instead, a new sonic-mgmt test case is on the way to cover these two CLI. Will create PR for the sonic-mgmt new test case later.

### DIFF
--- a/sonic_installer/common.py
+++ b/sonic_installer/common.py
@@ -31,6 +31,8 @@ def run_command(command):
     if proc.returncode != 0:
         sys.exit(proc.returncode)
 
+    return out.rstrip("\n")
+
 # Run bash command and return output, raise if it fails
 def run_command_or_raise(argv, raise_exception=True):
     click.echo(click.style("Command: ", fg='cyan') + click.style(' '.join(argv), fg='green'))
--- a/sonic_installer/common.py
+++ b/sonic_installer/common.py
@@ -31,8 +31,6 @@ def run_command(command):
     if proc.returncode != 0:
         sys.exit(proc.returncode)
 
-    return out.rstrip("\n")
-
 # Run bash command and return output, raise if it fails
 def run_command_or_raise(argv, raise_exception=True):
     click.echo(click.style("Command: ", fg='cyan') + click.style(' '.join(argv), fg='green'))
--- a/sonic_installer/main.py
+++ b/sonic_installer/main.py
@@ -134,7 +134,7 @@ def get_docker_tag_name(image):
 
 
 def echo_and_log(msg, priority=LOG_NOTICE, fg=None):
-    if priority >= LOG_ERR:
+    if priority == LOG_ERR:
         # Print to stderr if priority is error
         click.secho(msg, fg=fg, err=True)
     else:
@@ -872,17 +872,17 @@ def upgrade_docker(container_name, url, cleanup_image, skip_check, tag, warm):
     # this is image_id for image with "latest" tag
     image_id_latest = get_container_image_id(image_latest)
 
-    for id in image_id_all:
-        if id != image_id_latest:
-            # Unless requested, the previoud docker image will be preserved
-            if not cleanup_image and id == image_id_previous:
-                continue
-            run_command("docker rmi -f %s" % id)
+    if cleanup_image:
+        # Unless requested, the previoud docker image will be preserved
+        for id in image_id_all:
+            if id != image_id_latest and id == image_id_previous:
+                run_command("docker rmi -f %s" % id)
+                break
 
     exp_state = "reconciled"
     state = ""
     # post warm restart specific procssing for swss, bgp and teamd dockers, wait for reconciliation state.
-    if warm_configured is True or warm:
+    if warm_app_names and (warm_configured is True or warm):
         count = 0
         for warm_app_name in warm_app_names:
             state = ""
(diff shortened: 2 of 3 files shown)


stepanblyschak (approved):

yxieca:
@Junchao-Mellanox can you add unit test?

Junchao-Mellanox:
Hi Ying, these two CLIs runs a lot docker/sonic commands which makes it hard to be unit tested. I have a sonic-mgmt PR to cover the test: https://github.com/sonic-net/sonic-mgmt/pull/6054 . Would that be ok?

Junchao-Mellanox:
Hi @yxieca , could you please provide your comment?

sonic_installer/common.py
```diff
@@ -31,6 +31,8 @@ def run_command(command):
     if proc.returncode != 0:
         sys.exit(proc.returncode)
 
+    return out.rstrip("\n")
```
qiluo-msft, line 34: >return out.rstrip("\n") [](http://example.com/codeflow?start=4&length=23)

To make coverage easier, do you really need this change?

sonic_installer/main.py
```diff
@@ -938,7 +939,8 @@ def rollback_docker(container_name):
     version_tag = ""
     for id in image_id_all:
         if id != image_id_previous:
-            version_tag = get_docker_tag_name(id)
```
qiluo-msft, line 941: >get_docker_tag_name [](http://example.com/codeflow?start=26&length=19)

you can change the implementation of `get_docker_tag_name()`, and this function should be easy to unit test, you still need some mock.

sonic_installer/main.py
```diff
@@ -872,17 +873,17 @@ def upgrade_docker(container_name, url, cleanup_image, skip_check, tag, warm):
...
-    for id in image_id_all:
-        if id != image_id_latest:
-            # Unless requested, the previoud docker image will be preserved
-            if not cleanup_image and id == image_id_previous:
-                continue
-            run_command("docker rmi -f %s" % id)
+    if cleanup_image:
+        # Unless requested, the previoud docker image will be preserved
+        for id in image_id_all:
+            if id != image_id_latest and id == image_id_previous:
```
qiluo-msft, line 879: >id != image_id_latest and id == image_id_previous [](http://example.com/codeflow?start=15&length=49)

This check is weird, seems like `image_id_latest != image_id_previous`, and not related to the loop var. Could you double check?

Junchao-Mellanox: will fix

sonic_installer/common.py
Junchao-Mellanox: I need the return value on line 942.

sonic_installer/main.py
Junchao-Mellanox: Hi qi, thanks for the suggestion. I would to rollback this change because I found the real issue is in sonic build system. Here is my analyze:

During my previous manual test, I found `get_docker_tag_name()` cannot return the correct tag name of docker.  The return value of `get_docker_tag_name()` does not equal to output of `docker images`.

However, I re-visit the logic today, and I found:
1. `get_docker_tag_name()` tries to get the docker tag by querying a docker label specified by sonic make file.
```
cmd = "docker inspect --format '{{.ContainerConfig.Labels.Tag}}' " + image
```
2. This label is added in slave.mk. See https://github.com/sonic-net/sonic-buildimage/blob/736c739bf492a73c60cfca000b853c040f53a104/slave.mk#L895 
```
	docker build --squash --no-cache \
...
		--label Tag=$(SONIC_IMAGE_VERSION) \
```
3. The real docker tag is added in build_debian.sh. See https://github.com/sonic-net/sonic-buildimage/blob/736c739bf492a73c60cfca000b853c040f53a104/files/build_templates/sonic_debian_extension.j2#L697
```
sudo LANG=C DOCKER_HOST="$DOCKER_HOST" chroot $FILESYSTEM_ROOT docker tag {{imagename}}:latest {{imagename}}:"${SONIC_IMAGE_VERSION}"
```

In theory, item#2 and item#3 shall get the same value of ${SONIC_IMAGE_VERSION}. But it doesn't. I got from build log:
1. For item#2, it prints:
```
docker-database.gz.log:Step 23/25 : LABEL Tag=syslog-rate-limit.0-dirty-20220804.115836
```
2. For item#3, it prints:
```
sonic-mellanox.bin.log:+ sudo LANG=C DOCKER_HOST= chroot ./fsroot-mellanox docker tag docker-database:latest docker-database:syslog-rate-limit.0-dirty-20220804.165951
```

As you can see, value of SONIC_IMAGE_VERSION is different in the same build. I suppose SONIC_IMAGE_VERSION  value shall always be the same in the same build.

sonic_installer/common.py
Junchao-Mellanox: Since I will rollback line 942, I don't need this change any more.

qiluo-msft (dismissed):

liat-grozovik:
@qiluo-msft  could you please help to approve again following the improved coverage?

qiluo-msft (approved):

PR #2281, opened 2022-07-31
Vnet_route_check Vxlan tunnel route update.

siqbal1986:
Modified Vnet_route_check to ignore the Vxlan tunnel routes which uses default VRF.

#### What I did
Updated the vnet_route_check script to handle the vxlan tunnel routes.
#### How I did it
By checking the scope of the tunnel route, if it is default then it is treated in a different way. Tunnel routes coming from App_db are compared with ASIC_db default_VRF routes and their presence is verified.
For routes coming from ASIC_db with default_VRF, the comparison is not done with the routes form App_db as ASIC_DB carries all other the routes.
For normal vnet routes, the functionality of the script is unaffected. 
#### How to verify it
run the script after adding a vxlan tunnel.

### DIFF
--- a/scripts/vnet_route_check.py
+++ b/scripts/vnet_route_check.py
@@ -213,7 +213,7 @@ def get_vnet_routes_from_app_db():
     for vnet_route_db_key in vnet_routes_db_keys:
         vnet_route_list = vnet_route_db_key.split(':',1)
         vnet_name = vnet_route_list[0]
-        vnet_route = vnet_route_list[1:]
+        vnet_route = vnet_route_list[1]
 
         if vnet_name not in vnet_routes:
             vnet_routes[vnet_name] = {}
@@ -296,7 +296,7 @@ def check_routes_with_default_vrf(vnet_name, vnet_attrs, routes_1, routes):
     for vnet_route in vnet_attrs['routes']:
         ispresent = False
         for vnet_name_other, vnet_attrs_other in routes_1.items():
-            if vnet_route[0] in vnet_attrs_other['routes']:
+            if vnet_route in vnet_attrs_other['routes']:
                 ispresent = True
         if not ispresent:
             if vnet_name not in routes:
--- a/scripts/vnet_route_check.py
+++ b/scripts/vnet_route_check.py
@@ -47,7 +47,7 @@ Format of differences output:
 
 RC_OK = 0
 RC_ERR = -1
-
+default_vrf_oid = ""
 
 report_level = syslog.LOG_ERR
 write_to_syslog = True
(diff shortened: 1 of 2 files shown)


prsunny:
We should not ignore vnet routes in the script. Scrip must handle the vnet routes in default scope and proceed with the check

prsunny:
can you please add some tests and also fix the test failure?

scripts/vnet_route_check.py
```diff
@@ -219,15 +219,24 @@ def get_vnet_routes_from_app_db():
             vnet_routes[vnet_name] = {}
             vnet_routes[vnet_name]['routes'] = []
 
-            intf = vnet_intfs[vnet_name][0]
-            vnet_routes[vnet_name]['vrf_oid'] = vnet_vrfs.get(intf, 'None')
+            if vnet_name not in vnet_intfs:
+                # this route has no vnet_intf and may be part of default VRF.
+                vnet_table = swsscommon.Table(db, 'VNET_TABLE')
+                scope = vnet_table.get(vnet_name)[1][1][1]
```
prsunny, line 225: what is this index [1][1][1] ? Can you provide an example?

scripts/vnet_route_check.py
```diff
@@ -237,10 +246,12 @@ def get_vnet_routes_from_asic_db():
 
     vnet_vrfs = get_vrf_entries()
     vnet_vrfs_oids = [vnet_vrfs[k] for k in vnet_vrfs]
+    vnet_vrfs_oids.append(default_vrf_oid) 
 
     vnet_intfs = get_vnet_intfs()
 
     vrf_oid_to_vnet_map = {}
+    vrf_oid_to_vnet_map[default_vrf_oid] = 'default_VRF'
```
prsunny, line 254: I don't see this being used. Can you confirm?

scripts/vnet_route_check.py
```diff
@@ -326,12 +350,17 @@ def main():
     # Don't run VNET routes consistancy logic if there is no VNET configuration
     if not check_vnet_cfg():
         return rc
+    asic_db = swsscommon.DBConnector('ASIC_DB', 0)
+    virtual_router = swsscommon.Table(asic_db, 'ASIC_STATE:SAI_OBJECT_TYPE_VIRTUAL_ROUTER')
+    default_vrf_oid = ""
+    if virtual_router.getKeys() != []:
+        default_vrf_oid = virtual_router.getKeys()[0]
```
prsunny, line 357: please make it as a global variable so you don't have to pass it in multiple places and use as "self.xxx"

scripts/vnet_route_check.py
siqbal1986: this looking up the scope=default but i think I can improve it.

scripts/vnet_route_check.py
siqbal1986: This is used as a placeholder, as the data structure keeps track of the routes by means of the vnet interface.

scripts/vnet_route_check.py
siqbal1986: ill do so.

siqbal1986: done

prsunny (approved):

prsunny (approved):

PR #2289, opened 2022-08-02
Fix version in db_migrator  for `PORT_QOS_MAP|global`

bingwang-ms:
#### What I did
This PR is to fix the db_migrator version for table `PORT_QOS_MAP|global`.

Because the migrator version for table `PORT_QOS_MAP|global` is `version_2_0_0` in `202012` branch (PR #2215), we should use the same version in `master` and `202205` branch to be compatible.

#### How I did it
Change the migrator version for table `PORT_QOS_MAP|global` from `version_3_0_5` to `version_2_0_0` .

#### How to verify it
Verified by UT.
```
collected 78 items                                                                                                                                                                                    

tests/db_migrator_test.py ..............................................................................                                                                                        [100%]

===================================================================================== 78 passed in 21.88 seconds ======================================================================================
```

### DIFF
--- a/scripts/db_migrator.py
+++ b/scripts/db_migrator.py
@@ -44,7 +44,7 @@ class DBMigrator():
                      none-zero values.
               build: sequentially increase within a minor version domain.
         """
-        self.CURRENT_VERSION = 'version_3_0_6'
+        self.CURRENT_VERSION = 'version_3_0_5'
 
         self.TABLE_NAME      = 'VERSIONS'
         self.TABLE_KEY       = 'DATABASE'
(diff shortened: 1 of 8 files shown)


yxieca (approved):

scripts/db_migrator.py
```diff
@@ -632,8 +632,26 @@ def version_1_0_6(self):
...
+    def version_2_0_0(self):
+        """
+        Version 2_0_0
+        """
+        log.log_info('Handling version_2_0_0')
+        self.migrate_port_qos_map_global()
+        self.set_version('version_2_0_1')
+        return 'version_2_0_1'
+
+    def version_2_0_1(self):
```
vaibhavhd, line 648: Here version 2_0_1 is making jump to version 3_0_0. However in 202012 branch version 2_0_1 makes jump to 2_0_2.

May be we should bring in changes made in https://github.com/sonic-net/sonic-utilities/pull/2277/files to master? That will keep different branches in sync.

vaibhavhd: Unrelated to your changes, but general comment:
Inserting versions in between two existing versions also has another challenge - it causes two images (one from past and one latest) to have same DB versions, but different DBs.
For example, some images built with DB version 3_0_0 will have none of 202012 migration, and some will have all of the 202012 migration logic. This causes inconsistencies.

vaibhavhd: Concluded in offline discussion. Solution is to port changes from 2277 to master after this PR merges.

vaibhavhd (approved):

PR #2295, opened 2022-08-04
Fix GCU bug when backend service modifying config

wen587:
Modifying PORT `admin_status` will result in backend service changing the running config. e.g. buffermgrd will modify `BUFFER_PG|EthernetXX|3-4` when one of the EthernetXX in `PORT` table admin down.
This PR is to skip comparison of config that is possiblly being modified by backend service

#### What I did
Fixes https://github.com/sonic-net/sonic-buildimage/issues/11576
#### How I did it
Add a workaround to only compare config without backend service impact.
#### How to verify it
Manual test on specific platform and check operation success.

### DIFF
--- a/generic_config_updater/change_applier.py
+++ b/generic_config_updater/change_applier.py
@@ -71,6 +71,11 @@ class ChangeApplier:
 
     def __init__(self):
         self.config_db = get_config_db()
+        self.backend_tables = [
+            "BUFFER_PG",
+            "BUFFER_PROFILE",
+            "FLEX_COUNTER_TABLE"
+        ]
         if (not ChangeApplier.updater_conf) and os.path.exists(UPDATER_CONF_FILE):
             with open(UPDATER_CONF_FILE, "r") as s:
                 ChangeApplier.updater_conf = json.load(s)
@@ -142,12 +147,21 @@ class ChangeApplier:
         ret = self._services_validate(run_data, upd_data, upd_keys)
         if not ret:
             run_data = self._get_running_config()
+            upd_data = self.get_config_without_backend_tables(upd_data)
+            run_data = self.get_config_without_backend_tables(run_data)
             if upd_data != run_data:
                 self._report_mismatch(run_data, upd_data)
                 ret = -1
         if ret:
             log_error("Failed to apply Json change")
         return ret
+
+
+    def get_config_without_backend_tables(self, data):
+        cropped_data = {
+            k: v for k, v in data.items() if k not in self.backend_tables
+        }
+        return cropped_data
 
 
     def _get_running_config(self):
--- a/generic_config_updater/change_applier.py
+++ b/generic_config_updater/change_applier.py
@@ -147,8 +147,8 @@ class ChangeApplier:
         ret = self._services_validate(run_data, upd_data, upd_keys)
         if not ret:
             run_data = self._get_running_config()
-            upd_data = self.get_config_without_backend_tables(upd_data)
-            run_data = self.get_config_without_backend_tables(run_data)
+            self.remove_backend_tables_from_config(upd_data)
+            self.remove_backend_tables_from_config(run_data)
             if upd_data != run_data:
                 self._report_mismatch(run_data, upd_data)
                 ret = -1
@@ -157,11 +157,9 @@ class ChangeApplier:
         return ret
 
 
-    def get_config_without_backend_tables(self, data):
-        cropped_data = {
-            k: v for k, v in data.items() if k not in self.backend_tables
-        }
-        return cropped_data
+    def remove_backend_tables_from_config(self, data):
+        for key in self.backend_tables:
+            data.pop(key, None)
 
 
     def _get_running_config(self):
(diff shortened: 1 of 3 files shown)


wen587:
Hi @liuh-80 , could you help review if only these two tables `[BUFFER_PG, BUFFER_PROFILE]` will be impacted by buffermgrd?

qiluo-msft:
@DavidZagury Could you help review this PR?

generic_config_updater/change_applier.py
```diff
@@ -71,6 +71,11 @@ class ChangeApplier:
 
     def __init__(self):
         self.config_db = get_config_db()
+        self.backend_tables = [
+            "BUFFER_PG",
+            "BUFFER_PROFILE",
+            "FLEX_COUNTER_TABLE"
```
qiluo-msft, line 77: >FLEX_COUNTER_TABLE [](http://example.com/codeflow?start=13&length=18)

Which process will modify FLEX_COUNTER_TABLE?

wen587: I found it is being modified after load_minigraph, but not sure if other config changes would modify this table.
Maybe flex_counter_manager https://github.com/sonic-net/sonic-swss/blob/master/orchagent/flex_counter/flex_counter_manager.cpp#L112?
```
admin@vlab-01:~$ redis-cli monitor | grep " \[4 "  | grep 'HSET\|HMSET'
...
1652252491.657340 [4 127.0.0.1:48348] "HMSET" "FLEX_COUNTER_TABLE|PORT" "FLEX_COUNTER_STATUS" "enable"
1652252491.665481 [4 127.0.0.1:48348] "HMSET" "FLEX_COUNTER_TABLE|RIF" "FLEX_COUNTER_STATUS" "enable"
1652252491.673144 [4 127.0.0.1:48348] "HMSET" "FLEX_COUNTER_TABLE|QUEUE" "FLEX_COUNTER_STATUS" "enable"
1652252491.677734 [4 127.0.0.1:48348] "HMSET" "FLEX_COUNTER_TABLE|PFCWD" "FLEX_COUNTER_STATUS" "enable"
1652252491.689264 [4 127.0.0.1:48348] "HMSET" "FLEX_COUNTER_TABLE|PG_WATERMARK" "FLEX_COUNTER_STATUS" "enable"
1652252491.692459 [4 127.0.0.1:48348] "HMSET" "FLEX_COUNTER_TABLE|PG_DROP" "FLEX_COUNTER_STATUS" "enable"
1652252491.695281 [4 127.0.0.1:48348] "HMSET" "FLEX_COUNTER_TABLE|QUEUE_WATERMARK" "FLEX_COUNTER_STATUS" "enable"
1652252491.699036 [4 127.0.0.1:48348] "HMSET" "FLEX_COUNTER_TABLE|BUFFER_POOL_WATERMARK" "FLEX_COUNTER_STATUS" "enable"
1652252491.703090 [4 127.0.0.1:48348] "HMSET" "FLEX_COUNTER_TABLE|PORT_BUFFER_DROP" "FLEX_COUNTER_STATUS" "enable"
1652252491.706964 [4 127.0.0.1:48348] "HMSET" "FLEX_COUNTER_TABLE|ACL" "FLEX_COUNTER_DELAY_STATUS" "false" "FLEX_COUNTER_STATUS" "enable" "POLL_INTERVAL" "10000"
```

qiluo-msft (approved):
LGTM.

generic_config_updater/change_applier.py
```diff
@@ -150,6 +157,13 @@ def apply(self, change):
         return ret
 
 
+    def get_config_without_backend_tables(self, data):
+        cropped_data = {
+            k: v for k, v in data.items() if k not in self.backend_tables
+        }
+        return cropped_data
```
DavidZagury, line 164: I don't see a reason to iterate over the entire database to remove a specific list of keys, we can directly remove the tables that should not be tested.

Something like this:
```suggestion
    def get_config_without_backend_tables(self, data):
        for key in self.backend_tables:
            data.pop(key, None)
```

DavidZagury:
@qiluo-msft I tested this PR on my setup and it fixed the issue. I did add one comment that I think could have a slight improvement in the performance of this solution .

@stephenxs do you know of any other tables that should be ignored during the GCU validation check?

generic_config_updater/change_applier.py
wen587: Modified. Please review again. Thanks

DavidZagury:
LGTM

DavidZagury (approved):
LGTM

stephenxs (approved):

stephenxs:
> @qiluo-msft I tested this PR on my setup and it fixed the issue. I did add one comment that I think could have a slight improvement in the performance of this solution .
> 
> @stephenxs do you know of any other tables that should be ignored during the GCU validation check?

Hi
I'm not aware of other tables that can be updated by manager daemons. Now that we have an infra to ignore such tables, we can add them to the list in case we find some new.

neethajohn (approved):

qiluo-msft (approved):

PR #2299, opened 2022-08-07
Convert IPv6 addresses to lowercase in apply-patch

dbarashinvd:
Fixes https://github.com/sonic-net/sonic-buildimage/issues/11622

#### What I did
Convert IPv6 addresses to lowercase in apply-patch for remove op

#### How I did it
python regex on 'remove' op in JSON patch input file

#### How to verify it
manual test of created bug

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1357,6 +1357,20 @@ def apply_patch(ctx, patch_file_path, format, dry_run, ignore_non_yang_tables, i
             patch_as_json = json.loads(text)
             patch = jsonpatch.JsonPatch(patch_as_json)
 
+        # convert IPv6 addresses to lowercase
+        for patch_line in patch:
+            if 'remove' == patch_line['op']:
+                match = re.search(r"(?P<prefix>/INTERFACE/\w+\|)(?P<ipv6_address>([a-fA-F0-9]{0,4}[:~]|::){1,7}[a-fA-F0-9]{0,4})"
+                                    "(?P<suffix>.*)", str.format(patch_line['path']))
+                if match:
+                    prefix = match.group('prefix')
+                    ipv6_address_str = match.group('ipv6_address')
+                    suffix = match.group('suffix')
+                    ipv6_address_str = ipv6_address_str.lower()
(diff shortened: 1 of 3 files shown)


qiluo-msft:
@isabelmsft Please help review this PR.

config/main.py
```diff
@@ -1357,6 +1357,20 @@ def apply_patch(ctx, patch_file_path, format, dry_run, ignore_non_yang_tables, i
             patch_as_json = json.loads(text)
             patch = jsonpatch.JsonPatch(patch_as_json)
 
+        # convert IPv6 addresses to lowercase
+        for l in patch:
```
isabelmsft, line 1361: Can we make `l` more descriptive?

dbarashinvd: done.
changed to patch_line

dprital:
@isabelmsft - comment was addressed. Can you please approve ?

isabelmsft (approved):

liat-grozovik (approved):

wen587:
@dbarashinvd 
There are some issues when GCU tries to remove the interface IP in tests, such as add rack test.

In the initial config, the INTERFACE's IPv6 was all loaded in uppercase. 
```
    "INTERFACE": {
        "Ethernet68": {},
        "Ethernet68|10.0.0.34/31": {},
        "Ethernet68|FC00::45/126": {},
        "Ethernet72": {},
        "Ethernet72|10.0.0.36/31": {},
        "Ethernet72|FC00::49/126": {},
```
GCU will never be able to remove that IP because IPv6 was always translated to lowercase due to this PR. It reported error `can't remove a non-existent object`, thus making GCU fail.

This PR is to deal with this issue: https://github.com/sonic-net/sonic-buildimage/issues/11622. 

Although config CLI always translates uppercase to lowercase when adding an IP, user can have two choices to remove that IP: One is to use config CLI to remove that IP no matter uppercase or lowercase. Another way is to use GCU. In order to use GCU, the user has to check the IP format saved in ConfigDB because GCU operation does differentiate between uppercase and lowercase.
@qiluo-msft  for vis.

PR #2305, opened 2022-08-10
Add verification for override

wen587:
#### What I did
Add Yang verification for config override-config-table
#### How I did it
Make 3 step verification: running config, golden input, final config
#### How to verify it
Run unit test.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1860,7 +1860,7 @@ def override_config_table(db, input_config_db, dry_run):
 
     updated_config = update_config(current_config, config_input)
 
-    yang_enabled = is_yang_config_verification_enabled(config_db)
+    yang_enabled = device_info.is_yang_config_validation_enabled(config_db)
     if yang_enabled:
         # The ConfigMgmt will load YANG and running
         # config during initialization.
--- a/config/main.py
+++ b/config/main.py
@@ -12,7 +12,6 @@ import subprocess
 import sys
 import time
 import itertools
-import sonic_yang
 import copy
 
 from collections import OrderedDict
@@ -47,7 +46,7 @@ from . import nat
 from . import vlan
 from . import vxlan
 from . import plugins
-from .config_mgmt import ConfigMgmtDPB
+from .config_mgmt import ConfigMgmtDPB, ConfigMgmt
 from . import mclag
 from . import syslog
 
@@ -79,7 +78,6 @@ INTF_KEY = "interfaces"
 DEFAULT_GOLDEN_CONFIG_DB_FILE = '/etc/sonic/golden_config_db.json'
 
 INIT_CFG_FILE = '/etc/sonic/init_cfg.json'
-YANG_DIR = '/usr/local/yang-models'
 
 DEFAULT_NAMESPACE = ''
 CFG_LOOPBACK_PREFIX = "Loopback"
@@ -1864,13 +1862,18 @@ def override_config_table(db, input_config_db, dry_run):
 
     yang_enabled = is_yang_config_verification_enabled(config_db)
     if yang_enabled:
-        sy = sonic_yang_with_loaded_models(YANG_DIR)
-        # Validate running config
-        validate_config_by_yang(sy, current_config, "current_config")
+        # The ConfigMgmt will load YANG and validate
+        # runnning config during initialization.
+        try:
+            cm = ConfigMgmt()
+        except Exception as ex:
+            click.secho("Failed to validate running config. Error: {}".format(ex), fg="magenta")
+            sys.exit(1)
+
         # Validate input config
-        validate_config_by_yang(sy, config_input, "config_input")
+        validate_config_by_cm(cm, config_input, "config_input")
         # Validate updated whole config
-        validate_config_by_yang(sy, updated_config, "updated_config")
+        validate_config_by_cm(cm, updated_config, "updated_config")
 
     if dry_run:
         print(json.dumps(updated_config, sort_keys=True,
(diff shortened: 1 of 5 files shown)


config/main.py
```diff
@@ -1852,27 +1855,71 @@ def override_config_table(db, input_config_db, dry_run):
...
+
+    updated_config = update_config(current_config, config_input)
+
+    yang_enabled = is_yang_config_verification_enabled(config_db)
+    if yang_enabled:
+        sy = sonic_yang_with_loaded_models(YANG_DIR)
+        # Validate running config
+        validate_config_by_yang(sy, current_config)
+        # Validate input config
+        validate_config_by_yang(sy, config_input)
```
isabelmsft, line 1871: Why do we need to validate the input config and also the updated whole config? Does updated whole config not include input config?

qiluo-msft: It is an enforcement to input config. The assumption is that input config should be self-contained, and do not need anything more in current_config on device.

With this assumption, we could more easily to tell a process failure is due to invalid input or sonic config error or logic bug.

qiluo-msft:
@praveen-li  Could you help review?

config/main.py
```diff
@@ -1852,27 +1855,71 @@ def override_config_table(db, input_config_db, dry_run):
...
+
+
+def validate_config_by_yang(sy, config_json):
+    try:
+        tmp_config_json = copy.deepcopy(config_json)
+        sy.loadData(tmp_config_json)
+        sy.validate_data_tree()
+    except sonic_yang.SonicYangException as ex:
+        click.secho("Failed to validate config. Error: {}".format(ex), fg="red")
+        sys.exit(1)
```
qiluo-msft, line 1900: These 2 functions are for common purpose. Possible to reuse SonicYangExtMixin or extend libraries?

config/main.py
```diff
@@ -1852,27 +1855,71 @@ def override_config_table(db, input_config_db, dry_run):
...
-        for table in config_input:
-            config_db.delete_table(table)
-        click.echo("Overriding input config to configDB ...")
-        data = sonic_cfggen.FormatConverter.output_to_db(config_input)
-        config_db.mod_config(data)
-        click.echo("Overriding completed. No service is restarted.")
+        override_config_db(config_db, config_input)
+
+
+def is_yang_config_verification_enabled(config_db):
```
qiluo-msft, line 1882: >is_yang_config_verification_enabled [](http://example.com/codeflow?start=4&length=35)

This may be reused by @isabelmsft. Find a better place?

config/main.py
wen587: You mean `loadData` and `validate_data_tree`? If so, I think it is not necessary to combine two into one.

config/main.py
wen587: Draft a PR here. https://github.com/sonic-net/sonic-buildimage/pull/11715

config/main.py
qiluo-msft: I do not mean to combine two functions into one. Is it possible to move these functions to a common library and make it reusable? I see they are similar to the features in `SonicYangExtMixin`.

config/main.py
isabelmsft: In that case, should we add some logging to `validate_config_by_yang()` (or in between calls) so we know which config caused the error?

qiluo-msft: Agree with @isabelmsft, after adding this new logging, also need test the logging.

wen587: Added.

config/main.py
wen587: Test by using class ConfigMgmt, there are some unwanted logs:
```
def validate_config_by_yang(sy, config_json, jname):
    tmp_config_json = copy.deepcopy(config_json)
    cm = ConfigMgmt()
    try:
        cm.loadData(tmp_config_json)
        cm.validateConfigData():
    except Exception as ex:
         click.secho("Failed to validate {}. Error: {}".format(jname, ex), fg="magenta")
         sys.exit(1)
---below 1) and 2) together are whole print---
  ---1) Default class print---
    sonic_yang(6):Note: Below table(s) have no YANG models: CONSOLE_SWITCH, DEVICE_NEIGHBOR_METADATA, DHCP_SERVER, RESTAPI, SNMP, SNMP_COMMUNITY
    sonic_yang(6):Note: Below table(s) have no YANG models: CONSOLE_SWITCH, DEVICE_NEIGHBOR_METADATA, DHCP_SERVER, RESTAPI, SNMP, SNMP_COMMUNITY
    sonic_yang(6):Note: Below table(s) have no YANG models: CONSOLE_SWITCH, DEVICE_NEIGHBOR_METADATA, DHCP_SERVER, RESTAPI, SNMP, SNMP_COMMUNITY
    libyang[0]: Invalid value "denable" in "synchronous_mode" element. (path: /sonic-device_metadata:sonic-device_metadata/DEVICE_METADATA/localhost/synchronous_mode)
    sonic_yang(3):Data Loading Failed:Invalid value "denable" in "synchronous_mode" element.
  ---2) My Print---
    Failed to validate config_input. Error: Data Loading Failed
    Invalid value "denable" in "synchronous_mode" element.
```

config/main.py
```diff
@@ -1852,27 +1853,70 @@ def override_config_table(db, input_config_db, dry_run):
...
+    sonic_cfggen.FormatConverter.to_serialized(current_config)
+
+    updated_config = update_config(current_config, config_input)
+
+    yang_enabled = is_yang_config_verification_enabled(config_db)
+    if yang_enabled:
+        # The ConfigMgmt will load YANG and validate
+        # runnning config during initialization.
+        try:
+            cm = ConfigMgmt()
```
qiluo-msft, line 1868: >ConfigMgmt [](http://example.com/codeflow?start=17&length=10)

There is an optional parameter `allowTablesWithoutYang=True`. Should it be False?

config/main.py
qiluo-msft: Let's track the verbose output issue as a github issue at low priority.

config/main.py
qiluo-msft: Can you calling function in that PR now?

config/main.py
```diff
@@ -1852,27 +1853,70 @@ def override_config_table(db, input_config_db, dry_run):
...
+        except Exception as ex:
+            click.secho("Failed to validate running config. Error: {}".format(ex), fg="magenta")
```
isabelmsft, line 1870: This try statement initializes class and loads data, but I don't believe validation is done until you call `validate_config_by_cm` 
Should the error message be about initialization and loading instead of validation?

config/main.py
wen587: Done. https://github.com/sonic-net/sonic-utilities/issues/2322

config/main.py
wen587: No. Current call that funtion will fail PR checker. Because the VM being tested now has not included that PR.

config/main.py
wen587: I prefer no. That could be set to `False` when YANG covers all CONFIG_DB. Otherwise it will fail if CONFIG_DB has config without yang.

config/main.py
wen587: You are right. It only load_data during the initialization. I will add one line for validation.

qiluo-msft (approved):

Issue #2312, opened 2022-08-15
"show int status" command fails on multi-asic/chassis devices.


judyjoseph:
#### Description

The sanity was failing for chassis & multi-asic devices with the following error in pre-sanity 
```
E                   "msg": "Command failed rc = 1, out = Invalid port 'Ethernet-IB0'\n, err = Traceback (most recent call last):\n  File \"/usr/local/bin/intfutil\", line 838, in <module>\n    main()\n  File \"/usr/local/bin/intfutil\", line 821, in main\n    interface_stat.display_intf_status()\n  File \"/usr/local/bin/intfutil\", line 448, in display_intf_status\n    self.get_intf_status()\n  File \"/usr/local/lib/python3.9/dist-packages/utilities_common/multi_asic.py\", line 151, in wrapped_run_on_all_asics\n    func(self,  *args, **kwargs)\n  File \"/usr/local/bin/intfutil\", line 535, in get_intf_status\n    self.table += self.generate_intf_status()\n  File \"/usr/local/bin/intfutil\", line 479, in generate_intf_status\n    port_oper_speed_get(self.db, key),\n  File \"/usr/local/bin/intfutil\", line 204, in port_oper_speed_get\n    optics_type = port_optics_get(db, intf_name, PORT_OPTICS_TYPE)\n  File \"/usr/local/bin/intfutil\", line 224, in port_optics_get\n    if is_rj45_port(intf_name):\n  File \"/usr/local/lib/python3.9/dist-packages/utilities_common/platform_sfputil_helper.py\", line 134, in is_rj45_port\n    physical_port = logical_port_name_to_physical_port_list(port_name)[0]\nTypeError: 'NoneType' object is not subscriptable\n"
```

The "show int status" command also fails 

```
admin@str2--lc1-1:~$ show int status
Traceback (most recent call last):
  File "/usr/local/bin/intfutil", line 838, in <module>
    main()
  File "/usr/local/bin/intfutil", line 821, in main
    interface_stat.display_intf_status()
  File "/usr/local/bin/intfutil", line 448, in display_intf_status
    self.get_intf_status()
  File "/usr/local/lib/python3.9/dist-packages/utilities_common/multi_asic.py", line 151, in wrapped_run_on_all_asics
    func(self,  *args, **kwargs)
  File "/usr/local/bin/intfutil", line 535, in get_intf_status
    self.table += self.generate_intf_status()
[...]

PR #2313, opened 2022-08-16
Fix issue: exception in is_rj45_port in multi ASIC env

stephenxs:
#### What I did
Fix issue #2312. On some platforms, `logical_port_name_to_physical_port_list` doesn't return a valid port name list, which causes exception in `is_rj45_port`.
Fix it by catching the exception and treating it as a non-RJ45 port.

### DIFF
--- a/utilities_common/platform_sfputil_helper.py
+++ b/utilities_common/platform_sfputil_helper.py
@@ -131,10 +131,10 @@ def is_rj45_port(port_name):
         if not platform_porttab_mapping_read:
             platform_sfputil_read_porttab_mappings()
 
-        physical_port = logical_port_name_to_physical_port_list(port_name)[0]
         try:
+            physical_port = logical_port_name_to_physical_port_list(port_name)
             port_type = platform_chassis.get_port_or_cage_type(physical_port)
-        except NotImplementedError as e:
+        except Exception as e:
             port_type = None
 
         return port_type == platform_sfp_base.SFP_PORT_TYPE_BIT_RJ45
--- a/utilities_common/platform_sfputil_helper.py
+++ b/utilities_common/platform_sfputil_helper.py
@@ -133,7 +133,8 @@ def is_rj45_port(port_name):
 
         try:
             physical_port = logical_port_name_to_physical_port_list(port_name)
-            port_type = platform_chassis.get_port_or_cage_type(physical_port)
+            if physical_port:
+                port_type = platform_chassis.get_port_or_cage_type(physical_port[0])
         except Exception as e:
             port_type = None
 


stephenxs:
@prgeor @judyjoseph could you please review it? thanks.

utilities_common/platform_sfputil_helper.py
```diff
@@ -131,10 +131,10 @@ def is_rj45_port(port_name):
         if not platform_porttab_mapping_read:
             platform_sfputil_read_porttab_mappings()
 
-        physical_port = logical_port_name_to_physical_port_list(port_name)[0]
         try:
+            physical_port = logical_port_name_to_physical_port_list(port_name)
             port_type = platform_chassis.get_port_or_cage_type(physical_port)
```
judyjoseph, line 136: This API, get_port_or_cage_type() I find is not implemented by all vendors which will result in failures and setting port_type=None. This will be an issues still.

Can we just check if this API logical_port_name_to_physical_port_list(port_name), returns back None -- if so set the port_type=None ?

stephenxs: Thanks 
API get_port_or_cage_type() returning None is not an issue. In this case we just treat it as a normal xSFP port.
We expect a valid type only for RJ45 ports.
If we are confident that the API logical_port_name_to_physical_port_list will never raise an exception in any case, we can just check the return value without putting it into a try block
But I am not sure whether it’s safe enough so I suggest keeping as it is.
How do you think?
@judyjoseph

judyjoseph: Ok, it is fine then -- a value of None is treated as a normal xSFP port works ok.

judyjoseph: Another question **physicalPort** is of "list" datatype
will the API get_port_or_cage_type accept list datatype as input parameter?

stephenxs: > Another question **physicalPort** is of "list" datatype will the API get_port_or_cage_type accept list datatype as input parameter?

Nice catch. Fixed and verified on both RJ45 and xSFP ports.
```
admin@mtbc-sonic-03-2700:~$ show int st
     Interface            Lanes    Speed    MTU    FEC    Alias            Vlan    Oper    Admin    Type    Asym PFC
--------------  ---------------  -------  -----  -----  -------  --------------  ------  -------  ------  ----------
     Ethernet0              0,1      50G   9100    N/A    etp1a           trunk      up       up    RJ45         off
     Ethernet2              2,3      50G   9100    N/A    etp1b           trunk      up       up    RJ45         off
     Ethernet4              4,5      50G   9100    N/A    etp2a          routed    down     down     N/A         off
     Ethernet6              6,7      50G   9100    N/A    etp2b          routed    down     down     N/A         off
```

judyjoseph (approved):

PR #2314, opened 2022-08-16
Added missing commands for release 202111 in the command reference do…

kannankvs:
…cument

Signed-off-by: kannankvs <kannan_kvs@dell.com>


wen587:
Hi, need to mention that GCU is experimental in 202111 and may not function well. 
GCU is fully supporte in 202205.

wen587:
Hi, it seems the total documentation was removed fully and then added back, which make reviewer cannot check the file diff from here https://github.com/sonic-net/sonic-utilities/pull/2314/files.
Can we keep the original format and just add what is missing? That should be easier to compare the diff from the website.

kannankvs:
Closing this PR due to incorrect commits for the file

PR #2327, opened 2022-08-23
Handle non-front-panel ports in is_rj45_port

stephenxs:
Handle the cases that a port is not a front-panel port in is_rj45_port
Add mock test to cover the logic

### DIFF
--- a/utilities_common/platform_sfputil_helper.py
+++ b/utilities_common/platform_sfputil_helper.py
@@ -4,7 +4,6 @@ import click
 
 from . import cli as clicommon
 from sonic_py_common import multi_asic, device_info
-from sonic_py_common.interface import front_panel_prefix, SONIC_INTERFACE_PREFIXES
 
 platform_sfputil = None
 platform_chassis = None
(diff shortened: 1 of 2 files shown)


anamehra:
@abdosi , @judyjoseph please add this to 202205. Thanks

judyjoseph (approved):

utilities_common/platform_sfputil_helper.py
```diff
@@ -133,7 +133,7 @@ def is_rj45_port(port_name):
```
judyjoseph, line 133: Did a bit of debugging and see that the below code section 
```
        if not platform_porttab_mapping_read:
            platform_sfputil_read_porttab_mappings()
```
is actually creating this logs which @mlok-nokia was pointing out
```
Aug 25 05:50:23.693785 str2--lc1-1 ERR Chassis: :- initializeGlobalConfig: SonicDBConfig Global config is already initialized
```

stephenxs: thanks. I'll check it.

stephenxs: @judyjoseph if possible, can you share more details, like call stack? I don't have a multi-chassis testbed, it will be time-consuming for me to find it out. thank you.

PR #2333, opened 2022-08-25
[VRF]Adding CLI checks to ensure Vrf is valid in interface bind and static route commands

dgsudharsan:
#### What I did
Added CLI checks to verify if Vrf is created and valid when used in interface bind and static route commands

#### How I did it
Check for presence of Vrf in VRF table or in case if it is mgmt vrf, verify if mgmt vrf is enabled/

#### How to verify it
Added UT to verify it.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -369,6 +369,19 @@ def get_interface_ipaddresses(config_db, interface_name):
 
     return ipaddresses
 
+def is_vrf_exists(config_db, vrf_name):
+    """Check if VRF exists
+    """
+    keys = config_db.get_keys("VRF")
+    if vrf_name in keys:
+        return True
+    elif vrf_name == "mgmt":
+        entry = config_db.get_entry("MGMT_VRF_CONFIG", "vrf_global")
+        if entry and entry.get("mgmtVrfEnabled") == "true":
+           return True
+
+    return False
+
 def is_interface_bind_to_vrf(config_db, interface_name):
     """Get interface if bind to vrf or not
     """
@@ -986,6 +999,7 @@ def cli_sroute_to_config(ctx, command_str, strict_nh = True):
     nexthop_str = None
     config_entry = {}
     vrf_name = ""
+    config_db = ctx.obj['config_db']
 
     if "nexthop" in command_str:
         idx = command_str.index("nexthop")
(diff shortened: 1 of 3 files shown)


dgsudharsan:
@venkatmahalingam @prsunny Can you please review?

config/main.py
```diff
@@ -998,6 +1012,8 @@ def cli_sroute_to_config(ctx, command_str, strict_nh = True):
         if 'prefix' in prefix_str and 'vrf' in prefix_str:
             # prefix_str: ['prefix', 'vrf', Vrf-name, ip]
             vrf_name = prefix_str[2]
+            if not is_vrf_exists(config_db, vrf_name):
```
prsunny, line 1015: Do we have check to ensure vrf is removed only after all interface bindings are removed? Otherwise this is only a check in the add-path.

dgsudharsan: This is only a check in the add path.

dgsudharsan: In fact the current vrf removal logic removes all these bindings implicitly using the function del_interface_bind_to_vrf

prsunny (approved):

config/main.py
```diff
@@ -369,6 +369,19 @@ def get_interface_ipaddresses(config_db, interface_name):
...
 
+def is_vrf_exists(config_db, vrf_name):
+    """Check if VRF exists
+    """
+    keys = config_db.get_keys("VRF")
+    if vrf_name in keys:
+        return True
+    elif vrf_name == "mgmt":
+        entry = config_db.get_entry("MGMT_VRF_CONFIG", "vrf_global")
+        if entry and entry.get("mgmtVrfEnabled") == "true":
```
venkatmahalingam, line 380: How do we prevent from using non-existing VRF in the INTERFACE table during config load? this seems to address only in the Click CLI.

If we can do the YANG level validation for all north bound interfaces (CLick, config load..etc), it would be an one time check in the back-end rather than doing it on all front-end (e.g this check in Click), what is your thought?

dgsudharsan: Hi @venkatmahalingam 
Your comment is applicable for any table in config_db.json.
This PR is focussed on only the Click CLI. I believe we have a separate discussion on using yang for validation which is the work in progress and would take sometime. 
For now It is expected of the end user to use Click CLI to make any incremental configuration changes.
Let me know if the changes look good.

venkatmahalingam: Any update we do with config-db should be YANG validated is what the expectation but the infra is not ready yet, we can merge this code for now.

venkatmahalingam (approved):

dgsudharsan:
@yxieca While cherry-picking please ensure https://github.com/sonic-net/sonic-utilities/pull/2309 is cherry-picked before this to avoid build failures

PR #1947, opened 2021-11-24
[sfputil] Firmware download/upgrade CLI support for QSFP-DD

prgeor:
Signed-off-by: Prince George <prgeor@microsoft.com>

#### What I did
Added following sfputil CLIs to support firmware upgrade on QSFP-DD transceivers:-
1. sfputil show fwversion [OPTIONS] <port_name>
2. sfputil firmware [OPTIONS] COMMAND [ARGS] ...
    a. commit - Commit the running firmware
    b. download - Download firmware 
    c. run - Run the firmware
    d. unlock - Unlock firmware download feature
    e. upgrade - Upgrade firmware (combines a,b,c and d as one CLI) 

#### How to verify it
Firmware upgrade done successfully on Inphi, Molex and Centera QSFP-DD modules

#### New command output (if the output of a command-line utility has changed)
Capture

PR #2349, opened 2022-09-02
[sfputil] Firmware download/upgrade CLI support for QSFP-DD (#1947)

CliveNi:
#### What I did
* Checking that the running image is switched or not after CDB_run during firmware upgrade process.

#### How I did it
* Call api.get_module_fw_info() for checking the running and committed image are difference.

#### How to verify it
* Run sfputil firmware upgrade with transceivers of different vendor.


aravindmani-1:
@CliveNi Could you please upload UT logs?.

aravindmani-1 (commented):
Could you please provide the UT logs done inside the switch?.

CliveNi:
> @CliveNi Could you please upload UT logs?.

Adding unit test for new function, disabling a test condition in commit due to it will spend around 1min, it is enabled when capturing log.
Some environment issues cause few unrelated function failure.
[sfputil_test-results.txt](https://github.com/sonic-net/sonic-utilities/files/9505896/sfputil_test-results.txt)

aravindmani-1 (dismissed):

aravindmani-1:
@prgeor Could you please review this PR?..

sfputil/main.py
```diff
@@ -1112,6 +1112,64 @@ def run_firmware(port_name, mode):
...
+    sfp = platform_chassis.get_sfp(physical_port)
+
+    try:
+        api = sfp.get_xcvr_api()
+    except NotImplementedError:
+        click.echo("This functionality is currently not implemented for this platform")
+        sys.exit(ERROR_NOT_IMPLEMENTED)
+
+    try:
+        MAX_WAIT = 600 # 60s timeout.
```
prgeor, line 1132: can we read from the module what is advertised instead of hardcoding here?

sfputil/main.py
```diff
@@ -1112,6 +1112,64 @@ def run_firmware(port_name, mode):
...
+            (ImageA, ImageARunning, ImageACommitted, ImageAValid,
+             ImageB, ImageBRunning, ImageBCommitted, ImageBValid) = fw_info['result']
+
+            if (ImageARunning == 1) and (ImageBRunning == 1):       # Both imageA and B is running.
+                click.echo("FW info error : Both imageA and B show running!")
+                status = -1 # Abnormal status.
+            elif (ImageACommitted == 1) and (ImageBCommitted == 1): # Both imageA and B is committed.
+                click.echo("FW info error : Both imageA and B show committed!")
+                status = -1 # Abnormal status.
+            elif (ImageARunning == 1) and (ImageAValid == 1):       # ImageA is running, but also invalid.
```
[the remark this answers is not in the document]
CliveNi: Renaming variables for more suitable its purpose. Picture is the definition of CMIS5.0, 1 for invalid.
CDB_Info

sfputil/main.py
```diff
@@ -1112,6 +1112,64 @@ def run_firmware(port_name, mode):
...
+                click.echo("FW info error : ImageA shows running, but also shows invalid!")
+                status = -1 # Abnormal status.
+            elif (ImageBRunning == 1) and (ImageBValid == 1):       # ImageB is running, but also invalid.
```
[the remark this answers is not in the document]
CliveNi: Same as above.

sfputil/main.py
```diff
@@ -1112,6 +1112,64 @@ def run_firmware(port_name, mode):
...
+            cnt += 1
+
+        if fw_info['status'] == True:
+            (ImageA, ImageARunning, ImageACommitted, ImageAValid,
+             ImageB, ImageBRunning, ImageBCommitted, ImageBValid) = fw_info['result']
+
+            if (ImageARunning == 1) and (ImageBRunning == 1):       # Both imageA and B is running.
+                click.echo("FW info error : Both imageA and B show running!")
+                status = -1 # Abnormal status.
+            elif (ImageACommitted == 1) and (ImageBCommitted == 1): # Both imageA and B is committed.
```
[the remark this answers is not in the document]
CliveNi: Agree, these two cases are low correlation with this function. Already removed.

sfputil/main.py
CliveNi: There is no definition of max duration for CDB_run cmd until CMIS5.2.
CDB_Duration

sfputil/main.py
```diff
@@ -1112,6 +1112,61 @@ def run_firmware(port_name, mode):
...
+            (ImageA, ImageARunning, ImageACommitted, ImageAInvalid,
+             ImageB, ImageBRunning, ImageBCommitted, ImageBInvalid) = fw_info['result']
+
+            if (ImageARunning == 1) and (ImageAInvalid == 1):       # ImageA is running, but also invalid.
+                click.echo("FW info error : ImageA shows running, but also shows invalid!")
+                status = -1 # Abnormal status.
+            elif (ImageBRunning == 1) and (ImageBInvalid == 1):     # ImageB is running, but also invalid.
+                click.echo("FW info error : ImageB shows running, but also shows invalid!")
+                status = -1 # Abnormal status.
+            elif (ImageARunning == 1) and (ImageACommitted == 0):   # ImageA is running, but not committed.
```
prgeor, line 1152: ImageACommitted check is NOT correct here because a running image need NOT  be committed.

sfputil/main.py
prgeor: @CliveNi why 60 secs?

sfputil/main.py
```diff
@@ -1112,6 +1112,61 @@ def run_firmware(port_name, mode):
...
+                click.echo("FW images switch successful : ImageA is running")
+                status = 1  # run_firmware is done. 
+            elif (ImageBRunning == 1) and (ImageBCommitted == 0):   # ImageB is running, but not committed.
```
prgeor, line 1155: commit_firmware() happens after the run_firmware() so why are you checking for ImageBCommitted?

sfputil/main.py
```diff
@@ -1112,6 +1112,61 @@ def run_firmware(port_name, mode):
 
     return status
 
+def is_fw_switch_done(port_name):
```
prgeor, line 1115: I am not sure why this API is required because there is already a return status from run_firmware() to indicate a failure. An imageA could be running while the committed Image could be Image B. For eg.

1. Image A is running and its is the committed image
2. User updates Image B
3. Runs the image B

Now the running image is B but the committed image is A

sfputil/main.py
```diff
@@ -1280,6 +1335,10 @@ def upgrade(port_name, filepath):
 
     click.echo("Firmware run in mode 1 successful")
 
+    if is_fw_switch_done(port_name) != 1:
+        click.echo('Failed to switch firmware images!')
+        sys.exit(EXIT_FAIL)
+
```
prgeor, line 1341: why are you checking for committed image inside this function when the commit_firmware() happens at line 1342

prgeor (changes requested):

sfputil/main.py
CliveNi: "run_firmware" indicates the status which read before reset to target image. This status only means module received and done the judgement for "run_firmware" cmd, but not means module already jumped into target image.
"run_firmware" will maybe involved non-volatile memory operation in module according to module implementation, so that it will maybe spend several or tens of seconds. That will cause commit failed if sends "commit_firmware" once "run_firmware" done without "is_fw_switch_done".
imageCheckPurpose

sfputil/main.py
CliveNi: Yes, so that (Running == 1) and (Committed == 0) will return success, isn't it?

sfputil/main.py
CliveNi: is_fw_switch_done() executes between run_firmware() and commit_firmware() in upgrade() and assuming the running and committed image is same before upgrade(). run_firmware() will change the running image, so that checking running and committed image are different will be a easy way for judging image switch done.

sfputil/main.py
CliveNi: For more compatible with different module, timeout time should be the worst case. 60s is calculated for different implementation.
Feasible implementation as below, 
1. Dual image(A and B) is supported, and fw only executable at specific image.
2. Dual image(A and B) is supported, and fw executable at both image.
3. Triple image(A, B and factory) is supported, and fw only executable at specific image.
3. Triple image(A, B and factory) is supported, and fw executable at all image.
4. Single image is supported.

Case 1 and 3 will spend the most time to switch image. It will need to copy one image to non-volatile buffer, then interchange two images with each other. That will need 3 cycles erasing and programming on non-volatile memory.

We can roughly assume all of fw component in module around 1MBytes. A common M0~M4 MCU needs around 40ms per page(2KBytes) for non-volatile erasing and programming, so that the calculation as below.
1MBytes / 2KBytes * 40ms * 3cycles = 61.44s

sfputil/main.py
```diff
@@ -1112,6 +1112,61 @@ def run_firmware(port_name, mode):
...
+        click.echo("This functionality is currently not implemented for this platform")
+        sys.exit(ERROR_NOT_IMPLEMENTED)
+
+    try:
+        MAX_WAIT = 600 # 60s timeout.
+        fw_info = api.get_module_fw_info()
+        cnt = 0
+        is_busy = 1 if (fw_info['status'] == False) and (fw_info['result'] is not None) else 0
+        while is_busy and cnt < MAX_WAIT:
+            time.sleep(0.1)
```
qinchuanares, line 1137: May not need to check this frequently. It also brings a lot of workload to the module. Checking every 1 or 2 seconds should be OK without causing too much time delay.

sfputil/main.py
```diff
@@ -1112,6 +1112,61 @@ def run_firmware(port_name, mode):
...
+        api = sfp.get_xcvr_api()
+    except NotImplementedError:
+        click.echo("This functionality is currently not implemented for this platform")
+        sys.exit(ERROR_NOT_IMPLEMENTED)
+
+    try:
+        MAX_WAIT = 600 # 60s timeout.
+        fw_info = api.get_module_fw_info()
+        cnt = 0
+        is_busy = 1 if (fw_info['status'] == False) and (fw_info['result'] is not None) else 0
```
qinchuanares, line 1135: Could you show examples of  fw_info 1) when a module is busy and 2) when module shows faulty status?

prgeor (commented):
@CliveNi can you take care of the comment from @qinchuanares ? Rest all looks fine.

sfputil/main.py
qinchuanares: Just curious, if a firmware size is >> 1MB (e.g. close to 2 MB), does that mean 60 sec  wait time would not be enough?

CliveNi: Exactly, 1MB is a roughly number, but I think the margin is enough for different DSP solution now. Actually, it only uses around 500KB totally in our 400/800G implementation.

sfputil/main.py
CliveNi: Agree, already revised.

sfputil/main.py
CliveNi: "fw_info" is returned by "get_module_fw_info()" which is in cmis.py of "sonic-platform-common". It will return 3 combination patterns of status and result as below.
  1. {['status']: False, ['result']: None} : **Cdb** features **not supported** by switch, **or wrong chkcode** got from module. Do **NOT need to recheck** in these cases.
  2. {['status']: False, ['result']: 0} : I2C NACK or other interface issue. **Recheck** module fw info **until timeout** in this case.
  3. {['status']: True, ['result']: (ImageA & B information)} : **Valid information** got from module, **go ahead** to handle the return value.

A unit test function "test_is_fw_switch_done" in sfputil_test.py will also test all these cases.

prgeor (approved):

prgeor:
@yxieca please cherry pick to 202205
@StormLiangMS please cherry pick to 202211

PR #2350, opened 2022-09-05
[minigraph] add option to specify golden path in load_minigraph

wen587:
#### What I did
Add an option for load_minigraph to specify golden path
#### How I did it
Add an option for load_minigraph
#### How to verify it
Unit test
sudo config load_minigraph -h
Usage: config load_minigraph [OPTIONS]

Reconfigure based on minigraph.

Options:
-y, --yes
-n, --no_service_restart Do not restart docker services
-t, --traffic_shift_away Keep device in maintenance with TSA
-?, -h, --help Show this message and exit.
sudo config load_minigraph -h
Usage: config load_minigraph [OPTIONS]

Reconfigure based on minigraph.

Options:
-y, --yes
-n, --no_service_restart Do not restart docker services
-t, --traffic_shift_away Keep device in maintenance with TSA
**-p, --golden_config_path TEXT  specify Golden Config path**
-?, -h, --help Show this message and exit.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1718,8 +1718,9 @@ def load_mgmt_config(filename):
                 expose_value=False, prompt='Reload config from minigraph?')
 @click.option('-n', '--no_service_restart', default=False, is_flag=True, help='Do not restart docker services')
 @click.option('-t', '--traffic_shift_away', default=False, is_flag=True, help='Keep device in maintenance with TSA')
+@click.option('-p', '--golden_config_path', help='specify Golden Config path')
 @clicommon.pass_db
-def load_minigraph(db, no_service_restart, traffic_shift_away):
+def load_minigraph(db, no_service_restart, traffic_shift_away, golden_config_path):
     """Reconfigure based on minigraph."""
     log.log_info("'load_minigraph' executing...")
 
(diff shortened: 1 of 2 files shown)


qiluo-msft:
>         if os.path.isfile(DEFAULT_GOLDEN_CONFIG_DB_FILE):

The checked file should be changed according `golden_config_path`

---
Refers to: config/main.py:1796 in 416c29a. [](commit_id = 416c29a235db68310e207006d4fc5e83a0e52daf, deletion_comment = False)

config/main.py
```diff
@@ -1797,8 +1798,15 @@ def load_minigraph(db, no_service_restart, traffic_shift_away):
...
     # Load golden_config_db.json
-    if os.path.isfile(DEFAULT_GOLDEN_CONFIG_DB_FILE):
-        override_config_by(DEFAULT_GOLDEN_CONFIG_DB_FILE)
+    if golden_config_path:
+        if not os.path.isfile(golden_config_path):
+            click.secho("Cannot find '{}'!".format(golden_config_path),
+                        fg='magenta')
+            raise click.Abort()
+        override_config_by(golden_config_path)
+    else:
```
qiluo-msft, line 1807: >else [](http://example.com/codeflow?start=4&length=4)

You are treating below 2 usages the same.
1. `config load_minigraph --golden_config_path`
2. `config load_minigraph`

The user expectation may be
1. User want to override, but do not provide a gulden config path, so just

wen587: I am forcing users to add an argument when add `--golden_config_path`.
`config load_minigraph --golden_config_path` will return an error: 
`Error: --golden_config_path option requires an argument`

> 2. User does not want to override.

Are we chaning the code behavior? Previously, `load_minigraph` will check DEFAULT_GOLDEN_PATH and override if it exists. Do you mean to change to override only when `--golden_config_path` is set?

qiluo-msft: Let's merge current PR, backport, ask client to explicitly use the option, and you submit another PR to achieve the expected behavior.

qiluo-msft (approved):

PR #2370, opened 2022-09-14
Update the DBmigrator to support persistent loglevel during warm-upgrade

EdenGri:
This PR needs to be merged only after the following PR: https://github.com/sonic-net/sonic-buildimage/pull/11996

#### What I did
Add the ability to the user to save the loglevel and make it persistent to reboot with the support in warm-upgrade. (Not need to add special support for cold/fast reboot since the LOGLEVEL DB flushes during fast/cold reboot).

#### How I did it
Add a new version to the db_migrator that will move the logger tables from the LOGLEVEL DB to the CONFIG DB

#### How to verify it
1. start with a switch that has an old master image (old master means that the VERSION in the config db is 3_0_5 or lower)
2. install this image to the switch
3. run warm-reboot
4. verify that the LOGLEVEL DB does not contain logger tables -> run redis-cli -n 3 keys "*" (expect to see only the JINJA2CACHE)
5. verify that the logger tables are in the config db -> run redis-cli -n 4 keys "LOGGER|*"


qiluo-msft:
Could you refine the PR title?

yxieca (approved):

EdenGri:
> 

Done, thanks

qiluo-msft (approved):

tests/db_migrator_input/config_db/reclaiming-buffer-warmreboot-expected.json
```diff
@@ -2043,6 +2043,6 @@
         "admin_status": "up"
     },
     "VERSIONS|DATABASE": {
-        "VERSION": "version_3_0_5"
+        "VERSION": "version_3_0_6"
```
stephenxs, line 2046:
```suggestion
        "VERSION": "version_3_0_3"
```
We should not update db version in `expected` config_db every time a new db version is introduced.
There is an infrastructure to tolerate the case that the target database version in `expected` config_db file isn't the latest database version. The logic is to advance the DB version in the `expected` config_db iff it is the latest version that the feature handles. Check `advance_version_for_expected_database` for details.
Reclaiming buffer unit test does leverage this infra, so theoretically, we should not touch it every time.
However, the version was somehow updated by some previous PR, which breaks the logic and requires the version to be updated every time a new db version is introduced, which is inconvenient to developers.
Now that you are going to introduce new db version, let's change it back to the original `expected` version.

EdenGri: Done, thanks

stephenxs (approved):

qiluo-msft (approved):

yxieca (approved):

qiluo-msft:
bandit checker is not 'Required', so please go ahead and merge it.

tests/db_migrator_input/config_db/reclaiming-buffer-warmreboot-expected.json
```diff
@@ -2043,6 +2043,6 @@
         "admin_status": "up"
     },
     "VERSIONS|DATABASE": {
-        "VERSION": "version_3_0_5"
+        "VERSION": "version_3_0_3"
```
vaibhavhd, line 2046: Why is this change needed? Is this relevant to persistance loglevel design?

qiluo-msft: Yes, relevant to persistance loglevel design.

vaibhavhd: I see. I am unable to understand, can you explain why tests for reclaiming-buffer need to be changed for persistence log-level design?

stephenxs: This is to change it back. Refer [this comment](https://github.com/sonic-net/sonic-utilities/pull/2370#discussion_r992945987) for details

vaibhavhd (approved):

PR #2238, opened 2022-06-28
Fix DBConfig not initialize issue in pfcwd

liuh-80:
#### What I did
Fix pfcwd connect DB with exception issue: https://github.com/Azure/sonic-buildimage/issues/11269

pfcwd implicit depends on InterfaceAliasConverter() to initialize DB config, however following PR change InterfaceAliasConverter() behavior to lazy initialize, then pfcwd failed when try connect to DB without initialize DB config:

https://github.com/Azure/sonic-utilities/pull/2183

#### How I did it
Load  DB config in pfcwd.

#### How to verify it
Pass all UT.

PR #2372, opened 2022-09-14
Fix DBConfig not initialize issue in pfcwd (#2238)

gechiang:
#### What I did
Fix pfcwd connect DB with exception issue: https://github.com/Azure/sonic-buildimage/issues/11269

pfcwd implicit depends on InterfaceAliasConverter() to initialize DB config, however following PR change InterfaceAliasConverter() behavior to lazy initialize, then pfcwd failed when try connect to DB without initialize DB config:

https://github.com/Azure/sonic-utilities/pull/2183

#### How I did it
Load  DB config in pfcwd.

#### How to verify it
Pass all UT.

### DIFF
--- a/pfcwd/main.py
+++ b/pfcwd/main.py
@@ -9,6 +9,7 @@ from sonic_py_common.multi_asic import get_external_ports
 from tabulate import tabulate
 from utilities_common import multi_asic as multi_asic_util
 from utilities_common import constants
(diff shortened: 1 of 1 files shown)


gechiang:
since this one can be picked cleanly, will let the repo owner cherry-pick it instead of me raising a PR for this...

qiluo-msft (approved):

PR #2389, opened 2022-09-20
Fix show vxlan tunnel command display issue #11902

skbhava:
Issue:
Only one vni to vlan map entry in the output of show vxlan tunnel command Fix:
Fix:
Added fix to display all vni to vlan map entries for "show vxlan tunnel" and "show vxlan name <tunnel-name> commands.
Fixes #11902


skbhava:
Open with some mistake

PR #2401, opened 2022-09-26
Add UT to check sonic installer does not depend on database

liuh-80:
### Description of PR
Add new test cases to test sonic-installer does not depends on database docker.

### Type of change
- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [x] Test case(new/improvement)

Add new test cases to test sonic-installer does not depends on database docker.

#### How did you do it?
Add new test case to cover user scenarios.

Run new UT make sure they are all pass.
Make sure all current UT not break during merge validation.

#### Any platform specific information?
N/A

### DIFF
--- a/tests/installer_dependency_test.py
+++ b/tests/installer_dependency_test.py
@@ -34,4 +34,4 @@ def test_sonic_installer_not_depends_on_database_docker():
     except RuntimeError:
         exception_happen = True
 
-    assert result.exception_happen == True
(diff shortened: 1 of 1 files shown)


qiluo-msft:
In the section "How did you verify/test it?", you mentioned the positive testcase. How about the negative case? Previously sonic installer does depends on database, so the new testcase should catch the bug.

PR #2404, opened 2022-09-27
[VxLAN]Fix Vxlan delete command to throw error when there are references

dgsudharsan:
Fix for https://github.com/sonic-net/sonic-buildimage/issues/12063
#### What I did
Check for VNET references for vxlan before deleting vxlan object

#### How I did it
Check VNET table and throw error if vxlan is referenced

#### How to verify it
Added UT to verify it.

### DIFF
--- a/config/vxlan.py
+++ b/config/vxlan.py
@@ -38,6 +38,10 @@ def del_vxlan(db, vxlan_name):
     """Del VXLAN"""
     ctx = click.get_current_context()
 
+    vxlan_keys = db.cfgdb.get_keys('VXLAN_TUNNEL')
+    if vxlan_name not in vxlan_keys:
+        ctx.fail("Vxlan tunnel {} does not exist".format(vxlan_name))
+
     vxlan_keys = db.cfgdb.get_keys('VXLAN_EVPN_NVO')
     if not vxlan_keys:
       vxlan_count = 0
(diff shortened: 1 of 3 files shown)


config/vxlan.py
```diff
@@ -56,6 +60,12 @@ def del_vxlan(db, vxlan_name):
     if(vxlan_count > 0):
         ctx.fail("Please delete all VLAN VNI mappings.")  
 
+    vnet_table = db.cfgdb.get_table('VNET')
+    vnet_keys = vnet_table.keys()
+    for vnet_key in vnet_keys:
+        if ('vxlan_tunnel' in vnet_table[vnet_key] and vnet_table[vnet_key]['vxlan_tunnel'] == vxlan_name):
```
ITJamie, line 66: It might be worth including the `vnet_key` var in the error message to let the user know which vnets need to be removed?

volodymyrsamotiy (approved):

prsunny (approved):

config/vxlan.py
dgsudharsan: @ITJamie There might be multiple vnets pointing to a vxlan tunnel. In such case the  error message would be very verbose to the user. Additionally there are no commands to remove the VNET today in SONiC. The user has to modify the config_db.json and do reboot or config reload.
Lastly if we take other similar messages we don't print the references (e.g. Adding a portchannel member to a vlan will print "Ethernetx is part of portchannel!". 
Aligning with all existing error messages and the restrictions around VNET I believe the error message in current form should be sufficient

PR #2405, opened 2022-09-27
YANG Validation for ConfigDB Updates: Fix Decorator Bug

isabelmsft:
#### What I did
In ValidatedConfigDBConnector decorator code, create a new instance of ConfigDBConnector so that original instance is not modified.
#### How to verify it
Tested locally - ensure that GCU is used when YANG Config Validation is enabled, and ADHOC_VALIDATION is False

### DIFF
--- a/config/validated_config_db_connector.py
+++ b/config/validated_config_db_connector.py
@@ -4,7 +4,6 @@ from jsonpointer import JsonPointer
 from sonic_py_common import device_info
 from generic_config_updater.generic_updater import GenericUpdater, ConfigFormat
 from generic_config_updater.gu_common import EmptyTableError, genericUpdaterLogging
-from swsscommon.swsscommon import ConfigDBConnector
 
 class ValidatedConfigDBConnector(object):
     
--- a/config/validated_config_db_connector.py
+++ b/config/validated_config_db_connector.py
@@ -1,5 +1,4 @@
 import jsonpatch
-import types
 from jsonpointer import JsonPointer
 
 from sonic_py_common import device_info
@@ -20,9 +19,9 @@ class ValidatedConfigDBConnector(object):
         if not yang_enabled:
             return self.connector.__getattribute__(name)
         if name == "set_entry":
-            return validated_set_entry
+            return self.validated_set_entry
         if name == "delete_table":
-            return validated_delete_table
+            return self.validated_delete_table
 
     def make_path_value_jsonpatch_compatible(self, table, key, value):
         if type(key) == tuple:
@@ -35,7 +34,7 @@ class ValidatedConfigDBConnector(object):
 
     def create_gcu_patch(self, op, table, key=None, value=None):
         if key:
-            path, value = make_path_value_jsonpatch_compatible(table, key, value)
+            path, value = self.make_path_value_jsonpatch_compatible(table, key, value)
         else: 
             path = "/{}".format(table)
 
@@ -50,7 +49,7 @@ class ValidatedConfigDBConnector(object):
         return gcu_patch
 
     def validated_delete_table(self, table):
-        gcu_patch = create_gcu_patch("remove", table)
+        gcu_patch = self.create_gcu_patch("remove", table)
         format = ConfigFormat.CONFIGDB.name
         config_format = ConfigFormat[format.upper()]
         try:
@@ -65,11 +64,11 @@ class ValidatedConfigDBConnector(object):
         else:
             op = "remove"
     
-        gcu_patch = create_gcu_patch(op, table, key, value)
+        gcu_patch = self.create_gcu_patch(op, table, key, value)
         format = ConfigFormat.CONFIGDB.name
         config_format = ConfigFormat[format.upper()]
 
         try:
             GenericUpdater().apply_patch(patch=gcu_patch, config_format=config_format, verbose=False, dry_run=False, ignore_non_yang_tables=False, ignore_paths=None)
         except EmptyTableError:
-            validated_delete_table(table)
+            self.validated_delete_table(table)
--- a/config/validated_config_db_connector.py
+++ b/config/validated_config_db_connector.py
@@ -66,4 +66,4 @@ def validated_set_entry(self, table, key, value):
     try:
         GenericUpdater().apply_patch(patch=gcu_patch, config_format=config_format, verbose=False, dry_run=False, ignore_non_yang_tables=False, ignore_paths=None)
     except EmptyTableError:
-        validated_delete_table(table)
+        validated_delete_table(self, table)
--- a/config/validated_config_db_connector.py
+++ b/config/validated_config_db_connector.py
@@ -10,13 +10,10 @@ class ValidatedConfigDBConnector(object):
     
     def __init__(self, config_db_connector):
         self.connector = config_db_connector
+        self.yang_enabled = device_info.is_yang_config_validation_enabled(self.connector)
 
     def __getattr__(self, name):
-        config_db_connector = ConfigDBConnector()
-        config_db_connector.connect()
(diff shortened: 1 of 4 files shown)


qiluo-msft:
> def validated_delete_table(table):

Need to add `self` as first parameter. ref: [https://stackoverflow.com/questions/1647586/is-it-possible-to-change-an-instances-method-implementation-without-changing-al#comment89043742_1647616](https://stackoverflow.com/questions/1647586/is-it-possible-to-change-an-instances-method-implementation-without-changing-al#comment89043742_1647616)

---
Refers to: config/validated_config_db_connector.py:45 in 454d740. [](commit_id = 454d7407c59e165af0a19ade496d2debdcbd29e0, deletion_comment = False)

qiluo-msft:
> def validated_set_entry(table, key, value):

The same of `self` comment as above.

---
Refers to: config/validated_config_db_connector.py:55 in 454d740. [](commit_id = 454d7407c59e165af0a19ade496d2debdcbd29e0, deletion_comment = False)

qiluo-msft:
>         GenericUpdater().apply_patch(patch=gcu_patch, config_format=config_format, verbose=False, dry_run=False, ignore_non_yang_tables=False, ignore_paths=None)

Not an immediate ask, but we have plan to change the API, so apply patch will reuse the original `config_db_connector`. Currently its implementation create a temporary connector inside.

---
Refers to: config/validated_config_db_connector.py:50 in 454d740. [](commit_id = 454d7407c59e165af0a19ade496d2debdcbd29e0, deletion_comment = False)

liuh-80:
Here is an example about use reflection:
http://www.sefidian.com/2021/06/06/python-__getattr__-and-__getattribute__-magic-methods/

class ConfigDBConnector(object):
    def __init__(self, name):
        self.name = name
    def connect(self, name):
        print("ConfigDBConnector::connect")
    def gettable(self, table):
        print("ConfigDBConnector::gettable {} {}".format(table, self.name))
    def showtable(self, table):
        print("ConfigDBConnector::showtable {} {}".format(table, self.name))


# LayeredConfigDBConnectorDecorator can't be ConfigDBConnector
# because __getattr__ will check inherited method first
class LayeredConfigDBConnectorDecorator(object):
        def __init__(self, config_db_connector):
            self.connector = config_db_connector
        def deco_gettable(self, table):
            print("LayeredConfigDBConnectorDecorator::gettable {}".format(table))
            self.connector.gettable(table)
        def __getattr__(self, name):
            print("LayeredConfigDBConnectorDecorator __getattr__ {}".format(name))
            if name == "gettable":
                return self.deco_gettable
            originalMethod = self.connector.__getattribute__(name)
            return originalMethod

conn = ConfigDBConnector("123")
deco = LayeredConfigDBConnectorDecorator(conn)
deco.gettable("test")
deco.showtable("test")



Here is output:

>>> conn = ConfigDBConnector("123")
>>> deco = LayeredConfigDBConnectorDecorator(conn)
>>> deco.gettable("test")
LayeredConfigDBConnectorDecorator __getattr__ gettable
**LayeredConfigDBConnectorDecorator::gettable test** <<== "gettable" been decorated
ConfigDBConnector::gettable test 123
>>> deco.showtable("test")
LayeredConfigDBConnectorDecorator __getattr__ showtable
ConfigDBConnector::showtable test 123


The limit of this solution are:
1. The decorator class can't be an instance of ConfigDBConnector.
2. Performance impact: every API call need call __getattr__ first.

isabelmsft:
> Here is an example about use reflection: http://www.sefidian.com/2021/06/06/python-__getattr__-and-__getattribute__-magic-methods/
> 
> ```python
> class ConfigDBConnector(object):
>     def __init__(self, name):
>         self.name = name
>     def connect(self, name):
>         print("ConfigDBConnector::connect")
>     def gettable(self, table):
>         print("ConfigDBConnector::gettable {} {}".format(table, self.name))
>     def showtable(self, table):
>         print("ConfigDBConnector::showtable {} {}".format(table, self.name))
> 
> 
> # LayeredConfigDBConnectorDecorator can't be ConfigDBConnector
> # because __getattr__ will check inherited method first
> class LayeredConfigDBConnectorDecorator(object):
>         def __init__(self, config_db_connector):
>             self.connector = config_db_connector
>         def deco_gettable(self, table):
>             print("LayeredConfigDBConnectorDecorator::gettable {}".format(table))
>             self.connector.gettable(table)
>         def __getattr__(self, name):
>             print("LayeredConfigDBConnectorDecorator __getattr__ {}".format(name))
>             if name == "gettable":
>                 return self.deco_gettable
>             originalMethod = self.connector.__getattribute__(name)
>             return originalMethod
> 
> conn = ConfigDBConnector("123")
> deco = LayeredConfigDBConnectorDecorator(conn)
> deco.gettable("test")
> deco.showtable("test")
> ```
> 
> Here is output:
> 
> > > > conn = ConfigDBConnector("123")
> > > > deco = LayeredConfigDBConnectorDecorator(conn)
> > > > deco.gettable("test")
> > > > LayeredConfigDBConnectorDecorator **getattr** gettable
> > > > **LayeredConfigDBConnectorDecorator::gettable test** <<== "gettable" been decorated
> > > > ConfigDBConnector::gettable test 123
> > > > deco.showtable("test")
> > > > LayeredConfigDBConnectorDecorator **getattr** showtable
> > > > ConfigDBConnector::showtable test 123
> 
> The limit of this solution are:
> 
> 1. The decorator class can't inherit ConfigDBConnector. if not follow this getattr will call itself and throw following error:
> 
> LayeredConfigDBConnectorDecorator **getattr** showtable Traceback (most recent call last): File "", line 1, in File "", line 11, in **getattribute** File "", line 11, in **getattribute** File "", line 11, in **getattribute** [Previous line repeated 992 more times] File "", line 5, in **getattribute** RecursionError: maximum recursion depth exceeded while calling a Python object
> 
> 2. Performance impact: every API call need call **getattr** first.

Thank you @liuh-80 ! I refactored the decorator based on your input

config/validated_config_db_connector.py
```diff
@@ -4,13 +4,18 @@
...
 def ValidatedConfigDBConnector(config_db_connector):
     yang_enabled = device_info.is_yang_config_validation_enabled(config_db_connector) 
-    if yang_enabled:
-        config_db_connector.set_entry = validated_set_entry
-        config_db_connector.delete_table = validated_delete_table
-    return config_db_connector
+    if not yang_enabled:
+        return config_db_connector
+
+    validated_config_db_connector = ConfigDBConnector()
```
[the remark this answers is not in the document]
isabelmsft: Thanks for the suggestion and Hua's input, I fixed the code to decorator this behavior

config/validated_config_db_connector.py
```diff
@@ -4,60 +4,71 @@
...
-        GenericUpdater().apply_patch(patch=gcu_patch, config_format=config_format, verbose=False, dry_run=False, ignore_non_yang_tables=False, ignore_paths=None)
-    except EmptyTableError:
-        validated_delete_table(table)
+    def __init__(self, config_db_connector):
+        self.connector = config_db_connector
+
+    def __getattr__(self, name):
+        config_db_connector = ConfigDBConnector()
+        config_db_connector.connect()
+        yang_enabled = device_info.is_yang_config_validation_enabled(config_db_connector)
```
qiluo-msft, line 17: >is_yang_config_validation_enabled [](http://example.com/codeflow?start=35&length=33)

Move the flag checking into `__init__()`, so only check once during the life time of `config` command.

And you can use `self.connector` instead of a new `ConfigDBConnector`.

isabelmsft: Thanks for the suggestion, I moved the flag checking

qiluo-msft (approved):

PR #2452, opened 2022-10-25
YANG Validation for ConfigDB Updates: TACPLUS, TACPLUS_SERVER, AAA, VLAN_SUB_INTERFACE tables + decorated validated_mod_entry

isabelmsft:
#### What I did
-Add YANG validation using GCU for writes to TACPLUS, TACPLUS_SERVER, AAA, VLAN_SUB_INTERFACE  in ConfigDB
-decorate mod_entry with validated_mod_entry to include YANG validation 
-process all values in GCU JsonPatch, convert to string

#### How I did it
Using same method as https://github.com/sonic-net/sonic-utilities/pull/2190/files, extend to TACPLUS, TACPLUS_SERVER, AAA, VLAN_SUB_INTERFACE use cases
#### How to verify it
Verified testing CLI on virtual switch

### DIFF
--- a/config/aaa.py
+++ b/config/aaa.py
@@ -20,7 +20,7 @@ def add_table_kv(table, entry, key, val):
     config_db.connect()
     try:
         config_db.mod_entry(table, entry, {key:val})
-    except Exception as e:
+    except ValueError as e:
         ctx = click.get_current_context()
         ctx.fail("Invalid ConfigDB. Error: {}".format(e))
 
@@ -34,7 +34,7 @@ def del_table_key(table, entry, key):
             del data[key]
         try:
             config_db.set_entry(table, entry, data)
-        except Exception as e:
+        except (ValueError, JsonPatchConflict) as e:
             ctx = click.get_current_context()
             ctx.fail("Invalid ConfigDB. Error: {}".format(e))
 
--- a/config/aaa.py
+++ b/config/aaa.py
@@ -2,8 +2,11 @@ import click
 import ipaddress
 import re
 from swsscommon.swsscommon import ConfigDBConnector
+from .validated_config_db_connector import ValidatedConfigDBConnector
+from jsonpatch import JsonPatchConflict
 import utilities_common.cli as clicommon
 
+ADHOC_VALIDATION = True
 RADIUS_MAXSERVERS = 8
 RADIUS_PASSKEY_MAX_LEN = 65
 VALID_CHARS_MSG = "Valid chars are ASCII printable except SPACE, '#', and ','"
@@ -13,19 +16,27 @@ def is_secret(secret):
 
 
 def add_table_kv(table, entry, key, val):
-    config_db = ConfigDBConnector()
+    config_db = ValidatedConfigDBConnector(ConfigDBConnector())
     config_db.connect()
-    config_db.mod_entry(table, entry, {key:val})
+    try:
+        config_db.mod_entry(table, entry, {key:val})
+    except Exception as e:
+        ctx = click.get_current_context()
+        ctx.fail("Invalid ConfigDB. Error: {}".format(e))
 
 
 def del_table_key(table, entry, key):
-    config_db = ConfigDBConnector()
+    config_db = ValidatedConfigDBConnector(ConfigDBConnector())
     config_db.connect()
     data = config_db.get_entry(table, entry)
     if data:
         if key in data:
             del data[key]
-        config_db.set_entry(table, entry, data)
+        try:
+            config_db.set_entry(table, entry, data)
+        except Exception as e:
+            ctx = click.get_current_context()
+            ctx.fail("Invalid ConfigDB. Error: {}".format(e))
 
 @click.group()
 def aaa():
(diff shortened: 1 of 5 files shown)


config/validated_config_db_connector.py
```diff
@@ -17,33 +19,79 @@ def __getattr__(self, name):
                 return self.validated_set_entry
             if name == "delete_table":
                 return self.validated_delete_table
+            if name == "mod_entry":
+                return self.validated_mod_entry
         return self.connector.__getattribute__(name)
 
     def make_path_value_jsonpatch_compatible(self, table, key, value):
+        def stringify_value():
+            nonlocal value
```
qiluo-msft, line 28: >value [](http://example.com/codeflow?start=21&length=5)

Just move `value` to parameter list, so this function could be a normal function of the class.

config/validated_config_db_connector.py
```diff
@@ -17,33 +19,79 @@ def __getattr__(self, name):
...
-        else: 
-            path = "/{}".format(table)
-
+    def create_gcu_patch(self, op, table, key=None, value=None, mod_entry=False):
         gcu_json_input = []
-        gcu_json = {"op": "{}".format(op),
-                    "path": "{}".format(path)}
-        if op == "add":
-            gcu_json["value"] = value
+        if op == "add" and not self.get_table(table):
```
qiluo-msft, line 47: >if op == "add" and not self.get_table(table): [](http://example.com/codeflow?start=8&length=45)

Add some block level comment for each block of code.

config/validated_config_db_connector.py
```diff
@@ -17,33 +19,79 @@ def __getattr__(self, name):
...
+        
+        if mod_entry:
+            key_start = key
+            value_copy = copy.deepcopy(value)
+            for key_end, cleaned_value in value_copy.items():
+                key = [key_start, key_end]
+                value = cleaned_value
+                add_patch_entry()
+        else:
+            add_patch_entry()
```
qiluo-msft, line 81: >add_patch_entry [](http://example.com/codeflow?start=12&length=15)

For set_entry, is it directly translated to jsonpatch `replace` op?

config/aaa.py
```diff
@@ -13,19 +16,27 @@ def is_secret(secret):
 
 
 def add_table_kv(table, entry, key, val):
-    config_db = ConfigDBConnector()
+    config_db = ValidatedConfigDBConnector(ConfigDBConnector())
     config_db.connect()
-    config_db.mod_entry(table, entry, {key:val})
+    try:
+        config_db.mod_entry(table, entry, {key:val})
+    except Exception as e:
```
qiluo-msft, line 23: >Exception [](http://example.com/codeflow?start=11&length=9)

I guess you are handling the case that Invalid ConfigDB before you apply patch. If yes, possible to use a more specific exception type?

Let's assume yang model is fancy enough to capture `val` is not invalid, it will also throw and catch here. So the error message may be misleading.

qiluo-msft:
@preetham-singh Could you help review?

config/validated_config_db_connector.py
isabelmsft: set_entry and mod_entry both use jsonpatch `add` op. 

The difference between set_entry and mod_entry is that set_entry will remove extra fields in the db which are not in the data. So this difference is captured by differing paths in the jsonpatch. The path for mod_entry is more granular.

config/aaa.py
isabelmsft: Yes, we are handling the case that ConfigDB is invalid before applying patch. But this implementation also works for if yang model catches that `val` is invalid. When we print the exception, as long as the constraint is reflected properly in the yang model, we can get a pretty clear error message. 

For example:
```
admin@vlab-01:~$ sudo config aaa authentication failthrough huh
Patch Applier: Patch: [{"op": "add", "path": "/AAA/authentication/failthrough", "value": "huh"}]
Usage: config aaa authentication failthrough [OPTIONS] OPTION
Try "config aaa authentication failthrough -h" for help.

Error: Invalid ConfigDB. Error: Given patch will produce invalid config. Error: Data Loading Failed
Value "huh" does not satisfy the constraint "false|true|False|True" (range, length, or pattern).
```

We have the option to set specific error-message in YANG models as well. This would require updating the YANG models. I would like to start doing this for scenarios where we have multiple issues throwing the same exception type.

config/validated_config_db_connector.py
isabelmsft: Changed function to be normal function of the class

config/validated_config_db_connector.py
isabelmsft: Added comments

config/aaa.py
isabelmsft: Changed to make exception handling more specific

config/validated_config_db_connector.py
isabelmsft: set_entry is not able to be directly translated to jsonpatch `replace` op because `replace` does not allow for addition of a field that does not already exist. 

For example, if we try to add a new portchannel, we will get the following error:
```
admin@vlab-01:~$ sudo config apply-patch patch.json
Patch Applier: Patch: [{"op": "replace", "path": "/PORTCHANNEL/PortChannel04", "value": {"admin_status": "up", "mtu": "9100", "lacp_key": "auto", "min_links": "1"}}]
Failed to apply patch



Error: can't replace a non-existent object 'PortChannel04'
```

qiluo-msft:
I see a new lgtm alert.

qiluo-msft (approved):

qiluo-msft (approved):

PR #2473, opened 2022-11-02
[dualtor][202012] Send arp/ndp packets from vlan_mac if the device is dualtor

vaibhavhd:
#### What I did

`fast-reboot-dump.pty` script collects fdb entries, will send out arp request for these fdb entries to keep them active longer.

However, the script sends these requests with switch mac.

Which has negative impacts on dualtor devices where the arp requests should come from vlan_mac, instead of router_mac

#### How I did it

Check the device subtype from metadata. If it is dualtor, use vlan_mac, otherwise use interface mac as the src address for ARP packets.

### DIFF
--- a/scripts/fast-reboot-dump.py
+++ b/scripts/fast-reboot-dump.py
@@ -13,6 +13,7 @@ import syslog
 import traceback
 import ipaddress
 from builtins import str #for unicode conversion in python2
+from swsscommon import swsscommon
 
 
 ARP_CHUNK = binascii.unhexlify('08060001080006040001') # defines a part of the packet for ARP Request
@@ -266,6 +267,11 @@ def send_garp_nd(neighbor_entries, map_mac_ip_per_vlan):
     src_ifs = {map_mac_ip_per_vlan[vlan_name][dst_mac] for vlan_name, dst_mac, _ in neighbor_entries}
     src_mac_addrs = {src_if:get_iface_mac_addr(src_if) for src_if in src_ifs}
 
+    config_db = swsscommon.ConfigDBConnector()
+    config_db.connect()
+    device_metadata = config_db.get_table('DEVICE_METADATA')
+    subtype = device_metadata['localhost'].get('subtype', '')
+
     # open raw sockets for all required interfaces
     sockets = {}
     for src_if in src_ifs:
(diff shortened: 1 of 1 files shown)


scripts/fast-reboot-dump.py
```diff
@@ -266,6 +267,11 @@ def send_garp_nd(neighbor_entries, map_mac_ip_per_vlan):
     src_ifs = {map_mac_ip_per_vlan[vlan_name][dst_mac] for vlan_name, dst_mac, _ in neighbor_entries}
     src_mac_addrs = {src_if:get_iface_mac_addr(src_if) for src_if in src_ifs}
 
+    config_db = swsscommon.ConfigDBConnector()
+    config_db.connect()
+    device_metadata = config_db.get_table('DEVICE_METADATA')
+    subtype = device_metadata['localhost'].get('subtype', '')
```
yxieca, line 273: Is subtype case insensitive?

scripts/fast-reboot-dump.py
```diff
@@ -275,10 +281,17 @@ def send_garp_nd(neighbor_entries, map_mac_ip_per_vlan):
     # send arp/ndp packets
     for vlan_name, dst_mac, dst_ip in neighbor_entries:
         src_if = map_mac_ip_per_vlan[vlan_name][dst_mac]
+        if subtype.lower() != 'dualtor':
```
yxieca, line 284: This is technically ok, but contains maybe safer?

vaibhavhd: Sure, addressed this now.

yxieca (approved):

scripts/fast-reboot-dump.py
```diff
@@ -275,10 +281,17 @@ def send_garp_nd(neighbor_entries, map_mac_ip_per_vlan):
     # send arp/ndp packets
     for vlan_name, dst_mac, dst_ip in neighbor_entries:
         src_if = map_mac_ip_per_vlan[vlan_name][dst_mac]
+        if subtype and 'dualtor' in subtype.lower():
+            # for non dualtor devices, src mac will be router_mac
+            src_mac = src_mac_addrs[src_if]
+        else:
+            # for dualtor devices, the src mac will be vlan_mac
+            src_mac = get_iface_mac_addr(vlan_name)
```
prsunny, line 289: Isn't this the other way round?  This else case seems to be for non-dualtor.

vaibhavhd: Absolutely, yes :) Fixed this in latest. Finding out a way to test this change.

prsunny (approved):

yxieca (approved):

PR #2477, opened 2022-11-04
[GCU] Fix JsonPointerFilter bug

wen587:
Refer to https://github.com/sonic-net/sonic-utilities/pull/2273
Before the fix, the prefix pattern `Ethernet12|*` would match `Ethernet124|0` and `Ethernet12|0`. 
The suffix pattern `*|10` would match `Ethernet1|10` and `Ethernet1|11110`.

#### What I did
Fix a bug of JsonPointerFilter in GCU that will make the wrong selection of candidates.
#### How I did it
Add the patch addressing '|' in the prefix and suffix filter to avoid wrong selection.
#### How to verify it
Add UT.

### DIFF
--- a/generic_config_updater/patch_sorter.py
+++ b/generic_config_updater/patch_sorter.py
@@ -399,10 +399,10 @@ class JsonPointerFilter:
         if token == "*":
             matching_keys = config.keys()
         elif token.startswith("*|"):
-            suffix = token[2:]
+            suffix = token[1:]
             matching_keys = [key for key in config.keys() if key.endswith(suffix)]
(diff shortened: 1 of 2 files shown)


qiluo-msft (approved):

ganglyu (approved):

isabelmsft (approved):

qiluo-msft:
@wen587 Will you backport to 202205 branch?

wen587:
> @wen587 Will you backport to 202205 branch?

I have added the label for 202205. I will help if there is any conflict.

PR #2480, opened 2022-11-04
bugfix[2024] vnet route check exit code fix.

siqbal1986:
closes https://github.com/sonic-net/sonic-utilities/issues/2424

#### What I did
Fixed the return value for vnet route check if vxlan tunnel is present but no route mismatch is happening.
#### How I did it
return 0 of there is no route diff
#### How to verify it
add a vxlan tunnel without routes and run the script. it should return 0.


prsunny:
could you please add a test?

prsunny (approved):

qiluo-msft:
What does 2024 mean?

siqbal1986:
@qiluo-msft thats the bug id this one fixes.

PR #2482, opened 2022-11-06
Add "f" option for executing reboot force mode

Liran-Ar:
**Why I did it**
The vendor platform reboot, which is called from the reboot script, supports force mode, and for now there is no option to execute it via the reboot script, so this change will enable it.

**What I did**
Add an "f" option in reboot_script for executing reboot force mode.
Add this option to help command as well.

**How to verify it**
Execute the reboot script with "-f" and see that the vendor platform reboot executes the force mode indeed.

### DIFF
--- a/scripts/reboot
+++ b/scripts/reboot
@@ -121,6 +121,7 @@ function show_help_and_exit()
     echo " "
     echo "    Available options:"
     echo "        -h, -? : getting this help"
+    echo "        -f     : execute reboot force"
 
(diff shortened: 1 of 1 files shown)


liat-grozovik (commented):
please check that the help command has this option as well.

Liran-Ar:
The option has been added, thanks.

liat-grozovik (approved):

PR #2483, opened 2022-11-07
[GCU] Get config thru cfggen

wen587:
#### What I did
Getting configs from CLI is very slow, need to get it from sonic-cffgen directly. 
#### How I did it
Change CLI to sonic-cfggen
#### How to verify it
No failure UT.


wen587:
keep original design

PR #2500, opened 2022-11-15
[GCU] Add RemoveCreateOnlyDependency Validator/Generator

wen587:
#### What I did
Refer https://github.com/sonic-net/sonic-utilities/pull/2273 

Efficiency Improvement:
- Add RemoveCreateOnlyDependency Generator and Validator. 

What is the problem?
* Lanes is create-only field, so the parent `/PORT/Ethernet124` for example needs to be deleted then added.
* Deletion of the `/PORT/Ethernet124` requires deletion of all of its dependencies.
* In the DUT there are around 18 different dependencies
* The search algorithm to find the moves to apply is not very efficient in this case, what it does is find the first dependency delete it. Then it will try to add the dependency back because it thinks this will bring us closer to the final goal config. But then it realizes adding the config back, will bring us back to a visited state. So we move on and try to delete another dependency, after that the search algorithm will try to add the first dependency, but this time the config state was not visited before because that's a new state.

Check the following to better understand:
initial state:
```
port ethernet124
dep1
dep2
dep3
```
Deletion dep1
```
port ethernet124
dep2
dep3
```
Adding dep1 back will rejected, because it will move us back to the initial state
Try and delete dep2
```
port ethernet124
dep3
```
Try and add dep1
```
port ethernet124
dep1
dep3
```
Now the above is a valid state, but it is useless. If you can expand that to 18 dependencies we end up with a lot of tries that are useless.

Solution is to add a custom validator that understand that once a dependency is deleted, do not add it back until we delete the `/PORT/Ethernet124`

Also add a generator for the deletion to make it faster to pick the dependencies to remove.

#### How I did it
Added new validator/generator for handling the lane replacement case. The validator/generator understands that for which create-only fields and their dependencies need to be removed.

#### How to verify it
Unit Test.


generic_config_updater/patch_sorter.py
```diff
@@ -544,6 +544,66 @@ def _get_default_value_from_settings(self, parent_path, field_name):
...
+
+        simulated_port = self.path_addressing.get_from_path(simulated_config, f"/PORT/{port_name}")
+
+        if simulated_port is None:
+            return True
+
+        current_admin_status = self.path_addressing.get_from_path(current_config, f"/PORT/{port_name}/admin_status")
+        simulated_admin_status = self.path_addressing.get_from_path(simulated_config, f"/PORT/{port_name}/admin_status")
+        if current_admin_status != simulated_admin_status and current_admin_status != "up":
+            return False
```
wen587, line 595: I am not sure if the admin_status comparison is needed

generic_config_updater/gu_common.py
```diff
@@ -54,8 +54,7 @@ def get_config_db_as_json(self):
         return json.loads(text)
 
     def _get_config_db_as_text(self):
-        # TODO: Getting configs from CLI is very slow, need to get it from sonic-cffgen directly
-        cmd = "show runningconfiguration all"
+        cmd = "sonic-cfggen -d --print-data"
```
qiluo-msft, line 57: >cmd = "sonic-cfggen -d --print-data" [](http://example.com/codeflow?start=8&length=36)

This is a valid performance improvement, and not related to "LaneReplacement Generator".

Could you extract it into a separate PR? also add some performance measurement result.

generic_config_updater/patch_sorter.py
```diff
@@ -544,6 +544,66 @@ def _get_default_value_from_settings(self, parent_path, field_name):
 
         return None
 
+class LaneReplacementMoveValidator:
```
qiluo-msft, line 547: >LaneReplacementMoveValidator [](http://example.com/codeflow?start=6&length=28)

Checking the implementation, I do not see the relationship of "lane replacement", I am also thinking it could be generalized and not limited to PORT table.

generic_config_updater/gu_common.py
wen587: I've made it a separate PR but got the coverage check failure.
That is because, in our test, the function is being mocked. If we do performance measurement, I think we need to put that in config CLI test.

Local manual test for comparison.
```
admin@vlab-01:~/time$ python check.py  | grep "\-\-\-"
show run all --- 0.7292575836181641 seconds ---
sonic-cfggen --- 0.0 seconds ---
```

qiluo-msft: Offline discussed, we need to create an issue that "show runningconfiguration all" is slow, and we need to fix in its own implementation.

wen587: My previous time check is wrong. Please check the below metric.

```
--- SINGLE RUN
show SINGLE --- 0.7178833484649658 seconds ---
sonic-cfggen SINGLE --- 0.3109145164489746 seconds ---
--- 10 RUNS
show AVERAGE --- 0.7603521585464478 seconds ---
sonic-cfggen AVERAGE --- 0.28761484622955324 seconds ---
```

generic_config_updater/patch_sorter.py
```diff
@@ -544,6 +544,119 @@ def _get_default_value_from_settings(self, parent_path, field_name):
...
+        if current_field == target_field:
+            return True
+
+        simulated_member = self.path_addressing.get_from_path(
+            simulated_config, f"/{table_to_check}/{member_name}")
+
+        if simulated_member is None:
+            return True
+
+        if table_to_check == "PORT":
```
wen587, line 639: Hi @qiluo-msft , still need to confirm admin_status need to be set with "down" for PORT["lanes"] change.

qiluo-msft: Yes. PORT["lanes"] are created only field. If the user require to change this field, we need to delete port and create new port. It is already a disruptive change, set `admin_status` to `down` will make it super safe on all platform, and does not make things worse.

qiluo-msft (approved):

wen587:
Should also add to 202211.

PR #2503, opened 2022-11-17
[QoS] Introduce delay to the qos reload flow

DavidZagury:
Fix issue with qos reload sometime causes error from orchagent "Failed to remove dscp_to_tc map"
Fix issue with qos reload with dry run clears qos configurations

#### What I did
Added a delay mechanism to the qos reload flow so that after qos clear we will wait for the APP DB tables to be clear before continuing to the reload phase.

Disabled qos clear when in dry run.

#### How I did it
After the buffer tables have been deleted there will be check whether the APP DB table BUFFER_POOL_TABLE is deleted. Only after it will be deleted we will continue with the reload flow.


stephenxs (dismissed):

stephenxs (approved):

keboliu:
@neethajohn would you please help to review this PR?

liat-grozovik:
@neethajohn kindly reminder to review and provide feedback

config/main.py
```diff
@@ -2591,15 +2611,16 @@ def _update_buffer_calculation_model(config_db, model):
...
     """Reload QoS configuration"""
     if ports:
         log.log_info("'qos reload --ports {}' executing...".format(ports))
         _qos_update_ports(ctx, ports, dry_run, json_data)
         return
 
     log.log_info("'qos reload' executing...")
-    _clear_qos()
+    if not dry_run:
+        _clear_qos(delay = no_delay)
```
neethajohn, line 2623: The usage is confusing. no_delay flag is set to True in load_minigraph, so the user expects that there will be no delay but the code eventually translates that no_delay to 'delay' and introduces a wait.
IMO, the call should be _clear_qos(no_delay=no_delay) and execute the wait only if '--no-delay' flag is not called in the command 'config qos reload'
The current usage of adding a worst case 30s time to load_minigraph which should be avoided

config/main.py
```diff
@@ -728,7 +728,24 @@ def storm_control_delete_entry(port_name, storm_type):
...
+
+    while not empty and time.time() - start < timeout:
+        current_profiles = app_db.keys(app_db.APPL_DB, table)
+        if not current_profiles:
+            empty = True
+        else:
+            time.sleep(interval)
+    if not empty:
+        click.echo("Operation not completed successfully, please save and reload configuration.")
+    return empty
```
neethajohn, line 745: No need to return here since the calling function does not handle the return value

DavidZagury: > No need to return here since the calling function does not handle the return value

@neethajohn it wasn't needed for the function call, but I used it on the unit tests.
Do you have any suggestion on how to test it without the return?

config/main.py
DavidZagury: @neethajohn  You are correct, that was suppose to be the other way around. Fixed the mistake, thank you.

neethajohn (approved):

PR #2504, opened 2022-11-17
[generate_dump] Optimize the execution time of the 'show techsupport' script to 5-10% by reducing calls to the 'tar append' operation

vadymhlushko-mlnx:
Signed-off-by: Vadym Hlushko <vadymh@nvidia.com>

#### What I did
Optimize the execution time of the 'show techsupport' script to 5-10%.

#### How I did it
The `show techsupport` CLI command calls the `generate_dump` bash script. In the script, there are a many functions that do the next scenario:
```
1. Run some CLI command
2. Save output from step 1 to the temporary file
3. Append the temporary file from step 2 to the `/var/dump/sonic_dump_XXXX.tar` file
4. Delete the temporary file from step 2
```

This PR removes the `3 and 4 step` from those functions and creates a new function `save_to_tar()` which will add to `.tar` archive the whole directory with temporary files (which means it will not spawn a `tar -v -rhf ...` process for each temporary file)

#### How to verify it
Run the `time show techsupport` CLI command and compare the execution time to the original script, the execution time will be decreased by 5-10%.

### DIFF
--- a/scripts/generate_dump
+++ b/scripts/generate_dump
@@ -106,7 +106,6 @@ save_bcmcmd() {
     local filename=$2
     local filepath="${LOGDIR}/$filename"
     local do_gzip=${3:-false}
-    local tarpath="${BASE}/dump/$filename"
     local timeout_cmd="timeout --foreground ${TIMEOUT_MIN}m"
     local cmd=$(escape_quotes "$cmd")
     if [ ! -d $LOGDIR ]; then
@@ -141,12 +140,9 @@ save_bcmcmd() {
     fi
     if $do_gzip; then
         gzip ${filepath} 2>/dev/null
-        tarpath="${tarpath}.gz"
         filepath="${filepath}.gz"
     fi
-    ($TAR $V -rhf $TARFILE -C $DUMPDIR "$tarpath" \
-        || abort "${EXT_TAR_FAILED}" "tar append operation failed. Aborting to prevent data loss.") \
-        && $RM $V -rf "$filepath"
+
     end_t=$(date +%s%3N)
     echo "[ save_bcmcmd:$cmd ] : $(($end_t-$start_t)) msec" >> $TECHSUPPORT_TIME_INFO
 }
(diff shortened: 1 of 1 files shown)


stepanblyschak (approved):

scripts/generate_dump
```diff
@@ -1298,11 +1300,7 @@ save_warmboot_files() {
     else
         mkdir -p $TARDIR
         $CP $V -rf /host/warmboot $TARDIR
-
-        ($TAR $V --warning=no-file-removed  -rhf $TARFILE -C $DUMPDIR --mode=+rw \
-            $BASE/warmboot \
-            || abort "${EXT_TAR_FAILED}" "Tar append operation failed. Aborting for safety.") \
-            && $RM $V -rf $TARDIR
+        chmod ugo+rw -R $DUMPDIR/$BASE/warmboot
```
qiluo-msft, line 1303: >chmod ugo+rw -R $DUMPDIR/$BASE/warmboot [](http://example.com/codeflow?start=8&length=39)

Could you add some code comment?

scripts/generate_dump
```diff
@@ -260,13 +254,35 @@ save_cmd() {
...
+# Returns:
+#  None
+###############################################################################
+save_to_tar() {
+    trap 'handle_error $? $LINENO' ERR
+    local start_t=$(date +%s%3N)
+    local end_t=0
+
+    cd $DUMPDIR
+    $TAR $V -rhf $TARFILE "$BASE"
```
qiluo-msft, line 280: >$TAR [](http://example.com/codeflow?start=4&length=4)

Could you delete files after tar?

scripts/generate_dump
vadymhlushko-mlnx: In line `1302` of the previous code, it will add the `warmboot` folder to archive with `--mode=+rw`, so I did the same but with the different command

scripts/generate_dump
vadymhlushko-mlnx: It will be done later on in function `handle_exit()`

vadymhlushko-mlnx:
@wen587 kind reminder to review

wen587 (approved):

yxieca:
@vadymhlushko-mlnx this change cannot be cherry-picked to 202205 cleanly. Please create separate PR.

PR #2515, opened 2022-11-22
Port 202012 DB migration changes to newer branches

vaibhavhd:
#### What I did

Fixes https://github.com/sonic-net/sonic-buildimage/issues/12643


dgsudharsan (approved):

tjchadaga:
@qiluo-msft - please help revert this PR as it is causing https://github.com/sonic-net/sonic-buildimage/issues/12955

vaibhavhd:
> @qiluo-msft This needs to be reverted only in master and not in 202205 since 202205 still has swsssdk.

I think revert is not needed since now we have fixed this as part of https://github.com/sonic-net/sonic-utilities/pull/2544

PR #2518, opened 2022-11-22
[drop counters] Fix CLI script for unconfigured PGs

liorghub:
#### What I did
Since PG counters are created only if they are configured in the switch, it is not enough to relay only on the first entry in the DB when building the output table of watermarkstat script.
We need to go over all configured counters, check what is the max configured and build the table accordingly.

#### How I did it
Iterate all configured PG buffers for all ports and find the max index.
Build the output table according to the max index.

#### How to verify it
Run test "pgdropstat_test.py" including this PR and observe it passes.

- N/A

- N/A

### DIFF
--- a/scripts/pg-drop
+++ b/scripts/pg-drop
@@ -116,16 +116,23 @@ class PgDropStat(object):
 
         self.header_list = ['Port']
         header_map = pg_drop_type["obj_map"]
-        single_key = list(header_map.keys())[0]
-        header_len = len(header_map[single_key])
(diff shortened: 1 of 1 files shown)


nazariig (approved):

liat-grozovik:
@liorghub  on which branches this need to be fixed? please specify so we can add the right labels.

liorghub:
@liat-grozovik This PR should be merged to branch 202211 as well.

liat-grozovik (approved):

PR #2524, opened 2022-11-24
Preserve copp tables through DB migration

arfeigin:
This PR should be merged together with https://github.com/sonic-net/sonic-swss/pull/2548 and is required to 202205 and 202211.
This PR implements [fastboot] Preserve CoPP table HLD to improve fastboot flow (https://github.com/sonic-net/SONiC/pull/1107).

#### What I did
Preserve COPP table contents through DB migration. (Mellanox only)

#### How I did it
Skipped deleting of COPP tables in DB migrator.

#### How to verify it
Observe COPP table contents are preserved right after reboot.

### DIFF
--- a/scripts/db_migrator.py
+++ b/scripts/db_migrator.py
@@ -868,6 +868,10 @@ class DBMigrator():
                 new_cfg = {**init_cfg, **curr_cfg}
                 self.configDB.set_entry(init_cfg_table, key, new_cfg)
 
+        # Avoiding copp table migration is platform specific at the moment as I understood this might cause issues for some
+        # vendors, probably Broadcom. This change can be checked with any specific vendor and if this works fine the platform
+        # condition can be modified and extend. If no vendor has an issue with not clearing copp tables the condition can be
+        # removed together with calling to migrate_copp_table function.
         if self.asic_type != "mellanox":
             self.migrate_copp_table()
         if self.asic_type == "broadcom" and 'Force10-S6100' in self.hwsku:            
--- a/scripts/db_migrator.py
+++ b/scripts/db_migrator.py
@@ -868,7 +868,8 @@ class DBMigrator():
                 new_cfg = {**init_cfg, **curr_cfg}
                 self.configDB.set_entry(init_cfg_table, key, new_cfg)
 
-        self.migrate_copp_table()
+        if self.asic_type != "mellanox":
+            self.migrate_copp_table()
         if self.asic_type == "broadcom" and 'Force10-S6100' in self.hwsku:            
             self.migrate_mgmt_ports_on_s6100()
         else:


stepanblyschak (dismissed):

dgsudharsan (approved):

scripts/db_migrator.py
```diff
@@ -868,7 +868,8 @@ def common_migration_ops(self):
                 new_cfg = {**init_cfg, **curr_cfg}
                 self.configDB.set_entry(init_cfg_table, key, new_cfg)
 
-        self.migrate_copp_table()
+        if self.asic_type != "mellanox":
+            self.migrate_copp_table()
```
venkatmahalingam, line 872: Looks like this change can be platform agnostic, the change looks good for now as we don't know the behavior on other platforms. Please add a comment here so that as and when we validate this change for other platforms, we can make the changes accordingly here.

scripts/db_migrator.py
```diff
@@ -868,7 +868,8 @@ def common_migration_ops(self):
                 new_cfg = {**init_cfg, **curr_cfg}
                 self.configDB.set_entry(init_cfg_table, key, new_cfg)
 
-        self.migrate_copp_table()
+        if self.asic_type != "mellanox":
```
rck-innovium, line 871: It is not clear if there is any differences in the calls made to syncd/SAI. Can you explain why this check needs to be platform specific? Please also explain the difference in terms of SAI API calls?

svshah-intel: yeah, would like to know as well. Much of the coppmgr code is same for both the cases except handling dup entries, not clear what differs.

arfeigin: This check is platform specific since it was brought to my attention that avoiding clearing the copp tables might cause issues on Broadcom platforms. I am not familiar with their implementation and from my point of view this can be platform agnostic as it was commented in the review. I guess the issue is with SAI handling overwrites or writing to create-only attributes.
The coppmgr code is platform agnostic as mentioned.
I will add a comment in the code according to @venkatmahalingam comment and in case you see that this change works fine with other vendors as well they can be added to the condition check. In case no vendor has any issues with that the condition check can be removed and then just delete calling migrate_copp_table function.

scripts/db_migrator.py
arfeigin: Thanks, added the comment.

liat-grozovik (approved):

dgsudharsan (approved):

venkatmahalingam (approved):

scripts/db_migrator.py
svshah-intel: I would have thought copporch listening to AppDb should never get update for new created entries and should never get create for updated entries, so not sure what scenario creates writing to create-only attributes?

PR #2532, opened 2022-11-29
Add support of secure warm-boot

davidpil2002:
#### What I did
Add support of secure warm-boot to SONiC. 
Basically, warm-boot is supporting to load a new kernel without doing full/cold boot.
That is by loading a new kernel and exec with kexec Linux command. As a result of that, even when the Secure Boot feature is enabled, still a user or a malicious user can load an unsigned kernel, so to avoid that we added the support of the secure warm boot.
More Description about this feature can be found in the Secure Boot HLD: https://github.com/sonic-net/SONiC/pull/1028

#### How I did it
In general, Linux support it, so I enabled this support by doing the follow steps:
1. I added some special flags in Linux Kernel when user build the sonic-buildimage with secure boot feature enabled.
2. I added a flag "-s" to the kexec command
Note: more details in the HLD above.
#### How to verify it
- Good flow
1. manually just install with sonic-installed a new secure image (a SONiC image that was build with Secure Boot flag enabled)
2. after the secure image is installed, do:
`warm-reboot`
3. Check now that the new kernel is really loaded and switched.
- Bad flow:
1. Do the same steps 1-2 as a good flow but with an insecure image (SONiC image that was built without setting Secure Boot enabled)
2. After the insecure image is installed, and triggered warm-boot you should get an error that the new unsigned kernel from the unsecured image was not loaded.

Automation test - TBD

no changed
no changed

NOTE:
PR of the Secure Boot [sonic-buildimage]: https://github.com/sonic-net/sonic-buildimage/pull/12692
PR of Secure Boot [sonic-linux-kernel]: https://github.com/sonic-net/sonic-linux-kernel/pull/298
This is an old paper: https://lwn.net/Articles/603116/
but explaining why it's critical to add this support.

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -608,15 +608,17 @@ if [[ "$sonic_asic_type" == "mellanox" ]]; then
     fi
 fi
 
-# check if secure boot is enable in UEFI 
-SECURE_UPGRADE_ENABLED=$(bootctl status 2>/dev/null | grep -c "Secure Boot: enabled")
 
 if is_secureboot && grep -q aboot_machine= /host/machine.conf; then
     load_aboot_secureboot_kernel
-elif [ ${SECURE_UPGRADE_ENABLED} -eq 1 ]; then
-    load_kernel_secure
 else
-    load_kernel
+    # check if secure boot is enable in UEFI 
+    SECURE_UPGRADE_ENABLED=$(bootctl status 2>/dev/null | grep -c "Secure Boot: enabled")
+    if [ ${SECURE_UPGRADE_ENABLED} -eq 1 ]; then
+        load_kernel_secure
+    else
+        load_kernel
+    fi
 fi
 
 init_warm_reboot_states
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -447,6 +447,13 @@ function load_kernel() {
     /sbin/kexec -a -l "$KERNEL_IMAGE" --initrd="$INITRD" --append="$BOOT_OPTIONS"
 }
 
+function load_kernel_secure() {
+    # Load kernel into the memory secure
+    # -s flag is for enforcing the new load kernel(vmlinuz) to be signed and verify.
+    # not using -a flag, this flag can fallback to an old kexec load that do not support Secure Boot verification
+    /sbin/kexec -l "$KERNEL_IMAGE" --initrd="$INITRD" --append="$BOOT_OPTIONS" -s
+}
+
 function unload_kernel()
 {
     # Unload the previously loaded kernel if any loaded
@@ -597,9 +604,13 @@ if [[ "$sonic_asic_type" == "mellanox" ]]; then
     fi
 fi
 
+# check if secure boot is enable in UEFI 
+SECURE_UPGRADE_ENABLED=$(bootctl status 2>/dev/null | grep -c "Secure Boot: enabled")
 
 if is_secureboot && grep -q aboot_machine= /host/machine.conf; then
     load_aboot_secureboot_kernel
+elif [ ${SECURE_UPGRADE_ENABLED} -eq 1 ]; then
+    load_kernel_secure
 else
     load_kernel
 fi
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -442,16 +442,20 @@ function load_aboot_secureboot_kernel() {
         swipath=$next_image kexec=true loadonly=true ENV_EXTRA_CMDLINE="$BOOT_OPTIONS" bash -
 }
 
+function invoke_kexec() {
+    /sbin/kexec -l "$KERNEL_IMAGE" --initrd="$INITRD" --append="$BOOT_OPTIONS" $@
+}
+
 function load_kernel() {
     # Load kernel into the memory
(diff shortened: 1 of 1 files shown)


stepanblyschak (dismissed):

liat-grozovik:
@davidpil2002  can you please resolve conflicts?
@qiluo-msft can you please review or assign someone to review?

liat-grozovik:
@qiluo-msft  @xumia could you please help to review this PR as part of the secured boot/update feature?

davidpil2002:
> @davidpil2002 can you please handle conflicts?

sure, the conflict was that was added when using `kexec` a flag named `-a` `(--kexec-syscall-auto),` I read about it to be sure, and it not so secure to use it.
So, when the flow its **without** secure boot I will not touch the flag, but when the flow its using Secure boot **enabled**, I will not add this flag.

can be seen more info here:
https://www.spinics.net/lists/kexec/msg30932.html
https://man7.org/linux/man-pages/man8/kexec.8.html
" Try the new KEXEC_FILE_LOAD syscall first and when it is not supported or the
 kernel does not understand the supplied image fall back to the old KEXEC_LOAD
 interface."

The old KEXEC_LOAD its not supporting Secure Boot, its not including a Linux patch that verify the kernel when loading an new kernel image

liat-grozovik:
@stepanblyschak , @qiluo-msft could you please help to review?

scripts/fast-reboot
```diff
@@ -447,6 +447,13 @@ function load_kernel() {
     /sbin/kexec -a -l "$KERNEL_IMAGE" --initrd="$INITRD" --append="$BOOT_OPTIONS"
 }
 
+function load_kernel_secure() {
+    # Load kernel into the memory secure
+    # -s flag is for enforcing the new load kernel(vmlinuz) to be signed and verify.
+    # not using -a flag, this flag can fallback to an old kexec load that do not support Secure Boot verification
+    /sbin/kexec -l "$KERNEL_IMAGE" --initrd="$INITRD" --append="$BOOT_OPTIONS" -s
```
stepanblyschak, line 454: Would it be beneficial if we extend ```load_kernel``` with argument whether to perform sign verification instead of copy/paste most of the kexec parameters here?

scripts/fast-reboot
```diff
@@ -597,9 +604,13 @@ if [[ "$sonic_asic_type" == "mellanox" ]]; then
     fi
 fi
 
+# check if secure boot is enable in UEFI 
+SECURE_UPGRADE_ENABLED=$(bootctl status 2>/dev/null | grep -c "Secure Boot: enabled")
```
stepanblyschak, line 608: I think this var is not needed in aboot secure boot case or it might confuse that we have both SECURE_UPGRADE_ENABLED and is_secureboot

scripts/fast-reboot
davidpil2002: in my previous implementation, I extended the command, but now it's not possible because was added a new flag that is not secure.
so we need to add and remove flags when for the SB use case.

scripts/fast-reboot
davidpil2002: `is_secureboot `is based on `aboot `secure boot - "Arista secure boot" was implemented a different process to secure boot the system, not related to our approach and our HLD review to the community.
their flag `is_secureboot` is based on the information from /proc/cmdline, and in our flow this information it's not including the info about secure boot enabled.

In our generic flow, we are reading the info about secure boot state from `bootctl `and that it's according to the flag in UEFI/BIOS, it's an accurate way to know that the system is really loaded with SB enabled/disabled.

scripts/fast-reboot
stepanblyschak: How about? To avoid duplicating kexec line
```bash
function invoke_kexec() {
  /sbin/kexec -l "$KERNEL_IMAGE" --initrd="$INITRD" --append="$BOOT_OPTIONS" $@
}

function load_kernel() {
   invoke_kexec() -a
}

function load_kernel_secure() {
  invoke_kexec() -s
}
```

scripts/fast-reboot
stepanblyschak: Right, that means SECURE_UPGRADE_ENABLED is relevant only in case the platform is not aboot platform, correct? Or same new approach can be used for aboot?

scripts/fast-reboot
davidpil2002: I think this one can work, I will add a commit

davidpil2002 (commented):
fix last comment.

scripts/fast-reboot
davidpil2002: well, if aboot its not with secure boot they still need load kernel regular, and I dont like to create a complex flow or duplicate code.
I moved now the bootctl call in the else case, so its a practical and easy solution and should be a robust solutions for all the vendors.
pls approve or suggest other solution
thanks

stepanblyschak (approved):

xumia (approved):

StormLiangMS:
@davidpil2002 could  you raise separate PR to 202211?

davidpil2002:
> @davidpil2002 could you raise separate PR to 202211?

yes, I will

liat-grozovik:
@davidpil2002 can you please refer the 202211 PR to this one?

davidpil2002:
> @davidpil2002 can you please refer the 202211 PR to this one?

sure, I will.
(probably next week)

PR #2537, opened 2022-12-05
[show] Add bgpraw to show run all

wen587:
#### What I did
Add bgpraw output to `show runningconfiguration all`
```
Requirements:
1. current `show runningconfig` will print all the ConfigDB in a json format, we need to add a new key-value into the json output "bgpraw" with a long string value
2. The long string value should be the output of `vtysh -c "show run"`. It is normally multiline string, may include special characters like \". Need to make sure the escaping properly
3. We do not need to insert the key-value into ConfigDB is not existing there
4. If ConfigDB already has the key-value, we do not need to override it by vtysh command output
5. Not break multi-asic use
```
#### How I did it
Generate bgpraw output then append it to `show runnningconfiguration all`'s output
#### How to verify it
Mannual test
```
admin@vlab-01:~$ show run all
{
    "ACL_TABLE": {
......
    "WRED_PROFILE": {
        "AZURE_LOSSLESS": {
            "ecn": "ecn_all",
            "green_drop_probability": "5",
            "green_max_threshold": "2097152",
            "green_min_threshold": "1048576",
            "red_drop_probability": "5",
            "red_max_threshold": "2097152",
            "red_min_threshold": "1048576",
            "wred_green_enable": "true",
            "wred_red_enable": "true",
            "wred_yellow_enable": "true",
            "yellow_drop_probability": "5",
            "yellow_max_threshold": "2097152",
            "yellow_min_threshold": "1048576"
        }
    }
}
```
```
admin@vlab-01:~$ show run all
{
    "ACL_TABLE": {
......
    "WRED_PROFILE": {
        "AZURE_LOSSLESS": {
            "ecn": "ecn_all",
            "green_drop_probability": "5",
            "green_max_threshold": "2097152",
            "green_min_threshold": "1048576",
            "red_drop_probability": "5",
            "red_max_threshold": "2097152",
            "red_min_threshold": "1048576",
            "wred_green_enable": "true",
            "wred_red_enable": "true",
            "wred_yellow_enable": "true",
            "yellow_drop_probability": "5",
            "yellow_max_threshold": "2097152",
            "yellow_min_threshold": "1048576"
        }
    },
    "bgpraw": "Building configuration...\n\nCurrent configuration......end\n"
}
```

### DIFF
--- a/show/main.py
+++ b/show/main.py
@@ -132,7 +132,7 @@ def run_command(command, display_cmd=False, return_cmd=False):
 
 def get_cmd_output(cmd):
     proc = subprocess.Popen(cmd, text=True, stdout=subprocess.PIPE)
-    return proc.communicate()[0]
+    return proc.communicate()[0], proc.returncode
 
 # Lazy global class instance for SONiC interface name to alias conversion
 iface_alias_converter = lazy_object_proxy.Proxy(lambda: clicommon.InterfaceAliasConverter())
@@ -1389,7 +1389,10 @@ def runningconfiguration():
 def all(verbose):
     """Show full running configuration"""
     cmd = ['sonic-cfggen', '-d', '--print-data']
-    stdout = get_cmd_output(cmd)
+    stdout, rc = get_cmd_output(cmd)
+    if rc:
+        click.echo("Failed to get cmd output '{}':rc {}".format(cmd, rc))
+        raise click.Abort()
 
     try:
         output = json.loads(stdout)
@@ -1397,9 +1400,11 @@ def all(verbose):
         click.echo("Failed to load output '{}':{}".format(cmd, e))
         raise click.Abort()
 
-    if 'bgpraw' in output or not multi_asic.is_multi_asic():
+    if not multi_asic.is_multi_asic():
         bgpraw_cmd = [constants.RVTYSH_COMMAND, '-c', 'show running-config']
-        bgpraw = get_cmd_output(bgpraw_cmd)
+        bgpraw, rc = get_cmd_output(bgpraw_cmd)
+        if rc:
+            bgpraw = ""
         output['bgpraw'] = bgpraw
     click.echo(json.dumps(output, indent=4))
 
--- a/show/main.py
+++ b/show/main.py
@@ -1385,7 +1385,7 @@ def all(verbose):
     """Show full running configuration"""
     cmd = "sonic-cfggen -d --print-data"
     output = json.loads(run_command(cmd, display_cmd=verbose, return_cmd=True))
-    if not multi_asic.is_multi_asic():
+    if 'bgpraw' in output or not multi_asic.is_multi_asic():
         bgpraw_cmd = "sudo {} -c 'show running-config'".format(constants.RVTYSH_COMMAND)
         bgpraw = run_command(bgpraw_cmd, display_cmd=verbose, return_cmd=True)
         output['bgpraw'] = bgpraw
--- a/show/main.py
+++ b/show/main.py
@@ -130,7 +130,7 @@ def run_command(command, display_cmd=False, return_cmd=False):
         sys.exit(rc)
 
 def get_cmd_output(cmd):
-    proc = subprocess.Popen(cmd, shell=True, text=True, stdout=subprocess.PIPE)
+    proc = subprocess.Popen(cmd, text=True, stdout=subprocess.PIPE)
     stdout, stderr = proc.communicate()
     if proc.returncode != 0:
         click.echo("Command failed '{}': {}".format(cmd, stderr))
@@ -1391,7 +1391,7 @@ def runningconfiguration():
 @click.option('--verbose', is_flag=True, help="Enable verbose output")
 def all(verbose):
     """Show full running configuration"""
-    cmd = "sonic-cfggen -d --print-data"
+    cmd = ['sonic-cfggen', '-d', '--print-data']
     stdout = get_cmd_output(cmd)
 
     try:
@@ -1401,7 +1401,7 @@ def all(verbose):
         raise click.Abort()
 
     if 'bgpraw' in output or not multi_asic.is_multi_asic():
-        bgpraw_cmd = "{} -c 'show running-config'".format(constants.RVTYSH_COMMAND)
+        bgpraw_cmd = [constants.RVTYSH_COMMAND, '-c', 'show running-config']
         bgpraw = get_cmd_output(bgpraw_cmd)
         output['bgpraw'] = bgpraw
     click.echo(json.dumps(output, indent=4))
--- a/show/main.py
+++ b/show/main.py
@@ -19,6 +19,7 @@ from utilities_common.db import Db
 from datetime import datetime
 import utilities_common.constants as constants
 from utilities_common.general import load_db_config
+from json.decoder import JSONDecodeError
 
 # mock the redis for unit test purposes #
 try:
@@ -131,11 +132,7 @@ def run_command(command, display_cmd=False, return_cmd=False):
 
 def get_cmd_output(cmd):
     proc = subprocess.Popen(cmd, text=True, stdout=subprocess.PIPE)
-    stdout, stderr = proc.communicate()
-    if proc.returncode != 0:
-        click.echo("Command failed '{}': {}".format(cmd, stderr))
-        raise click.Abort()
-    return stdout
+    return proc.communicate()[0]
 
 # Lazy global class instance for SONiC interface name to alias conversion
 iface_alias_converter = lazy_object_proxy.Proxy(lambda: clicommon.InterfaceAliasConverter())
@@ -1396,7 +1393,7 @@ def all(verbose):
 
     try:
         output = json.loads(stdout)
-    except Exception as e:
+    except JSONDecodeError as e:
         click.echo("Failed to load output '{}':{}".format(cmd, e))
         raise click.Abort()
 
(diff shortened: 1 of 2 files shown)


show/main.py
```diff
@@ -1384,7 +1384,12 @@ def runningconfiguration():
 def all(verbose):
     """Show full running configuration"""
     cmd = "sonic-cfggen -d --print-data"
-    run_command(cmd, display_cmd=verbose)
+    output = json.loads(run_command(cmd, display_cmd=verbose, return_cmd=True))
+    if not multi_asic.is_multi_asic():
+        bgpraw_cmd = "sudo {} -c 'show running-config'".format(constants.RVTYSH_COMMAND)
```
qiluo-msft, line 1389: >sudo [](http://example.com/codeflow?start=22&length=4)

Could you check if really need `sudo`? #Closed

show/main.py
```diff
@@ -1384,7 +1384,12 @@ def runningconfiguration():
 def all(verbose):
     """Show full running configuration"""
     cmd = "sonic-cfggen -d --print-data"
-    run_command(cmd, display_cmd=verbose)
+    output = json.loads(run_command(cmd, display_cmd=verbose, return_cmd=True))
```
ganglyu, line 1387: Should we check the output of run_command?

show/main.py
```diff
@@ -1384,7 +1384,12 @@ def runningconfiguration():
...
+    if not multi_asic.is_multi_asic():
+        bgpraw_cmd = "sudo {} -c 'show running-config'".format(constants.RVTYSH_COMMAND)
+        bgpraw = run_command(bgpraw_cmd, display_cmd=verbose, return_cmd=True)
+        output['bgpraw'] = bgpraw
```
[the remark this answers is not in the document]
wen587: The existing check is for hwproxy use when they are not sure if bgpraw has already been included in `show run all`
For sonic show command, sonic-cfggen will not generate bgpraw output. We can ignore the check, I think.

show/main.py
wen587: You mean the format of output or bgpraw in output?

qiluo-msft, line 1387: >display_cmd [](http://example.com/codeflow?start=41&length=11)

`run_command` will internally call `click.echo()` or `sys.exit()`, so there will be side-effect on corner cases.

show/main.py
arlakshm: I think if we use `rvtysh` we dont need need `sudo`. I think if we use sudo then the user has have RW privilages

show/main.py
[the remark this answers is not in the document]
arlakshm: I think the current implementation is correct. This way we  always update with the latest configuration from the bgp container.

show/main.py
```diff
@@ -1384,7 +1384,12 @@ def runningconfiguration():
 def all(verbose):
     """Show full running configuration"""
     cmd = "sonic-cfggen -d --print-data"
-    run_command(cmd, display_cmd=verbose)
+    output = json.loads(run_command(cmd, display_cmd=verbose, return_cmd=True))
+    if not multi_asic.is_multi_asic():
+        bgpraw_cmd = "sudo {} -c 'show running-config'".format(constants.RVTYSH_COMMAND)
+        bgpraw = run_command(bgpraw_cmd, display_cmd=verbose, return_cmd=True)
```
arlakshm, line 1390: if the bgp container is down, bgpraw will contain error messge. Do we want to store that or make `output['bgpraw'] `empty?

arlakshm (commented):
as comments

show/main.py
qiluo-msft: Empty running-config value for bgp down is a good suggestion.

wen587: The run_command was calling subprocess. The output will be empty though subprocess calling is error. 
So, the `bgpraw` will be empty if bgp container is down.

show/main.py
ganglyu: geoffrey@vm-04:~$ python3
Python 3.6.9 (default, Nov 25 2022, 14:10:45) 
[GCC 8.4.0] on linux
>>> import json
>>> json.loads("")
  File "/usr/lib/python3.6/json/__init__.py", line 354, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python3.6/json/decoder.py", line 339, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python3.6/json/decoder.py", line 357, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
>>>

wen587: Maybe that should be checked in sonic-cfggen?
sonic-cfggen guarantees that `--print-data` could be json.dump.

show/main.py
```diff
@@ -1383,8 +1391,20 @@ def runningconfiguration():
 @click.option('--verbose', is_flag=True, help="Enable verbose output")
 def all(verbose):
     """Show full running configuration"""
-    cmd = "sonic-cfggen -d --print-data"
-    run_command(cmd, display_cmd=verbose)
+    cmd = ['sonic-cfggen', '-d', '--print-data']
+    stdout = get_cmd_output(cmd)
```
qiluo-msft, line 1395: >get_cmd_output [](http://example.com/codeflow?start=13&length=14)

Original behavior is that even rc is non-zero, the command output is printed as output. Do you  want to keep the behavior?

show/main.py
```diff
@@ -1383,8 +1391,20 @@ def runningconfiguration():
...
+
+    try:
+        output = json.loads(stdout)
+    except Exception as e:
```
qiluo-msft, line 1399: >Exception [](http://example.com/codeflow?start=11&length=9)

Do not catch general exception type in normal practice.. Could you use a more specific type?

show/main.py
```diff
@@ -1383,8 +1388,20 @@ def runningconfiguration():
 @click.option('--verbose', is_flag=True, help="Enable verbose output")
 def all(verbose):
     """Show full running configuration"""
-    cmd = "sonic-cfggen -d --print-data"
-    run_command(cmd, display_cmd=verbose)
+    cmd = ['sonic-cfggen', '-d', '--print-data']
+    stdout = get_cmd_output(cmd)
```
qiluo-msft, line 1392: >stdout [](http://example.com/codeflow?start=4&length=6)

suggest function return `(stdout, rc)`. After calling this function, check `rc` first.

show/main.py
```diff
@@ -1383,8 +1388,20 @@ def runningconfiguration():
...
+
+    try:
+        output = json.loads(stdout)
+    except JSONDecodeError as e:
+        click.echo("Failed to load output '{}':{}".format(cmd, e))
+        raise click.Abort()
+
+    if 'bgpraw' in output or not multi_asic.is_multi_asic():
+        bgpraw_cmd = [constants.RVTYSH_COMMAND, '-c', 'show running-config']
+        bgpraw = get_cmd_output(bgpraw_cmd)
```
qiluo-msft, line 1402: >bgpraw [](http://example.com/codeflow?start=8&length=6)

return `(stdout, rc)`. After calling this function, check `rc` first. If rc non zero, output['bgpraw'] should be empty string.

show/main.py
wen587: @qiluo-msft  If we follow the original behavior, we don't need to check the rc. As it will be empty string by default if rc is non-zero.
```
def run_command(command, display_cmd=False, return_cmd=False):
    if display_cmd:
        click.echo(click.style("Command: ", fg='cyan') + click.style(command, fg='green'))

    # No conversion needed for intfutil commands as it already displays
    # both SONiC interface name and alias name for all interfaces.
    if clicommon.get_interface_naming_mode() == "alias" and not command.startswith("intfutil"):
        clicommon.run_command_in_alias_mode(command)
        raise sys.exit(0)

    proc = subprocess.Popen(command, shell=True, text=True, stdout=subprocess.PIPE)

    while True:
        if return_cmd:
            output = proc.communicate()[0]                   <===============
            return output
```

show/main.py
wen587: Same as above, it will be empty string by default if rc is non-zero.

show/main.py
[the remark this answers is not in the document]
qiluo-msft: The decision on whether override ConfigDB or not is not determined yet. Let's keep it simple as
```
    if not multi_asic.is_multi_asic():

```

qiluo-msft (approved):

ganglyu (approved):

isabelmsft (approved):

yxieca:
This change causes build failure in 202205 branch. @wen587 will create separate PR for 202205 branch.

yxieca:
@wen587 #2609 caused test failure when advancing submodule head. I reverted it for now.

liuh-80:
Cherry-pick PR merged: https://github.com/sonic-net/sonic-utilities/pull/2639

stephenxs:
Hi @wen587 @qiluo-msft 
I would like to understand the expected behavior if BGP docker is not running?

wen587:
Hi @stephenxs , it does print the rest config in this PR. I saw it is another PR break this behavior: https://github.com/sonic-net/sonic-utilities/pull/3148
It is calling run_bgp_show_command which will fail if bgp down.

stephenxs:
> Hi @stephenxs , it does print the rest config in this PR. I saw it is another PR break this behavior: #3148 It is calling run_bgp_show_command which will fail if bgp down.

Thank you @wen587 for clarifying it. I'll check PR #3148

PR #2538, opened 2022-12-05
Add CLI command for ECMP calculator

liorghub:
#### What I did
I have added a new CLI command for calling ECMP calculator.

#### How I did it
Add new CLI command. Command handler calls script /usr/bin/ecmp_calc.py

#### How to verify it
Call command and verify its output.
```
root@sonic:/home/admin# show ip ecmp-egress-port --ingress-port Ethernet64 --packet /tmp/packet.json
Egress port: Ethernet0
root@sonic:/home/admin# 
```

### DIFF
--- a/show/main.py
+++ b/show/main.py
@@ -66,6 +66,8 @@ from . import syslog
 PLATFORM_JSON = 'platform.json'
 HWSKU_JSON = 'hwsku.json'
 PORT_STR = "Ethernet"
+ECMP_CALC = '/usr/bin/ecmp_calc.py'
+CONTAINER_NAME = 'syncd'
 
 VLAN_SUB_INTERFACE_SEPARATOR = '.'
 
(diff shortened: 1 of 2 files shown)


liat-grozovik:
@prsunny could you please review or suggest someone who can?

liat-grozovik (changes requested):
missing tests for new code. please add
missing changes in command reference guide. please add

prsunny:
This is a vendor specific utility. I thought we can directly use the utility to get the hash result. I don't think we need to expose it via a generic CLI command.

liat-grozovik:
> 

In this case, the utility is already available and readme file explain directly how it can be used.
I was under the impression you would like to have the CLI and if the utility available it will be running.
But if this is not required, we can close this PR and leave with the utility itself.

PR #2545, opened 2022-12-08
[GCU] Prohibit removal of PFC_WD POLL_INTERVAL field

isabelmsft:
#### What I did
-Add new infrastructure to GCU for validation of moves that YANG models are unable to capture
-Prohibit deletion of PFC_WD POLL_INTERVAL field

#### How I did it
- Ensure JSON Diff does not have POLL_INTERVAL deletion
- If diff reflects request for POLL_INTERVAL deletion, raise an Exception and do not allow patch application to go through

#### How to verify it
Attempt to delete POLL_INTERVAL field using GCU

Please note that this is different than the create-only extension, which is described here: https://github.com/sonic-net/SONiC/blob/master/doc/config-generic-update-rollback/Json_Patch_Ordering_using_YANG_Models_Design.md#313-using-yang-models-to-host-flags-for-move-validation
create-only is programmed as a MoveValidator in GCU, but that is different than the use case necessary for this PR. We don't want to validate each intermediate move to get to the final state, as is done by MoveValidators- we just need to validate the initial request. Additionally, MoveValidators continue searching for a different move that is 'valid' if it finds one invalid - possibly by deleting the parent and then readding other children. This is not the behavior we desire. We only need a check of the initial request (reflected in Json diff).  

```
admin@vlab-01:~$ sudo config apply-patch patch.json
Patch Applier: Patch: [{"op": "remove", "path": "/PFC_WD/GLOBAL/POLL_INTERVAL"}]
Patch Applier: Validating all JsonPatch operations are permitted on the specified fields
Patch Applier: The patch was sorted into 1 change:
Patch Applier:   * [{"op": "remove", "path": "/PFC_WD/GLOBAL/POLL_INTERVAL"}]
Patch Applier: Applying 1 change in order:
Patch Applier:   * [{"op": "remove", "path": "/PFC_WD/GLOBAL/POLL_INTERVAL"}]
Patch Applier: Verifying patch updates are reflected on ConfigDB.
Patch Applier: Patch application completed.
Patch applied successfully.
```
```
admin@vlab-01:~$ sudo config apply-patch patch.json
Patch Applier: Patch: [{"op": "remove", "path": "/PFC_WD/GLOBAL/POLL_INTERVAL"}]
Patch Applier: Validating all JsonPatch operations are permitted on the specified fields
Failed to apply patch

Error: Given patch operation is invalid. Operation: remove is illegal on field: /PFC_WD/GLOBAL/POLL_INTERVAL
```

### DIFF
--- a/generic_config_updater/gu_common.py
+++ b/generic_config_updater/gu_common.py
@@ -149,7 +149,7 @@ class ConfigWrapper:
         # illegal_operations_to_fields_map['remove'] yields a list of fields for which `remove` is an illegal operation 
         illegal_operations_to_fields_map = {'add':[],
                                             'replace': [],
-                                            'remove': ['/PFC_WD/GLOBAL/POLL_INTERVAL']}
+                                            'remove': ['/PFC_WD/GLOBAL/POLL_INTERVAL', '/PFC_WD/GLOBAL']}
         for operation, field_list in illegal_operations_to_fields_map.items():
             for field in field_list:
                 if any(op['op'] == operation and field in op['path'] for op in patch):
--- a/generic_config_updater/gu_common.py
+++ b/generic_config_updater/gu_common.py
@@ -152,7 +152,7 @@ class ConfigWrapper:
                                             'remove': ['/PFC_WD/GLOBAL/POLL_INTERVAL', '/PFC_WD/GLOBAL']}
         for operation, field_list in illegal_operations_to_fields_map.items():
             for field in field_list:
-                if any(op['op'] == operation and field in op['path'] for op in patch):
+                if any(op['op'] == operation and field == op['path'] for op in patch):
                     raise IllegalPatchOperationError("Given patch operation is invalid. Operation: {} is illegal on field: {}".format(operation, field))
 
     def validate_lanes(self, config_db):
--- a/generic_config_updater/gu_common.py
+++ b/generic_config_updater/gu_common.py
@@ -16,6 +16,9 @@ SYSLOG_IDENTIFIER = "GenericConfigUpdater"
 class GenericConfigUpdaterError(Exception):
     pass
 
+class IllegalPatchOperationError(ValueError):
+    pass
+
 class EmptyTableError(ValueError):
     pass
 
@@ -135,6 +138,22 @@ class ConfigWrapper:
             return False, ex
 
         return True, None
+
+    def validate_field_operation(self, old_config, target_config):
+        """
+        Some fields in ConfigDB are restricted and may not allow third-party addition, replacement, or removal. 
+        Because YANG only validates state and not transitions, this method helps to JsonPatch operations/transitions for the specified fields. 
+        """
+        patch = jsonpatch.JsonPatch.from_diff(old_config, target_config)
+        
+        # illegal_operations_to_fields_map['remove'] yields a list of fields for which `remove` is an illegal operation 
+        illegal_operations_to_fields_map = {'add':[],
+                                            'replace': [],
+                                            'remove': ['/PFC_WD/GLOBAL/POLL_INTERVAL']}
+        for operation, field_list in illegal_operations_to_fields_map.items():
+            for field in field_list:
+                if any(op['op'] == operation and field in op['path'] for op in patch):
+                    raise IllegalPatchOperationError("Given patch operation is invalid. Operation: {} is illegal on field: {}".format(operation, field))
 
     def validate_lanes(self, config_db):
         if "PORT" not in config_db:
--- a/generic_config_updater/generic_updater.py
+++ b/generic_config_updater/generic_updater.py
@@ -47,6 +47,10 @@ class PatchApplier:
         # Generate target config
         self.logger.log_notice("Simulating the target full config after applying the patch.")
         target_config = self.patch_wrapper.simulate_patch(patch, old_config)
+        
+        # Validate all JsonPatch operations on specified fields
+        self.logger.log_notice("Validating all JsonPatch operations are permitted on the specified fields")
+        self.config_wrapper.validate_field_operation(old_config, target_config)
 
         # Validate target config does not have empty tables since they do not show up in ConfigDb
         self.logger.log_notice("Validating target config does not have empty tables, " \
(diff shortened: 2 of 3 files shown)


generic_config_updater/gu_common.py
```diff
@@ -136,6 +139,22 @@ def validate_config_db_config(self, config_db_as_json):
...
+        """
+        Some fields in ConfigDB are restricted and may not allow third-party addition, replacement, or removal. 
+        Because YANG only validates state and not transitions, this method helps to JsonPatch operations/transitions for the specified fields. 
+        """
+        patch = jsonpatch.JsonPatch.from_diff(old_config, target_config)
+        
+        # illegal_operations_to_fields_map['remove'] yields a list of fields for which `remove` is an illegal operation 
+        illegal_operations_to_fields_map = {'add':[],
+                                            'replace': [],
+                                            'remove': ['/PFC_WD/GLOBAL/POLL_INTERVAL']}
```
wen587, line 152: I saw in [sonic-pfcwd.yang](https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-pfcwd.yang#L75) that POLL_INTERVAL will only be valid if its table's key is GLOBAL.

Is it guaranteed that GLOBAL will always have the POLL_INTERVAL field?
In that case, we can add a statement in YANG file if we want to avoid deletion of the POLL_INTERVAL.

isabelmsft: @neethajohn , can you help confirm if there will ever be a situation where GLOBAL will be present without POLL_INTERVAL field?

@wen587 If this approach is possible, then we must also prohibit removal of GLOBAL. So I think we also need to consider if PFC_WD must always have GLOBAL.

isabelmsft: @bingwang-ms , Could you help address the question of " if there will ever be a situation where GLOBAL will be present without POLL_INTERVAL field?"

bingwang-ms: > @bingwang-ms , Could you help address the question of " if there will ever be a situation where GLOBAL will be present without POLL_INTERVAL field?"

Yes, I think so. In the `GLOBAL` table, we can also have another entry named `BIG_RED_SWITCH`. If PFC watchdog is not enabled, the `POLL_INTERVAL` can be absent, while the `BIG_RED_SWITCH` can be present. So the `GLOBAL` table is present without `POLL_INTERVAL`.

isabelmsft: Thank you @bingwang-ms .

In this case, @wen587 , does implementation as-is look alright to you?

wen587: Hi @bingwang-ms, does that mean `POLL_INTERVAL` could be deleted if PFC watchdog state changes from enable to disable?
Hi @isabelmsft , if Bing confirmed it cannot be deleted. I think it is good to go.

generic_config_updater/gu_common.py
```diff
@@ -136,6 +139,22 @@ def validate_config_db_config(self, config_db_as_json):
...
+        """
+        patch = jsonpatch.JsonPatch.from_diff(old_config, target_config)
+        
+        # illegal_operations_to_fields_map['remove'] yields a list of fields for which `remove` is an illegal operation 
+        illegal_operations_to_fields_map = {'add':[],
+                                            'replace': [],
+                                            'remove': ['/PFC_WD/GLOBAL/POLL_INTERVAL', '/PFC_WD/GLOBAL']}
+        for operation, field_list in illegal_operations_to_fields_map.items():
+            for field in field_list:
+                if any(op['op'] == operation and field in op['path'] for op in patch):
```
wen587, line 155: Hi @isabelmsft , I think we should use an exact match for field comparison: `field == op['path']`
In current pfcwd yang model, the GLOBAL table only contians one field: POLL_INTERVAL. But I notice it could also contains other fields. https://github.com/sonic-net/sonic-utilities/blob/master/tests/mock_tables/asic0/config_db.json#L119

If we use field **in** op['path'], it will block all other fields change(such as BIG_RED_SWITCH).

isabelmsft: thank you for the suggestion, changed to exact match

generic_config_updater/gu_common.py
bingwang-ms: Yes, I confirm that.

wen587 (approved):

vadymhlushko-mlnx:
Hi @isabelmsft, because of this PR, in the [submodule update PR](https://github.com/sonic-net/sonic-buildimage/pull/13482) the Azure pipelines test failing - [failure URL](https://www.testbed-tools.org/scheduler/testplan/63cec453e429c1efea2d6bb1)

Could you please take a look?

isabelmsft:
Hi Vadym,

Yes, I'm taking a  look now.

Thank you,
Isabel

isabelmsft:
> Hi @isabelmsft, because of this PR, in the [submodule update PR](https://github.com/sonic-net/sonic-buildimage/pull/13482) the Azure pipelines test failing - [failure URL](https://www.testbed-tools.org/scheduler/testplan/63cec453e429c1efea2d6bb1)
> 
> Could you please take a look?

Hi @vadymhlushko-mlnx , the submodule update PR test should be failing due to another PR, and Jingwen helped address the other issue in this PR https://github.com/sonic-net/sonic-utilities/pull/2623

vadymhlushko-mlnx:
> > Hi @isabelmsft, because of this PR, in the [submodule update PR](https://github.com/sonic-net/sonic-buildimage/pull/13482) the Azure pipelines test failing - [failure URL](https://www.testbed-tools.org/scheduler/testplan/63cec453e429c1efea2d6bb1)
> > Could you please take a look?
> 
> Hi @vadymhlushko-mlnx , the submodule update PR test should be failing due to another PR, and Jingwen helped address the other issue in this PR #2623

It seems like even after [PR #2623](https://github.com/sonic-net/sonic-utilities/pull/2623) the issue is still here.
Could you please take a look at [submodule update PR #13508](https://github.com/sonic-net/sonic-buildimage/pull/13508)

yxieca:
@isabelmsft please create separate PR for 202205 branch.

isabelmsft:
> @isabelmsft please create separate PR for 202205 branch.
@yxieca created separate PR for 202205:
https://github.com/sonic-net/sonic-utilities/pull/2776

PR #2548, opened 2022-12-09
[route_check]: Ignore ASIC only SOC IPs

theasianpianist:
#### What I did
For active-active devices, it's expected that tunnel routes for SOC IPs are programmed only to the ASIC and not to the kernel. This allows gRPC traffic (which uses kernel routing) to continue using the direct link to communicate with the SOC, but external traffic destined for the SOC IP (which is routed by the ASIC) will use the tunnel route if present to maintain connectivity.

#### How I did it
- Remove any SOC IP routes from the list of ASIC only routes when checking
- Restructure route_check test cases for better granularity
- Add test case to cover SOC IPs in ASIC only

#### How to verify it
Run `route_check_test.py` and verify that it passes


scripts/route_check.py
```diff
@@ -489,6 +493,45 @@ def filter_out_standalone_tunnel_routes(routes):
...
+    For active-active cables, we want the tunnel route for SOC IPs
+    to only be programmed to the ASIC and not to the kernel. This is to allow
+    gRPC connections coming from ycabled to always use the direct link (since this
+    will use the kernel routing table), but still provide connectivity to any external
+    traffic in case of a link issue (since this traffic will be forwarded by the ASIC).
+    """
+    config_db = swsscommon.ConfigDBConnector()
+    config_db.connect()
+
+    if not is_dualtor(config_db):
```
yxieca, line 519: Is it a concern that this skip is applied to both active-active and active-standby?

zjswhhh: `active-standby` wouldn't have a `soc_ipv4` entry?

theasianpianist: I can also change `get_soc_ips` to make sure the cable type is active-active

yxieca (approved):

zjswhhh (dismissed):
lgtm

tests/mock_tables/config_db.json
```diff
@@ -831,8 +831,7 @@
         "mac": "1d:34:db:16:a6:00",
         "platform": "x86_64-mlnx_msn3800-r0",
         "peer_switch": "sonic-switch",
-        "type": "ToRRouter",
-        "subtype": "DualToR"
```
prsunny, line 835: Why remove this? We need to test for DualToR, right?

theasianpianist: I updated the test so that we can mock this dynamically per test case, it's configured for the dual ToR cases individually: https://github.com/theasianpianist/sonic-utilities/blob/de5c0332defd6585f9096e0970ef145bc706204c/tests/route_check_test_data.py#L288
https://github.com/theasianpianist/sonic-utilities/blob/de5c0332defd6585f9096e0970ef145bc706204c/tests/route_check_test_data.py#L336

prsunny (approved):

lolyu (approved):

PR #2560, opened 2022-12-15
[db_migrator] Fix migration of Loopback data: handle all Loopback interfaces

vaibhavhd:
#### What I did
Fix the issue where cross branch upgrades (base DB version 1_0_1) lead to a OA crash due to a duplicate IP2ME route being added when there are more than one Loopback interfaces.

The issue happens as in current implementation `lo` is hardcoded to be replaced as `Loopback0`.
When the base image's APP DB has more than one IP assigned to `lo` interface, upon migration, all the IPs are assinged to same loopback `Loopback0`. This is incorrect, as in newer images different IPs are assinged to distinct Loopback interfaces.

Eg. If the base image APP DB contains:
```
{
    "INTF_TABLE:lo:10.1.0.32/32" : {
        "scope": "global",
        "family": "IPv4"
    },
    "INTF_TABLE:lo:10.20.8.199/32" : {
        "scope": "global",
        "family": "IPv4"
    }
}
```

In current implementation, this will be INCORRECTLY migrated as:
```
{
    "INTF_TABLE:Loopback0" : {"NULL": "NULL"},
    "INTF_TABLE:Loopback0:10.1.0.32/32" : {
        "scope": "global",
        "family": "IPv4"
    },
    "INTF_TABLE:Loopback0:10.20.8.199/32" : {
        "scope": "global",
        "family": "IPv4"
    }
}
```

The expected migration would look like:
```
{
    "INTF_TABLE:Loopback0" : {"NULL": "NULL"},
    "INTF_TABLE:Loopback0:10.1.0.32/32" : {
        "scope": "global",
        "family": "IPv4"
    },
    "INTF_TABLE:Loopback1" : {"NULL": "NULL"},
    "INTF_TABLE:Loopback1:10.20.8.199/32" : {
        "scope": "global",
        "family": "IPv4"
    },
}
```

#### How I did it

Use source-of-truth data from CONFIG_DB to assign IPs (v4 and v6) to different Loopback interfaces, and add new entries into APPL_DB.
New entries for new loopback interfaces are created accordingly, and the field/values are also copied into new entries.

#### How to verify it
Verified on a physical testbed that this change fixes the OA crash issue.
Also added a unit test to catch this issue in PR tests.

New log from migration:
```
Dec 15 00:34:30.011565 str-7260cx3-acs-1 NOTICE db_migrator: Set table INTF_TABLE:Loopback0:FC00:1::32/128 with field family and value IPv6
Dec 15 00:34:30.011773 str-7260cx3-acs-1 NOTICE db_migrator: Set table INTF_TABLE:Loopback0:FC00:1::32/128 with field scope and value global

Dec 15 00:34:30.011880 str-7260cx3-acs-1 NOTICE db_migrator: Set table INTF_TABLE:Loopback1:2001:506:28:500::1/128 with field family and value IPv6
Dec 15 00:34:30.011991 str-7260cx3-acs-1 NOTICE db_migrator: Set table INTF_TABLE:Loopback1:2001:506:28:500::1/128 with field scope and value global

Dec 15 00:34:30.012089 str-7260cx3-acs-1 NOTICE db_migrator: Set table INTF_TABLE:Loopback0:10.1.0.32/32 with field family and value IPv4
Dec 15 00:34:30.012177 str-7260cx3-acs-1 NOTICE db_migrator: Set table INTF_TABLE:Loopback0:10.1.0.32/32 with field scope and value global

Dec 15 00:34:30.012280 str-7260cx3-acs-1 NOTICE db_migrator: Set table INTF_TABLE:Loopback1:10.20.8.199/32 with field family and value IPv4
Dec 15 00:34:30.012385 str-7260cx3-acs-1 NOTICE db_migrator: Set table INTF_TABLE:Loopback1:10.20.8.199/32 with field scope and value global
```

### DIFF
--- a/scripts/db_migrator.py
+++ b/scripts/db_migrator.py
@@ -178,24 +178,33 @@ class DBMigrator():
         if self.appDB is None:
             return
 
-        data = self.appDB.keys(self.appDB.APPL_DB, "INTF_TABLE:*")
+        # Get Lo interface corresponding to IP(v4/v6) address from CONFIG_DB.
+        configdb_data = self.configDB.get_keys('LOOPBACK_INTERFACE')
+        lo_addr_to_int = dict()
+        for int_data in configdb_data:
+            if type(int_data) == tuple and len(int_data) > 1:
+                intf_name = int_data[0]
+                intf_addr = int_data[1]
+                lo_addr_to_int.update({intf_addr: intf_name})
 
-        if data is None:
+        lo_data = self.appDB.keys(self.appDB.APPL_DB, "INTF_TABLE:lo*")
+        if lo_data is None:
             return
+        for lo_row in lo_data:
+            # Example of lo_row: 'INTF_TABLE:lo:10.1.0.32/32'
+            # Example of lo_dict: {'family': 'IPv4', 'scope': 'global'}
+            lo_dict = self.appDB.get_all(self.appDB.APPL_DB, lo_row)
+            lo_addr = lo_row.split('INTF_TABLE:lo:')[1]
+            lo_name = lo_addr_to_int.get(lo_addr)
+            lo_table = "INTF_TABLE:" + lo_name + ":" + lo_addr
+            for key, value in lo_dict.items():
+                log.log_notice("Set table {} with field {} and value {}".format(
+                    lo_table, key, value))
+                self.appDB.set(self.appDB.APPL_DB, lo_table, key, value)
+            # delete the old row with name as 'lo' as new row has been added
+            self.appDB.delete(self.appDB.APPL_DB, lo_row)
 
         if_db = []
-        for key in data:
-            if_name = key.split(":")[1]
-            if if_name == "lo":
-                self.appDB.delete(self.appDB.APPL_DB, key)
-                key = key.replace(if_name, "Loopback0")
-                log.log_info('Migrating lo entry to ' + key)
-                self.appDB.set(self.appDB.APPL_DB, key, 'NULL', 'NULL')
-
-            if '/' not in key:
-                if_db.append(key.split(":")[1])
-                continue
-
         data = self.appDB.keys(self.appDB.APPL_DB, "INTF_TABLE:*")
         for key in data:
             if_name = key.split(":")[1]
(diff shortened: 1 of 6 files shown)


yxieca (approved):

scripts/db_migrator.py
```diff
@@ -178,24 +178,33 @@ def migrate_intf_table(self):
...
+
+        lo_data = self.appDB.keys(self.appDB.APPL_DB, "INTF_TABLE:lo*")
+        if lo_data is None:
             return
+        for lo_row in lo_data:
+            # Example of lo_row: 'INTF_TABLE:lo:10.1.0.32/32'
+            # Example of lo_dict: {'family': 'IPv4', 'scope': 'global'}
+            lo_dict = self.appDB.get_all(self.appDB.APPL_DB, lo_row)
+            lo_addr = lo_row.split('INTF_TABLE:lo:')[1]
+            lo_name = lo_addr_to_int.get(lo_addr)
```
prsunny, line 198: A logical difference from previous implementation is that, if config_db data is empty, previously we hardcode to 'Loopback0' whereas here it will come as empty which can cause swss crash. For safe side, I suggest if 'lo_name' is empty use the previous logic.

scripts/db_migrator.py
```diff
@@ -178,24 +178,33 @@ def migrate_intf_table(self):
...
+            lo_table = "INTF_TABLE:" + lo_name + ":" + lo_addr
+            for key, value in lo_dict.items():
+                log.log_notice("Set table {} with field {} and value {}".format(
+                    lo_table, key, value))
+                self.appDB.set(self.appDB.APPL_DB, lo_table, key, value)
+            # delete the old row with name as 'lo' as new row has been added
+            self.appDB.delete(self.appDB.APPL_DB, lo_row)
```
prsunny, line 205: You are deleting the iterating element in the loop. Not sure if it can cause issues. Why don't we use the same logic as before and just replace the `if_name `from your new dict - `lo_addr_to_int`?

scripts/db_migrator.py
vaibhavhd: Sure, addressed this in latest.

scripts/db_migrator.py
vaibhavhd: Updated this in latest - removed the inner loop (handling of field, value). In the updated logic, only name `lo` is replaced with name from config db.

prsunny (approved):

PR #2569, opened 2022-12-27
[Mellanox] Change severity to NOTICE in Mellanox buffer migrator when unable to fetch DEVICE_METADATA due to empty CONFIG_DB during initialization

stephenxs:
#### What I did
It is expected that db_migrator is not able to fetch DEVICE_METADATA when it is invoked before the CONFIG_DB is initialized. In this case, we should not use ERROR to log the message since it's not an error.
Change the severity to NOTICE

#### How I did it

Change the severity.

#### How to verify it

Manually test.

### DIFF
--- a/scripts/mellanox_buffer_migrator.py
+++ b/scripts/mellanox_buffer_migrator.py
@@ -101,7 +101,7 @@ class MellanoxBufferMigrator():
             self.sku = device_data.get('hwsku')
             self.ready = True
         if not self.platform or not self.sku:
-            log.log_error("Trying to get DEVICE_METADATA from DB but doesn't exist, skip migration")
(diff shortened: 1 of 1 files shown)


Junchao-Mellanox (approved):

liat-grozovik (commented):
but in this case migration is not fully skipped. we assume that it will be done later?

stephenxs:
> but in this case migration is not fully skipped. we assume that it will be done later?

for the first db_migrator is called when the config_db is empty, it is skipped. for the second time db_migrator is called, it will not and will run the migrating logic. from the log we also see that (2nd time db_migrator run)

liat-grozovik (approved):

PR #2577, opened 2023-01-03
Extend fast-reboot STATE_DB entry timer

arfeigin:
Due to an issue of fallback from to cold-boot when using upgrade with fast-reboot combined with FW upgrade a short term solution is to extend the timer.
An issue will be opened on this case with the full details, link will be added to the PR.
Long term solution of using fast-reboot finalizer replacing the timer is in work.

#### What I did
Extend the timer of STATE DB fast-reboot entry to 210 seconds.

#### How I did it
Modify the TTL of the entry set in STATE DB.

#### How to verify it
Verify the entry was preserved for 210 seconds.

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -532,7 +532,7 @@ case "$REBOOT_TYPE" in
         check_warm_restart_in_progress
         BOOT_TYPE_ARG=$REBOOT_TYPE
         trap clear_boot EXIT HUP INT QUIT TERM KILL ABRT ALRM
-        sonic-db-cli STATE_DB SET "FAST_REBOOT|system" "1" "EX" "180" &>/dev/null
+        sonic-db-cli STATE_DB SET "FAST_REBOOT|system" "1" "EX" "210" &>/dev/null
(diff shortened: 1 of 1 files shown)


dprital (approved):

liat-grozovik:
@vaibhavhd could you please review this one? it should be a mitigation plan until the next improvement on fastboot will get in.

dprital:
@vaibhavhd , as discussed, can you please review and add your comments to this PR ? (Thanks in advance)

vaibhavhd:
Please revert this when long-term fix is added.

For long-term solution:
Since fast-reboot is running on warmboot infra now, I think we don't need two different finalizers.

In that sense, can you 1) do away with the timer for fast-reboot flag. 2) Address this in warmreboot-finalizer? 
https://github.com/sonic-net/sonic-buildimage/blob/master/files/image_config/warmboot-finalizer/finalize-warmboot.sh

vaibhavhd (approved):

prsunny (approved):

PR #2578, opened 2023-01-04
[202205] Extend fast-reboot STATE_DB entry timer

arfeigin:
This is a PR against 202205 based on the following PR: https://github.com/sonic-net/sonic-utilities/pull/2577
Due to an issue of fallback from to cold-boot when using upgrade with fast-reboot combined with FW upgrade a short term solution is to extend the timer.
An issue will be opened on this case with the full details, link will be added to the PR.
Long term solution of using fast-reboot finalizer replacing the timer is in work.

#### What I did
Extend the timer of STATE DB fast-reboot entry to 210 seconds.

#### How I did it
Modify the TTL of the entry set in STATE DB.

#### How to verify it
Verify the entry was preserved for 210 seconds.


dprital (approved):

liat-grozovik:
@prsunny , @vaibhavhd could you please review?

PR #2595, opened 2023-01-13
Revert "sonic-utilities: Update config reload() to verify formatting of an input file"

ganglyu:
#### What I did
There are use cases like config reload /dev/stdin, for example [L2 Switch mode · sonic-net/SONiC Wiki (github.com)](https://github.com/sonic-net/SONiC/wiki/L2-Switch-mode). The original PR would read input file twice, so /dev/stdin does not work.

#### How I did it
Reverts https://github.com/sonic-net/sonic-utilities/pull/2529

#### How to verify it
Run unit test


StormLiangMS (approved):
LGTM

PR #2600, opened 2023-01-16
Remove TODO comment which is no longer relevant

liorghub:
#### What I did
I removed TODO comment which is no longer relevant.

TODO can be removed as there is no need to delete FDB table.
Recently, fast boot was changed to use this function and in fast boot we don’t want to clean FDB. 
In warmboot we do need to clean FDB, however, it is being performed in a later stage when FDB flush event occurs.

#### How I did it
Remove comment.

#### How to verify it
Remove comment from python module, no functional effect.

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -727,8 +727,6 @@ for service in ${SERVICES_TO_STOP}; do
             sonic-db-cli FLEX_COUNTER_DB FLUSHDB > /dev/null
         fi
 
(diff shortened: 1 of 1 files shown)


dprital (approved):

liat-grozovik (approved):

liat-grozovik:
@vaibhavhd @yxieca  FYI

yxieca (approved):

PR #2610, opened 2023-01-20
[202106][show] Add bgpraw to show run all

wen587:
https://github.com/sonic-net/sonic-utilities/pull/2537
The only difference between this PR and original PR is just one line of import.
`from sonic_py_common import multi_asic`


wen587:
low pri on this branch supprot

PR #2619, opened 2023-01-21
[GCU] Add PFC_WD RDMA validator

isabelmsft:
#### What I did
Add PFC_WD RDMA validator to ensure GCU is being used on allowable platform and version for PFC_WD.
The allowable platforms and versions are determined by the last two rows of this table:
image

#### How to verify it
Attempt to use GCU on vs and modify PFC_WD table, unit tests
GCU patch application improperly goes through
```
Failed to apply patch

Error: Modification of /PFC_WD table is illegal due to corresponding validating function rdma_config_update_validator returning false
```


generic_config_updater/generic_config_updater.conf.json
```diff
@@ -58,19 +73,19 @@
             "validate_commands": [ ]
         },
         "rsyslog": {
-            "validate_commands": [ "generic_config_updater.services_validator.rsyslog_validator" ]
```
qiluo-msft, line 61: `validate_commands` is used for post health check after ConfigDB modification. What you want to achieve is to check the modification before it is applied to ConfigDB.

generic_config_updater/generic_config_updater.conf.json
```diff
@@ -48,7 +48,22 @@
...
+            "table_modification_validators": [ "generic_config_updater.validators.is_mellanox_device_validator" ]
+	},
+	"WRED_PROFILE": {
+            "table_modification_validators": [ "generic_config_updater.validators.is_mellanox_device_validator" ]
+	},
+	"QUEUE": {
+            "table_modification_validators": [ "generic_config_updater.validators.is_mellanox_device_validator" ]
+	},
+	"BUFFER_PROFILE": {
+            "table_modification_validators": [ "generic_config_updater.validators.is_mellanox_device_validator" ]
```
qiluo-msft, line 65: Offline discussed, this is not flexible to express all the rules: ASIC vendor, ASIC generation, and sonic branches.

generic_config_updater/gu_common.py
```diff
@@ -155,6 +159,32 @@ def validate_field_operation(self, old_config, target_config):
...
+
+        if os.path.exists(UPDATER_CONF_FILE):
+            with open(UPDATER_CONF_FILE, "r") as s:
+                updater_conf = json.load(s)
+        else:
+            raise GenericConfigUpdaterError("GCU Config file not found") 
+        
+        for element in patch:
+            path = element["path"]
+            table = re.search(r'\/([^\/]+)(\/|$)', path).group(1)
```
wen587, line 178: What does it match? Could you add an commented example?

generic_config_updater/generic_config_updater.conf.json
isabelmsft: I created a new conf file specific to validators that need to be run before the change is applied to ConfigDB

generic_config_updater/generic_config_updater.conf.json
isabelmsft: Previously, the design only allowed for "AND" of all validators
I added a new section to allow for "OR" check of validators
Extending the "OR" section to be a list of lists would allow for (X AND Y AND Z) OR (S AND T) checks as well for even greater flexibility

generic_config_updater/gu_common.py
isabelmsft: This matches the table name "PFC_WD" and removes any forward slashes at the beginning or end. I added a comment to clarify as well

generic_config_updater/gcu_table_modification_validators.conf.json
```diff
@@ -0,0 +1,39 @@
+{
+    "README": [
+        "table_modification_validators provides, module & method name as ",
+        "    <module name>.<method name>",
+        "NOTE: module name could have '.'",
+        " ",
```
wen587, line 6: Do we need testcase for this file? Like tests in `service_validator_test.py`?

generic_config_updater/gcu_table_modification_validators.conf.json
```diff
@@ -0,0 +1,39 @@
...
+        "method name",
+        "",
+        "e.g. 'show.acl.test_acl'",
+        "",
+	"table_modification_validators_and for a given table defines a list of validators that all must pass for modification to the table to be allowed",
+	"table_modification_validators_or for a given table defines a list of validators, at least one of which must pass for modification of the table to be allowed",
+        ""
+    ],
+    "tables": {
+	"PFC_WD": {
```
qiluo-msft, line 17: could you reformat? mixing tabs and spaces now.

generic_config_updater/generic_config_updater.conf.json
qiluo-msft: Could you keep original name `services_validator`? there is nothing wrong with old code.

isabelmsft: I added the new validator `is_mellanox_device_validator` to the originally named `services_validator.py` file, so I changed to `validator.py` 

If `is_mellanox_device_validator` would be better placed somewhere else, I will move it and leave the original file name `services_validator.py`

generic_config_updater/gcu_table_modification_validators.conf.json
```diff
@@ -0,0 +1,39 @@
...
+            "table_modification_validators_and": [ "generic_config_updater.validators.is_mellanox_device_validator" ],
```
qiluo-msft, line 18: >table_modification_validators_and [](http://example.com/codeflow?start=13&length=33)

There is no design doc for these new field `table_modification_validators_and` `table_modification_validators_or`. It is easy to understand if either of them has value, but what is the semantics if both of them appear?

generic_config_updater/generic_config_updater.conf.json
qiluo-msft: You can place new validator in `generic_config_updater.move_validator`, and still keep old code.

isabelmsft: This new validator is not a MoveValidator, which validates every intermediary move. The preexisting MoveValidators come in during patch sorting.  For MoveValidators, if validation fails, the PatchSorter algorithm will continue searching for a new valid move. For this use case, we don't need to start patch sorting at all.

Please see a comment I made on a previous PR https://github.com/sonic-net/sonic-utilities/pull/2545 describing why we are adding this new infra for one-shot validation at the start (as opposed to MoveValidators for every intermediary move).

We don't want to validate each intermediate move to get to the final state, as is done by MoveValidators- we just need to validate the initial request. MoveValidators continue searching for a different move that is 'valid' if it finds one invalid - possibly by deleting the parent and then readding other children. This is not the behavior we desire. We only need a check of the initial request (reflected in Json diff).

generic_config_updater/gcu_table_modification_validators.conf.json
isabelmsft: I added additional unit testing for the new RDMA-specific validator as suggested offline

generic_config_updater/gcu_table_modification_validators.conf.json
isabelmsft: I changed back to just `table_modification_validators` 
So now, all listed validators must pass in order for modification to the table to be allowed. This is the same logic as MoveValidators, but MoveValidators checks on every intermediary move during patch sorting.

generic_config_updater/gcu_table_modification_validators.conf.json
isabelmsft: Changed to all spaces

generic_config_updater/gcu_table_modification_validators.conf.json
```diff
@@ -0,0 +1,21 @@
...
+        "method name",
+        "",
+        "e.g. 'show.acl.test_acl'",
+        "",
+        "table_modification_validators for a given table defines a list of validators that all must pass for modification to the table to be allowed",
+        ""
+    ],
+    "tables": {
+        "PFC_WD": {
+            "table_modification_validators": [ "generic_config_updater.validators.rdma_config_update_validator" ]
```
qiluo-msft, line 17: >generic_config_updater.validators.rdma_config_update_validator [](http://example.com/codeflow?start=48&length=62)

generic_config_updater.field_operation_validator.rdma_validator

generic_config_updater/generic_config_updater.conf.json
qiluo-msft: Could you keep original name services_validator? there is nothing wrong with old code.

generic_config_updater/gu_common.py
```diff
@@ -162,6 +166,32 @@ def validate_field_operation(self, old_config, target_config):
                 if any(op['op'] == operation and field == op['path'] for op in patch):
                     raise IllegalPatchOperationError("Given patch operation is invalid. Operation: {} is illegal on field: {}".format(operation, field))
 
+        def _invoke_validating_function(cmd):
+            # cmd is in the format as <package/module name>.<method name>
+            method_name = cmd.split(".")[-1]
+            module_name = ".".join(cmd.split(".")[0:-1])
+            module = importlib.import_module(module_name, package=None)
+            method_to_call = getattr(module, method_name)
+            return method_to_call()
```
qiluo-msft, line 175: >method_to_call [](http://example.com/codeflow?start=19&length=14)

There is potential risk to run ambiguous code. We can proactively check the module, and method_names. Maybe we can have allowlist or similar?

generic_config_updater/gu_common.py
```diff
@@ -162,6 +166,32 @@ def validate_field_operation(self, old_config, target_config):
...
+
+        if os.path.exists(GCU_TABLE_MOD_CONF_FILE):
+            with open(GCU_TABLE_MOD_CONF_FILE, "r") as s:
+                gcu_table_modification_conf = json.load(s)
+        else:
+            raise GenericConfigUpdaterError("GCU table modification validators config file not found") 
+
+        for element in patch:
+            path = element["path"]
+            table = re.search(r'\/([^\/]+)(\/|$)', path).group(1) # This matches the table name in the path, eg. PFC_WD without the forward slashes
```
qiluo-msft, line 185: ># This matches the table name in the path, eg. PFC_WD without the forward slashes [](http://example.com/codeflow?start=66&length=81)

Regex is hard to read. Would you provide more code comment, like sample match case.

qiluo-msft, line 185: >group [](http://example.com/codeflow?start=57&length=5)

Better be cautious about mismatch. Check matching result before getting group item.

generic_config_updater/field_operation_validators.py
```diff
@@ -0,0 +1,21 @@
...
+
+def rdma_config_update_validator():
+    version_info = device_info.get_sonic_version_info()
+    build_version = version_info.get('build_version')
+    asic_type = version_info.get('asic_type')
+
+    if (asic_type != 'mellanox' and asic_type != 'broadcom' and asic_type != 'cisco-8000'):
+        return False
+
+    if len(build_version) >= 6:
```
qiluo-msft, line 11: >6 [](http://example.com/codeflow?start=29&length=1)

Magic number. Let's make version comparison more robust and easy to read.

generic_config_updater/generic_config_updater.conf.json
isabelmsft: Yes, I changed it back to services_validator. I initially changed it away because I added a non-service validator to this file, but this new non-service validator is now in a new file `field_operation_validators.py`

generic_config_updater/gu_common.py
isabelmsft: Yes, I added code to ensure the module_name is `generic_config_updater.field_operation_validators` and the `method_name` contains the string `validator`

generic_config_updater/gu_common.py
isabelmsft: yes, I added a sample match case in the comment

isabelmsft: We should not be able to reach this part of the code if there is a mismatch.

When creating the jsonpatch before this part of the code is reached, the path must be a string that matches the defined regex. Else, an error will be thrown. 
https://github.com/sonic-net/sonic-utilities/blob/master/config/main.py#L1391

I did add extra code to check for mismatch again if that would be more complete.

generic_config_updater/field_operation_validators.py
isabelmsft: I updated the version comparison

generic_config_updater/field_operation_validators.py
```diff
@@ -0,0 +1,25 @@
...
+    for substring in version_substrings:
+        if substring.isdigit() and (substring.startswith("202") or substring.startswith("201")):
+            branch_int = int(substring)
+            break
+
+    if branch_int is None:
+        return False
+
+    if asic_type == 'cisco-8000':
+        return branch_int >= 202012
```
qiluo-msft, line 23: >branch_int [](http://example.com/codeflow?start=15&length=10)

You will get something like `20201231`. It is always larger.

qiluo-msft: yyyymmdd string is not int in concept. better to use string comparison, or parsed to datetime and compare.

isabelmsft: I changed to string comparison

qiluo-msft (approved):

wen587 (approved):
lgtm

yxieca:
@isabelmsft please create separate PR for 202205 branch.

isabelmsft:
> @isabelmsft please create separate PR for 202205 branch.

Created separate PR for 202205 https://github.com/sonic-net/sonic-utilities/pull/2781/files

yxieca:
The backport change was reverted due to extra unit test failing with the backport.

StormLiangMS:
@isabelmsft cherry-pick failures, please create separate PR for 202211 branch.

PR #2620, opened 2023-01-23
Update override_config_table for multi-asic platforms

judyjoseph:
#### What I did
Update override_config_table for multi-asic platforms to apply the overriding config in golden_config_db.json to all namespaces in addition to the host namespace.

This will be removed later as we support the complete depreciation of minigraph and the golden_config_db.json file support the DB schema for multi-asic platforms. 

#### How I did it
Check the db client list which contains the db connector to DB in multiple namespaces and do yang validation and config override.

#### How to verify it
Validated with MACSEC_PROFILE present in the golden_config_db.json file and checked that the MACSEC_PROFILE gets added/updated in the config_db's for host as well as the namespaces.

```
admin@str2--lc1-1:~$ sudo config load_minigraph -y -o
Disabling container monitoring ...
Stopping SONiC target ...
Running command: /usr/local/bin/sonic-cfggen -H -m -j /etc/sonic/init_cfg.json   --write-to-db
Running command: /usr/local/bin/sonic-cfggen -H -m -j /etc/sonic/init_cfg.json  -n asic0 --write-to-db
/usr/local/lib/python3.9/dist-packages/minigraph.py:796: FutureWarning: The behavior of this method will change in future versions. Use specific 'len(elem)' or 'elem is not None' test instead.
  if voqinbandintfs:
Running command: /usr/local/bin/sonic-cfggen -H -m -j /etc/sonic/init_cfg.json  -n asic1 --write-to-db
/usr/local/lib/python3.9/dist-packages/minigraph.py:796: FutureWarning: The behavior of this method will change in future versions. Use specific 'len(elem)' or 'elem is not None' test instead.
  if voqinbandintfs:
Running command: /usr/local/bin/sonic-cfggen -d -y /etc/sonic/sonic_version.yml -t /usr/share/sonic/templates/sonic-environment.j2,/etc/sonic/sonic-environment
Running command: config qos reload --no-dynamic-buffer --no-delay
Running command: /usr/local/bin/sonic-cfggen -n asic0 -d --write-to-db -t /usr/share/sonic/device/0/buffers.json.j2,config-db -t /usr/share/sonic/device/0/qos.json.j2,config-db -y /etc/sonic/sonic_version.yml
Running command: /usr/local/bin/sonic-cfggen -n asic1 -d --write-to-db -t /usr/share/sonic/device/1/buffers.json.j2,config-db -t /usr/share/sonic/device/1/qos.json.j2,config-db -y /etc/sonic/sonic_version.yml
Running command: pfcwd start_default
Running command: config override-config-table /etc/sonic/golden_config_db.json
Working in Host namespace :
  - Removing configDB overriden table first ...
  - Overriding input config to configDB ...
  - Overriding completed. No service is restarted.
Working in asic0 namespace :
  - Removing configDB overriden table first ...
  - Overriding input config to configDB ...
  - Overriding completed. No service is restarted.
Working in asic1 namespace :
  - Removing configDB overriden table first ...
  - Overriding input config to configDB ...
  - Overriding completed. No service is restarted.
Restarting SONiC target ...
Enabling container monitoring ...
Reloading Monit configuration ...
Reinitializing monit daemon
Please note setting loaded from minigraph will be lost after system reboot. To preserve setting, run `config save`.
admin@str2--lc1-1:~$ sudo config save -y
Running command: /usr/local/bin/sonic-cfggen -n asic0 -d --print-data > /etc/sonic/config_db0.json
Running command: /usr/local/bin/sonic-cfggen -n asic1 -d --print-data > /etc/sonic/config_db1.json
```

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1893,9 +1893,23 @@ def override_config_table(db, input_config_db, dry_run):
                     fg='magenta')
         sys.exit(1)
 
-    # Do the yang validation and config override for host namespace and 
+
+    # List of Tables that needs to be populated in host and asic namespaces
+    tables_per_namespace = {"MACSEC_PROFILE"}
+
+    # Get the namespace db connector where we really need to override the config.
+    # if any of the table in config_input is part of the list tables_per_namespace
+    # we override config in all namespaces, else do it only in host namespace.
+    db_list = {}
+    db_list[DEFAULT_NAMESPACE] = db.cfgdb
+    for table in config_input:
+        if table in tables_per_namespace:
+            db_list = db.cfgdb_clients
+            break
+
+    # Do the yang validation and config override for host namespace and
     # other namespaces in case of multi-asic platform
-    for ns, config_db in db.cfgdb_clients.items():
+    for ns, config_db in db_list.items():
         # Read config from configDB
         current_config = config_db.get_config()
         # Serialize to the same format as json input
@@ -1923,8 +1937,9 @@ def override_config_table(db, input_config_db, dry_run):
             print(json.dumps(updated_config, sort_keys=True,
                              indent=4, cls=minigraph_encoder))
         else:
-            override_config_db(ns, config_db, config_input)
-
+            namespace = "Host" if ns is DEFAULT_NAMESPACE else ns
+            click.echo("Working in {} namespace :".format(namespace))
+            override_config_db(config_db, config_input)
 
 def validate_config_by_cm(cm, config_json, jname):
     tmp_config_json = copy.deepcopy(config_json)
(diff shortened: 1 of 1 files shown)


config/main.py
```diff
@@ -1893,36 +1893,37 @@ def override_config_table(db, input_config_db, dry_run):
...
-    if dry_run:
-        print(json.dumps(updated_config, sort_keys=True,
-                         indent=4, cls=minigraph_encoder))
-    else:
-        override_config_db(config_db, config_input)
+        if dry_run:
+            print(json.dumps(updated_config, sort_keys=True,
+                             indent=4, cls=minigraph_encoder))
+        else:
+            override_config_db(ns, config_db, config_input)
```
qiluo-msft, line 1926: >ns [](http://example.com/codeflow?start=31&length=2)

Seems the new parameter is only used for echo. In this case, you do not need an extra parameter, you can just echo before the function call "override_config_db in namespace {}"

config/main.py
```diff
@@ -1893,36 +1893,37 @@ def override_config_table(db, input_config_db, dry_run):
...
-        # config during initialization.
-        try:
-            cm = ConfigMgmt()
-            cm.validateConfigData()
-        except Exception as ex:
-            click.secho("Failed to validate running config. Error: {}".format(ex), fg="magenta")
-            sys.exit(1)
+    # Do the yang validation and config override for host namespace and 
+    # other namespaces in case of multi-asic platform
+    for ns, config_db in db.cfgdb_clients.items():
```
qiluo-msft, line 1898: >for [](http://example.com/codeflow?start=4&length=3)

I understand that `MACSEC_PROFILE` will be in all namespaces in addition to the host namespace. However this is right in general.

Can we limit the loop only for `MACSEC_PROFILE` table?

config/main.py
```diff
@@ -1943,18 +1944,18 @@ def update_config(current_config, config_input):
     return updated_config
 
 
-def override_config_db(config_db, config_input):
+def override_config_db(ns, config_db, config_input):
```
qiluo-msft, line 1947: >override_config_db [](http://example.com/codeflow?start=4&length=18)

Could you turn "How to verify it" into a unit testcase?

config/main.py
judyjoseph: Updated not to add an additional "ns" parameter

config/main.py
judyjoseph: I have updated the logic to have a list of tables which is there per namespace, if any of the tables in the config_input json is present in that list, we do override config for all namespaces. Else only for host. Currently MACSEC_PROFILE is there in that table list.

this logic can be generalized to any table depending on how we plan to go with Virtual connector, to do this action based on table, and have it give back which connector to use, or which namespace the config needs to be overridden.

judyjoseph:
closing this in favor of https://github.com/sonic-net/sonic-utilities/pull/2738

PR #2621, opened 2023-01-23
Remove timer from FAST_REBOOT STATE_DB entry and use finalizer

arfeigin:
This should come along with sonic-buildimage PR (https://github.com/sonic-net/sonic-buildimage/pull/13484) implementing fast-reboot finalizing logic in finalize-warmboot script and other submodules PRs utilizing the change.

This PR should come along with the following PRs as well:
https://github.com/sonic-net/sonic-swss-common/pull/742
https://github.com/sonic-net/sonic-platform-daemons/pull/335
https://github.com/sonic-net/sonic-sairedis/pull/1196

This set of PRs solves the issue https://github.com/sonic-net/sonic-buildimage/issues/13251

#### What I did
Remove the timer used to clear fast-reboot entry from state-db, instead it will be cleared by fast-reboot finalize function implemented inside finalize-warmboot script (which will be invoked since fast-reboot is using warm-reboot infrastructure).

As well instead of having "1" as the value for fast-reboot entry in state-db and deleting it when done it is now modified to set enable/disable according to the context.

As well all scripts reading this entry should be modified to the new value options.

#### How I did it
Removed the timer usage in the fast-reboot script and adding fast-reboot finalize logic to warm-reboot in the linked PR.
Use "enable/disable" instead of "1" as the entry value.

#### How to verify it
Run fast-reboot and check that the state-db entry for fast-reboot is being deleted after finalizing fast-reboot and not by an expiring timer.


stepanblyschak (dismissed):

stepanblyschak (approved):

scripts/db_migrator.py
```diff
@@ -850,9 +850,21 @@ def version_3_0_6(self):
     def version_4_0_0(self):
         """
         Version 4_0_0.
-        This is the latest version for master branch
         """
         log.log_info('Handling version_4_0_0')
+        # Update state-db fast-reboot entry to enable if set to enable fast-reboot finalizer when using upgrade with fast-reboot
+        fastreboot_state = self.stateDB.get(self.stateDB.STATE_DB, 'FAST_REBOOT|system', '1')
+        if fastreboot_state == 'true':
```
vaibhavhd, line 857: Do you not want to set the entry to `disable` when the fastreboot_state is not true?

scripts/fast-reboot
```diff
@@ -532,7 +532,7 @@ case "$REBOOT_TYPE" in
         check_warm_restart_in_progress
         BOOT_TYPE_ARG=$REBOOT_TYPE
         trap clear_boot EXIT HUP INT QUIT TERM KILL ABRT ALRM
-        sonic-db-cli STATE_DB SET "FAST_REBOOT|system" "1" "EX" "210" &>/dev/null
+        sonic-db-cli STATE_DB SET "FAST_REBOOT|system" "enable" &>/dev/null
```
vaibhavhd, line 535: Regarding the naming of system wide flag, please check-- https://github.com/sonic-net/sonic-sairedis/pull/1196#discussion_r1099245793

scripts/db_migrator.py
```diff
@@ -45,7 +45,7 @@ def __init__(self, namespace, socket=None):
                      none-zero values.
               build: sequentially increase within a minor version domain.
         """
-        self.CURRENT_VERSION = 'version_4_0_0'
+        self.CURRENT_VERSION = 'version_4_0_1'
```
vaibhavhd, line 48: I think we can't take this to 202205 easily. Will there be separate PR for 202205?

arfeigin: Yes, I am working on 202205 seperate PR.
Advancing 3_0_5 to 3_0_6 but it looks like it is not working smooth.

vaibhavhd (dismissed):

stepanblyschak (approved):

sonic-utilities-data/templates/service_mgmt.sh.j2
```diff
@@ -51,7 +51,8 @@ function check_warm_boot()
 
 function check_fast_boot()
 {
-    if [[ $($SONIC_DB_CLI STATE_DB GET "FAST_REBOOT|system") == "1" ]]; then
+    SYSTEM_FAST_REBOOT=`sonic-db-cli STATE_DB hget "FAST_RESTART_ENABLE_TABLE|system" enable`
```
vaibhavhd, line 54: Suggest to still use `$FAST_RESTART_ENABLE_TABLE` instead of `sonic-db-cli`

vaibhavhd (approved):

PR #2623, opened 2023-01-24
Ignore bgpraw in GCU applier

wen587:
#### What I did
`show run all` output will include bgpraw for business needs. GCU ipv6 test will update BGP_NEIGHBOR table which caused `bgpraw` content change, which will make the `apply-patch` operation fail. The solution is to add `bgpraw` to ignored tables.
#### How I did it
Add new added `bgpraw` table to ignored backend table.
#### How to verify it
Existing Unit test and local E2E GCU test.

### DIFF
--- a/generic_config_updater/change_applier.py
+++ b/generic_config_updater/change_applier.py
@@ -76,6 +76,7 @@ class ChangeApplier:
     def __init__(self):
         self.config_db = get_config_db()
         self.backend_tables = [
+            "bgpraw",
             "BUFFER_PG",
             "BUFFER_PROFILE",
             "FLEX_COUNTER_TABLE"


isabelmsft (approved):

vadymhlushko-mlnx:
Hi @wen587, it seems like because of this PR - the Azure pipeline tests not passing for [submodule update PR #13508](https://github.com/sonic-net/sonic-buildimage/pull/13508)

wen587:
> Hi @wen587, it seems like because of this PR - the Azure pipeline tests not passing for [submodule update PR #13508](https://github.com/sonic-net/sonic-buildimage/pull/13508). Could you please take a look?

Hi @vadymhlushko-mlnx, I will merge the fix later. https://github.com/sonic-net/sonic-utilities/pull/2628

PR #2626, opened 2023-01-24
[202205][show_bfd] add local discriminator in show bfd command

baorliu:
Double commit of PR https://github.com/sonic-net/sonic-utilities/pull/2616


baorliu:
don't need for 202205, 202211

PR #2627, opened 2023-01-24
[202211][show_bfd] add local discriminator in show bfd command

baorliu:
Double commit of PR https://github.com/sonic-net/sonic-utilities/pull/2616


prsunny:
@baorliu , we dont need PRs for 202205, 202211 etc. We can label those and if there are conflicts, you can raise separate PRs. Mostly it is required for 202012

baorliu:
don't need for 202205, 202211

PR #2628, opened 2023-01-25
[GCU] Ignore bgpraw table in GCU operation

wen587:
#### What I did
After the previous fix https://github.com/sonic-net/sonic-utilities/pull/2623 , GCU still fails in the rollback operation. The `bgpraw` table should be discard in all GCU operation. 
Thus, I change `get_config_db_as_json` function to crop out "bgpraw" table.
#### How I did it
Pop "bgpraw" table if exists.
#### How to verify it
Unittest

### DIFF
--- a/generic_config_updater/gu_common.py
+++ b/generic_config_updater/gu_common.py
@@ -54,7 +54,9 @@ class ConfigWrapper:
 
     def get_config_db_as_json(self):
         text = self._get_config_db_as_text()
-        return json.loads(text)
+        config_db_json = json.loads(text)
+        config_db_json.pop("bgpraw", None)
+        return config_db_json
 
     def _get_config_db_as_text(self):
         # TODO: Getting configs from CLI is very slow, need to get it from sonic-cffgen directly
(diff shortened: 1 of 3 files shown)


isabelmsft (approved):

isabelmsft (commented):
Please add UT coverage

generic_config_updater/gu_common.py
```diff
@@ -57,8 +57,8 @@ def get_config_db_as_json(self):
         return json.loads(text)
 
     def _get_config_db_as_text(self):
-        # TODO: Getting configs from CLI is very slow, need to get it from sonic-cffgen directly
-        cmd = "show runningconfiguration all"
```
qiluo-msft, line 61: >"show runningconfiguration all" [](http://example.com/codeflow?start=14&length=31)

The original cmd is better because it is the formal CLI  interface. Event today the implementation is the same, but in future, the formal CLI will work better with multi-asic.

Instead, you can remove unwanted fields in `get_config_db_as_json` function.

qiluo-msft (approved):
Please also check with other reviewers.

isabelmsft (approved):

qiluo-msft:
@wen587 Please make PR description more specific. What is the "previous fix"?

qiluo-msft:
@wen587 Please be aware that I am reverting your code in https://github.com/sonic-net/sonic-utilities/pull/2631 (not merged yet). Please sync with me when you plan to merge this one.

PR #2631, opened 2023-01-26
Revert "Ignore bgpraw in GCU applier"

qiluo-msft:
Reverts sonic-net/sonic-utilities#2623


isabelmsft (approved):

wen587 (changes requested):
Let's merge this one for fix
https://github.com/sonic-net/sonic-utilities/pull/2628

PR #2633, opened 2023-01-27
[sai_failure_dump]Invoking dump during SAI failure

dgsudharsan:
HLD: https://github.com/sonic-net/SONiC/pull/1212

#### What I did
Added logic in techsupport script to collect SAI failure dump

#### How I did it
Collected the SAI failure dumps if present.

#### How to verify it
Invoke techsupport and verify.

### DIFF
--- a/scripts/generate_dump
+++ b/scripts/generate_dump
@@ -1059,21 +1059,26 @@ collect_mellanox() {
     local sai_dump_folder="/tmp/saisdkdump"
     local sai_dump_filename="${sai_dump_folder}/sai_sdk_dump_$(date +"%m_%d_%Y_%I_%M_%p")"
 
-    ${CMD_PREFIX}docker exec syncd mkdir -p $sai_dump_folder
-    ${CMD_PREFIX}docker exec syncd saisdkdump -f $sai_dump_filename
+    if [[ "$( docker container inspect -f '{{.State.Running}}' syncd )" == "true" ]]; then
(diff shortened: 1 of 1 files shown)


scripts/generate_dump
```diff
@@ -1053,21 +1053,26 @@ collect_mellanox() {
     local sai_dump_folder="/tmp/saisdkdump"
     local sai_dump_filename="${sai_dump_folder}/sai_sdk_dump_$(date +"%m_%d_%Y_%I_%M_%p")"
 
-    ${CMD_PREFIX}docker exec syncd mkdir -p $sai_dump_folder
-    ${CMD_PREFIX}docker exec syncd saisdkdump -f $sai_dump_filename
-
-    if [ $? != 0 ]; then
-        echo "Failed to collect saisdkdump."
-    fi
+    if [[ "$( docker container inspect -f '{{.State.Running}}' syncd )" == "true" ]]; then
```
prsunny, line 1056: Why is this change part of this PR? This is not related to the feature, correct?
Suggest revert this change.

dgsudharsan: We have saisdkdump run as part of regular techsupport flow which will be called during auto techsupport. Since during orchagent abort, syncd will not be running, this flow will throw errors. That's the main reason we are introducing this feature and since we are taking sai failure dump, a check is added here to ensure that saisdkdump is not called in regular techsupport flow when syncd is not running.

volodymyrsamotiy (approved):

prsunny (approved):

PR #2634, opened 2023-01-28
Work around unpopulated hwsku in db_migrator.py

rjthomson:
Handle the case where hwsku is None.
See sonic-net/sonic-buildimage#12083 and sonic-net/sonic-buildimage#11416

#### How I did it
`db_migrator.py` can fail on Arista devices if the hwsku field is not populated.
Check that the field is not None before trying to read the value.

#### How to verify it
Tested on Arista products, the error no longer occurs.


maipbui:
Fixed in https://github.com/sonic-net/sonic-utilities/pull/2821/files

PR #2638, opened 2023-01-31
[GCU] protect loopback0 from deletion

wen587:
#### What I did
Refer to https://github.com/sonic-net/sonic-buildimage/issues/11171, protect loopback0 from deletion
#### How I did it
Add patch checker to fail the validation when remove loopback0
#### How to verify it
Unit test

### DIFF
--- a/tests/generic_config_updater/gu_common_test.py
+++ b/tests/generic_config_updater/gu_common_test.py
@@ -75,7 +75,7 @@ class TestConfigWrapper(unittest.TestCase):
         config_wrapper = gu_common.ConfigWrapper()
         config_wrapper.validate_field_operation(old_config, target_config)
 
-    def test_validate_field_operation_legal__loopback0(self):
+    def test_validate_field_operation_legal__rm_loopback1(self):
         old_config = {
             "LOOPBACK_INTERFACE": {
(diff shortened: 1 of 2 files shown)


isabelmsft (dismissed):

tests/generic_config_updater/gu_common_test.py
```diff
@@ -69,18 +69,54 @@ def setUp(self):
...
+            "LOOPBACK_INTERFACE": {
+                "Loopback0": {},
+                "Loopback0|10.1.0.32/32": {},
+                "Loopback1": {},
+                "Loopback1|10.1.0.33/32": {}
+            }
+        }
+        target_config = {
+            "LOOPBACK_INTERFACE": {
+                "Loopback0": {},
```
qiluo-msft, line 89: >Loopback0 [](http://example.com/codeflow?start=17&length=9)

Seem you are testing removing Loopback1. Do you intent to removing Loopback0?

qiluo-msft: Still applicable.

isabelmsft (approved):

qiluo-msft (approved):

PR #2642, opened 2023-02-01
Add CLI configuration options for teamd retry count feature

saiarcot895:
Add a SONiC CLI to more easily configure the retry count for port channels. This effectively acts like a wrapper around the underlying teamdctl command.

Also add a python script that'll be installed into /usr/local/bin/teamd_increase_retry_count.py that will detect if the peer device likely supports this teamd feature (based on LLDP neighbor info) and increases the teamd retry count to 5 in preparation for warm upgrade. This script requires sudo to run.

This is tied to sonic-net/sonic-buildimage#13453.

Signed-off-by: Saikrishna Arcot <sarcot@microsoft.com>

```
admin@vlab-02:~$ sudo config portchannel retry-count get PortChannel101
3
admin@vlab-02:~$ sudo config portchannel retry-count get PortChannel999
Usage: config portchannel retry-count get [OPTIONS] <portchannel_name>
Try "config portchannel retry-count get -h" for help.

Error: PortChannel999 is not present.
admin@vlab-02:~$ sudo config portchannel retry-count set PortChannel101 3
admin@vlab-02:~$ sudo config portchannel retry-count set PortChannel101 5
```


yxieca (approved):

config/main.py
```diff
@@ -2199,6 +2199,83 @@ def del_portchannel_member(ctx, portchannel_name, port_name):
...
+            ctx.fail("{} is not present.".format(portchannel_name))
+
+    try:
+        proc = subprocess.Popen(["teamdctl", portchannel_name, "state", "item", "get", "runner.enable_retry_count_feature"], text=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        output, err = proc.communicate()
+        if proc.returncode != 0:
+            ctx.fail("Unable to determine if the retry count feature is enabled or not: {}".format(err.strip()))
+        if output.strip() != "true":
+            ctx.fail("Retry count feature is not enabled!")
+    except FileNotFoundError:
```
vaibhavhd, line 2231: Why catch `FileNotFoundError`? Also please catch other exceptions that might be raised due to Popen and communicate. Good idea to also add catch-all handler at the end.

config/main.py
```diff
@@ -2199,6 +2199,83 @@ def del_portchannel_member(ctx, portchannel_name, port_name):
...
+            ctx.fail("{} is invalid!, name should have prefix '{}' and suffix '{}'"
+                     .format(portchannel_name, CFG_PORTCHANNEL_PREFIX, CFG_PORTCHANNEL_NO))
+
+        # Dont proceed if the port channel does not exist
+        if is_portchannel_present_in_db(db, portchannel_name) is False:
+            ctx.fail("{} is not present.".format(portchannel_name))
+
+    try:
+        proc = subprocess.Popen(["teamdctl", portchannel_name, "state", "item", "get", "runner.enable_retry_count_feature"], text=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        output, err = proc.communicate()
```
vaibhavhd, line 2226: This is a blocking call, right? May be we should consider adding a timeout. And then also handling that timeout.

config/main.py
```diff
@@ -2199,6 +2199,83 @@ def del_portchannel_member(ctx, portchannel_name, port_name):
...
+        if proc.returncode != 0:
+            ctx.fail("Unable to determine if the retry count feature is enabled or not: {}".format(err.strip()))
+        if output.strip() != "true":
+            ctx.fail("Retry count feature is not enabled!")
+    except FileNotFoundError:
+        ctx.fail("Unable to get the retry count: teamdctl could not be run")
+
+    try:
+        proc = subprocess.Popen(["teamdctl", portchannel_name, "state", "item", "get", "runner.retry_count"], text=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        output, err = proc.communicate()
```
vaibhavhd, line 2236: Same comment as above regarding blocking call and timeout.

config/main.py
```diff
@@ -2199,6 +2199,83 @@ def del_portchannel_member(ctx, portchannel_name, port_name):
...
+def set_portchannel_retry_count(ctx, portchannel_name, retry_count):
+    """Set the retry count for a port channel"""
+    # Dont proceed if the port channel name is not valid
+    if is_portchannel_name_valid(portchannel_name) is False:
+        ctx.fail("{} is invalid!, name should have prefix '{}' and suffix '{}'"
+                 .format(portchannel_name, CFG_PORTCHANNEL_PREFIX, CFG_PORTCHANNEL_NO))
+
+    db = ValidatedConfigDBConnector(ctx.obj['db'])
+
+    if ADHOC_VALIDATION:
```
vaibhavhd, line 2256: Why is `ADHOC_VALIDATION` necessary in this context?

scripts/fast-reboot
```diff
@@ -663,6 +663,17 @@ if [[ "$REBOOT_TYPE" = "fast-reboot" ]]; then
     fi
 fi
 
+TEAMD_INCREASE_RETRY_COUNT=0
+if [[ "$REBOOT_TYPE" = "warm-reboot" || "$REBOOT_TYPE" = "fastfast-reboot" ]]; then
+    /usr/local/bin/teamd_increase_retry_count.py --probe-only || TEAMD_RETRY_COUNT_PROBE_RC=$?
+    if [[ TEAMD_RETRY_COUNT_PROBE_RC -ne 0 ]]; then
+        debug "Warning: Not all neighbors support teamd retry count feature"
```
vaibhavhd, line 670: This warning is not always true when return code is != 0. For example, the script can fail for reasons other than retry count support being unavailable.

scripts/fast-reboot
```diff
@@ -663,6 +663,17 @@ if [[ "$REBOOT_TYPE" = "fast-reboot" ]]; then
     fi
 fi
 
+TEAMD_INCREASE_RETRY_COUNT=0
+if [[ "$REBOOT_TYPE" = "warm-reboot" || "$REBOOT_TYPE" = "fastfast-reboot" ]]; then
+    /usr/local/bin/teamd_increase_retry_count.py --probe-only || TEAMD_RETRY_COUNT_PROBE_RC=$?
+    if [[ TEAMD_RETRY_COUNT_PROBE_RC -ne 0 ]]; then
```
vaibhavhd, line 669: Are you missing a `$` in variable usage? I think it should be `if [[ $TEAMD_RETRY_COUNT_PROBE_RC -ne 0 ]]; then`

scripts/teamd_increase_retry_count.py
```diff
@@ -0,0 +1,236 @@
...
+            sys.exit(2)
+            return False
+    else:
+        signal.signal(signal.SIGUSR1, abortTeamdChanges)
+        signal.signal(signal.SIGTERM, abortTeamdChanges)
+        retryCountGetProcess = subprocess.run(["config", "portchannel", "retry-count", "get", portChannels[0]])
+        if retryCountGetProcess.returncode == 0:
+            # Currently running on SONiC version with teamd retry count feature
+            for portChannel in portChannels:
+                subprocess.run(["config", "portchannel", "retry-count", "set", portChannel, "5"])
```
vaibhavhd, line 209: I suggest to define retry-count 5 and 3 as constants on top of file, and use them here.

scripts/teamd_increase_retry_count.py
```diff
@@ -0,0 +1,236 @@
...
+    while not revertTeamdRetryCountChanges:
+        for port, packet in packets:
+            sendp(packet, iface=port)
+        time.sleep(15)
+    if revertTeamdRetryCountChanges:
+        for port, packet in revertPackets:
+            sendp(packet, iface=port)
+
+def abortTeamdChanges(signum, frame):
+    print("Got signal {}, reverting teamd retry count change".format(signum))
```
vaibhavhd, line 137: Is the `print` usage intentional? I think we should use `logger`.

scripts/teamd_increase_retry_count.py
```diff
@@ -0,0 +1,236 @@
...
+                portChannelChecked = True
+                if "SONiC" not in peerInfo["descr"]:
+                    print("WARNING: Peer device is not a SONiC device; skipping")
+                    failedPortChannels.append(portChannel)
+                    break
+
+                sendReadyEvent = Event()
+
+                # Start sniffing thread
+                lacpThread = LacpPacketListenThread(portName, config["ports"][portName]["runner"]["partner_lacpdu_info"]["system"], sendReadyEvent)
```
vaibhavhd, line 174: This access of config can potentially result in KeyError

scripts/teamd_increase_retry_count.py
```diff
@@ -0,0 +1,236 @@
...
+
+def getLldpNeighbors():
+    process = subprocess.run(["lldpctl", "-f", "json"], capture_output=True)
+    return json.loads(process.stdout)
+
+def craftLacpPacket(portChannelConfig, portName, isResetPacket=False, newVersion=True):
+    portConfig = portChannelConfig["ports"][portName]
+    actorConfig = portConfig["runner"]["actor_lacpdu_info"]
+    partnerConfig = portConfig["runner"]["partner_lacpdu_info"]
+    l2 = Ether(dst="01:80:c2:00:00:02", src=portConfig["ifinfo"]["dev_addr"], type=0x8809)
```
vaibhavhd, line 80: Why the dst is hardcoded?

scripts/teamd_increase_retry_count.py
```diff
@@ -0,0 +1,236 @@
...
+    l3 = scapy.contrib.lacp.SlowProtocol(subtype=0x01) 
+    l4 = LACPRetryCount()
+    if newVersion:
+        l4.version = 0xf1
+    else:
+        l4.version = 0x1
```
vaibhavhd, line 86: This will override any version other than 0x1. For eg., when/if teamd follow version 0x2 for LACPDU, then this logic will forcefully set it to 0x1.

scripts/teamd_increase_retry_count.py
```diff
@@ -0,0 +1,236 @@
...
+    l4.partner_port_number = partnerConfig["port"]
+    l4.partner_state = partnerConfig["state"]
+    if newVersion:
+        l4.actor_retry_count = 5 if not isResetPacket else 3
+        l4.partner_retry_count = 3
+    packet = l2 / l3 / l4
+    return packet
+
+def getPortChannels():
+    configDb = DBConnector("CONFIG_DB", 0)
```
vaibhavhd, line 106: CONFIGDB will return all expected POs.
What you want is all admin/oper UP portchannels. For that app_db is the source.

You might also want to check if the BGP is UP on these LAGs.

Otherwise, the script will attempt to send LACPDUs to the neighbors that are not established, or/and on ports that are admin/oper down.

scripts/fast-reboot
```diff
@@ -687,6 +698,10 @@ if [ -x ${LOG_SSD_HEALTH} ]; then
     ${LOG_SSD_HEALTH}
 fi
 
+if [[ ( "$REBOOT_TYPE" = "warm-reboot" || "$REBOOT_TYPE" = "fastfast-reboot" ) && "$TEAMD_INCREASE_RETRY_COUNT" -eq 1 ]]; then
```
vaibhavhd, line 701: I suggest to use syntax of `${REBOOT_TYPE}` over `$REBOOT_TYPE`. The former is more safer and predictable.

scripts/fast-reboot
saiarcot895: That was my initial thinking as well, but then I saw the use of this syntax in other parts of this file (above and below this section), and it worked, so I followed that syntax to match that style. I don't entirely know why it works (is there some feature in bash that checks to see that the string is a variable if it's not a number?), but I think it is better to have it with the `$`.

I'll change it after verifying it doesn't break anything.

scripts/teamd_increase_retry_count.py
saiarcot895: This is a special Ethernet address used for slow protocols.

config/main.py
saiarcot895: `FileNotFoundError` is specifically for `teamdctl` (and other teamd binaries) not existing in the `sonic-utilities` test environment. I'll add a catch-all handler as well.

config/main.py
saiarcot895: Timeout added, along with handler.

config/main.py
saiarcot895: Timeout added, along with handler.

config/main.py
saiarcot895: I was following the same pattern as the portchannel member getter and setter, but I see now that this is mainly meant for YANG validation. I don't _think_ this will need to be tested through YANG, so I removed that if-check.

scripts/teamd_increase_retry_count.py
saiarcot895: Constants defined and used.

scripts/teamd_increase_retry_count.py
saiarcot895: Switched to logger, with some messages still being explicitly being printed to console (provided the minimum log level is met). This is so that it's easy for someone to run just this script and see if it passed or not.

scripts/teamd_increase_retry_count.py
saiarcot895: Checks added.

scripts/fast-reboot
saiarcot895: Changed.

scripts/teamd_increase_retry_count.py
saiarcot895: Added some checks to verify that at least some level of keys are returned from `teamd`.

scripts/fast-reboot
saiarcot895: Updated the message to say that support is unknown for one or more neighbor devices.

scripts/teamd_increase_retry_count.py
saiarcot895: If teamd does add support for version 0x2 LACPDUs and default to using that version, we would anyways need a transition mechanism of some kind to handle both 0x1 and 0x2 (assuming there's no built-in mechanism). I think if that does happen, we can add support for that here.

scripts/fast-reboot
yxieca: I think you also need to set it to 0 before line 668? I agree it is better with $.

scripts/fast-reboot
```diff
@@ -663,6 +663,17 @@ if [[ "$REBOOT_TYPE" = "fast-reboot" ]]; then
     fi
 fi
 
+TEAMD_INCREASE_RETRY_COUNT=0
+if [[ "${REBOOT_TYPE}" = "warm-reboot" || "${REBOOT_TYPE}" = "fastfast-reboot" ]]; then
+    /usr/local/bin/teamd_increase_retry_count.py --probe-only || TEAMD_RETRY_COUNT_PROBE_RC=$?
+    if [[ TEAMD_RETRY_COUNT_PROBE_RC -ne 0 ]]; then
+        debug "Warning: Retry count feature support unknown for one or more neighbor devices; assuming that it's not available"
+        TEAMD_INCREASE_RETRY_COUNT=0
```
yxieca, line 671: Are we supposed to bail out from here?

saiarcot895: For the images we deploy to our internal environment, this will be changed to bail out and fail the warm-reboot if SONiC neighbors are not seen. For the public version, since we can't require using SONiC-only neighbors for warm reboot, this will only be a warning.

yxieca (dismissed):

scripts/fast-reboot
```diff
@@ -635,6 +644,22 @@ init_warm_reboot_states
...
+TEAMD_INCREASE_RETRY_COUNT=0
+if [[ "${REBOOT_TYPE}" = "warm-reboot" || "${REBOOT_TYPE}" = "fastfast-reboot" ]]; then
+    TEAMD_RETRY_COUNT_PROBE_RC=0
+    /usr/local/bin/teamd_increase_retry_count.py --probe-only || TEAMD_RETRY_COUNT_PROBE_RC=$?
+    if [[ $TEAMD_RETRY_COUNT_PROBE_RC -ne 0 ]]; then
+        if [[ "${REQUIRE_TEAMD_RETRY_COUNT}" = "yes" ]]; then
+            error "Could not confirm that all neighbor devices are running SONiC with the retry count feature"
+        else
+            debug "Warning: Retry count feature support unknown for one or more neighbor devices; assuming that it's not available"
+            TEAMD_INCREASE_RETRY_COUNT=0
```
vaibhavhd, line 656: Is this needed to be set to 0 again? The variable is initially set to 0 anyway, right?

scripts/fast-reboot
```diff
@@ -663,6 +688,10 @@ if [[ "$REBOOT_TYPE" = "fast-reboot" ]]; then
     fi
 fi
 
+if [[ ( "${REBOOT_TYPE}" = "warm-reboot" || "${REBOOT_TYPE}" = "fastfast-reboot" ) && "${TEAMD_INCREASE_RETRY_COUNT}" -eq 1 ]]; then
```
vaibhavhd, line 691: I think this feature is needed even for fast-reboot?
For fast-reboot it will help avoid LAG flaps too - if the boot up is slow enough that 90s is reached before even syncd does ASIC reset call. So essentially, preventing LAG flap other than where it should really happen.

vaibhavhd: We can make decision on this later too.

scripts/teamd_increase_retry_count.py
```diff
@@ -0,0 +1,320 @@
...
+
+def getCmdOutput(cmd):
+    proc = subprocess.Popen(cmd, stdout=subprocess.PIPE)
+    return proc.communicate()[0], proc.returncode
+
+def main(probeOnly=False):
+    if os.geteuid() != 0:
+        log.log_error("Root privileges required for this operation", also_print_to_console=True)
+        sys.exit(1)
+        return False
```
vaibhavhd, line 212: This return will never be reached  after sys.exit

scripts/teamd_increase_retry_count.py
```diff
@@ -0,0 +1,320 @@
...
+    portChannels = getPortChannels()
+    if not portChannels:
+        return True
```
vaibhavhd, line 215: Why is this main returning True/False? What is checking on these return values?

scripts/teamd_increase_retry_count.py
```diff
@@ -0,0 +1,320 @@
...
+
+def getPortChannels():
+    applDb = DBConnector("APPL_DB", 0)
+    configDb = DBConnector("CONFIG_DB", 0)
+    portChannelTable = Table(applDb, "LAG_TABLE")
+    portChannels = portChannelTable.getKeys()
+    activePortChannels = []
+    for portChannel in portChannels:
+        state = portChannelTable.get(portChannel)
+        if not state[0]:
```
vaibhavhd, line 99: Risk of index error here?

scripts/teamd_increase_retry_count.py
```diff
@@ -0,0 +1,320 @@
...
+                    log.log_warning("WARNING: No LLDP info available for {}; skipping".format(portName))
+                    continue
+                interfaceLldpInfo = interfaceLldpInfo[0][portName]
+                peerName = list(interfaceLldpInfo["chassis"].keys())[0]
+                peerInfo = interfaceLldpInfo["chassis"][peerName]
+                if "descr" not in peerInfo:
+                    log.log_warning("WARNING: No peer description available via LLDP for {}; skipping".format(portName))
+                    continue
+                portChannelChecked = True
+                if "SONiC" not in peerInfo["descr"]:
```
vaibhavhd, line 241: Minor: Check w/ ignore case here might be better?

vaibhavhd (commented):
LGTM overall. Posted minor concerns.

scripts/fast-reboot
```diff
@@ -125,6 +128,12 @@ function parseOptions()
             u )
                 SSD_FW_UPDATE_BOOT_OPTION=yes
                 ;;
+            n )
```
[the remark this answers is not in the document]
saiarcot895: I'll have the default be yes in the internal images, since we want to make sure this feature is present internally. Because of this, I have two options that'll set this value consistently, regardless of whether a public image or internal image is being tested.

scripts/fast-reboot
```diff
@@ -635,6 +644,22 @@ init_warm_reboot_states
 
 setup_control_plane_assistant
 
+TEAMD_INCREASE_RETRY_COUNT=0
+if [[ "${REBOOT_TYPE}" = "warm-reboot" || "${REBOOT_TYPE}" = "fastfast-reboot" ]]; then
+    TEAMD_RETRY_COUNT_PROBE_RC=0
+    /usr/local/bin/teamd_increase_retry_count.py --probe-only || TEAMD_RETRY_COUNT_PROBE_RC=$?
+    if [[ $TEAMD_RETRY_COUNT_PROBE_RC -ne 0 ]]; then
+        if [[ "${REQUIRE_TEAMD_RETRY_COUNT}" = "yes" ]]; then
+            error "Could not confirm that all neighbor devices are running SONiC with the retry count feature"
```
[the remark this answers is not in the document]
saiarcot895: I think `error` does the exit as well?

scripts/fast-reboot
```diff
@@ -635,6 +644,22 @@ init_warm_reboot_states
 
 setup_control_plane_assistant
 
+TEAMD_INCREASE_RETRY_COUNT=0
+if [[ "${REBOOT_TYPE}" = "warm-reboot" || "${REBOOT_TYPE}" = "fastfast-reboot" ]]; then
+    TEAMD_RETRY_COUNT_PROBE_RC=0
+    /usr/local/bin/teamd_increase_retry_count.py --probe-only || TEAMD_RETRY_COUNT_PROBE_RC=$?
+    if [[ $TEAMD_RETRY_COUNT_PROBE_RC -ne 0 ]]; then
```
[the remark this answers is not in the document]
saiarcot895: Changed.

scripts/fast-reboot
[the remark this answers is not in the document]
saiarcot895: Looks like it doesn't, not sure why I didn't see this not failing before. Added an exit with a new error code here.

scripts/fast-reboot
saiarcot895: That's true, removed this line.

scripts/teamd_increase_retry_count.py
saiarcot895: Removed the extra returns.

scripts/teamd_increase_retry_count.py
saiarcot895: I think libswsscommon returns whether the call is a success or failure in the first index of the tuple that's returned, but regardless, I added a check to see if an empty tuple/list was returned.

vaibhavhd (dismissed):

vaibhavhd (approved):

PR #2643, opened 2023-02-01
[sonic_installer] remove subprocess with shell=True

maipbui:

#### What I did
#### How I did it
#### How to verify it
Pass UT


sonic_installer/main.py
```diff
@@ -449,7 +448,8 @@ def setup_swapmem(self):
         with open(swapfile, 'wb') as fd:
             os.posix_fallocate(fd.fileno(), 0, self.swap_mem_size * SWAPAllocator.MiB_TO_BYTES_FACTOR)
         os.chmod(swapfile, 0o600)
-        run_command(f'mkswap {swapfile}; swapon {swapfile}')
+        run_command(['mkswap', swapfile])
```
qiluo-msft, line 451: >run_command [](http://example.com/codeflow?start=8&length=11)

The behavior changes. if first command fails, it will exit.

sonic_installer/main.py
```diff
@@ -604,8 +604,10 @@ def install(url, force, skip_platform_check=False, skip_migration=False, skip_pa
             migrate_sonic_packages(bootloader, binary_image_version)
 
     # Finally, sync filesystem
-    run_command("sync;sync;sync")
-    run_command("sleep 3")  # wait 3 seconds after sync
+    run_command(["sync"])
```
qiluo-msft, line 607: >run_command [](http://example.com/codeflow?start=4&length=11)

The behavior changes. if first command fails, it will exit.

sonic_installer/main.py
```diff
@@ -911,7 +913,7 @@ def upgrade_docker(container_name, url, cleanup_image, skip_check, tag, warm):
     # Restore to previous cold restart setting
     if warm_configured is False and warm:
         if container_name == "swss" or container_name == "bgp" or container_name == "teamd":
-            run_command("config warm_restart disable %s" % container_name)
+            run_command(["config", "warm_restart", "disable", "%s" % container_name])
```
qiluo-msft, line 916: >"%s" % container_name [](http://example.com/codeflow?start=62&length=21)

Use container_name directly?

sonic_installer/main.py
```diff
@@ -949,11 +951,11 @@ def rollback_docker(container_name):
             break
 
     # make previous image as latest
-    run_command("docker tag %s:%s %s:latest" % (image_name, version_tag, image_name))
+    run_command(["docker", "tag", "%s:%s" % (image_name, version_tag), "%s:latest" % (image_name)])
     if container_name == "swss" or container_name == "bgp" or container_name == "teamd":
         echo_and_log("Cold reboot is required to restore system state after '{}' rollback !!".format(container_name), LOG_ERR)
     else:
-        run_command("systemctl restart %s" % container_name)
+        run_command(["systemctl", "restart", "%s" % container_name])
```
qiluo-msft, line 958: >"%s" % container_name [](http://example.com/codeflow?start=45&length=21)

Use container_name directly?

sonic_installer/bootloader/aboot.py
```diff
@@ -126,8 +126,8 @@ def set_next_image(self, image):
         return True
 
     def install_image(self, image_path):
-        run_command("/usr/bin/unzip -od /tmp %s boot0" % image_path)
-        run_command("swipath=%s target_path=/host sonic_upgrade=1 . /tmp/boot0" % image_path)
+        run_command(["/usr/bin/unzip", "-od", "/tmp", "%s" % image_path, "boot0"])
```
qiluo-msft, line 129: >"%s" % image_path [](http://example.com/codeflow?start=54&length=17)

Use image_path directly?

sonic_installer/common.py
```diff
@@ -20,10 +20,14 @@
 DOCKERDIR_NAME = 'docker'
 
 # Run bash command and print output to stdout
-def run_command(command):
-    click.echo(click.style("Command: ", fg='cyan') + click.style(command, fg='green'))
-
-    proc = subprocess.Popen(command, shell=True, text=True, stdout=subprocess.PIPE)
+def run_command(command, devnull=False):
```
qiluo-msft, line 23: >devnull=False [](http://example.com/codeflow?start=25&length=13)

Use `stdout=subprocess.PIPE` will simplify the code.

qiluo-msft:
@staphylo Could you help review?

sonic_installer/common.py
```diff
@@ -20,10 +20,14 @@
...
+    command_str = join(command)
```
Junchao-Mellanox, line 24: should we reject a command when its type is string? Or, should we automatically shlex.split it if input command is a string?

sonic_installer/bootloader/uboot.py
```diff
@@ -23,12 +23,12 @@ class UbootBootloader(OnieInstallerBootloader):
...
     def get_installed_images(self):
         images = []
-        proc = subprocess.Popen("/usr/bin/fw_printenv -n sonic_version_1", shell=True, text=True, stdout=subprocess.PIPE)
+        proc = subprocess.Popen(["/usr/bin/fw_printenv", "-n", "sonic_version_1"], text=True, stdout=subprocess.PIPE)
         (out, _) = proc.communicate()
         image = out.rstrip()
         if IMAGE_PREFIX in image:
             images.append(image)
-        proc = subprocess.Popen("/usr/bin/fw_printenv -n sonic_version_2", shell=True, text=True, stdout=subprocess.PIPE)
+        proc = subprocess.Popen(["/usr/bin/fw_printenv", "-n", "sonic_version_2"], text=True, stdout=subprocess.PIPE)
```
Junchao-Mellanox, line 31: Should we force user use `run_command` instead of `subprocess.Popen`?

sonic_installer/main.py
maipbui: I think second command will run regardless of whether or not the first one succeeds? https://stackoverflow.com/questions/25669540/what-is-the-difference-between-double-ampersand-and-semicolon-in-linux

sonic_installer/main.py
maipbui: I think second command will run regardless of whether or not the first one succeeds? https://stackoverflow.com/questions/25669540/what-is-the-difference-between-double-ampersand-and-semicolon-in-linux

sonic_installer/common.py
maipbui: devnull is for these original commands
`run_command("docker exec -i teamd pkill -USR1 teamd > /dev/null")`
`run_command("docker kill %s > /dev/null" % container_name)`

sonic_installer/main.py
qiluo-msft: The new implementation will exit if the `mkswap` failed. [https://github.com/sonic-net/sonic-utilities/blob/b81734289b7ca361b365cbf1c845d31b0a81657e/sonic_installer/common.py#L32](https://github.com/sonic-net/sonic-utilities/blob/b81734289b7ca361b365cbf1c845d31b0a81657e/sonic_installer/common.py#L32)

sonic_installer/main.py
qiluo-msft: The new implementation will exit if the first `sync` fail. Considering this use case, it is a minor issue.

sonic_installer/bootloader/aboot.py
```diff
@@ -126,8 +126,14 @@ def set_next_image(self, image):
...
-        run_command("/usr/bin/unzip -od /tmp %s boot0" % image_path)
-        run_command("swipath=%s target_path=/host sonic_upgrade=1 . /tmp/boot0" % image_path)
+        run_command(["/usr/bin/unzip", "-od", "/tmp", image_path, "boot0"])
+        env = os.environ.copy()
+        env.update({
+            'swipath': image_path,
+            'target_path': '/host',
+            'sonic_upgrade': '1'
+        })
+        run_command(["/bin/bash", "/tmp/boot0"], env=env)
```
qiluo-msft, line 136: >/bin/bash [](http://example.com/codeflow?start=22&length=9)

Will `.` work? No need to assume `bash`.

qiluo-msft: Or just run boot0 as command.

Staphylo: AFAIK `boot0` won't be executable so it will need to be executed via an interpreter.
`.` will not work as it's the equivalent to `source` which just runs the script in the current shell environment.
If anything `/bin/bash` should be replaced by `/bin/sh` to preserve the old behavior.

maipbui: Is there any difference b/w running `/tmp/boot0` using `/bin/bash` and `/bin/sh`? The goal of this PR is to remove the use of `shell=True`, if remain using `/bin/sh` then I think we cannot achieve this goal as the shell is invoked explicitly via `shell=True`.

sonic_installer/bootloader/aboot.py
```diff
@@ -126,8 +126,8 @@ def set_next_image(self, image):
         return True
 
     def install_image(self, image_path):
-        run_command("/usr/bin/unzip -od /tmp %s boot0" % image_path)
-        run_command("swipath=%s target_path=/host sonic_upgrade=1 . /tmp/boot0" % image_path)
+        run_command(["/usr/bin/unzip", "-od", "/tmp", "%s" % image_path, "boot0"])
+        run_command(["swipath=%s" % image_path, "target_path=/host sonic_upgrade=1", ".", "/tmp/boot0"])
```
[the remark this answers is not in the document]
maipbui: Thanks for your explanation!

sonic_installer/bootloader/uboot.py
maipbui: we can but it can change original behavior that it will print the running command to the output and exit if the command fails.

https://github.com/sonic-net/sonic-utilities/blob/5e53d83deb955d335a7bce1dbf2c1b4545ada3c8/sonic_installer/common.py#L25

https://github.com/sonic-net/sonic-utilities/blob/5e53d83deb955d335a7bce1dbf2c1b4545ada3c8/sonic_installer/common.py#L36

sonic_installer/common.py
```diff
@@ -19,11 +19,20 @@
 WORKDIR_NAME = 'work'
 DOCKERDIR_NAME = 'docker'
 
+def is_list_of_strings(command):
+    if isinstance(command, list) and all(isinstance(item, str) for item in command):
```
qiluo-msft, line 23: >if [](http://example.com/codeflow?start=4&length=2)

You do not need if-else. Just return the AND result.

sonic_installer/bootloader/aboot.py
qiluo-msft: In this case, `shell=True` is expected. So we can keep original code, and add comment to suppress semgrep warning.

sonic_installer/bootloader/aboot.py
```diff
@@ -126,8 +126,8 @@ def set_next_image(self, image):
         return True
 
     def install_image(self, image_path):
-        run_command("/usr/bin/unzip -od /tmp %s boot0" % image_path)
-        run_command("swipath=%s target_path=/host sonic_upgrade=1 . /tmp/boot0" % image_path)
+        run_command(["/usr/bin/unzip", "-od", "/tmp", image_path, "boot0"])
+        run_command("swipath=%s target_path=/host sonic_upgrade=1 . /tmp/boot0" % image_path, shell=True)
```
qiluo-msft, line 130: >swipath [](http://example.com/codeflow?start=21&length=7)

Is it possible to use `env` parameter?

qiluo-msft, line 130: >. [](http://example.com/codeflow?start=66&length=1)

Is it possible to use string array as first argument? In your implementation, `is_list_of_strings` will fail if the first argument is string.

qiluo-msft (approved):

qiluo-msft (approved):

PR #2646, opened 2023-02-02
fixed bootloader method verify_image_sign to have dummy implementation…

ycoheNvidia:
…n in father class

#### What I did
Fixed bootloader issue in sonic-installer
#### How I did it
Added dummy implementation of verify_image_sign in bootloader.py
Added tests to check it's available
#### How to verify it
Run sonic-utilities tests and see that they pass
NONE
NONE

### DIFF
--- a/sonic_installer/main.py
+++ b/sonic_installer/main.py
@@ -579,11 +579,16 @@ def install(url, force, skip_platform_check=False, skip_migration=False, skip_pa
 
         # Calling verification script by default - signature will be checked if enabled in bios
         echo_and_log("Verifing image {} signature...".format(binary_image_version))
-        if not bootloader.verify_image_sign(image_path):
-            echo_and_log('Error: Failed verify image signature', LOG_ERR)
-            raise click.Abort()
-        else:
-            echo_and_log('Verification successful')
+        try:
+            if not bootloader.verify_image_sign(image_path):
+                echo_and_log('Error: Failed verify image signature', LOG_ERR)
+                raise click.Abort()
+            else:
+                echo_and_log('Verification successful')
+        except AttributeError:
+                echo_and_log("Skip Verifing image {} signature,".format(binary_image_version) +
+                             "method not implemented for the current bootloader type: {}".format(bootloader.__class__.__name__))
+                pass
 
         echo_and_log("Installing image {} and setting it as default...".format(binary_image_version))
         with SWAPAllocator(not skip_setup_swap, swap_mem_size, total_mem_threshold, available_mem_threshold):
--- a/sonic_installer/main.py
+++ b/sonic_installer/main.py
@@ -586,9 +586,9 @@ def install(url, force, skip_platform_check=False, skip_migration=False, skip_pa
             else:
                 echo_and_log('Verification successful')
         except AttributeError:
-                echo_and_log("Skip Verifing image {} signature,".format(binary_image_version) +
-                             "method not implemented for the current bootloader type: {}".format(bootloader.__class__.__name__))
-                pass
+            echo_and_log("Skip Verifing image {} signature,".format(binary_image_version) +
+            " method not implemented for the current bootloader type: {}".format(bootloader.__class__.__name__))
+            pass
 
         echo_and_log("Installing image {} and setting it as default...".format(binary_image_version))
         with SWAPAllocator(not skip_setup_swap, swap_mem_size, total_mem_threshold, available_mem_threshold):
--- a/tests/test_sonic_installer.py
+++ b/tests/test_sonic_installer.py
@@ -53,6 +53,13 @@ def test_install(run_command, run_command_or_raise, get_bootloader, swap, fs):
     print(result.output)
 
     assert result.exit_code != 0
+    mock_bootloader_verify_image_sign_not_implemented = mock_bootloader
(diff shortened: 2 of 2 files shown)


sonic_installer/bootloader/bootloader.py
```diff
@@ -75,6 +75,10 @@ def supports_package_migration(self, image):
         """tells if the image supports package migration"""
         return True
 
+    def verify_image_sign(self, image_path):
```
Staphylo, line 78: Outside of the scope of this fix but my main grief here is that there was already a method to check for secureboot images called `verify_secureboot_image` (see above).
I think your feature should use this method instead of adding a new one with the exact same purpose.
Also if you look closely most methods in the `Bootloader` class `raise NotImplementedError` to ensure child classes implement all the features (or explicitely skip some)`

ycoheNvidia: Sorry for your grief. I decided not to implement the method as verify_secureboot_image because of follow things:
1. From what I can see, the actual verify_secureboot_image is doing something completely different (checking that it has a certificate). Our method is doing image signature verification concerning secure upgrade
2. naming is not what we intend to do - this verification is regarding image signature, not secureboot (maybe in the general scope of it but not regarding the boot itself)
3. verify_secureboot_image is called with a force flag in main.py - we would like to avoid such thing, as it prevents the whole secure notion
Do you suggest in Bootloader the method will raise NotImplementedError and child classes will have empty implementation of it?

Staphylo (approved):

sonic_installer/bootloader/bootloader.py
```diff
@@ -75,6 +75,10 @@ def supports_package_migration(self, image):
...
+        """verify image signature is valid"""
+        return True
```
qiluo-msft, line 80: >True [](http://example.com/codeflow?start=15&length=4)

I do not understand the semantics of the return value, and why using `True` as default value. A unsecure normal image is considered not signed image, right?

ycoheNvidia: unsecured normal image is considered not signed, yes

liat-grozovik:
@qiluo-msft can you approve so we can merge ?

sonic_installer/bootloader/bootloader.py
prgeor: @ycoheNvidia try/except would be more appropriate? 

```
        echo_and_log("Verifing image {} signature...".format(binary_image_version))
        try : 
               if not bootloader.verify_image_sign(image_path):
                        echo_and_log('Error: Failed verify image signature', LOG_ERR)
                        raise click.Abort()
                else:
                       echo_and_log('Verification successful')
           except AttributeError:
                     echo_and_log("Skip Verifing image {} signature...Not signed image".format(binary_image_version))
                      pass
                     
 ```

Comment (ycoheNvidia) [reply sonic_installer/bootloader/bootloader.py:80]: I am not sure which lines you are referring to @prgeor, can you please elaborate?

Comment (prgeor) [reply sonic_installer/bootloader/bootloader.py:80]: @ycoheNvidia @liat-grozovik instead of defining verify_image_sign() as NOP function in the bootloader.py can we try the except/catch block I showed above? So that if a bootloader does not implement verify_image_sign(), it won't crash here [https://github.com/sonic-net/sonic-utilities/blob/master/sonic_installer/main.py#L582](https://github.com/sonic-net/sonic-utilities/blob/master/sonic_installer/main.py#L582)

**Review by prgeor — APPROVED** [review]

File: sonic_installer/main.py [hunk 3]
```diff
@@ -579,11 +579,16 @@ def install(url, force, skip_platform_check=False, skip_migration=False, skip_pa
...
-            raise click.Abort()
-        else:
-            echo_and_log('Verification successful')
+        try:
+            if not bootloader.verify_image_sign(image_path):
+                echo_and_log('Error: Failed verify image signature', LOG_ERR)
+                raise click.Abort()
+            else:
+                echo_and_log('Verification successful')
+        except AttributeError:
```
qiluo-msft, line 588: >except AttributeError: [](http://example.com/codeflow?start=8&length=22)

This looks like a hack to workaround the logic error inside implementation.

prgeor: @qiluo-msft whats in your mind? How do you want the fix ?

liat-grozovik:
@qiluo-msft can you approve so we can merge ?

ycoheNvidia:
> @qiluo-msft can you approve so we can merge ?

I thought there was an agreement to use https://github.com/sonic-net/sonic-utilities/pull/2698 instead, since original PR 2337 was reverted. 2698 includes both Secure upgrade original code and the fixes included in this PR

liat-grozovik:
@ycoheNvidia can you please update the conflicts?
@qiluo-msft to avoid further conflicts can you please help to review?

ycoheNvidia:
I thought there was an agreement to use #2698 instead, since original PR 2337 was reverted. 2698 includes both Secure upgrade original code and the fixes included in this PR



> @ycoheNvidia can you please update the conflicts? @qiluo-msft to avoid further conflicts can you please help to review?

This PR is irrelevant without 2337, can we move with https://github.com/sonic-net/sonic-utilities/pull/2698 instead? it has both original secure upgrade code and the fixes introduced here

ycoheNvidia:
This solution is part of a new PR https://github.com/sonic-net/sonic-utilities/pull/2698

PR #2657, opened 2023-02-07
[clear] Replace shell=True, replace exit() by sys.exit()

maipbui:

#### What I did
Ref:
#### How I did it
#### How to verify it
Manual test, execute all `sonic-clear` commands
Add UT


clear/main.py
```diff
@@ -79,16 +79,14 @@ def get_command(self, ctx, cmd_name):
 # location (configdb?), so that we prevent the continous execution of this
 # bash oneliner. To be revisited once routing-stack info is tracked somewhere.
 def get_routing_stack():
-    command = "sudo docker ps | grep bgp | awk '{print$2}' | cut -d'-' -f3 | cut -d':' -f1"
+    cmd0 = ["sudo", "docker", "ps"]
+    cmd1 = ["grep", "bgp"]
+    cmd2 = ["awk", '{print$2}']
+    cmd3 = ["cut", "-d", '-', "-f3"]
```
qiluo-msft, line 85: >"-d", '-' [](http://example.com/codeflow?start=19&length=9)

Originally it is one arg.

clear/main.py
```diff
@@ -79,16 +79,14 @@ def get_command(self, ctx, cmd_name):
...
+    cmd4 = ["cut", "-d", ':', "-f1"]
```
qiluo-msft, line 86: >"-d", ':' [](http://example.com/codeflow?start=19&length=9)

Originally it is one arg.

clear/main.py
maipbui: the result is the same as original behavior, but I did change to one arg.

qiluo-msft: Just wandering, is there any diff between `-d-` and `-d'-'` ?

maipbui: if use direct command in linux, then no difference. But in this case `cmd3 = ["cut", "-d'-'", "-f3"]` will lead to an error, cut: the delimiter must be a single character. So we need to use either `cmd3 = ["cut", "-d-", "-f3"]` or `cmd3 = ["cut", "-d", '-', "-f3"]`

qiluo-msft: Make sense. Bash handles the quotation marks. `shell=False` bypassed the shell.

qiluo-msft (approved):

clear/bgp_frr_v6.py
```diff
@@ -69,22 +69,22 @@ def soft_in(ipaddress):
...
+        command = ['sudo', 'vtysh', '-c', "clear bgp ipv6 {} soft in".format(ipaddress)]
     else:
-        command = 'sudo vtysh -c "clear bgp ipv6 * soft in"'
+        command = ['sudo', 'vtysh', '-c', "clear bgp ipv6 * soft in"]
     run_command(command)
 
 
 # 'soft all' subcommand
-@neighbor.command('all')
+@soft.command('all')
```
qiluo-msft, line 79: >soft [](http://example.com/codeflow?start=1&length=4)

Thanks for the bug fix!

qiluo-msft (approved):

qiluo-msft (approved):

PR #2658, opened 2023-02-07
[config] replace shell=True and replace exit() by sys.exit()

maipbui:

#### What I did
Ref:
#### How I did it

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -245,8 +245,8 @@ def _get_device_type():
     TODO: move to sonic-py-common
     """
 
-    command = "{} -m -v DEVICE_METADATA.localhost.type".format(SONIC_CFGGEN_PATH)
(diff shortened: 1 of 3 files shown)


qiluo-msft (dismissed):

nazariig (dismissed):

config/main.py
```diff
@@ -2913,12 +2913,12 @@ def warm_restart_enable(ctx, module):
 @warm_restart.command('disable')
 @click.argument('module', metavar='<module>', default='system', required=False)
 @click.pass_context
-def warm_restart_enable(ctx, module):
+def warm_restart_disable(ctx, module):
```
qiluo-msft, line 2916: >warm_restart_disable [](http://example.com/codeflow?start=4&length=20)

Thanks you for fixing code bug!

qiluo-msft (approved):

PR #2662, opened 2023-02-08
[debug/undebug] replace shell=True

maipbui:

#### What I did
#### How I did it
#### How to verify it
Manual test, execute all debug/undebug bgp/zebra commands
Add UT


debug/main.py
```diff
@@ -1,9 +1,13 @@
 import click
 import subprocess
+from shlex import join
+
+vtysh_cmd = ['sudo', 'vtysh', '-c']
```
qiluo-msft, line 5: >vtysh_cmd [](http://example.com/codeflow?start=0&length=9)

Move this line to the part which it is used.

debug/main.py
```diff
@@ -35,89 +40,86 @@ def bgp():
...
     @bgp.command()
     @click.argument('additional', type=click.Choice(['segment']), required=False)
     def as4(additional):
         """BGP AS4 actions"""
-        command = 'sudo vtysh -c "debug bgp as4'
+        command = vtysh_cmd + [debug_bgp + "as4"]
         if additional is not None:
-            command += " segment"
-        command += '"'
+            command = vtysh_cmd + [debug_bgp + "as4 segment"]
```
qiluo-msft, line 52: >command = vtysh_cmd + [debug_bgp + "as4 segment"] [](http://example.com/codeflow?start=12&length=49)

You may consider unify this line with above, most of components are the same.

debug/main.py
```diff
@@ -35,89 +40,86 @@ def bgp():
...
     @click.argument('prefix_or_iface', required=False)
     def keepalives(prefix_or_iface):
         """BGP Neighbor Keepalives"""
-        command = 'sudo vtysh -c "debug bgp keepalives'
+        bgp_cmd = [debug_bgp + "keepalives"]
+        command = vtysh_cmd + bgp_cmd
         if prefix_or_iface is not None:
-            command += " " + prefix_or_iface
-        command += '"'
+            command = vtysh_cmd + bgp_cmd + [prefix_or_iface]
```
qiluo-msft, line 69: >command [](http://example.com/codeflow?start=12&length=7)

You may consider unify this line with above, most of components are the same.

debug/main.py
```diff
@@ -21,8 +25,9 @@ def cli():
     """SONiC command line - 'debug' command"""
     pass
 
-
-p = subprocess.check_output(["sudo vtysh -c 'show version'"], shell=True, text=True)
+debug_bgp = "debug bgp "
```
qiluo-msft, line 28: >debug_bgp [](http://example.com/codeflow?start=0&length=9)

Most of "debug bgp " is used in static string, if you extract, you need use `+` to concat. Is it more readable to keep original long string?

debug/main.py
```diff
@@ -21,8 +25,9 @@ def cli():
...
+debug_zebra = "debug zebra "
```
qiluo-msft, line 29: >debug_zebra [](http://example.com/codeflow?start=0&length=11)

The same.

debug/main.py
qiluo-msft: How about:
`command[-1] += " segment"`

debug/main.py
qiluo-msft: The same.

undebug/main.py
```diff
@@ -35,89 +37,85 @@ def bgp():
...
-        command += '"'
+            command[-1] += " segment"
         run_command(command)
 
     @bgp.command()
     @click.argument('prefix', required=True)
     def bestpath(prefix):
         """BGP bestpath"""
-        command = 'sudo vtysh -c "no debug bgp bestpath %s"' % prefix
+        command = ["sudo", "vtysh", "-c", "no debug bgp bestpath %s" % prefix]
```
qiluo-msft, line 56: >prefix [](http://example.com/codeflow?start=71&length=6)

Need to validate prefix. Suggest checking it only contains numbers, alphabets, `. : /`

qiluo-msft: This comment is still applicable.

undebug/main.py
```diff
@@ -35,89 +39,87 @@ def bgp():
...
     @bgp.command()
     @click.argument('prefix', required=False)
     def zebra(prefix):
         """BGP Zebra messages"""
-        command = 'sudo vtysh -c "no debug bgp zebra'
+        bgp_cmd = "no debug bgp zebra"
         if prefix is not None:
-            command += " prefix " + prefix
-        command += '"'
+            bgp_cmd += ' prefix ' + prefix
```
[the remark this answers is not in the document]
qiluo-msft: The pattern string seems duplicated in multiple files. Could you define a string as shared var?

qiluo-msft (approved):

PR #2663, opened 2023-02-08
[fwutil] replace shell=True

maipbui:

#### What I did
#### How I did it
#### How to verify it
Pass UT
Manual test

### DIFF
--- a/fwutil/lib.py
+++ b/fwutil/lib.py
@@ -236,7 +236,7 @@ class SquashFs(object):
     def get_current_image(self):
         cmd0 = ["sonic-installer", "list"]
         cmd1 = ["grep", 'Current: ']
-        cmd2 = ["cut", "-f2", "-d", ' ']
+        cmd2 = ["cut", "-f2", "-d "]
         output = check_output_pipe(cmd0, cmd1, cmd2)
 
         return output.rstrip(NEWLINE)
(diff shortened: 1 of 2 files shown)


qiluo-msft (approved):

fwutil/lib.py
```diff
@@ -233,14 +234,18 @@ def __init__(self):
...
+        output = check_output_pipe(cmd0, cmd1, cmd2)
 
         return output.rstrip(NEWLINE)
 
     def get_next_image(self):
-        cmd = "sonic-installer list | grep 'Next: ' | cut -f2 -d' '"
-        output = subprocess.check_output(cmd, stderr=subprocess.STDOUT, shell=True, text=True)
+        cmd0 = ["sonic-installer", "list"]
+        cmd1 = ["grep", 'Next: ']
+        cmd2 = ["cut", "-f2", "-d", ' ']
```
qiluo-msft, line 247: >"-d", ' ' [](http://example.com/codeflow?start=30&length=9)

Original it's one arg, right?

fwutil/lib.py
```diff
@@ -233,14 +234,18 @@ def __init__(self):
         self.overlay_mountpoint = self.OVERLAY_MOUNTPOINT_TEMPLATE.format(image_stem)
 
     def get_current_image(self):
-        cmd = "sonic-installer list | grep 'Current: ' | cut -f2 -d' '"
-        output = subprocess.check_output(cmd, stderr=subprocess.STDOUT, shell=True, text=True)
+        cmd0 = ["sonic-installer", "list"]
+        cmd1 = ["grep", 'Current: ']
+        cmd2 = ["cut", "-f2", "-d", ' ']
```
qiluo-msft, line 239: >"-d", ' ' [](http://example.com/codeflow?start=30&length=9)

Original it's one arg, right?

qiluo-msft (dismissed):

qiluo-msft (approved):

PR #2664, opened 2023-02-08
replace shell=True, replace xml, and replace exit()

maipbui:

#### What I did
The [xml.etree.ElementTree](https://docs.python.org/3/library/xml.etree.elementtree.html#module-xml.etree.ElementTree) module is not secure against maliciously constructed data.
Ref:
#### How I did it
Remove xml. Use [lxml](https://pypi.org/project/lxml/) XML parsers package that prevent potentially malicious operation.
#### How to verify it
Pass UT
Manual test


scripts/boot_part
```diff
@@ -76,16 +76,25 @@ def set_boot_partition(blkdev, index):
...
         ## Set GRUB bootable
-        out = runcmd('sudo grub-install --boot-directory="{0}" --recheck "{1}"'.format(mntpath, blkdev))
+        out = runcmd(['sudo', 'grub-install', '--boot-directory='+mntpath, "--recheck", blkdev])
         return out is not None
     finally:
         ## Cleanup
-        out = runcmd('sudo fuser -km {0} || sudo umount {0}'.format(mntpath))
-        logger.info('fuser out={0}'.format(out))
+        cmd1 = ['sudo', 'fuser', '-km', mntpath]
+        rc, _ = getstatusoutput_noshell(cmd1)
```
qiluo-msft, line 88: >rc [](http://example.com/codeflow?start=8&length=2)

Use `rc1` `rc2` accordingly, do not reuse var.

qiluo-msft, line 88: >_ [](http://example.com/codeflow?start=12&length=1)

Originally the out1 is also captured and logged, however the original log message is misleading.

You can logging both out1 and out2, no matter success or failure.

scripts/ipintutil
```diff
@@ -72,13 +72,12 @@ def get_if_admin_state(iface, namespace):
     """
     Given an interface name, return its admin state reported by the kernel
     """
-    cmd = "cat /sys/class/net/{0}/flags".format(iface)
+    cmd = ["cat", "/sys/class/net/{0}/flags".format(iface)]
     if namespace != constants.DEFAULT_NAMESPACE:
-        cmd = "sudo ip netns exec {} {}".format(namespace, cmd)
+        cmd = ["sudo", "ip", "netns", "exec", namespace, "cat", "/sys/class/net/{0}/flags".format(iface)]
```
qiluo-msft, line 77: >"cat", "/sys/class/net/{0}/flags [](http://example.com/codeflow?start=57&length=32)

You can reuse old cmd.

scripts/ipintutil
```diff
@@ -105,13 +104,12 @@ def get_if_oper_state(iface, namespace):
     """
     Given an interface name, return its oper state reported by the kernel.
     """
-    cmd = "cat /sys/class/net/{0}/carrier".format(iface)
+    cmd = ["cat", "/sys/class/net/{0}/carrier".format(iface)]
     if namespace != constants.DEFAULT_NAMESPACE:
-        cmd = "sudo ip netns exec {} {}".format(namespace, cmd)
+        cmd = ["sudo", "ip", "netns", "exec", namespace, "cat", "/sys/class/net/{0}/carrier".format(iface)]
```
qiluo-msft, line 109: >"cat", "/sys/class/net/{0}/carrier [](http://example.com/codeflow?start=57&length=34)

You can reuse old cmd.

scripts/lldpshow
```diff
@@ -80,9 +80,9 @@ class Lldpshow(object):
             lldp_interface_list = lldp_port if lldp_port is not None else self.lldp_interface[lldp_instace_num]
             # In detail mode we will pass interface list (only front ports) and get O/P as plain text
             # and in table format we will get xml output
-            lldp_cmd = 'sudo docker exec -i lldp{} lldpctl '.format(self.lldp_instance[lldp_instace_num]) + (
-                '-f xml' if not lldp_detail_info else lldp_interface_list)
-            p = subprocess.Popen(lldp_cmd, stdout=subprocess.PIPE, shell=True, text=True)
+            lldp_list = ['-f', 'xml'] if not lldp_detail_info else [lldp_interface_list]
```
qiluo-msft, line 83: >lldp_list [](http://example.com/codeflow?start=12&length=9)

-> lldpctl_args

scripts/lldpshow
```diff
@@ -121,7 +121,7 @@ class Lldpshow(object):
         if lldp_detail_info:
             return
         for lldpraw in self.lldpraw:
-            neis = ET.fromstring(lldpraw)
```
qiluo-msft, line 124: >lldpraw [](http://example.com/codeflow?start=33&length=7)

lldpraw.encode()

scripts/storyteller
```diff
@@ -28,31 +30,35 @@ reference_file = '/tmp/storyteller_time_reference'
...
 
-    return ' '.join(x for x in options)
+    return options
 
 
 def find_log(logpath, log, regex, after=0, before=0, context=0, field=0):
     options = build_options(after, before, context)
     if field <= 0:
-        cmd = 'find -L {}/{}* -newer {} | xargs ls -rt | xargs zgrep -a {} "{}"'.format(logpath, log, reference_file, options, regex)
+        cmd0 = ['find', '-L'] + glob.glob("{}/{}*".format(logpath, log)) + ["-newer", reference_file]
```
qiluo-msft, line 53: >glob.glob [](http://example.com/codeflow?start=32&length=9)

No need to glob?

scripts/storyteller
```diff
@@ -28,31 +30,35 @@ reference_file = '/tmp/storyteller_time_reference'
...
+        cmd1 = ["xargs", "ls", "-rt"]
+        cmd2 = ["xargs", "zgrep", "-a"] + options + [regex]
     else:
-        cmd = 'find -L {0}/{1}* -newer {2} | sort -rn -t . -k {3},{3} | xargs zgrep -a {4} "{5}"'.format(logpath, log, reference_file, field, options, regex)
+        cmd0 = ['find', '-L'] + glob.glob('{0}/{1}*'.format(logpath, log)) + ['-newer', reference_file]
```
qiluo-msft, line 57: >glob.glob [](http://example.com/codeflow?start=32&length=9)

No need to glob?

scripts/storyteller
maipbui: need to glob, otherwise the program's failed: `find: ‘/var/log/syslog*’: No such file or directory`

scripts/lldpshow
maipbui: fixed, thank you!

scripts/storyteller
qiluo-msft: I am worried about super long cmd args list. How about `find /var/log/ -name 'syslog*'` which does not require globing?

maipbui: I observed there's not much significant difference b/w the perf of the original code and the new code using glob even super long list. But your approach also works, fixed the code not to use glob.

qiluo-msft: Thanks for the fixing! To clarify, I am worried about super long cmd args, which may above some shell limit. I am not talking about performance.

qiluo-msft (approved):

qiluo-msft:
@liat-grozovik  @madhanmellanox Could you help review scripts/sonic_sku_create.py ?

qiluo-msft (approved):

scripts/lldpshow
```diff
@@ -80,9 +80,11 @@ class Lldpshow(object):
             lldp_interface_list = lldp_port if lldp_port is not None else self.lldp_interface[lldp_instace_num]
             # In detail mode we will pass interface list (only front ports) and get O/P as plain text
             # and in table format we will get xml output
-            lldp_cmd = 'sudo docker exec -i lldp{} lldpctl '.format(self.lldp_instance[lldp_instace_num]) + (
-                '-f xml' if not lldp_detail_info else lldp_interface_list)
-            p = subprocess.Popen(lldp_cmd, stdout=subprocess.PIPE, shell=True, text=True)
+            lldp_args = ['-f', 'xml'] if not lldp_detail_info else [lldp_interface_list]
```
qiluo-msft, line 83: >lldp_interface_list [](http://example.com/codeflow?start=68&length=19)

you can check `lldp_interface_list == ''` and assign different values to lldp_args.

So this line will be come a if-elsif-else block.

qiluo-msft (approved):

PR #2665, opened 2023-02-08
Add support for platform syncd pre shutdown plugin

anamehra:
#### Why I did it

Vendor platform may require running platform specific pre-shutdown routine before shutting down the syncd process which runs the SAI and vendor sdk instance.

#### How I did it
Added a platform script hook which will be executed if the plugin script is provided by the platform in device//plugins/

#### How to verify it
Add platform plugin script and run reboot


anamehra:
@abdosi , FYI. Thanks

anamehra:
Not required, will use the platform_reboot_pre_check

PR #2666, opened 2023-02-09
[show] replace shell=True, replace xml by lxml, replace exit by sys.exit

maipbui:

#### What I did
Ref:
#### How I did it
#### How to verify it
Pass UT
Manual test


show/main.py
```diff
@@ -951,10 +980,9 @@ def aging_time(ctx):
 @click.option('--verbose', is_flag=True, help="Enable verbose output")
 def route_map(route_map_name, verbose):
     """show route-map"""
-    cmd = 'sudo {} -c "show route-map'.format(constants.RVTYSH_COMMAND)
+    cmd = ['sudo', constants.RVTYSH_COMMAND, '-c', 'show route-map']
     if route_map_name is not None:
-        cmd += ' {}'.format(route_map_name)
-    cmd += '"'
+        cmd += [str(route_map_name)]
```
qiluo-msft, line 985: >route_map_name [](http://example.com/codeflow?start=20&length=14)

It should be appended to embedded `show route-map` command.

qiluo-msft: Could you make sure this is tested?

show/main.py
```diff
@@ -1042,10 +1070,9 @@ def route(args, namespace, display, verbose):
 @click.option('--verbose', is_flag=True, help="Enable verbose output")
 def prefix_list(prefix_list_name, verbose):
     """show ip prefix-list"""
-    cmd = 'sudo {} -c "show ip prefix-list'.format(constants.RVTYSH_COMMAND)
+    cmd = ['sudo', constants.RVTYSH_COMMAND, '-c', 'show ip prefix-list']
     if prefix_list_name is not None:
-        cmd += ' {}'.format(prefix_list_name)
-    cmd += '"'
+        cmd += [str(prefix_list_name)]
```
qiluo-msft, line 1075: >prefix_list_name [](http://example.com/codeflow?start=20&length=16)

append to embedded command

show/main.py
```diff
@@ -1090,10 +1117,9 @@ def ipv6():
 @click.option('--verbose', is_flag=True, help="Enable verbose output")
 def prefix_list(prefix_list_name, verbose):
     """show ip prefix-list"""
-    cmd = 'sudo {} -c "show ipv6 prefix-list'.format(constants.RVTYSH_COMMAND)
+    cmd = ['sudo', constants.RVTYSH_COMMAND, '-c', 'show ipv6 prefix-list']
     if prefix_list_name is not None:
-        cmd += ' {}'.format(prefix_list_name)
-    cmd += '"'
+        cmd += [str(prefix_list_name)]
```
qiluo-msft, line 1122: >prefix_list_name [](http://example.com/codeflow?start=20&length=16)

append to embedded command

show/main.py
```diff
@@ -1257,26 +1283,34 @@ def table(verbose):
...
         if process is not None:
-            cmd += " | grep '{}'".format(process)
+            cmd1 = ['grep', str(process)]
 
         if lines is not None:
-            cmd += " | tail -{}".format(lines)
+            cmd2 = ['tail', "-" + str(lines)]
 
-        run_command(cmd, display_cmd=verbose)
+        if cmd1 is None and cmd2 is None:
```
qiluo-msft, line 1306: >cmd1 [](http://example.com/codeflow?start=11&length=4)

You implemented a 2x2 conditions with 4 branches. This is not scalable if number of cmd increase. For example 10 cmd.

You can construct an array first, and feed the expanded array to function call. ref: [https://stackoverflow.com/a/7745986/2514803](https://stackoverflow.com/a/7745986/2514803)

qiluo-msft: or cmdn could be defined as array elements from the beginning.

show/main.py
```diff
@@ -1714,16 +1748,20 @@ def startupconfiguration():
 @click.option('--verbose', is_flag=True, help="Enable verbose output")
 def bgp(verbose):
     """Show BGP startup configuration"""
-    cmd = "sudo docker ps | grep bgp | awk '{print$2}' | cut -d'-' -f3 | cut -d':' -f1"
-    proc = subprocess.Popen(cmd, stdout=subprocess.PIPE, shell=True, text=True)
-    result = proc.stdout.read().rstrip()
+    cmd0 = ['sudo', 'docker', 'ps']
+    cmd1 = ['grep', 'bgp']
+    cmd2 = ['awk', '{print$2}']
+    cmd3 = ['cut', '-d', '-', '-f3']
```
qiluo-msft, line 1754: >'-d', '-' [](http://example.com/codeflow?start=19&length=9)

originally it is one arg.

show/main.py
```diff
@@ -1714,16 +1748,20 @@ def startupconfiguration():
...
+    cmd4 = ['cut', '-d', ':', "-f1"]
```
qiluo-msft, line 1755: >'-d', ':' [](http://example.com/codeflow?start=19&length=9)

originally it is one arg.

show/plugins/barefoot.py
```diff
@@ -37,9 +39,10 @@ def profile():
     
     # Print profile list
     click.echo('Available profile(s):')
-    subprocess.run('docker exec -it syncd find /opt/bfn -mindepth 1 '
-        r'-maxdepth 1 -type d -name install_\*_profile ' + opts + '| sed '
-        r's%/opt/bfn/install_\\\(.\*\\\)_profile%\\1%', shell=True)
+    cmd0 = ['docker', 'exec', '-it', 'syncd', 'find', '/opt/bfn', '-mindepth', '1',\
+            r'-maxdepth', '1', '-type', 'd', '-name', 'install_\*_profile', r'{}' % opts]
```
qiluo-msft, line 43: >r'{}' % opts [](http://example.com/codeflow?start=76&length=12)

What the diff if using `opts` directly?

show/plugins/barefoot.py
```diff
@@ -4,6 +4,7 @@
 import json
 import subprocess
 from sonic_py_common import device_info
+from sonic_py_common.general import getstatusoutput_noshell_pipe, check_output_pipe
 
 @click.group()
 def barefoot():
```
qiluo-msft, line 10: >barefoot [](http://example.com/codeflow?start=4&length=8)

This file need to reviewed by barefoot. Suggest further split into a smaller PR.

qiluo-msft:
> # Specific command-line utility for Mellanox platform

This file need to reviewed by Mellanox. Suggest further split into a smaller PR.

---
Refers to: show/plugins/mlnx.py:22 in 4d96b6e. [](commit_id = 4d96b6e5ae778fa7c846b8b3ab6d5c9b93316d46, deletion_comment = False)

show/main.py
```diff
@@ -130,6 +133,32 @@ def run_command(command, display_cmd=False, return_cmd=False):
     if rc != 0:
         sys.exit(rc)
 
+def run_command_pipe(*args, display_cmd=False, return_cmd=False):
```
qiluo-msft, line 136: >run_command_pipe [](http://example.com/codeflow?start=4&length=16)

Possible to unify `run_command` and `run_command_pipe` into one function `run_command`?

Or treat `run_command` as a special case, call the other with only one command in args, so most of the code are reused.

qiluo-msft:
>     clicommon.run_command(cmd + options, display_cmd=verbose)

Use string array instead of string for `run_command`.

---
Refers to: show/platform.py:111 in 4d96b6e. [](commit_id = 4d96b6e5ae778fa7c846b8b3ab6d5c9b93316d46, deletion_comment = False)

show/main.py
maipbui: fixed, thanks!

show/plugins/barefoot.py
maipbui: `r''` is to bypass escaping special characters. Using `opts` directly will escape special characters.

show/plugins/barefoot.py
maipbui: will do

maipbui:
> > ```
> > clicommon.run_command(cmd + options, display_cmd=verbose)
> > ```
> 
> Use string array instead of string for `run_command`.
> 
> Refers to: show/platform.py:111 in [4d96b6e](https://github.com/sonic-net/sonic-utilities/commit/4d96b6e5ae778fa7c846b8b3ab6d5c9b93316d46). [](commit_id = [4d96b6e](https://github.com/sonic-net/sonic-utilities/commit/4d96b6e5ae778fa7c846b8b3ab6d5c9b93316d46), deletion_comment = False)

this will be covered in another PR. I'll draft another PR.

show/main.py
qiluo-msft: To simplify:
`cmd[-1] += ' {}'.format(route_map_name)`

show/main.py
qiluo-msft: the same.

show/main.py
qiluo-msft: the same

show/plugins/barefoot.py
qiluo-msft: Could you be more specific?, `r` is applied to `{}` string only.

maipbui: for example, `r'\n'` will treated the string containing 2 characters: backslash `\` and letter `n`. But '\n' will be treated as a newline character.

qiluo-msft: >r [](http://example.com/codeflow?start=12&length=1)

This `r` is not needed.

qiluo-msft (approved):

show/main.py
```diff
@@ -1252,17 +1255,20 @@ def table(verbose):
...
 @cli.command()
 @click.argument('process', required=False)
-@click.option('-l', '--lines')
+@click.option('-l', '--lines', type=int)
 @click.option('-f', '--follow', is_flag=True)
 @click.option('--verbose', is_flag=True, help="Enable verbose output")
 def logging(process, lines, follow, verbose):
     """Show system log"""
+    if process and not re.match(r'^[a-zA-Z0-9\s]+$', process):
+        sys.exit('Process contains only number, alphabet, and whitespace.')
```
qiluo-msft, line 1264: >exit [](http://example.com/codeflow?start=12&length=4)

What is the reason to add this check?

qiluo-msft: Offline discussed, this check is not accurate. Let's revert back, and please use another future PR to focus on this fix.

qiluo-msft (approved):

qiluo-msft:
> class TestShowPlatform(object):

Thanks for cleaning up the codebase!

---
Refers to: tests/show_test.py:498 in a20435a. [](commit_id = a20435a753bf3db51995837010c862e585026536, deletion_comment = True)

qiluo-msft (approved):

PR #2676, opened 2023-02-15
[generate_dump]: Enhance show techsupport for Marvell platform

pavannaregundi:
Change-Id: Ibb5b5a860833e8d7834a1d72756d070c61f2ab37

#### What I did
Added support for marvell sdk commands in "show techsupport". 

#### How I did it
Modified the generate_dump script, to add marvell sdk commands under a proper platform check.

#### How to verify it
Verified the change by running "show techsupport" command and check all the newly added marvel sdk commands are collecting the correct information.

### DIFF
--- a/scripts/generate_dump
+++ b/scripts/generate_dump
@@ -1158,6 +1158,67 @@ collect_mellanox_dfw_dumps() {
 }
 
 ###############################################################################
+# Runs a given marvellcmd command in all namesapces in case of multi ASIC platform
+# Globals:
+#  NUM_ASICS
+# Arguments:
+#  cmd: The command to run. Make sure that arguments with spaces have quotes
+#  filename: the filename to save the output as in $BASE/dump
+#  do_gzip: (OPTIONAL) true or false. Should the output be gzipped
+# Returns:
+#  None
+###############################################################################
+save_marvellcmd() {
+    trap 'handle_error $? $LINENO' ERR
+
+    mkdir -p $LOGDIR/sdkdump
+    local cmd="docker exec syncd mrvlcmd -c \"$1\""
+    save_cmd "$cmd" "sdkdump/$2"
+}
+
+###############################################################################
+# Collect Marvell specific information
+# Globals:
+#  None
+# Arguments:
+#  None
+# Returns:
+#  None
+###############################################################################
+collect_marvell() {
+    trap 'handle_error $? $LINENO' ERR
+
+    save_marvellcmd "show version" "CPSS_version"
+    save_marvellcmd "debug-mode;xps-api call xpsDataIntegrityDumpSerInfo all" "SER_table"
+    save_marvellcmd "show traffic cpu rx statistic device 0" "CPU_stat"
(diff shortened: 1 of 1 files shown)


radha-danda:
@lguohan, can you help review and merge the PR please?

lguohan:
@qiluo-msft can you help?

scripts/generate_dump
```diff
@@ -1157,6 +1157,73 @@ collect_mellanox_dfw_dumps() {
...
+#  do_gzip: (OPTIONAL) true or false. Should the output be gzipped
+# Returns:
+#  None
+###############################################################################
+save_marvellcmd() {
+    trap 'handle_error $? $LINENO' ERR
+
+    mkdir -p $LOGDIR/sdkdump
+    if [ "$1" = "collect SER" ];then
+        local cmd="sudo cat /var/log/syslog | grep \"Chip MARVELL occurs\""
```
qiluo-msft, line 1176: >/var/log/syslog [](http://example.com/codeflow?start=28&length=15)

syslog file will be rotated at any time. So this grep may not work. Since all the syslog files are already dumped, why you need grep?

scripts/generate_dump
```diff
@@ -1157,6 +1157,73 @@ collect_mellanox_dfw_dumps() {
...
+    else
+        local cmd="~/mrvlcmd -c \"$1\""
```
qiluo-msft, line 1178: >~/ [](http://example.com/codeflow?start=19&length=2)

Do not assume current working directory. User could run from any directory.

scripts/generate_dump
pavannaregundi: Will changed this logic to collect logs from SDK.

scripts/generate_dump
pavannaregundi: Will change this, to run command directed using docker exec.

scripts/generate_dump
```diff
@@ -1157,6 +1157,67 @@ collect_mellanox_dfw_dumps() {
...
+#  cmd: The command to run. Make sure that arguments with spaces have quotes
+#  filename: the filename to save the output as in $BASE/dump
+#  do_gzip: (OPTIONAL) true or false. Should the output be gzipped
+# Returns:
+#  None
+###############################################################################
+save_marvellcmd() {
+    trap 'handle_error $? $LINENO' ERR
+
+    mkdir -p $LOGDIR/sdkdump
```
qiluo-msft, line 1174: >sdkdump [](http://example.com/codeflow?start=21&length=7)

Do you want to clean up this folder if there are old files, before you save new file?

pavannaregundi: Full path of the sdkdump will be something like below, where parenet directory(sonic_dump_sonic_XXXXXXX_XXXX) is always new each time we execute "show techsupport" command. So old file will not be present.
"/var/dump/sonic_dump_sonic_20230615_125930/dump/sdkdump/".

pavannaregundi:
@qiluo-msft Let me know if you have further comments.

qiluo-msft (approved):

PR #2686, opened 2023-02-20
Collect information for the issue #13791 asic_table.json is not rendered

stephenxs:
#### What I did

Collect information for issue [#13791](https://github.com/sonic-net/sonic-buildimage/issues/13791)
Besides redis server being temporarily inaccessible, another possible cause is that the template is not accessible.
So, copy the template into the dump folder in `show tech dump`

### DIFF
--- a/scripts/generate_dump
+++ b/scripts/generate_dump
@@ -1108,6 +1108,8 @@ collect_mellanox() {
       echo "HW Mgmt dump script $HW_DUMP_FILE does not exist"
     fi
 
+    # Temporarily collect information for debugging the issue that asic_table.json not rendered
+    save_file "/usr/share/sonic/templates/asic_table.j2" dump false
 }
 
 ###############################################################################


scripts/generate_dump
```diff
@@ -1108,6 +1108,8 @@ collect_mellanox() {
       echo "HW Mgmt dump script $HW_DUMP_FILE does not exist"
     fi
 
+    # Temporarily collect information for debugging the issue that asic_table.json not rendered
+    save_file "/usr/share/sonic/templates/asic_table.j2" dump false
```
qiluo-msft, line 1112: this is a static file, why collect?

stephenxs: regarding the issue [#13791](https://github.com/sonic-net/sonic-buildimage/issues/13791), one possible reason is that the template is not available when it is being rendered.
Even the probability is very slim, I would like to collect the template to rule out (or prove) it.
How do you think?

qiluo-msft: It is static file, and show tech already collect the version, so we could check the file in that image. There is no need to collect here.

stephenxs: ok. then closing it.

PR #2691, opened 2023-02-23
[db_migrator] Add missing attribute 'weight' to route entries in APPL DB

vaibhavhd:
#### What I did
Fixes: https://github.com/sonic-net/sonic-buildimage/issues/12625

#### How I did it

Check for missing attribute `weight` in APPLDB route entries. If found missing this attribute is added with empty value.

#### How to verify it
Verified on physical device. 201911 to 202205 upgrade worked fine.

### DIFF
--- a/scripts/db_migrator.py
+++ b/scripts/db_migrator.py
@@ -899,6 +899,17 @@ class DBMigrator():
         else:
             log.log_notice("Asic Type: {}, Hwsku: {}".format(self.asic_type, self.hwsku))
 
+        route_table = self.appDB.get_table("ROUTE_TABLE")
+        for route_prefix, route_attr in route_table.items():
+            if 'weight' not in route_attr:
+                if type(route_prefix) == tuple:
+                    # IPv6 route_prefix is returned from db as tuple
+                    route_key = "ROUTE_TABLE:" + ":".join(route_prefix)
+                else:
+                    # IPv4 route_prefix is returned from db as str
+                    route_key = "ROUTE_TABLE:{}".format(route_prefix)
+                self.appDB.set(self.appDB.APPL_DB, route_key, 'weight','')
+
     def migrate(self):
         version = self.get_version()
         log.log_info('Upgrading from version ' + version)
--- a/scripts/db_migrator.py
+++ b/scripts/db_migrator.py
@@ -579,6 +579,18 @@ class DBMigrator():
             self.configDB.set_entry('PORT_QOS_MAP', 'global', {"dscp_to_tc_map": dscp_to_tc_map_table_names[0]})
             log.log_info("Created entry for global DSCP_TO_TC_MAP {}".format(dscp_to_tc_map_table_names[0]))
 
+    def migrate_route_table_weights(self):
+        route_table = self.appDB.get_table("ROUTE_TABLE")
+        for route_prefix, route_attr in route_table.items():
+            if 'weight' not in route_attr:
+                if type(route_prefix) == tuple:
+                    # IPv6 route_prefix is returned from db as tuple
+                    route_key = "ROUTE_TABLE:" + ":".join(route_prefix)
+                else:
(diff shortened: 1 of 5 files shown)


scripts/db_migrator.py
```diff
@@ -899,6 +899,17 @@ def common_migration_ops(self):
...
+        route_table = self.appDB.get_table("ROUTE_TABLE")
+        for route_prefix, route_attr in route_table.items():
+            if 'weight' not in route_attr:
+                if type(route_prefix) == tuple:
+                    # IPv6 route_prefix is returned from db as tuple
+                    route_key = "ROUTE_TABLE:" + ":".join(route_prefix)
+                else:
+                    # IPv4 route_prefix is returned from db as str
+                    route_key = "ROUTE_TABLE:{}".format(route_prefix)
+                self.appDB.set(self.appDB.APPL_DB, route_key, 'weight','')
```
vaibhavhd, line 911: @prsunny please note that I have not initialized the values to `1.1.1.1`.  The failure was due to missing attributes, but not the value.

During reconciliation, fpmsyncd will still put the default (or the intended) value as the weight. Hardcoding the value to `1.1.1.1` can be thus avoided. Please let me know if you have any concerns.

yxieca: Isn't the default weight '1'?

vaibhavhd: The default value is indeed `1` - for every nexthop.

So, for example, the value of `weight` for 4 nexthops will be `1,1,1,1`.

Hardcoding here will require the logic to check how many nexthops are present in the route and then generate the default value based on that.

yxieca (approved):

scripts/db_migrator.py
```diff
@@ -899,6 +899,17 @@ def common_migration_ops(self):
         else:
             log.log_notice("Asic Type: {}, Hwsku: {}".format(self.asic_type, self.hwsku))
 
+        route_table = self.appDB.get_table("ROUTE_TABLE")
```
prsunny, line 902: can you move this to a dedicated function.

vaibhavhd: Sure, moved this logic to a method now.

scripts/db_migrator.py
```diff
@@ -899,6 +911,8 @@ def common_migration_ops(self):
         else:
             log.log_notice("Asic Type: {}, Hwsku: {}".format(self.asic_type, self.hwsku))
 
+        self.migrate_route_table_weights()
```
prsunny, line 914: since we have other functions as "`migrate_copp_table`", can we have a generic function like `migrate_route_table` so that a future changes can also be contained within this. What do you think?

vaibhavhd: Sure, that's a good idea. Updated now.

prsunny (approved):

scripts/db_migrator.py
```diff
@@ -579,6 +579,23 @@ def migrate_port_qos_map_global(self):
...
 
+    def migrate_route_table(self):
+        """
+        Handle route table migration. Migrations handled:
+        1. 'weight' attr in ROUTE object was introduced 202205 onwards.
+            Upgrade from older branch to 202205 will require this 'weight' attr to be added explicitly
+        """
+        route_table = self.appDB.get_table("ROUTE_TABLE")
+        for route_prefix, route_attr in route_table.items():
+            if 'weight' not in route_attr:
```
prsunny, line 590: Can you also check how the subnet routes looks like? IIRC, it may not have weights today.

vaibhavhd: This logic will still add the weight attr to these routes.
It did not create an issue though, as the bootup path took two diff directions: For all other routes which do not have nexthops - weight attribute added by my logic will be removed after reconciliation.

StormLiangMS:
@vaibhavhd cherry-pick conflict, pls help to create separate PR for 202211.

PR #2695, opened 2023-02-27
[Secure Boot] Fix issue with fast\warm boot on devices not supporting…

DavidZagury:
… UEFI

#### What I did
Fix issue that caused fast & warm reboot to not work if SB is disabled or not supported.

When SB was not enabled, we got an EXIT signal which caused the trap function to run and cancel the reboot.

#### How I did it
Change the way we check if SB is supported. The pervious check has thrown an error that the warm/fast script caught which cause it to run the trap. I use a different check that will not thorw an error.

#### How to verify it
Run warm or fast reboot

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -612,9 +612,9 @@ fi
 if is_secureboot && grep -q aboot_machine= /host/machine.conf; then
     load_aboot_secureboot_kernel
 else
-    # check if secure boot is enable in UEFI 
-    SECURE_UPGRADE_ENABLED=$(bootctl status 2>/dev/null | grep -c "Secure Boot: enabled")
-    if [ ${SECURE_UPGRADE_ENABLED} -eq 1 ]; then
+    # check if secure boot is enable in UEFI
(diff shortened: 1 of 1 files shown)


scripts/fast-reboot
```diff
@@ -612,9 +612,9 @@ fi
 if is_secureboot && grep -q aboot_machine= /host/machine.conf; then
     load_aboot_secureboot_kernel
 else
-    # check if secure boot is enable in UEFI 
-    SECURE_UPGRADE_ENABLED=$(bootctl status 2>/dev/null | grep -c "Secure Boot: enabled")
-    if [ ${SECURE_UPGRADE_ENABLED} -eq 1 ]; then
+    # check if secure boot is enable in UEFI
+    SECURE_UPGRADE_ENABLED=$(bootctl status 2>/dev/null)
```
qiluo-msft, line 616: > 2>/dev/null [](http://example.com/codeflow?start=43&length=12)

Do you really need redirect?

qiluo-msft (approved):

davidpil2002:
see that your fix passes the CI now, pls can you merge and update the pointer?
@DavidZagury

PR #2697, opened 2023-02-28
[GCU] Add vlanintf-validator

wen587:
#### What I did
Fix the bug of GCU vlan interface modification. It should call `ip neigh flush dev`  after removing interface ip. 
The fix is basically following config CLI's tradition.
#### How I did it
Add vlanintf service validator to check if extra step of `ip neigh flush` is needed.
#### How to verify it
GCU E2E test in dualtor testbed.


generic_config_updater/services_validator.py
```diff
@@ -101,3 +101,20 @@ def caclmgrd_validator(old_config, upd_config, keys):
...
+
+    old_keys = [ tuple(key.split("|"))
+            for key in old_vlan_intf if len(key.split("|")) == 2 ]
+    upd_keys = [ tuple(key.split("|"))
+            for key in upd_vlan_intf if len(key.split("|")) == 2 ]
+
+    deleted_keys = list(set(old_keys) - set(upd_keys))
+    for key in deleted_keys:
+        iface, iface_ip = key
+        rc = os.system(f"ip neigh flush dev {iface} {iface_ip}")
```
qiluo-msft, line 117: >system [](http://example.com/codeflow?start=16&length=6)

Add unit test if this is a bug fix.

generic_config_updater/services_validator.py
```diff
@@ -101,3 +101,20 @@ def caclmgrd_validator(old_config, upd_config, keys):
 
 def ntp_validator(old_config, upd_config, keys):
     return _service_restart("ntp-config")
+
+def vlanintf_validator(old_config, upd_config, keys):
+    old_vlan_intf = old_config.get("VLAN_INTERFACE", {})
+    upd_vlan_intf = upd_config.get("VLAN_INTERFACE", {})
+
+    old_keys = [ tuple(key.split("|"))
```
qiluo-msft, line 109: >old_keys [](http://example.com/codeflow?start=4&length=8)

Could you add some comments into each code block? For example, give some sample of old_keys and upd_keys, and explain your intention.

isabelmsft (approved):

isabelmsft:
The code changes look correct to me. But I noticed `services_validator.py` has extended beyond its original purpose to restart SONiC services when specific changes are being made. Maybe we can consider updating the naming convention if we keep the code structured like this.

qiluo-msft (approved):

PR #2698, opened 2023-02-28
Secure upgrade v2

ycoheNvidia:
What I did
Added support for secure upgrade

How I did it
It includes image signing during build (in sonic buildimage repo) and verification during image install (in sonic-utilities).
HLD can be found in the following PR: https://github.com/sonic-net/SONiC/pull/1024

How to verify it
Feature is used to allow image was not modified since built from vendor. During installation, image can be verified with a signature attached to it.
In order for image verification - image must be signed - need to provide signing key and certificate (paths in SECURE_UPGRADE_DEV_SIGNING_KEY and SECURE_UPGRADE_DEV_SIGNING_CERT in rules/config) during build , and during image install, need to enable secure boot flag in bios, and signing_certificate should be available in bios.

Feature dependencies
In order for this feature to work smoothly, need to have secure boot feature implemented as well.
The Secure boot feature will be merged in the near future.

sonic-buildimage PR: https://github.com/sonic-net/sonic-buildimage/pull/11862


Yarden-Z:
@qiluo-msft Can you please review and approve this PR?
Containing the original secure upgrade item and a fix for the bug found in aboot.

liat-grozovik:
@qiluo-msft , @Yarden-Z  @xumia please help to review this PR. this one must go into 202211 as well .

Yarden-Z (approved):

zhangyanzhao:
moved to 202305 release

liat-grozovik:
@qiluo-msft kindly reminder to review and provide feedback

sonic_installer/bootloader/bootloader.py
```diff
@@ -75,6 +75,10 @@ def supports_package_migration(self, image):
         """tells if the image supports package migration"""
         return True
 
+    def verify_image_sign(self, image_path):
```
qiluo-msft, line 78: >verify_image_sign [](http://example.com/codeflow?start=8&length=17)

A unsecure normal image is considered not signed image, and should return False.

ycoheNvidia: We want to verify image signature only of secure boot flag is enabled. This flow was already merged in previous PR  https://github.com/sonic-net/sonic-utilities/pull/2337 (that was reverted even though it had a fix ready in a day) and described and approved in the HLD as well (https://github.com/sonic-net/SONiC/pull/1024)

ycoheNvidia: @qiluo-msft please add your comments/approval

qiluo-msft: After checking HLD, my comment is still applicable.

qiluo-msft: If you mean "image verification not implemented", you should either throw NotImplmented exception or return a reasonable value to indicate the error. Silently "continue without verification" does not make sense.

ycoheNvidia: Modified the code as suggested, please review @qiluo-msft

zhangyanzhao:
@xumia can you please help to review this PR? Thanks.

liat-grozovik:
@qiluo-msft kindly reminder to follow up on changes based on your inputs.

sonic_installer/main.py
```diff
@@ -576,6 +577,16 @@ def install(url, force, skip_platform_check=False, skip_migration=False, skip_pa
                 "Aborting...", LOG_ERR)
             raise click.Abort()
 
+        # Calling verification script by default - signature will be checked if enabled in bios
+        echo_and_log("Verifing image {} signature...".format(binary_image_version))
+        try:
+            if not bootloader.verify_image_sign(image_path):
```
xumia, line 583: If the force option is enabled, can we skip the verify_image_sign, as descirption in line 514? The check logic should be the same behavior as line 567 (secure bot), right?

ycoheNvidia: @xumia we don't want to allow force option for this feature, it will reduce its security properties. This remark was addressed in the previous merged and reverted PR https://github.com/sonic-net/sonic-utilities/pull/2337 (at least as I can remember)

sonic_installer/main.py
```diff
@@ -576,6 +577,16 @@ def install(url, force, skip_platform_check=False, skip_migration=False, skip_pa
                 "Aborting...", LOG_ERR)
             raise click.Abort()
 
+        # Calling verification script by default - signature will be checked if enabled in bios
```
qiluo-msft, line 580: >Calling verification script by default [](http://example.com/codeflow?start=10&length=38)

You mentioned "We want to verify image signature only of secure boot flag is enabled." but the code is verifying image signature always.

Still silently "continue without verification" does not make sense.

sonic_installer/main.py
qiluo-msft: I agree with @xumia.
The security is protected by the secure boot process, not by the installer. If the user intention is to install it anyway, no matter it will boot or not, we should respect the intetion, right?

Yarden-Z: But this is not what was defined in the HLD for this feature.
If we would like to make a feature change for this item - it should go as a separate request.

scripts/verify_image_sign.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/sh
+image_file="${1}"
+cms_sig_file="sig.cms"
+lines_for_lookup=50
+SECURE_UPGRADE_ENABLED=0
+DIR="$(dirname "$0")"
+if [ -d "/sys/firmware/efi/efivars" ]; then
```
ycoheNvidia, line 7: call this specifically from main.py to decided if image verification is required

qiluo-msft (approved):

ycoheNvidia:
@qiluo-msft @liat-grozovik fixed minor print and issues - CI tests pass

qiluo-msft (approved):

StormLiangMS:
@DavidZagury  still got some question.

1. for device which support this secure upgrade
a. disabled, to boot unsecure image would be ok.   
b. enabled, unsecure image will be failed to boot. 
2. for device which doesn't support this secure upgrade feature.
a. unsecure image would be ok to boot/install.

DavidZagury:
> @DavidZagury  still got some question.
> 
> 1. for device which support this secure upgrade
> a. disabled, to boot unsecure image would be ok.   
> b. enabled, unsecure image will be failed to boot. 
> 2. for device which doesn't support this secure upgrade feature.
> a. unsecure image would be ok to boot/install.  
> 

1. Yes, if Secure Boot is enabled, only Secured OS can be installed and booted. Unsecured images will fail to be installed, and if were already pre-installed, will fail to boot. 
2. Yes. Devices that don't support Secure Boot will not check anything related to the security of the image, and any image it will be possible to install and boot any image, just as it were in the past.

PR #2699, opened 2023-02-28
[show][barefoot] replace shell=True

maipbui:

#### What I did
#### How I did it

### DIFF
--- a/show/plugins/barefoot.py
+++ b/show/plugins/barefoot.py
@@ -4,6 +4,7 @@ import click
 import json
 import subprocess
 from sonic_py_common import device_info
+from sonic_py_common.general import getstatusoutput_noshell_pipe, check_output_pipe
 
 @click.group()
 def barefoot():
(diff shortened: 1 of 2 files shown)


show/plugins/barefoot.py
```diff
@@ -37,9 +39,10 @@ def profile():
     
     # Print profile list
     click.echo('Available profile(s):')
-    subprocess.run('docker exec -it syncd find /opt/bfn -mindepth 1 '
-        r'-maxdepth 1 -type d -name install_\*_profile ' + opts + '| sed '
-        r's%/opt/bfn/install_\\\(.\*\\\)_profile%\\1%', shell=True)
+    cmd0 = ['docker', 'exec', '-it', 'syncd', 'find', '/opt/bfn', '-mindepth', '1',\
+            r'-maxdepth', '1', r'-type', 'd', r'-name', r'install_\*_profile', opts]
```
qiluo-msft, line 43: >r [](http://example.com/codeflow?start=12&length=1)

This `r` is not needed.

qiluo-msft, line 43: >r [](http://example.com/codeflow?start=31&length=1)

the same

qiluo-msft, line 43: >r [](http://example.com/codeflow?start=46&length=1)

the same

qiluo-msft:
@oleksandrx-kolomeiets Could you help review?

qiluo-msft (approved):

maipbui:
@oleksandrx-kolomeiets could you help review?

qiluo-msft (approved):

PR #2702, opened 2023-02-28
Fixed techsupport failure in generate_dump 'show bgp neighbors'

saurabh17g:
## What I did
Resolved Techsupport failure upon generate_dump (techsupport gives rc = 1) execution for 'show bgp ipv6 neighbors' and 'show ip bgp neighbors' cli.

## How I did it
fgrep returned non 0 RC in case bgp neighbors are not configured or link down/config abscent on peer etc. This results in script failure. So, I replaced fgrep with awk. This returns rc = 0 even when we dont have any results in filter.
In case of ipv4 (show ip bgp neighbors), it is observed that cli also gives v6 peers. Consequently neighbor_list_v4 has v4 and v6 peers. This generates 2 sets of files in techsupport (1 set for v4 and other for v6). however v6 neighbors do not have routes and are empty files. So I added extra filtering here using awk and neighbor_list_v4 will only have v4 neighbors.
How to verify it
case 1 : setup having route. Verified my script does not result any errors by rc checking.
I also checked techsuppport dump and observed no non 0 rc and no files with empty routes.
case 2 : setup not having neigbor/routes. Ensured rc is always 0.

## Previous command output (if the output of a command-line utility has changed)
show techsupport -r --since '5 minute ago' -> echo $? gives 1

## New command output (if the output of a command-line utility has changed)
show techsupport -r --since '5 minute ago' -> echo $? gives 0


kevinskwang:
So this issue just appear on branch 202012?

saurabh17g:
> So this issue just appear on branch 202012?

we tested on both (202205, 202012) branches but it appears to be on 202012.

kevinskwang (approved):

PR #2708, opened 2023-03-01
execute a platform-provided pre-reboot script before shutting down swss

jhli-cisco:
#### What I did
Give platform chance to do proper cleanup before shutting down swss in warm/fast-reboot by calling a platform-provided script

#### How I did it
update fast-reboot script.

#### How to verify it
execute fast-reboot/warm-reboot

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -3,6 +3,7 @@
 REBOOT_USER=$(logname)
 REBOOT_TIME=$(date)
 REBOOT_CAUSE_FILE="/host/reboot-cause/reboot-cause.txt"
+PLATFORM_REBOOT_PRE_CHECK="platform_reboot_pre_check"
 WARM_DIR=/host/warmboot
 REDIS_FILE=dump.rdb
 CONFIG_DB_FILE=/etc/sonic/config_db.json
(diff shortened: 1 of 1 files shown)


jhli-cisco:
vaibhavhd, can you please review?

scripts/fast-reboot
```diff
@@ -174,6 +175,10 @@ function initialize_pre_shutdown()
 
 function request_pre_shutdown()
 {
+    if [ -x ${DEVPATH}/${PLATFORM}/${PLATFORM_REBOOT_PRE_CHECK} ]; then
+        debug "Requesting platform reboot pre-check ..."
+    	${DEVPATH}/${PLATFORM}/${PLATFORM_REBOOT_PRE_CHECK} ${REBOOT_TYPE}
```
vaibhavhd, line 180: What is this pre-check script?

What should be desired action when this script returns non-zero value or fails?

If the intention is to abort the reboot procedure:
1. Such error should be logged by checking return code, and calling `debug` message.
2. This check may be moved to `reboot_pre_check` as that's where we decide if the device is healthy enough for a reboot.

jhli-cisco: This is a script that does platform specific cleanup. In our case, it currently terminates a platform specific process in syncd conatiner. It is best effort, so we dont care the return code of the script.

scripts/fast-reboot
```diff
@@ -3,6 +3,7 @@
 REBOOT_USER=$(logname)
 REBOOT_TIME=$(date)
 REBOOT_CAUSE_FILE="/host/reboot-cause/reboot-cause.txt"
+PLATFORM_REBOOT_PRE_CHECK="platform_reboot_pre_check"
```
vaibhavhd, line 6: Minor: suggest moving this down to where other `PLATFORM` variables are defined.

vaibhavhd (approved):

scripts/fast-reboot
jhli-cisco: Updated. Thank you very much.

rraghav-cisco:
@kevinwangsk , Pls merge, as @vaibhavhd has approved this.

jhli-cisco:
Hi Kevin, I am seeing "Only those with [write access](https://docs.github.com/articles/what-are-the-different-access-permissions) to this repository can merge pull requests." I dont see merge button.

PR #2714, opened 2023-03-03
[warm/fast-reboot] Backup logs from tmpfs to disk during fast/warm shutdown

vaibhavhd:
#### What I did
Goal: Preserve logs during TOR upgrades and shutdown

Need:

Below PRs moved logs from disk to tmpfs for specific hwskus.
Due to these changes, shutdown path logs are now lost.
The logs in shutdown path are crucial for debug purposes.

https://github.com/sonic-net/sonic-buildimage/pull/13805

https://github.com/sonic-net/sonic-buildimage/pull/13587

https://github.com/sonic-net/sonic-buildimage/pull/13587

#### How I did it

Check if logs are on tmpfs. If yes, backup logs from /var/log 

#### How to verify it
Verified on a physical device

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -806,6 +806,17 @@ fi
 
 # Reboot: explicitly call Linux native reboot under sbin
 debug "Rebooting with ${REBOOT_METHOD} to ${NEXT_SONIC_IMAGE} ..."
+
+LOGS_ON_TMPFS=0
+df --output=fstype /var/log* | grep 'tmpfs' || LOGS_ON_TMPFS=$?
+if [[ LOGS_ON_TMPFS -eq 0 ]]; then
+    debug "Backup shutdown logs to /host/logs_before_reboot"
+    mkdir -p /host/logs_before_reboot || /bin/true
+    # maxdepth 2: find files within 2 nested directories (eg. /var/log/ and /var/log/swss/)
+    # mmin 30: find files written in past 30 minutes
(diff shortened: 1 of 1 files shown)


scripts/fast-reboot
```diff
@@ -806,6 +806,17 @@ fi
 
 # Reboot: explicitly call Linux native reboot under sbin
 debug "Rebooting with ${REBOOT_METHOD} to ${NEXT_SONIC_IMAGE} ..."
+
+LOGS_ON_TMPFS=0
+df --output=fstype /var/log* | grep 'tmpfs' || LOGS_ON_TMPFS=$?
```
yxieca, line 811: Use grep -c instead?

vaibhavhd: Sure, updated now.

scripts/fast-reboot
```diff
@@ -806,6 +806,17 @@ fi
 
 # Reboot: explicitly call Linux native reboot under sbin
 debug "Rebooting with ${REBOOT_METHOD} to ${NEXT_SONIC_IMAGE} ..."
+
+LOGS_ON_TMPFS=0
+df --output=fstype /var/log* | grep -c 'tmpfs' || LOGS_ON_TMPFS=$?
```
yxieca, line 811: grep -c will always return 0, what you want is the output of grep -c

LOGS_ON_TMPFS=$(df --output=fstype /var/log* | grep -c 'tmpfs')

yxieca: And you want to check the count is greater than 0.

yxieca (approved):

PR #2715, opened 2023-03-04
Fix non-zero status exit on non secure boot system

kellyyeh:
#### What I did
Warm-reboot fails on kvm due to non-zero exit upon command 
`bootctl status 2>/dev/null | grep -c "Secure Boot: enabled"` 

#### How I did it
Added `|| true` to return 0 when previous command fails.
Added `CHECK_SECURE_UPGRADE_ENABLED` to check output of previous command
Added debug logs

#### How to verify it
Run warm-reboot on kvm and physical device when increased verbosity. Expects debug log to indicate secure/non secure boot. Successful warm reboot

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -615,17 +615,13 @@ if is_secureboot && grep -q aboot_machine= /host/machine.conf; then
 else
     # check if secure boot is enable in UEFI 
     CHECK_SECURE_UPGRADE_ENABLED=0
-    SECURE_UPGRADE_ENABLED=$(bootctl status 2>/dev/null | grep -c "Secure Boot: enabled" || true) || CHECK_SECURE_UPGRADE_ENABLED=$?
+    SECURE_UPGRADE_ENABLED=$(bootctl status 2>/dev/null | grep -c "Secure Boot: enabled") || CHECK_SECURE_UPGRADE_ENABLED=$?
     if [[ CHECK_SECURE_UPGRADE_ENABLED -ne 0 ]]; then
-        error "CHECK_SECURE_UPGRADE_ENABLED failed"
+        debug "Loading kernel without secure boot"
+        load_kernel
     else
-        if [[ ${SECURE_UPGRADE_ENABLED} -eq 1 ]]; then
-            debug "Loading kernel with secure boot"
-            load_kernel_secure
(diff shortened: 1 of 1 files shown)


scripts/fast-reboot
```diff
@@ -614,11 +614,18 @@ if is_secureboot && grep -q aboot_machine= /host/machine.conf; then
     load_aboot_secureboot_kernel
 else
     # check if secure boot is enable in UEFI 
-    SECURE_UPGRADE_ENABLED=$(bootctl status 2>/dev/null | grep -c "Secure Boot: enabled")
-    if [ ${SECURE_UPGRADE_ENABLED} -eq 1 ]; then
-        load_kernel_secure
+    CHECK_SECURE_UPGRADE_ENABLED=0
+    SECURE_UPGRADE_ENABLED=$(bootctl status 2>/dev/null | grep -c "Secure Boot: enabled" || true) || CHECK_SECURE_UPGRADE_ENABLED=$?
```
vaibhavhd, line 618: This seems incorrect. The first part will always pass (due to first `||`) and hence second part (after second `||` to capture error code) will not execute.

scripts/fast-reboot
```diff
@@ -614,11 +614,18 @@ if is_secureboot && grep -q aboot_machine= /host/machine.conf; then
...
+        error "CHECK_SECURE_UPGRADE_ENABLED failed"
     else
-        load_kernel
+        if [[ ${SECURE_UPGRADE_ENABLED} -eq 1 ]]; then
+            debug "Loading kernel with secure boot"
+            load_kernel_secure
+        else
+            debug "Loading kernel without secure boot"
+            load_kernel
+        fi
```
vaibhavhd, line 628: Two conditions inside else are not needed? When ` CHECK_SECURE_UPGRADE_ENABLED -ne 0` then load_kernel can be called, otherwise `load_kernel_secure` can be called.

yxieca (approved):

yxieca:
This PR is addressing issue introduced by #2532

yxieca:
This change need to be cherry-picked to 202211 after #2532

StormLiangMS:
@kellyyeh  could you raise PR for 202211 branch?

PR #2716, opened 2023-03-06
Check SONiC dependencies before installation.

liushilongbuaa:
#### What I did
SONiC related packages shouldn't be intalled from Pypi.
It is security compliance requirement.
Check SONiC related packages when using setup.py.


xumia:
@liushilongbuaa , please add the description, thanks.

xumia (approved):

qiluo-msft (approved):

qiluo-msft:
This file is only used in build time or unit test, not in runtime. We can safely override the coverage checker.

PR #2724, opened 2023-03-07
Remove timer from FAST_REBOOT STATE_DB entry and use finalizer

arfeigin:
This should come along with sonic-buildimage PR  implementing fast-reboot finalizing logic in finalize-warmboot script and other submodules PRs utilizing the change.

This PR should come along with the following PRs as well:

This set of PRs solves the issue https://github.com/sonic-net/sonic-buildimage/issues/13251

What I did
Remove the timer used to clear fast-reboot entry from state-db, instead it will be cleared by fast-reboot finalize function implemented inside finalize-warmboot script (which will be invoked since fast-reboot is using warm-reboot infrastructure).

As well instead of having "1" as the value for fast-reboot entry in state-db and deleting it when done it is now modified to set enable/disable according to the context.

As well all scripts reading this entry should be modified to the new value options.

How I did it
Removed the timer usage in the fast-reboot script and adding fast-reboot finalize logic to warm-reboot in the linked PR.
Use "enable: true/false" instead of "1" as the entry value.

How to verify it
Run fast-reboot and check that the state-db entry for fast-reboot is being deleted after finalizing fast-reboot and not by an expiring timer.

### DIFF
--- a/scripts/db_migrator.py
+++ b/scripts/db_migrator.py
@@ -46,7 +46,7 @@ class DBMigrator():
                      none-zero values.
               build: sequentially increase within a minor version domain.
         """
-        self.CURRENT_VERSION = 'version_3_0_5'
+        self.CURRENT_VERSION = 'version_3_0_6'
 
         self.TABLE_NAME      = 'VERSIONS'
         self.TABLE_KEY       = 'DATABASE'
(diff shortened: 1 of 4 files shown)


scripts/db_migrator.py
```diff
@@ -806,9 +806,28 @@ def version_3_0_4(self):
...
-        Current latest version. Nothing to do here.
+        Version 3_0_5
         """
         log.log_info('Handling version_3_0_5')
+        # Update state-db fast-reboot entry to enable if set to enable fast-reboot finalizer when using upgrade with fast-reboot
+        # since upgrading from previous version FAST_REBOOT table will be deleted when the timer will expire.
+        # reading FAST_REBOOT table can't be done with stateDB.get as it uses hget behind the scenes and the table structure is
+        # not using hash and won't work.
+        # FAST_REBOOT table exists only if fast-reboot was triggered.
+        keys = self.stateDB.keys(self.stateDB.STATE_DB, "FAST_REBOOT")
```
vaibhavhd, line 817: Same as https://github.com/sonic-net/sonic-utilities/pull/2621#discussion_r1130519315

tests/db_migrator_input/config_db/reclaiming-buffer-warmreboot-expected.json
```diff
@@ -2044,6 +2044,6 @@
         "admin_status": "up"
     },
     "VERSIONS|DATABASE": {
-        "VERSION": "version_3_0_5"
+        "VERSION": "version_3_0_3"
```
vaibhavhd, line 2047: UT cases are missing on this branch, is that intentional?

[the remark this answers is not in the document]
arfeigin: For some reason without this change some db migrator tests are failing.
The change is to align to the code on master and 202211 branches.

arfeigin: No, didn't notice it wasn't included will fix in few minutes.
Thanks for the catch.

arfeigin: UT added.

vaibhavhd (approved):

PR #2726, opened 2023-03-07
[sonic_package_manager] replace shell=True

maipbui:
#### What I did
#### How I did it
#### How to verify it
Pass UT

### DIFF
--- a/sonic_package_manager/service_creator/creator.py
+++ b/sonic_package_manager/service_creator/creator.py
@@ -5,7 +5,7 @@ import os
 import stat
 import subprocess
 from collections import defaultdict
-from typing import Dict, Type
+from typing import Dict, Type, List
 
 import jinja2 as jinja2
 from config.config_mgmt import ConfigMgmt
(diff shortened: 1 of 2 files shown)


sonic_package_manager/service_creator/creator.py
```diff
@@ -97,7 +97,7 @@ def remove_if_exists(path):
     log.info(f'removed {path}')
 
 
-def run_command(command: str):
+def run_command(command: List[str]):
```
qiluo-msft, line 100: Please be aware type hint does not mean runtime check: https://stackoverflow.com/questions/41692473/does-python-type-hint-annotations-cause-some-run-time-effects. Will you add a runtime check?

qiluo-msft:
@stepanblyschak Could you help review?

qiluo-msft (approved):

stepanblyschak (approved):

stepanblyschak (approved):

qiluo-msft (approved):

PR #2738, opened 2023-03-15
[config]Support multi-asic  Golden Config override

wen587:
#### What I did
Support multi-asic Golden Config Override
#### How I did it
Add ConfigMgmt support for ASIC validation. Modify override config cli to support multi-asic.
#### How to verify it
Unit test:
tests/config_override_test.py::TestConfigOverrideMultiasic::test_macsec_override PASSED [  8%]
tests/config_override_test.py::TestConfigOverrideMultiasic::test_device_metadata_table_rm PASSED [  8%]


tests/config_override_test.py
```diff
@@ -257,3 +258,74 @@ def teardown_class(cls):
...
+        importlib.reload(mock_multi_asic)
+        dbconnector.load_namespace_config()
+        return
+
+    def test_macsec_override(self):
+        def read_json_file_side_effect(filename):
+            return macsec_profile
+        db = Db()
+        cfgdb_clients = db.cfgdb_clients
+        macsec_profile = {}
```
judyjoseph, line 281: This  will work .. 

but I suggest to create a golden_config_db.json file in the tests directory  in the format we need ( where we have host + various namespaces -- and th emacsec profile data in each of asic0,asic1 etc namespaces )

judyjoseph:
Otherwise looks ok.
We need to validate this and add some test results too in Comments above

qiluo-msft:
LGTM. Keep it open for more reviewing.

tests/config_override_test.py
wen587: fixed.

config/main.py
```diff
@@ -1902,36 +1929,43 @@ def override_config_table(db, input_config_db, dry_run):
...
-            cm.validateConfigData()
-        except Exception as ex:
-            click.secho("Failed to validate running config. Error: {}".format(ex), fg="magenta")
-            sys.exit(1)
+        if multi_asic.is_multi_asic():
+            ns_config_input = config_input[ns]
+            generate_sysinfo(ns_config_input, ns)
+        else:
+            ns_config_input = config_input
+            generate_sysinfo(ns_config_input)
```
qiluo-msft, line 1945: >generate_sysinfo [](http://example.com/codeflow?start=12&length=16)

Is it better to fill missing mac/platform during `override_config_db()` ?

If the original ConfigDB already has mac/platform, we should keep them even we override other fields.

config/main.py
```diff
@@ -1875,6 +1875,33 @@ def override_config_by(golden_config_path):
...
+        asic_role = device_metadata.get('localhost', {}).get('sub_role')
+        switch_type = device_metadata.get('localhost', {}).get('switch_type')
+
+        if ((switch_type is not None and switch_type.lower() == "chassis-packet") or
+                (asic_role is not None and asic_role.lower() == "backend")):
+            mac = device_info.get_system_mac(namespace=ns)
+        else:
+            mac = device_info.get_system_mac()
+    else:
+        mac = device_info.get_system_mac()
```
qiluo-msft, line 1895: I notice code duplication of https://github.com/sonic-net/sonic-buildimage/blob/d74055e12ca0ac4f44342e74ea097f03720663f7/src/sonic-config-engine/sonic-cfggen#L382-L388.
Possible to reuse code?

config/main.py
wen587: Do you mean the generated sysinfo should have less priority than existing one?
Though I think they should be the same.

judyjoseph (approved):
Checked on a multi-asic device running 2205. Find the config ( eg: macsec_profile) getting added in host/asic namespaces

config/main.py
wen587: Will submit another PR for sysinfo generation about override.

judyjoseph (approved):

config/main.py
qiluo-msft: Could you remove code dup in next PR?

qiluo-msft (approved):

wen587:
Remove Chassis tag because there is bug in this PR and block subodule. Will create an updated PR after the revert.

PR #2744, opened 2023-03-17
[202205] Fast-reboot: preserve connected routes and set teamd timer to minimum.

stepanblyschak:
#### What I did
Added a script to filter routes: preserve default routes (was already done as part of fast-reboot script) and connected routes.
Set teamd timer to minimal allowed value (1 second) for fast-reboot.
Both made in order to shorten dataplane downtime.

#### How I did it
fast-reboot-filter-routes.py was added to preserve connected and default routes and is being called from fast-reboot script.
teamd-timer is set when setting fast-reboot.

#### How to verify it
Community fast-reboot test.

### DIFF
--- a/scripts/fast-reboot-filter-routes.py
+++ b/scripts/fast-reboot-filter-routes.py
@@ -7,7 +7,6 @@ import utilities_common.cli as clicommon
 import syslog
 import traceback
 import click
-from builtins import str #for unicode conversion in python2
 from swsscommon.swsscommon import ConfigDBConnector
 
 ROUTE_IDX = 1
@@ -22,7 +21,7 @@ def get_connected_routes():
             sys.exit(ret)
         if output is not None:
             route_info = json.loads(output)
-            for route, info in route_info.items():
+            for route in route_info.keys():
                 connected_routes.append(route)
     except Exception:
         ctx = click.get_current_context()
--- a/scripts/fast-reboot-filter-routes.py
+++ b/scripts/fast-reboot-filter-routes.py
@@ -1,0 +1,92 @@
+#!/usr/bin/env python3
+
+import json
+import sys
+import os
+import utilities_common.cli as clicommon
+import syslog
+import traceback
+import click
+from builtins import str #for unicode conversion in python2
+from swsscommon.swsscommon import ConfigDBConnector
+
+ROUTE_IDX = 1
+
+def get_connected_routes():
+    cmd = 'sudo vtysh -c "show ip route connected json"'
+    connected_routes = []
+    try:
+        output, ret = clicommon.run_command(cmd, return_cmd=True)
+        if ret != 0:
+            click.echo(output.rstrip('\n'))
+            sys.exit(ret)
+        if output is not None:
+            route_info = json.loads(output)
+            for route, info in route_info.items():
+                connected_routes.append(route)
+    except Exception:
+        ctx = click.get_current_context()
+        ctx.fail("Unable to get connected routes from bgp")
+    
+    return connected_routes
+
+def get_route(db, route):
+    key = 'ROUTE_TABLE:%s' % route
+    val = db.keys(db.APPL_DB, key)
+    if val:
+        return val[0].split(":", 1)[ROUTE_IDX]
+    else:
+        return None
+
+def generate_default_route_entries():
+    db = ConfigDBConnector()
+    db.db_connect(db.APPL_DB)
+
+    default_routes = []
+
+    ipv4_default = get_route(db, '0.0.0.0/0')
+    if ipv4_default is not None:
+        default_routes.append(ipv4_default)
+
+    ipv6_default = get_route(db, '::/0')
+    if ipv6_default is not None:
+        default_routes.append(ipv6_default)
+
+    return default_routes
+
+def filter_routes(preserved_routes):
+    db = ConfigDBConnector()
+    db.db_connect(db.APPL_DB)
+
+    key = 'ROUTE_TABLE:*'
+    routes = db.keys(db.APPL_DB, key)
+
+    for route in routes:
+        stripped_route = route.split(":", 1)[ROUTE_IDX]
+        if stripped_route not in preserved_routes:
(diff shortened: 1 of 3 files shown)


stepanblyschak:
@arfeigin to remove draft indication once all testing is done

dprital:
@prsunny , @vaibhavhd , Can you please review ?

scripts/fast-reboot-filter-routes.py
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env python3
+
+import json
+import sys
+import os
+import utilities_common.cli as clicommon
+import syslog
+import traceback
+import click
+from builtins import str #for unicode conversion in python2
```
vaibhavhd, line 10: Looks like you are enforcing the env to be `python3` (top of file). Is Python2 handling still needed?

scripts/fast-reboot-filter-routes.py
```diff
@@ -0,0 +1,92 @@
...
+        if ret != 0:
+            click.echo(output.rstrip('\n'))
+            sys.exit(ret)
+        if output is not None:
+            route_info = json.loads(output)
+            for route, info in route_info.items():
+                connected_routes.append(route)
+    except Exception:
+        ctx = click.get_current_context()
+        ctx.fail("Unable to get connected routes from bgp")
```
vaibhavhd, line 29: What in intended behavior when we hit this exception? If this returns non-zero code to bash, fast-reboot script will still continue as the call to filer-routes is after `set +e` in bash script. Do you still want the fast-reboot to proceed when routes are not filtered?

scripts/fast-reboot-filter-routes.py
```diff
@@ -0,0 +1,92 @@
...
+    cmd = 'sudo vtysh -c "show ip route connected json"'
+    connected_routes = []
+    try:
+        output, ret = clicommon.run_command(cmd, return_cmd=True)
+        if ret != 0:
+            click.echo(output.rstrip('\n'))
+            sys.exit(ret)
+        if output is not None:
+            route_info = json.loads(output)
+            for route, info in route_info.items():
```
vaibhavhd, line 25: Since `info` value is not being used, suggest to use `route_info.keys()`

scripts/fast-reboot-filter-routes.py
```diff
@@ -0,0 +1,92 @@
...
+ROUTE_IDX = 1
+
+def get_connected_routes():
+    cmd = 'sudo vtysh -c "show ip route connected json"'
+    connected_routes = []
+    try:
+        output, ret = clicommon.run_command(cmd, return_cmd=True)
+        if ret != 0:
+            click.echo(output.rstrip('\n'))
+            sys.exit(ret)
```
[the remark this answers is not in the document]
arfeigin: Fast-reboot requires BGP is enabled, this will return non-zero code only in case BGP is not used and hence is not relevant for fast-reboot if I understand correctly.
However this won't fast-reboot to fail but it might make it exceed the dataplane downtime if filtering failed and switch has rather large amount of routes stored.

scripts/fast-reboot-filter-routes.py
arfeigin: Fast-reboot requires BGP is enabled, this will return non-zero code only in case BGP is not used and hence is not relevant for fast-reboot if I understand correctly.
However this won't fast-reboot to fail but it might make it exceed the dataplane downtime if filtering failed and switch has rather large amount of routes stored.

scripts/fast-reboot-filter-routes.py
```diff
@@ -0,0 +1,92 @@
...
+
+def filter_routes(preserved_routes):
+    db = ConfigDBConnector()
+    db.db_connect(db.APPL_DB)
+
+    key = 'ROUTE_TABLE:*'
+    routes = db.keys(db.APPL_DB, key)
+
+    for route in routes:
+        stripped_route = route.split(":", 1)[ROUTE_IDX]
```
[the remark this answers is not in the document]
arfeigin: IPv6 routes are stored in the way: ROUTE_TABLE:::/0
So using '1' when splitting will not break the logic.

scripts/fast-reboot-filter-routes.py
```diff
@@ -0,0 +1,92 @@
...
+    key = 'ROUTE_TABLE:%s' % route
+    val = db.keys(db.APPL_DB, key)
+    if val:
+        return val[0].split(":", 1)[ROUTE_IDX]
+    else:
+        return None
+
+def generate_default_route_entries():
+    db = ConfigDBConnector()
+    db.db_connect(db.APPL_DB)
```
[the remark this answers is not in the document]
arfeigin: Thanks for the heads-up.

scripts/fast-reboot-filter-routes.py
arfeigin: Modified and validated.

scripts/fast-reboot-filter-routes.py
arfeigin: Done.

scripts/fast-reboot-filter-routes.py
```diff
@@ -22,7 +21,7 @@ def get_connected_routes():
             sys.exit(ret)
         if output is not None:
             route_info = json.loads(output)
-            for route, info in route_info.items():
+            for route in route_info.keys():
```
vaibhavhd, line 24: Will this work? I though keys() in Python3 returns type `dict_keys` that can't be iterated upon? I think you need to force convert this to list: `list(route_info.keys())`

scripts/fast-reboot-filter-routes.py
vaibhavhd: I will leave it to you. To me it sounds like this will impact dataplane downtime as the PR is fixing that when the connected routes aren't preserved. So if you wish you could set this as a required criteria - by moving the call the preserve before set +e. Just FYI.

vaibhavhd (approved):

PR #2758, opened 2023-03-27
Revert "Convert IPv6 addresses to lowercase in apply-patch (#2299)"

wen587:
This reverts commit 28b6ba5fc11f65abaf421c70159a605d233eda41.

There are some issues when GCU tries to remove the interface IP in some tests, such as add rack test.

In the initial config, the INTERFACE's IPv6 was all loaded in uppercase by default.

    "INTERFACE": {
        "Ethernet68": {},
        "Ethernet68|10.0.0.34/31": {},
        "Ethernet68|FC00::45/126": {},
        "Ethernet72": {},
        "Ethernet72|10.0.0.36/31": {},
        "Ethernet72|FC00::49/126": {},
GCU will never be able to remove that IP because IPv6 was always translated to lowercase due to #2299 . It reported the error can't remove a non-existent object, thus making GCU fail.

#2299  is to deal with this issue: https://github.com/sonic-net/sonic-buildimage/issues/11622.

Although config CLI always translates uppercase to lowercase when adding an IP, the user can have two choices to remove that IP: One is to use config CLI to remove that IP no matter uppercase or lowercase. Another way is to use GCU. In order to use GCU, the user has to check the IP format saved in ConfigDB because GCU operation does differentiate between uppercase and lowercase.
#### What I did
Revert #2299

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1362,20 +1362,6 @@ def apply_patch(ctx, patch_file_path, format, dry_run, ignore_non_yang_tables, i
             patch_as_json = json.loads(text)
             patch = jsonpatch.JsonPatch(patch_as_json)
 
-        # convert IPv6 addresses to lowercase
-        for patch_line in patch:
-            if 'remove' == patch_line['op']:
-                match = re.search(r"(?P<prefix>/INTERFACE/\w+\|)(?P<ipv6_address>([a-fA-F0-9]{0,4}[:~]|::){1,7}[a-fA-F0-9]{0,4})"
-                                    "(?P<suffix>.*)", str.format(patch_line['path']))
-                if match:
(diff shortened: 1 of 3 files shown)


qiluo-msft:
@dbarashinvd  Could you help review this PR?

dbarashinvd:
> @dbarashinvd Could you help review this PR?

can you put me as reviewer ? I cant approve or add myself.

qiluo-msft:
@dbarashinvd If you permission is limited, you can comment with "Approve".

dbarashinvd:
Approve

yxieca:
@qiluo-msft @wangxin have we adapted any test according to #2299? if so, we need to revert the test chagne too.

PR #2760, opened 2023-03-28
Fast-reboot: set teamd timer to minimum, preserve connected routes

arfeigin:
Part of https://github.com/sonic-net/sonic-buildimage/pull/14583
Similar to https://github.com/sonic-net/sonic-utilities/pull/2744

#### What I did
Added a script to filter routes: preserve default routes (was already done as part of fast-reboot script) and connected routes.
Set teamd timer to minimal allowed value (1 second) for fast-reboot.
Both made in order to shorten dataplane downtime.

#### How I did it
fast-reboot-filter-routes.py was added to preserve connected and default routes and is being called from fast-reboot script.
teamd-timer is set when setting fast-reboot.

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -550,6 +550,7 @@ case "$REBOOT_TYPE" in
         BOOT_TYPE_ARG=$REBOOT_TYPE
         trap clear_boot EXIT HUP INT QUIT TERM KILL ABRT ALRM
         sonic-db-cli STATE_DB HSET "FAST_RESTART_ENABLE_TABLE|system" "enable" "true" &>/dev/null
+        config warm_restart teamsyncd_timer 1
         config warm_restart enable system
         ;;
     "warm-reboot")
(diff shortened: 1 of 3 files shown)


scripts/fast-reboot
```diff
@@ -667,14 +668,15 @@ fi
...
-            if string.match(k, 'ROUTE_TABLE:') and not string.match(k, 'ROUTE_TABLE:0.0.0.0/0') and not string.match(k, 'ROUTE_TABLE:::/0') then \
-                redis.call('del', k)
-            end
-        end
-    " 0 > /dev/null
+    # Clear all routes except of default and connected routes for faster reconciliation time.
+    debug "Clearing routes..."
+    FILTER_ROUTES=0
+    python /usr/local/bin/fast-reboot-filter-routes.py || FILTER_ROUTES=$?
+    if [[ FILTER_ROUTES -ne 0 ]]; then
```
vaibhavhd, line 675: 1. This error handling is not added to 202205 branch PR. Any specific reason?
2. Again highlighting that this call is after `set +e` - so if filtering routes fail, device will still proceed to reboot. This was discussed earlier on 202205 PR and was acceptable to MLNX.

vaibhavhd (approved):
Non-blocking question added.

liat-grozovik (approved):

PR #2765, opened 2023-03-29
[GCU]Fix bug in GCU vlanintf_validator

wen587:
#### What I did
Fix the bug in vlanintf_validator
#### How I did it
ret will be 0 if success. Fix the statement.
#### How to verify it
E2E test


StormLiangMS (approved):
LGTM

PR #2768, opened 2023-03-30
[202012][DBMigrator] Update db_migrator to support EdgeZoneAggregator Buffer Config for T0s

developfast:
#### What I did
Updated dbmigrator to support warm upgrade to new buffer config for EdgeZoneAggregator neighbors.

#### How I did it
Created a new version checkpoint for the change and made appropriate changes to the ConfigDB.

#### How to verify it
Unit tests pass and tested on device in lab


scripts/db_migrator.py
```diff
@@ -691,6 +745,26 @@ def set_version(self, version=None):
         entry = { self.TABLE_FIELD : version }
         self.configDB.set_entry(self.TABLE_NAME, self.TABLE_KEY, entry)
 
+    def is_curr_version_greater(curr_version=self.CURRENT_VERSION, compare_version=None):
```
vaibhavhd, line 748: This needs to be removed now?

vaibhavhd (approved):

PR #2778, opened 2023-04-04
[config]config reload should generate sysinfo if missing

wen587:
#### What I did
Missing platform and mac in CONFIG_DB will result in container failure. We should make the `config reload` generate those info if missing.
#### How I did it
Add missing sys info if config_db.json doesn't contain it.
#### How to verify it
Unit test

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1560,6 +1560,17 @@ def reload(db, filename, yes, load_sysinfo, no_service_restart, force, file_form
             click.echo("The config file {} doesn't exist".format(file))
             continue
 
+        if file_format == 'config_db':
+            file_input = read_json_file(file)
+
+            platform = file_input.get("DEVICE_METADATA", {}).\
+                get("localhost", {}).get("platform", {})
+            mac = file_input.get("DEVICE_METADATA", {}).\
+                get("localhost", {}).get("mac", {})
+
+            if not platform or not mac:
+                load_sysinfo = True
+
         if load_sysinfo:
             try:
                 command = "{} -j {} -v DEVICE_METADATA.localhost.hwsku".format(SONIC_CFGGEN_PATH, file)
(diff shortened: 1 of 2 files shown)


config/main.py
```diff
@@ -1560,6 +1560,17 @@ def reload(db, filename, yes, load_sysinfo, no_service_restart, force, file_form
             click.echo("The config file {} doesn't exist".format(file))
             continue
 
+        if file_format == 'config_db':
```
qiluo-msft, line 1563: Is this feature already implement by argument `load_sysinfo`?

wen587: Yes. The purpose of the PR is to keep the CLI cmd identical for Golden Config reload.
From NDM team's knowledge, they don't have such info from their side.

config/main.py
```diff
@@ -1561,6 +1561,17 @@ def reload(db, filename, yes, load_sysinfo, no_service_restart, force, file_form
...
+            file_input = read_json_file(file)
+
+            platform = file_input.get("DEVICE_METADATA", {}).\
+                get("localhost", {}).get("platform", {})
```
qiluo-msft, line 1568: >{} [](http://example.com/codeflow?start=53&length=2)

Default should be None or empty string

config/main.py
```diff
@@ -1561,6 +1561,17 @@ def reload(db, filename, yes, load_sysinfo, no_service_restart, force, file_form
...
+            mac = file_input.get("DEVICE_METADATA", {}).\
+                get("localhost", {}).get("mac", {})
```
qiluo-msft, line 1570: >{} [](http://example.com/codeflow?start=48&length=2)

Default should be None or empty string

config/main.py
```diff
@@ -1561,6 +1561,17 @@ def reload(db, filename, yes, load_sysinfo, no_service_restart, force, file_form
...
+
+            if not platform or not mac:
```
qiluo-msft, line 1572: >if [](http://example.com/codeflow?start=12&length=2)

For extreme corner case:
1. platform is None and mac is valid
2. platform is valid and mac is None
you can continue with load_sysinfo, please add some log_warning for future troubleshooting purpose.

qiluo-msft (approved):

PR #2780, opened 2023-04-04
[ci] Switch to using regular swss-common artifacts

saiarcot895:
#### What I did

Switch to using regular swss-common artifacts

The swss-common pipeline is now built for Bullseye.


developfast (approved):
lgtm

PR #2786, opened 2023-04-10
Update pcieutil error message on loading common pcie module

cytsao1:
Currently, pcieutil will print the following message to console every time it is called if loading the common module:
root@sonic:/home/cisco# pcieutil show
Failed to load platform Pcie module. Error : No module named 'sonic_platform.pcie', fallback to load Pcie common utility.
==============================Display PCIe Device===============================

#### What I did
Update the error message to only be logged in syslog, rather than print to console.

#### How to verify it
Run any pcieutil command

```
root@sonic:/home/cisco# pcieutil show
Failed to load platform Pcie module. Error : No module named 'sonic_platform.pcie', fallback to load Pcie common utility.
==============================Display PCIe Device===============================
......
```

```
root@sonic:/home/cisco# pcieutil show
==============================Display PCIe Device===============================
.....
```

### DIFF
--- a/pcieutil/main.py
+++ b/pcieutil/main.py
@@ -54,7 +54,7 @@ def load_platform_pcieutil():
         from sonic_platform.pcie import Pcie
         platform_pcieutil = Pcie(platform_path)
     except ImportError as e:
-        log.log_warning("Failed to load platform Pcie module. Error : {}, fallback to load Pcie common utility.".format(str(e)), True)
+        log.log_warning("Failed to load platform Pcie module. Warning : {}, fallback to load Pcie common utility.".format(str(e)))
         try:
             from sonic_platform_base.sonic_pcie.pcie_common import PcieUtil
             platform_pcieutil = PcieUtil(platform_path)
--- a/tests/pcieutil_test.py
+++ b/tests/pcieutil_test.py
@@ -3,6 +3,7 @@ import os
 from unittest import mock
 
 from click.testing import CliRunner
+from io import StringIO
 
 test_path = os.path.dirname(os.path.abspath(__file__))
 modules_path = os.path.dirname(test_path)
@@ -201,13 +202,12 @@ class TestPcieUtil(object):
         result = runner.invoke(pcieutil.cli.commands["pcie-aer"].commands["correctable"], ["-d", "0:1.0"])
         assert result.output == pcieutil_pcie_aer_correctable_dev_output
 
-    def test_load_pcie_module_warning(self):
-        runner = CliRunner()
-        sys_path = sys.path
-        sys.path = ['']
-        result = runner.invoke(pcieutil.cli.commands["show"])
-        assert pcieutil_load_module_warning_msg not in result.output
-        sys.path = sys_path
+    def test_load_pcie_module_warning(self): 
+        stdout = sys.stdout
+        sys.stdout = result = StringIO()
+        pcieutil.load_platform_pcieutil()
+        sys.stdout = stdout
+        assert pcieutil_load_module_warning_msg not in result.get_value()
 
     @classmethod
     def teardown_class(cls):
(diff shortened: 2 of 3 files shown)


cytsao1:
@abdosi Can this be reviewed?

prgeor:
@cytsao1 please add code coverage for your changes

pcieutil/main.py
```diff
@@ -54,7 +54,7 @@ def load_platform_pcieutil():
         from sonic_platform.pcie import Pcie
         platform_pcieutil = Pcie(platform_path)
     except ImportError as e:
-        log.log_warning("Failed to load platform Pcie module. Error : {}, fallback to load Pcie common utility.".format(str(e)), True)
+        log.log_warning("Failed to load platform Pcie module. Warning : {}, fallback to load Pcie common utility.".format(str(e)))
```
prgeor, line 57: @cytsao1 Log warning won't show up in console?

cytsao1:
> please add code coverage for your changes

@prgeor Do you mean to update this test file? https://github.com/sonic-net/sonic-utilities/blob/master/tests/pcieutil_test.py
or something else? This code should be invoked by any pcieutil command.

prgeor:
> > please add code coverage for your changes
> 
> @prgeor Do you mean to update this test file? https://github.com/sonic-net/sonic-utilities/blob/master/tests/pcieutil_test.py or something else? This code should be invoked by any pcieutil command.

@cytsao1 yes

abdosi (approved):

tests/pcieutil_test.py
```diff
@@ -199,6 +202,16 @@ def test_aer_option_device(self):
...
 
+    def test_load_pcie_module_warning(self): 
+        stdout = sys.stdout
+        sys.stdout = result = StringIO()
+        try:
+            pcieutil.load_platform_pcieutil()
+        except ImportError:
+            pass
+        sys.stdout = stdout
+        assert pcieutil_load_module_warning_msg not in result.getvalue()
```
prgeor, line 213: @cytsao1 what does this ASSERT confirm? There could be platform which may not see this import Error and in which case this ASSERT may fail?

cytsao1:
Can someone with permissions merge this?

bmridul:
@abdosi @prgeor 
Pls help merge this.

abdosi:
@prgeor can we merge this ?

tests/pcieutil_test.py
prgeor: @cytsao1 `result` value is empty string at the start, so when the test run at line 209, `result` is not updated at all and at line 213 ofcourse `pcieutil_load_module_warning_msg` won't match `result` value, so i don't understand why even run `pcieutil.load_platform_pcieutils()` at line 209?

cytsao1: sys.stdout is updated to be same IO stream as result, so running load_platform_pcieutil output will be captured in result instead of output to console stdout.
After, the stdout is set back to the sys.stdout stream.

cytsao1: Example of how it would look without this PR
```
>>> import sys
>>> from io import StringIO
>>> import pcieutil.main as pcieutil
>>> stdout = sys.stdout
>>> sys.stdout = result = StringIO()
>>> pcieutil.load_platform_pcieutil()
>>> sys.stdout = stdout
>>> result.getvalue()
"Failed to load platform Pcie module. Error : No module named 'sonic_platform.pcie', fallback to load Pcie common utility.\n"
```

abdosi:
> @prgeor can we merge this ?

@prgeor you comment is addressed above. Do we need more clarification ?

tests/pcieutil_test.py
prgeor: @cytsao1 thanks for clarifying....did not realize that stdout was captured in a string output...

prgeor (approved):

PR #2787, opened 2023-04-11
Support to display the SONiC OS Version in the command show version

xumia:
#### What I did
Support to display the SONiC OS Version in the command show version.
It will be used to display the version info in the SONiC command "show version". The version is used to do the FIPS certification. We do not do the FIPS certification on a specific release, but on the SONiC OS Version.

```
SONiC Software Version: SONiC.master-13812.218661-7d94c0c28
SONiC OS Version: 11
Distribution: Debian 11.6
Kernel: 5.10.0-18-2-amd64
```

#### How I did it
The device info is in sonic-net/sonic-buildimage, see PR: https://github.com/sonic-net/sonic-buildimage/pull/14601
The submodule change can be merged to sonic-buildimage, after the PR 14601 merged.

### DIFF
--- a/show/main.py
+++ b/show/main.py
@@ -1297,6 +1297,7 @@ def version(verbose):
     sys_date = datetime.now()
 
     click.echo("\nSONiC Software Version: SONiC.{}".format(version_info['build_version']))
+    click.echo("\nSONiC OS Version: {}".format(version_info['sonic_os_version']))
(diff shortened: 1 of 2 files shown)


qiluo-msft (commented):
LGTM

xumia:
@lguohan , could you please help review it? Thanks.

lguohan (approved):

PR #2791, opened 2023-04-13
[GCU] Complete RDMA Platform Validation Checks

isabelmsft:
#### What I did
Add RDMA validator to ensure GCU is being used on allowable platform and version for RDMA use cases.
The allowable platforms and versions are determined by this table: 
image

The last two rows were completed and infrastructure set up in the earlier PR https://github.com/sonic-net/sonic-utilities/pull/2619

#### How to verify it
Attempt to use GCU on vs and modify PFC_WD table
GCU patch application improperly goes through
```
Failed to apply patch

Error: Modification of /PFC_WD table is illegal due to corresponding validating function rdma_config_update_validator returning false
```


generic_config_updater/field_operation_validators.py
```diff
@@ -20,7 +51,48 @@ def rdma_config_update_validator():
     if branch_version is None:
         return False
 
-    if asic_type == 'cisco-8000':
-        return branch_version >= "20201200"
-    else:
-        return branch_version >= "20181100"
+    # PFCWD enable/disable scenario
+    if "pfc_wd" in field:
```
neethajohn, line 55: suggest to build a dictionary of the fields and pre-populate it with all the restrictions and write a generic function to evaluate it.
something similar to the eg. below
field = {'pfcwd': {'all': '201811',
                         'cisco-8000': '202012'
                         }}
                         '

isabelmsft: Thanks for the suggestion, I added this to the gcu_field_operation_validators.conf.json file

generic_config_updater/gu_common.py
```diff
@@ -184,17 +184,18 @@ def _invoke_validating_function(cmd):
 
         for element in patch:
             path = element["path"]
+            operation = element["op"]
             match = re.search(r'\/([^\/]+)(\/|$)', path) # This matches the table name in the path, eg if path if /PFC_WD/GLOBAL, the match would be PFC_WD
             if match is not None:
-                table = match.group(1)
+                table = match.group(1).lower()
```
qiluo-msft, line 190: >lower() [](http://example.com/codeflow?start=39&length=7)

Why change to lowercases?

generic_config_updater/gcu_field_operation_validators.conf.json
```diff
@@ -10,11 +10,113 @@
...
+                        }
+                    },
+                    "PG headroom modification": {
+                        "fields": [
+                            "xoff"
+                        ],
+                        "operations": ["replace"],
+                        "platforms": {
+                            "spc1": "20191100",
+                            "td2": "",
```
qiluo-msft, line 89: >"td2": "", [](http://example.com/codeflow?start=28&length=10)

I see "NA" and "Unknown" in the table, but they are both implemented as "". Is it a concern?

generic_config_updater/field_operation_validators.py
```diff
@@ -1,12 +1,50 @@
...
+GCU_TABLE_MOD_CONF_FILE = f"{SCRIPT_DIR}/gcu_field_operation_validators.conf.json"
+
+def get_asic_name():
+    asic = "unknown"
+    
+    if device_info.get_sonic_version_info()['asic_type'] == 'cisco-8000':
+        asic = "cisco-8000"
+    elif device_info.get_sonic_version_info()['asic_type'] == 'mellanox':
+        GET_HWSKU_CMD = "sonic-cfggen -d -v DEVICE_METADATA.localhost.hwsku"
+        spc1_hwskus = [ 'ACS-MSN2700', 'ACS-MSN2740', 'ACS-MSN2100', 'ACS-MSN2410', 'ACS-MSN2010', 'Mellanox-SN2700', 'Mellanox-SN2700-D48C8' ]
```
qiluo-msft, line 18: >spc1_hwskus [](http://example.com/codeflow?start=8&length=11)

These string are hard-coded so not future proof. Better to move into gcu_field_operation_validators.conf.json.

generic_config_updater/field_operation_validators.py
```diff
@@ -1,12 +1,50 @@
...
+        proc = subprocess.Popen(GET_HWSKU_CMD, shell=True, universal_newlines=True, stdout=subprocess.PIPE)
+        output, err = proc.communicate()
+        hwsku = output.rstrip('\n')
+        if hwsku.lower() in [spc1_hwsku.lower() for spc1_hwsku in spc1_hwskus]:
+            asic = "spc1"
+    elif device_info.get_sonic_version_info()['asic_type'] == 'broadcom':
+        command = ["sudo", "lspci"]
```
qiluo-msft, line 25: >command [](http://example.com/codeflow?start=8&length=7)

Runtime check is heavy. We could build the mapping offline, and move add the hareware skus into gcu_field_operation_validators.conf.json.

Then in runtime, you just check hwsku.

generic_config_updater/field_operation_validators.py
```diff
@@ -1,12 +1,50 @@
...
+    
+    return asic
 
-def rdma_config_update_validator():
+def rdma_config_update_validator(path, operation):
     version_info = device_info.get_sonic_version_info()
     build_version = version_info.get('build_version')
-    asic_type = version_info.get('asic_type')
+    asic = get_asic_name()
+    path = path.lower()
```
qiluo-msft, line 43: >lower [](http://example.com/codeflow?start=16&length=5)

Why change to lowercases?

generic_config_updater/field_operation_validators.py
```diff
@@ -1,12 +1,50 @@
...
 
-    if (asic_type != 'mellanox' and asic_type != 'broadcom' and asic_type != 'cisco-8000'):
+    # For paths like /BUFFER_PROFILE/pg_lossless_50000_300m_profile/xoff, remove pg_lossless_50000_300m from the path so that we can clearly determine which fields are modifiable
+    cleaned_path = "/".join([part for part in path.split("/") if not any(char.isdigit() for char in part)])
+    if asic == "unknown":
```
qiluo-msft, line 47: >== [](http://example.com/codeflow?start=12&length=2)

You can compare and return earlier.

generic_config_updater/field_operation_validators.py
```diff
@@ -1,12 +1,50 @@
...
-def rdma_config_update_validator():
+def rdma_config_update_validator(path, operation):
     version_info = device_info.get_sonic_version_info()
     build_version = version_info.get('build_version')
-    asic_type = version_info.get('asic_type')
+    asic = get_asic_name()
+    path = path.lower()
 
-    if (asic_type != 'mellanox' and asic_type != 'broadcom' and asic_type != 'cisco-8000'):
+    # For paths like /BUFFER_PROFILE/pg_lossless_50000_300m_profile/xoff, remove pg_lossless_50000_300m from the path so that we can clearly determine which fields are modifiable
```
qiluo-msft, line 45: >determine [](http://example.com/codeflow?start=141&length=9)

Not quite understand why removing help determine.

generic_config_updater/field_operation_validators.py
```diff
@@ -1,12 +1,50 @@
...
+    cleaned_path = "/".join([part for part in path.split("/") if not any(char.isdigit() for char in part)])
```
qiluo-msft, line 46: >split [](http://example.com/codeflow?start=51&length=5)

Manually split or join is dangerous, you do not consider escaping, etc. Suggest leverage some libraries, such as [https://pypi.org/project/jsonpointer/](https://pypi.org/project/jsonpointer/).

generic_config_updater/field_operation_validators.py
```diff
@@ -20,7 +58,38 @@ def rdma_config_update_validator():
...
 
-    if asic_type == 'cisco-8000':
-        return branch_version >= "20201200"
+    if os.path.exists(GCU_TABLE_MOD_CONF_FILE):
+        with open(GCU_TABLE_MOD_CONF_FILE, "r") as s:
+            gcu_field_operation_conf = json.load(s)
+    else:
+        raise GenericConfigUpdaterError("GCU table modification validators config file not found")
+
+    match = re.search(r'\/([^\/]+)(\/|$)', cleaned_path) # This matches the table name in the path, eg if path if /PFC_WD/GLOBAL, the match would be PFC_WD
```
qiluo-msft, line 67: >search [](http://example.com/codeflow?start=15&length=6)

Suggest use library instead of manually parsing.

generic_config_updater/gu_common.py
isabelmsft: I changed it back to default text which would be upper case, and also changed the config file we are matching against to have table names in upper case

generic_config_updater/gcu_field_operation_validators.conf.json
isabelmsft: It should not be a concern because we agreed that both "NA" and "Unknown" are considered unsupported. If the code reads "", we return False because this change is unsupported

generic_config_updater/field_operation_validators.py
isabelmsft: Moved to gcu_field_operation_validators.conf.json.

generic_config_updater/field_operation_validators.py
isabelmsft: Moved mapping to gcu_field_operation_validators.conf.json.

generic_config_updater/field_operation_validators.py
isabelmsft: For consistency, because all fields/paths other than the table in the conf file are lowercase. But in ConfigDB, some fields are uppercase (like AZURE_LOSSLESS)

generic_config_updater/field_operation_validators.py
isabelmsft: Moved comparison to the start of validator function

generic_config_updater/field_operation_validators.py
isabelmsft: Changed to use jsonpointer library

generic_config_updater/field_operation_validators.py
isabelmsft: Changed to use jsonpointer library

qiluo-msft:
Need to change code according to Neetha's test requirements.

qiluo-msft (approved):

generic_config_updater/gcu_field_operation_validators.conf.json
```diff
@@ -10,11 +10,128 @@
...
+            "field_operation_validators": [ "generic_config_updater.field_operation_validators.rdma_config_update_validator" ],
+            "validator_data": {
+                "rdma_config_update_validator": {
+                    "Dynamic threshold tuning": {
+                        "fields": [
+                            "dynamic_th"
+                        ],
+                        "operations": ["replace"],
+                        "platforms": {
+                            "spc1": "20181100",
```
wen587, line 88: Do we still need to check build number less than 20220531? Because GCU support is introduced in 202205.
It by default satisfies this rule.

tests/generic_config_updater/field_operation_validator_test.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+class TestValidateFieldOperation(unittest.TestCase):
+
+    @patch("generic_config_updater.field_operation_validators.get_asic_name", mock.Mock(return_value="unknown"))
+    def test_rdma_config_update_validator_unknown_asic(self):
+        patch_element = {"path": "/PFC_WD/Ethernet4/restoration_time", "op": "replace", "value": "234234"}
+        assert generic_config_updater.field_operation_validators.rdma_config_update_validator(patch_element) == False
+
+    @patch("sonic_py_common.device_info.get_sonic_version_info", mock.Mock(return_value={"build_version": "SONiC.20171131"}))
```
wen587, line 22: This should never happen in a real device. If it could use GCU feature, its  build should be 202205+

generic_config_updater/gcu_field_operation_validators.conf.json
isabelmsft: That's true, but I have this information in the conf file for completion and clarity. 

I left the comparison in field_operation_validators.py as-is because this version comparison is always going to be necessary for cases only supported in 202211+. Do you think it's better to not define lower versions in the conf file, and only run this version comparison check if the higher version is defined in the conf file? I figure for this implementation, I would have to add code just to check if the version is defined, which may be an extra unnecessary step.

tests/generic_config_updater/field_operation_validator_test.py
isabelmsft: Thanks, I changed to test BUFFER_POOL feature, which has restriction of 202211+

qiluo-msft: Suggest to implement the limitation according to the ground truth in the above table, which has no relation to GCU feature at all.

In theory GCU feature could be backported, then we do not need to change the logic in this PR again.

isabelmsft: That should be the current implementation in the feature code. I just changed this test case to reflect an actual production scenario with current GCU implementation only in 202205+

qiluo-msft (approved):

wen587 (dismissed):

qiluo-msft (approved):

qiluo-msft (approved):

StormLiangMS:
@isabelmsft revert this PR since it failed the PR test and blocked the submodule advance of sonic-utilities, pls check the issue and re-submit.

PR #2804, opened 2023-04-19
[sonic_cli_gen] Add YANG refine support

vadymhlushko-mlnx:
#### What I did
Add [YANG "refine"](https://www.rfc-editor.org/rfc/rfc7950.html#section-7.13.2) support for the `sonic_cli_gen` utility.
YANG refine is used to override the `description` and `mandatory` statements for YANG `grouping`.

#### How I did it
Modify the `sonic_cli_gen/yang_parser.py`

#### How to verify it
Modify the `sonic_cli_gen` UT


liat-grozovik:
@qiluo-msft kindly reminder to review.

liat-grozovik:
@qiluo-msft any comments or we can go a head and merge?

sonic_cli_gen/yang_parser.py
```diff
@@ -401,10 +402,10 @@ def get_mandatory(y_leaf: OrderedDict) -> bool:
             'leaf' 'mandatory' value
     """
 
-    if y_leaf.get('mandatory') is not None:
-        return True
+    if y_leaf.get('mandatory') is None:
```
qiluo-msft, line 405: >None [](http://example.com/codeflow?start=34&length=4)

Thanks for this bug fix! Could you separate it into a standalone PR?

praveen-li: Logic of both code is same. 
Though, assume  we have to wrap it in try-catch. In that case, what should be return value from catch block i.e. in exception case? 

-- If false then current block is good.
-- If true then new block is good.

I prefer, false, because by default, each node is not mandatory.

zhangyanzhao:
@qiluo-msft  will further check and merge. Thanks.

qiluo-msft (approved):

wen587 (approved):

StormLiangMS:
enhancement, no for cherry-pick

dgsudharsan:
@StormLiangMS  This is not a feature enhancement but required for yang models that are backported as fix to use refine keyword. Please help to cherry-pick

dgsudharsan:
@StormLiangMS  This change was not cherry-picked to 202305 but the label was added. Can you please check?

PR #2806, opened 2023-04-20
LAG keepalive script to reduce lacp session wait during warm-reboot

vaibhavhd:
#### What I did

A new mechanism is added here to to reduce LAG flap issue during hitless upgrades.

####  Problem being solved:

During warm upgrades T0 goes down and with that wait time for LACP session starts.
If the waittime to refresh LACP session is > 90s then T1 initiates LAG teardown, and as a result dataplane impact is seen.
This script makes sure that LACPDUs are sent in the going down path continuously.

How time is saved w/ this mechanism:

The lacpsession wait period earlier used to start from when teamd container goes down.
New lacpsession wait period starts when kexec in current kernel is issued, and new kernel boots up.

#### Implementation:

When warm-reboot starts, capture LACPDUs sent from all LAG member ports.
For this allow 60s of prep + collection time.
Start sending LACPDUs w/ ~1s interval.
The last LACPDU is sent after all containers are down and kexec is issued.

####  Results:

Tested this on different platforms and images. Some results for time saved:

BRCM: 201811 -> 202012 --- 18s
BRCM: 202012 -> 202012 --- 20s
MLNX: 201911 -> 202205 --- 10s

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -17,6 +17,7 @@ STRICT=no
 REBOOT_METHOD="/sbin/kexec -e"
 ASSISTANT_IP_LIST=""
 ASSISTANT_SCRIPT="/usr/local/bin/neighbor_advertiser"
+LAG_KEEPALIVE_SCRIPT="/usr/local/bin/lag_keepalive.py"
 WATCHDOG_UTIL="/usr/local/bin/watchdogutil"
 DEVPATH="/usr/share/sonic/device"
 PLATFORM=$(sonic-cfggen -H -v DEVICE_METADATA.localhost.platform)
@@ -681,6 +682,13 @@ fi
 
 # disable trap-handlers which were set before
 trap '' EXIT HUP INT QUIT TERM KILL ABRT ALRM
+
+# start sending LACPDUs to keep the LAGs refreshed
+# this is a non-blocking call, and the process will die in 300s
+debug "Starting lag_keepalive to send LACPDUs ..."
+timeout 300 python ${LAG_KEEPALIVE_SCRIPT} &
+# give the lag_keepalive script a chance to get ready (30s) and collect one lacpdu before going down (30s)
+sleep 60
 
 if [ -x ${LOG_SSD_HEALTH} ]; then
     debug "Collecting logs to check ssd health before ${REBOOT_TYPE}..."
--- a/scripts/lag_keepalive.py
+++ b/scripts/lag_keepalive.py
@@ -79,10 +79,10 @@ def main():
     while True:
         try:
             active_lag_members, lag_member_to_packet = get_lacpdu_per_lag_member()
-            if len(active_lag_members) != len(active_lag_members):
+            if len(active_lag_members) != len(lag_member_to_packet.keys()):
                 log_error("Failed to capture LACPDU packets for some lag members. " +\
                 "Active lag members: {}. LACPDUs captured for: {}".format(
-                    active_lag_members, active_lag_members.keys()))
+                    active_lag_members, lag_member_to_packet.keys()))
 
             log_info("ready to send LACPDU packets via {}".format(lag_member_to_packet.keys()))
         except Exception:
--- a/scripts/lag_keepalive.py
+++ b/scripts/lag_keepalive.py
@@ -1,0 +1,102 @@
+#!/usr/bin/env python3
+
+from scapy.config import conf
+conf.ipv6_enabled = False
+from scapy.all import sendp, sniff
+from swsssdk import ConfigDBConnector
+import time, threading, traceback
+import syslog
+
+SYSLOG_ID = 'lag_keepalive'
+
+
+def log_info(msg):
+    syslog.openlog(SYSLOG_ID)
+    syslog.syslog(syslog.LOG_INFO, msg)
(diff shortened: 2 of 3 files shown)


vaibhavhd:
@stepanblyschak please review. You might find this interesting, since you are also focused on reducing lacp session wait period during warm reboot. This is little hacky, but is taking advantage of the fact that packets can still be sent even if teamd and syncd are down, as the physical ports are still up, and sockets can still function.

It would help if you also test this in your setup and share the time saved in your tests.
This new mechanism saves ~10s for Nvidia platform where I tested.

scripts/lag_keepalive.py
```diff
@@ -0,0 +1,102 @@
...
+                continue
+        log_info("sent LACPDU packets via {}".format(lag_member_to_packet.keys()))
+        time.sleep(1)
+
+
+def main():
+    while True:
+        try:
+            active_lag_members, lag_member_to_packet = get_lacpdu_per_lag_member()
+            if len(active_lag_members) != len(active_lag_members):
```
saiarcot895, line 82: I think there's a typo here and in the log formatting below?

stepanblyschak:
@vaibhavhd This is an interesting finding. The original motivation of stopping teamd prior to syncd was to have an ability to send last LACPDU before syncd removes netdevs from kernel. Since, in your observation, kernel netdevs live until kexec, I believe this is no longer true. At least for Nvidia it is not true since - https://github.com/sonic-net/sonic-buildimage/pull/2572. But I did not know if other vendor syncd stop removes netdevs or not. Can we swap the teamd, syncd shutdown order to achive similar results?

BTW, a bit more general question. I know you're using "control plane assistant" to reply on ARP/NDP while DUT undergoes warm-reboot. Have you considered the same for LACP?

vaibhavhd:
> @vaibhavhd This is an interesting finding. The original motivation of stopping teamd prior to syncd was to have an ability to send last LACPDU before syncd removes netdevs from kernel. Since, in your observation, kernel netdevs live until kexec, I believe this is no longer true. At least for Nvidia it is not true since - [sonic-net/sonic-buildimage#2572](https://github.com/sonic-net/sonic-buildimage/pull/2572). But I did not know if other vendor syncd stop removes netdevs or not. Can we swap the teamd, syncd shutdown order to achive similar results?
> 
> BTW, a bit more general question. I know you're using "control plane assistant" to reply on ARP/NDP while DUT undergoes warm-reboot. Have you considered the same for LACP?

@stepanblyschak , I tried 201911 image on 2700 and I still see that the last lacpdu is sent when kexec is issued. Can you re-evaluate on your end? I think even if teamd/syncd sequence is swapped, this hack remains helpful as it can send LACPDU even after teamd is down.

scripts/lag_keepalive.py
vaibhavhd: Thanks for catching! Addressed now.

scripts/fast-reboot
```diff
@@ -682,6 +683,13 @@ fi
 # disable trap-handlers which were set before
 trap '' EXIT HUP INT QUIT TERM KILL ABRT ALRM
 
+# start sending LACPDUs to keep the LAGs refreshed
+# this is a non-blocking call, and the process will die in 300s
+debug "Starting lag_keepalive to send LACPDUs ..."
+timeout 300 python ${LAG_KEEPALIVE_SCRIPT} &
+# give the lag_keepalive script a chance to get ready (30s) and collect one lacpdu before going down (30s)
+sleep 60
```
stepanblyschak, line 691: Can we run the lag_keepalive.py script in the foreground and in the lag_keepalive.py script do fork() and run lag_keepalive() with a timeout? This way we only wait until LACPDUs are collected, probably much less than 60 sec.

stepanblyschak (commented):
I've tested these changed on 202211 and observe 9-10 sec improvement on Nvidia. Thanks.

scripts/fast-reboot
moshemos: @vaibhavhd  this is critical issue to our 202211 release, can you take care for review and merge?

scripts/lag_keepalive.py
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/env python3
+
+from scapy.config import conf
+conf.ipv6_enabled = False
+from scapy.all import sendp, sniff
+from swsssdk import ConfigDBConnector
```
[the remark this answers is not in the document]
vaibhavhd: Addressed this, thanks for catching it.

scripts/fast-reboot
vaibhavhd: @stepanblyschak I did not prefer to run this on foreground to avoid any odd chance of getting hung by the called process.

Additionally, the wait here has to be minimum 30s to collect LACPDU in worst case. The additional 30s is just a buffer. In my observation some platforms are very slow and just importing `from scapy.all import sendp, sniff` takes around 10-15s.
This wait can be optimized (if possible) in future PRs. What do you think?

stepanblyschak: @vaibhavhd Ok, agree

saiarcot895 (approved):

moshemos:
@vaibhavhd can you cherry pick to 202211?

PR #2808, opened 2023-04-25
Added platform plugin support in load_minigraph

anamehra:
Signed-off-by: anamehra anamehra@cisco.com

#### What I did
Added platform plugin hook to execute platform script after minigraph config is loaded in config db and before starting the services.

#### Why I did it
For the phy-based platforms, changing the port speed via minigraoh, e.g. 400G to 100G, also requires gearbox/phy config files to be modified in hwsku directory. Current gearbox config change is not supported by config db or minigraph and that causes a mismatch in speed configuration between config db and gearbox config. To handle this scenario, a platform plugin is added which could do the gearbox config modifications as required based on the minigraph config.

#### How I did it
Call platform plugin script, if present, before starting the services

#### How to verify it
Add a platform plugin script, platform_mg_post_check  in <platform>/plugins dir
Call config load_minigraph and validate if the script is invoked.
The following message is printed if the plugin script is present:
Running Platform plugin ............!


anamehra:
@abdosi , please review for load_minigraph platform plugin support

abdosi (approved):

anamehra:
202205 PR: https://github.com/sonic-net/sonic-utilities/pull/2831

gechiang:
Removed the "Chassis for 202205 branch" because this was already picked up in 202205 2 weeks ago and this label should be removed to not cause confusion...

PR #2816, opened 2023-05-02
[sonic_sku_create] remove shell=True, replace exit() with sys.exit()

maipbui:
#### What I did
Ref:
#### How I did it
#### How to verify it
Add UT

### DIFF
--- a/scripts/sonic_sku_create.py
+++ b/scripts/sonic_sku_create.py
@@ -263,7 +263,7 @@ class SkuCreate(object):
             pattern = '^Ethernet([0-9]{1,})'
             m = re.match(pattern,key)
             if m is None:
-                print("Port Name ",port_name, " is not valid, Exiting...", file=sys.stderr) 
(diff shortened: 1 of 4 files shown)


scripts/sonic_sku_create.py
```diff
@@ -263,8 +263,8 @@ def json_file_parser(self, json_file):
             pattern = '^Ethernet([0-9]{1,})'
             m = re.match(pattern,key)
             if m is None:
-                print("Port Name ",port_name, " is not valid, Exiting...", file=sys.stderr) 
-                exit(1)
+                print("Port Name ", key, " is not valid, Exiting...", file=sys.stderr)
```
qiluo-msft, line 266: >key [](http://example.com/codeflow?start=36&length=3)

Thanks for fixing the bug!

qiluo-msft:
@liat-grozovik This script is original developed by @madhanmellanox. Could you propose new reviewers?

qiluo-msft (approved):

PR #2824, opened 2023-05-04
[GCU] fix rdma check failure

wen587:
#### What I did
`if substring.isdigit() and re.match(r'^\d{8}$', substring):` In this version check, it only applies to this pattern: 20220531.XXX. It will fail the master and other build check. For example, `master.261851-010dc3957` is a master build_version string. It will fail version validation and return failure.

My thought to remove the check is that GCU is only supported in 202205 and later. Either `branch_version >= "20201200"` or `branch_version >= "20181100"` is already satisfied.
#### How I did it
Remove redundant check in GCU RDMA validator
#### How to verify it
Unit test

### DIFF
--- a/generic_config_updater/field_operation_validators.py
+++ b/generic_config_updater/field_operation_validators.py
@@ -3,24 +3,8 @@ import re
 
 def rdma_config_update_validator():
     version_info = device_info.get_sonic_version_info()
-    build_version = version_info.get('build_version')
     asic_type = version_info.get('asic_type')
 
     if (asic_type != 'mellanox' and asic_type != 'broadcom' and asic_type != 'cisco-8000'):
         return False
-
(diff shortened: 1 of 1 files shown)


isabelmsft:
Just requested review from @wen587 on the related PR https://github.com/sonic-net/sonic-utilities/pull/2791/files 
where we have the limiting version 202211, so we do still need to check branch version. 

But we do need to improve the check to consider build versions of the form `master.261851-010dc3957`

isabelmsft (approved):
Approving this PR, and I will improve the related PR https://github.com/sonic-net/sonic-utilities/pull/2791/files to account for build versions of the form `master.261851-010dc3957` and check for 202211+ versions where necessary.

qiluo-msft (approved):

PR #2832, opened 2023-05-10
Revert "[chassis]: remote cli commands infra for sonic chassis (#2701)"

arlakshm:
This reverts commit 7e24463f88ba64f485e8ae5c33e00ab3b152e073.

#### What I did
Revert #2701 because of arm build issues.


liat-grozovik (approved):

liat-grozovik:
@lguohan can we merge although the semgrep failure?
@arlakshm i believe we should also revert on 202211 as it is also included in that branch

PR #2835, opened 2023-05-13
[NTP] Configure NTP according to extended configuration

fastiuk:
hld [#1219](https://github.com/sonic-net/SONiC/pull/1296)
closes [#1171](https://github.com/sonic-net/SONiC/issues/1254)

#### What I did
Fix NTP CLI according to new extended configuration abilities

#### How I did it
Additionally configure `association_type` and `resolve_as` when adding new NTP servers.

#### How to verify it
* Add a new server
* Show NTP sync `show ntp`. It should show you are sync with new server
* Delete NTP server
* Show NTP sync `show ntp`. It should show you are unsync

Not changed


config/main.py
```diff
@@ -6515,7 +6515,9 @@ def add_ntp_server(ctx, ntp_ip_address):
         return
     else:
         try:
-            db.set_entry('NTP_SERVER', ntp_ip_address, {'NULL': 'NULL'})
+            db.set_entry('NTP_SERVER', ntp_ip_address,
+                         {'resolve_as': ntp_ip_address,
+                          'association_type': 'server'})
```
saiarcot895, line 6520: Just to verify, this would mean that pool NTP servers can be configured only via `config_db.json` directly, right?

fastiuk: Yes, right

saiarcot895 (approved):

bsun-sudo:
Just want to make sure. Should this PR support all new configurations from the [HLD](https://github.com/sonic-net/SONiC/pull/1296), or there will be a separate PR?

fastiuk:
> Just want to make sure. Should this PR support all new configurations from the [HLD](https://github.com/sonic-net/SONiC/pull/1296), or there will be a separate PR?

No, the new commands won't be supported by this PR

liat-grozovik:
could it be that the NTP command now looks different?
if so, https://github.com/sonic-net/sonic-utilities/blob/master/doc/Command-Reference.md#ntp-config-commands should be updated and aligned with the extra options

liat-grozovik (approved):

PR #2836, opened 2023-05-16
[config] Generate sysinfo as needed when override config

wen587:
ADO: 17921518

#### What I did
The generated Golden Config will not have knowledge of configs that are produced in run time, such as mac and platform.
Generate that info in Override Config if missing.
#### How I did it
Reuse the mac and platform in existing device runnning config and generate that if missing.
#### How to verify it
Unit test

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1824,6 +1824,43 @@ def override_config_by(golden_config_path):
     return
 
 
+def generate_sysinfo(cur_config, config_input, ns=None):
+    # Generate required sysinfo for Golden Config.
+    device_metadata = config_input.get('DEVICE_METADATA')
+
+    if not device_metadata or 'localhost' not in device_metadata:
+        return
+
+    mac = None
+    platform = None
+    cur_device_metadata = cur_config.get('DEVICE_METADATA')
+
+    # Reuse current config's mac and platform. Generate if absent
+    if cur_device_metadata is not None:
+        mac = cur_device_metadata.get('localhost', {}).get('mac')
+        platform = cur_device_metadata.get('localhost', {}).get('platform')
+
+    if not mac or not platform:
+        if ns:
+            asic_role = device_metadata.get('localhost', {}).get('sub_role')
+            switch_type = device_metadata.get('localhost', {}).get('switch_type')
+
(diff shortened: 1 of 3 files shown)


config/main.py
```diff
@@ -1824,6 +1824,43 @@ def override_config_by(golden_config_path):
     return
 
 
+def generate_sysinfo(cur_config, config_input, ns=None):
```
qiluo-msft, line 1827: >generate_sysinfo [](http://example.com/codeflow?start=4&length=16)

Could you add some function level comments?
At least mention that this function will change `config_input`, which is important to code reader.

config/main.py
```diff
@@ -1824,6 +1824,43 @@ def override_config_by(golden_config_path):
...
+    mac = None
+    platform = None
+    cur_device_metadata = cur_config.get('DEVICE_METADATA')
+
+    # Reuse current config's mac and platform. Generate if absent
+    if cur_device_metadata is not None:
+        mac = cur_device_metadata.get('localhost', {}).get('mac')
+        platform = cur_device_metadata.get('localhost', {}).get('platform')
+
+    if not mac or not platform:
```
qiluo-msft, line 1843: >if not mac or not platform: [](http://example.com/codeflow?start=4&length=27)

Could you separate mac and platform into 2 if-blocks? They seems not relating each other.

config/main.py
```diff
@@ -1863,8 +1900,10 @@ def override_config_table(db, input_config_db, dry_run):
             # Golden Config will use "localhost" to represent host name
             if ns == DEFAULT_NAMESPACE:
                 ns_config_input = config_input["localhost"]
+                generate_sysinfo(current_config, ns_config_input, ns)
             else:
                 ns_config_input = config_input[ns]
+                generate_sysinfo(current_config, ns_config_input, ns)
```
qiluo-msft, line 1906: >generate_sysinfo(current_config, ns_config_input, ns) [](http://example.com/codeflow?start=16&length=53)

Possible to move outside if-else? so code will be simplified a little bit.

qiluo-msft (approved):

wen587:
local test and can be clean picked to 202205.

PR #2839, opened 2023-05-17
[db_migrator] Add migration of FLEX_COUNTER_DELAY_STATUS during 1911->master upgrade + fast-reboot. Add UT.

vadymhlushko-mlnx:
#### What I did
Add migration of `FLEX_COUNTER_DELAY_STATUS` attribute of config_db `FLEX_COUNTER_TABLE` during the `SONiC to SONiC upgrade + fast-reboot` from older versions `201911 -> master`.

This change is required for the `fast-reboot` procedure because without it the counters will be created during the init flow which will waste a lot of resources and cause data plane degradation of more than 30 seconds.

#### How I did it
Modify the `db_migrator.py`.

#### How to verify it
Add UT.

### DIFF
--- a/scripts/db_migrator.py
+++ b/scripts/db_migrator.py
@@ -650,6 +650,20 @@ class DBMigrator():
                 # Set new cable length values
                 self.configDB.set(self.configDB.CONFIG_DB, "CABLE_LENGTH|AZURE", intf, EDGEZONE_AGG_CABLE_LENGTH)
 
+    def migrate_config_db_flex_counter_delay_status(self):
+        """
+        Migrate "FLEX_COUNTER_TABLE|*": { "value": { "FLEX_COUNTER_DELAY_STATUS": "false" } }
+        Set FLEX_COUNTER_DELAY_STATUS true in case of fast-reboot
+        """
+
+        flex_counter_objects = self.configDB.get_keys('FLEX_COUNTER_TABLE')
+        for obj in flex_counter_objects:
+            flex_counter = self.configDB.get_entry('FLEX_COUNTER_TABLE', obj)
+            delay_status = flex_counter.get('FLEX_COUNTER_DELAY_STATUS')
+            if delay_status is None or delay_status == 'false':
+                flex_counter['FLEX_COUNTER_DELAY_STATUS'] = 'true'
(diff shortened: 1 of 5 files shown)


stepanblyschak (approved):

keboliu:
@vaibhavhd would you please help to review?

scripts/db_migrator.py
```diff
@@ -811,6 +825,10 @@ def version_2_0_2(self):
         This is the latest version for 202012 branch
         """
         log.log_info('Handling version_2_0_2')
+
+        if self.stateDB.keys(self.stateDB.STATE_DB, "FAST_REBOOT|system"):
+            self.migrate_config_db_flex_counter_delay_status()
```
vaibhavhd, line 830: Why has this logic changed 202012 specific version? AFAIK, the change is needed for 202211 and above, so only branch specific versions should be changed. This will likely also require separate PRs for feature branches.

Also this is missing call to set a new version and also a placeholder function for new version.

vaibhavhd (commented):
Added a blocking comment to change vesions only for feature branches where this migration is relevant.

scripts/db_migrator.py
vadymhlushko-mlnx: You are right, we need these changes for every branch starting from 202111, 202205, 202211, master
I will create a separate PR's

vadymhlushko-mlnx: Here are the PR's:

[[202211][db_migrator] Add migration of FLEX_COUNTER_DELAY_STATUS during 1911->2211 upgrade + fast-reboot. Add UT.](https://github.com/sonic-net/sonic-utilities/pull/2838)
[[202205][db_migrator] Add migration of FLEX_COUNTER_DELAY_STATUS during 1911->2205 upgrade + fast-reboot. Add UT.](https://github.com/sonic-net/sonic-utilities/pull/2870)
[[202111][db_migrator] Add migration of FLEX_COUNTER_DELAY_STATUS during 1911->2111 upgrade + fast-reboot. Add UT.](https://github.com/sonic-net/sonic-utilities/pull/2869)

liat-grozovik:
@vaibhavhd could you please approve so we can merge?

vaibhavhd (approved):

PR #2847, opened 2023-05-24
Fix issue: out of range sflow polling interval is accepted and stored in config_db

Junchao-Mellanox:
#### What I did

Fixed issue: out of range sflow polling interval is accepted and stored in config_db.

Reproduce step:
```
1. Enable sflow feature:   config feature state sflow enabled
2. Enable sflow itself:   config sflow enable 
3. Configure out of range polling interval:  config sflow polling-interval 1. Error message is shown as expected
4. Save config:    config save -y
5. Check "SFLOW" section inside config_db
```

As the interval is invalid, the expected behavior is that the interval is not saved to redis. But we see the invalid value was written to redis.

#### How I did it

Change `click.echo` to `ctx.fail`

#### How to verify it

1. Manual test
2. Add a check in existing unit test case to cover the change

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -6506,7 +6506,7 @@ def polling_int(ctx, interval):
     """Set polling-interval for counter-sampling (0 to disable)"""
     if ADHOC_VALIDATION:
         if interval not in range(5, 301) and interval != 0:
-            click.echo("Polling interval must be between 5-300 (0 to disable)")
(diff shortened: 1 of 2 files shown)


keboliu (approved):

qiluo-msft (approved):

Junchao-Mellanox:
Hi @StormLiangMS , could you please help cherry-pick to 202211?

PR #2857, opened 2023-05-31
Complete Platform Validator PR

isabelmsft:
Original platform validator PR https://github.com/sonic-net/sonic-utilities/pull/2791/checks?check_run_id=12708751588 was merged and reverted due to unexpected test case behavior blocking submodule advance. The test case was fixed through https://github.com/sonic-net/sonic-mgmt/pull/8469. Now, this platform validator PR brings back the original changes

### DIFF
--- a/generic_config_updater/field_operation_validators.py
+++ b/generic_config_updater/field_operation_validators.py
@@ -1,10 +1,117 @@
-from sonic_py_common import device_info
+import os
 import re
+import json
+import jsonpointer
+import subprocess
+from sonic_py_common import device_info
+from .gu_common import GenericConfigUpdaterError
 
-def rdma_config_update_validator():
-    version_info = device_info.get_sonic_version_info()
-    asic_type = version_info.get('asic_type')
 
-    if (asic_type != 'mellanox' and asic_type != 'broadcom' and asic_type != 'cisco-8000'):
+SCRIPT_DIR = os.path.dirname(os.path.realpath(__file__))
+GCU_TABLE_MOD_CONF_FILE = f"{SCRIPT_DIR}/gcu_field_operation_validators.conf.json"
+
+def get_asic_name():
+    asic = "unknown"
+    
+    if os.path.exists(GCU_TABLE_MOD_CONF_FILE):
+        with open(GCU_TABLE_MOD_CONF_FILE, "r") as s:
+            gcu_field_operation_conf = json.load(s)
+    else:
+        raise GenericConfigUpdaterError("GCU table modification validators config file not found")
+    
+    asic_mapping = gcu_field_operation_conf["helper_data"]["rdma_config_update_validator"]
+    
+    if device_info.get_sonic_version_info()['asic_type'] == 'cisco-8000':
+        asic = "cisco-8000"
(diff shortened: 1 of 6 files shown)


qiluo-msft (approved):

generic_config_updater/field_operation_validators.py
```diff
@@ -1,10 +1,119 @@
...
+            gcu_field_operation_conf = json.load(s)
+    else:
+        raise GenericConfigUpdaterError("GCU table modification validators config file not found")
+    
+    asic_mapping = gcu_field_operation_conf["helper_data"]["rdma_config_update_validator"]
+    
+    if device_info.get_sonic_version_info()['asic_type'] == 'cisco-8000':
+        asic = "cisco-8000"
+    elif device_info.get_sonic_version_info()['asic_type'] == 'vs':
+        asic = "td2"
```
qiluo-msft, line 27: >asic [](http://example.com/codeflow?start=8&length=4)

This is not always true. Please check [https://github.com/sonic-net/sonic-buildimage/blob/220ea74cbb7e5cc783a27d7ca9f966899d81b7b3/platform/vs/docker-sonic-vs/Dockerfile.j2#L176](https://github.com/sonic-net/sonic-buildimage/blob/220ea74cbb7e5cc783a27d7ca9f966899d81b7b3/platform/vs/docker-sonic-vs/Dockerfile.j2#L176)

isabelmsft: Updated the PR to fully consider vs hwskus

generic_config_updater/field_operation_validators.py
```diff
@@ -1,10 +1,116 @@
...
+        output, err = proc.communicate()
+        hwsku = output.rstrip('\n')
+        if asic_type == 'mellanox' or asic_type == 'vs':
+            spc1_hwskus = asic_mapping["mellanox_asics"]["spc1"]
+            if hwsku.lower() in [spc1_hwsku.lower() for spc1_hwsku in spc1_hwskus]:
+                asic = "spc1"
+        if asic_type == 'broadcom' or asic_type == 'vs':
+            broadcom_asics = asic_mapping["broadcom_asics"]
+            for asic_shorthand, hwskus in broadcom_asics.items():
+                if asic != "unknown":
```
qiluo-msft, line 39: >asic [](http://example.com/codeflow?start=19&length=4)

Should you check unknown much earlier?

isabelmsft: asic is set to "unknown" at the start of the function. It remains unknown until one of the specified asic types in the table is identified. The purpose of this particular check is to stop iterating over all possible broadcom asics once the asic is already identified.

qiluo-msft: If it is already identified as spc1, you can return immediately.

isabelmsft: Changed to return spc1 immediately

qiluo-msft (approved):

isabelmsft:
I will open a new PR for 202205 branch, as this is not a clean cherry-pick

PR #2869, opened 2023-06-09
[202111][db_migrator] Add migration of FLEX_COUNTER_DELAY_STATUS during 1911->2111 upgrade + fast-reboot. Add UT.

vadymhlushko-mlnx:
#### What I did
Add migration of `FLEX_COUNTER_DELAY_STATUS` attribute of config_db `FLEX_COUNTER_TABLE` during the `SONiC to SONiC upgrade + fast-reboot` from older versions `201911 -> 202111`.

This change is required for the `fast-reboot` procedure because without it the counters will be created during the init flow which will waste a lot of resources and cause data plane degradation of more than 30 seconds.

#### How I did it
Modify the `db_migrator.py`.

#### How to verify it
Add UT.

### DIFF
--- a/scripts/db_migrator.py
+++ b/scripts/db_migrator.py
@@ -44,7 +44,7 @@ class DBMigrator():
                      none-zero values.
               build: sequentially increase within a minor version domain.
         """
-        self.CURRENT_VERSION = 'version_2_0_4'
+        self.CURRENT_VERSION = 'version_2_0_5'
 
         self.TABLE_NAME      = 'VERSIONS'
         self.TABLE_KEY       = 'DATABASE'
--- a/scripts/db_migrator.py
+++ b/scripts/db_migrator.py
@@ -447,6 +447,20 @@ class DBMigrator():
         self.migrate_qos_db_fieldval_reference_remove(qos_table_list, self.configDB, self.configDB.CONFIG_DB, '|')
         return True
 
+    def migrate_config_db_flex_counter_delay_status(self):
+        """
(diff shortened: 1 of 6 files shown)


liat-grozovik:
@vadymhlushko-mlnx  pls check failures

vaibhavhd:
I just noticed that 202012 and 202111 are creating same versions w/ different migration logic.

I think the deviation begun starting `version_2_0_0`. This is not expected as now same DB version will have different looking config (or other db) schema.

This almost sounds like an issue that was earlier seen for 202211 branch: https://github.com/sonic-net/sonic-utilities/pull/2470/files

Unfortuantely, like before, we did not lock new version for 202111 branch. This is not an issue just w/ this PR, rather this is a missed step in new branch creation.

I think we need to lock a version everytime a feature branch is created. This needs further discussion and agreement w/ branch owners.

vadymhlushko-mlnx:
> I just noticed that 202012 and 202111 are creating same versions w/ different migration logic.
> 
> I think the deviation begun starting `version_2_0_0`. This is not expected as now same DB version will have different looking config (or other db) schema.
> 
> This almost sounds like an issue that was earlier seen for 202211 branch: https://github.com/sonic-net/sonic-utilities/pull/2470/files
> 
> Unfortuantely, like before, we did not lock new version for 202111 branch. This is not an issue just w/ this PR, rather this is a missed step in new branch creation.
> 
> I think we need to lock a version everytime a feature branch is created. This needs further discussion and agreement w/ branch owners.

Could you please tell me with whom I could talk about this versioning issue?

PR #2884, opened 2023-06-21
Improve `route_check` to better handle duplicate updates

lolyu:
#### What I did
Reason for this PR is that there could be multiple route set updates from ASIC_DB when the mux toggles to standby. This PR aims to improve `route_check` to handle this situation.

Signed-off-by: Longxiang Lyu <lolv@microsoft.com>

#### How I did it
Few updates:
1. Replace `diff_sorted_lists` with `diff_lists`, which is better in handling duplicate entries and executed faster.
2. Remove redundant sort.
3. Add UT to cover duplicate updates.

#### How to verify it
UT and verify on dualtor environment

### DIFF
--- a/scripts/route_check.py
+++ b/scripts/route_check.py
@@ -192,38 +192,16 @@ def cmps(s1, s2):
     return 1
 
 
-def diff_sorted_lists(t1, t2):
+def diff_lists(t1, t2):
     """
-    helper to compare two sorted lists.
+    helper to compare two lists.
     :param t1: list 1
     :param t2: list 2
     :return (<t1 entries that are not in t2>, <t2 entries that are not in t1>)
     """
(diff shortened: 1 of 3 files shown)


yxieca (approved):

prsunny:
This seems to be a major change to existing script. Need to understand the motivation of this change as we use this script for many different validations? My point is, if its mux related, then handle that specifically instead of a generic change

lolyu:
> This seems to be a major change to existing script. Need to understand the motivation of this change as we use this script for many different validations? My point is, if its mux related, then handle that specifically instead of a generic change

Hi @prsunny, this PR doesn't add the mux validation, it aims to improve this script to better handle the duplicated updates and add a UT to validate the duplicated updates scenario

PR #2887, opened 2023-06-27
[db_migrator] Remove hardcoded config and migrate config from minigraph

vaibhavhd:
#### What I did

Microsoft ADO: 18217044

This PR is follow up to an old PR: https://github.com/sonic-net/sonic-utilities/pull/2515/

This PR addresses two issues:
1. Not migrating `RESTAPI`, `TELEMETRY`, `DEVICE_METADATA` entries if they are not supported in target image's minigraph.py.
2. Not migrating these entries if minigraph.xml file is missing.
3. Not maintaining the config for these tables in two different places. Currently db migrator has its own constants, and minigraph.py maintains its own.

#### How I did it
This change removes hardcoding config in migrator code, and migrating the config for `RESTAPI`, `TELEMETRY`, `DEVICE_METADATA` from minigraph generator.

#### How to verify it

Tested on a physical device.

This change works w/ another change to fix config-setup, database dependency https://github.com/sonic-net/sonic-buildimage/pull/14933


scripts/db_migrator.py
```diff
@@ -527,6 +535,9 @@ def migrate_vxlan_config(self):
 
     def migrate_restapi(self):
         # RESTAPI - add missing key
+        if not self.minigraph_data or 'RESTAPI' not in self.minigraph_data:
+            return
+        RESTAPI = self.minigraph_data['RESTAPI']
```
qiluo-msft, line 540: >RESTAPI [](http://example.com/codeflow?start=8&length=7)

RESTAPI -> restapi_data

scripts/db_migrator.py
```diff
@@ -51,6 +53,12 @@ def __init__(self, namespace, socket=None):
         self.TABLE_KEY       = 'DATABASE'
         self.TABLE_FIELD     = 'VERSION'
 
+        # load config data from minigraph to get the default/hardcoded values from minigraph.py
+        # this is to avoid duplicating the hardcoded these values in db_migrator
+        self.minigraph_data = None
+        if os.path.isfile(MINIGRAPH_FILE):
```
qiluo-msft, line 59: >isfile [](http://example.com/codeflow?start=19&length=6)

Could you add a testcase for missing MINIGRAPH_FILE? It would be even better if you can add a testcase for L2 switch mode

vaibhavhd: ack

scripts/db_migrator.py
```diff
@@ -51,6 +53,12 @@ def __init__(self, namespace, socket=None):
...
+            self.minigraph_data = parse_xml(MINIGRAPH_FILE)
```
yxieca, line 60: Do you need to protect against parse_xml throwing error?

vaibhavhd: Sure. Added now.

scripts/db_migrator.py
vaibhavhd: Added a new testcase for missing minigraph. We can discuss what other fields are expected/unexpected in L2 switch's case and add a test for that in future PR. What do you suggest?

scripts/db_migrator.py
vaibhavhd: Updated now

yxieca (dismissed):

tests/db_migrator_test.py
```diff
@@ -509,6 +509,25 @@ def test_warm_upgrade_to_2_0_2(self):
...
+        dbconnector.dedicated_dbs['CONFIG_DB'] = os.path.join(mock_db_path, 'config_db', 'cross_branch_upgrade_to_version_2_0_2_input')
+        import db_migrator
+        dbmgtr = db_migrator.DBMigrator(None)
+        # set minigraph_data to None to mimic the missing minigraph.xml scenario
+        dbmgtr.minigraph_data = None
+        dbmgtr.migrate()
+        dbconnector.dedicated_dbs['CONFIG_DB'] = os.path.join(mock_db_path, 'config_db', 'cross_branch_upgrade_without_mg_2_0_2_expected.json')
+        expected_db = Db()
+
+        expected_db
```
qiluo-msft, line 522: >expected_db [](http://example.com/codeflow?start=8&length=11)

remove this line?

scripts/db_migrator.py
qiluo-msft: Add a test for L2 switch in future is okay.

vaibhavhd: As per offline discussion: from unit test level the test for missing minigraph and absence of default entries is sufficient. L2 switch config and functionality needs to be tested separately as part of sonic-mgmt test. Raised a test gap for this: https://github.com/sonic-net/sonic-mgmt/issues/8777

tests/db_migrator_test.py
vaibhavhd: Removed here, and also in the old testcase.

qiluo-msft (approved):

qiluo-msft:
This commit could not be cleanly cherry-pick to 202012. Please raise another PR on that branch.

StormLiangMS:
caused UT failure in internal branch sync, @vaibhavhd 

{
  'values_changed': {
    "root['certs']['ca_crt']": {
      'new_value': '/etc/sonic/credentials/restapica.crt',
      'old_value': '/et...name'
    ]": {'new_value': 'client.restapi.sonic', 'old_value': 'client.restapi.sonic.gbl,vnetds.prod.int.azure-int.net'}}}

liuh-80:
Cherry-pick to 202012 PR: https://github.com/sonic-net/sonic-utilities/pull/2924

PR #2890, opened 2023-06-30
[db_migrator] Set docker_routing_config_mode to the value obtained from minigraph parser

vaibhavhd:
#### What I did

MSFT ADO: 24419953

This is to fix a bug where warm upgrade from old image (eg. 20181130) to new image (eg. 202012) does not update  `docker_routing_config_mode` to the new config expected the target OS.

For eg., in 201811 `DEVICE_METADATA.localhost.docker_routing_config_mode` is set to `unified`.
After upgrade to 202012 the value is not changed. However, the expectation in newer images is that the value is `separated`.

The move from `unified` to `separated` was done as part of an old change: https://github.com/sonic-net/sonic-buildimage/pull/2961

However, migration logic was not updated since then. Because of this miss, cross-branch warm-upgrade from 201811 to 2020212 to 202305 to latest will always keep the setting as `unified`.

#### How I did it

Added a common migration logic: update `docker_routing_config_mode` to the value from minigraph parser.

#### How to verify it

Added a new unit test. Updated old unit tests.


yxieca (approved):

qiluo-msft:
@vaibhavhd This commit could not be cleanly cherry-picked to 202012. Please submit another PR.

PR #2894, opened 2023-07-06
[202205][db_migrator] Set correct CURRENT_VERSION, extend UT

vadymhlushko-mlnx:
#### What I did
Set the correct value for the `db_migrator` class variable `CURRENT_VERSION`, because if the new DB version was introduced the `CURRENT_VERSION` variable should have the newest version value.

#### How I did it
Edit the `db_migrator.py`.

#### How to verify it
Extend existing UT.

### DIFF
--- a/tests/db_migrator_test.py
+++ b/tests/db_migrator_test.py
@@ -619,3 +619,4 @@ class TestFastUpgrade_to_3_0_7(object):
         dbmgtr.migrate()
         expected_db = self.mock_dedicated_config_db(db_after_migrate)
         assert not self.check_config_db(dbmgtr.configDB, expected_db.cfgdb)
+        assert dbmgtr.CURRENT_VERSION == expected_db.cfgdb.get_entry('VERSIONS', 'DATABASE')['VERSION']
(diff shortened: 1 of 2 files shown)


tests/db_migrator_test.py
```diff
@@ -619,3 +619,4 @@ def test_fast_reboot_upgrade_to_3_0_7(self):
         dbmgtr.migrate()
         expected_db = self.mock_dedicated_config_db(db_after_migrate)
         assert not self.check_config_db(dbmgtr.configDB, expected_db.cfgdb)
+        assert dbmgtr.CURRENT_VERSION == expected_db.cfgdb.get_entry('VERSIONS', 'DATABASE')['VERSION']
```
vaibhavhd, line 622: I think we need a dedicated test just for checking this assertion.

My goal is to use UT to always catch the issue that this PR is addressing -- missing update to CURRENT_VERSION

vaibhavhd (approved):

vaibhavhd:
I don't have permissions to merge to this branch. @yxieca can you help, please?

vaibhavhd:
MSFT ADO: 24471596

PR #2895, opened 2023-07-06
[db_migrator] Set correct CURRENT_VERSION, extend UT

vadymhlushko-mlnx:
#### What I did
Set the correct value for the `db_migrator` class variable `CURRENT_VERSION`, because if the new DB version was introduced the `CURRENT_VERSION` variable should have the newest version value.

#### How I did it
Edit the `db_migrator.py`.

#### How to verify it
Extend existing UT.

### DIFF
--- a/tests/db_migrator_test.py
+++ b/tests/db_migrator_test.py
@@ -675,3 +675,4 @@ class TestFastUpgrade_to_4_0_3(object):
         dbmgtr.migrate()
         expected_db = self.mock_dedicated_config_db(db_after_migrate)
         assert not self.check_config_db(dbmgtr.configDB, expected_db.cfgdb)
+        assert dbmgtr.CURRENT_VERSION == expected_db.cfgdb.get_entry('VERSIONS', 'DATABASE')['VERSION']
--- a/scripts/db_migrator.py
+++ b/scripts/db_migrator.py
@@ -47,7 +47,7 @@ class DBMigrator():
                      none-zero values.
               build: sequentially increase within a minor version domain.
         """
-        self.CURRENT_VERSION = 'version_4_0_2'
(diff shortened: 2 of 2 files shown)


tests/db_migrator_test.py
```diff
@@ -674,4 +674,5 @@ def test_fast_reboot_upgrade_to_4_0_3(self):
         dbmgtr = db_migrator.DBMigrator(None)
         dbmgtr.migrate()
         expected_db = self.mock_dedicated_config_db(db_after_migrate)
-        assert not self.check_config_db(dbmgtr.configDB, expected_db.cfgdb)
\ No newline at end of file
+        assert not self.check_config_db(dbmgtr.configDB, expected_db.cfgdb)
+        assert dbmgtr.CURRENT_VERSION == expected_db.cfgdb.get_entry('VERSIONS', 'DATABASE')['VERSION']
```
vaibhavhd, line 678: I think we need a dedicated test just for checking this assertion.

My goal is to use UT to always catch the issue that this PR is addressing -- missing update to CURRENT_VERSION

vadymhlushko-mlnx: I don't see how to achieve that, because not every UT mocks the config_db with a `VERSIONS` key

vaibhavhd: I see. May be we should always verify this assertion in the new UTs that we add. Specifically, in the changes where new version is introduced.

We can have that conversation later, let's take this bug fix for now.

vaibhavhd (approved):

vaibhavhd:
MSFT ADO : 24471596

bingwang-ms:
@StormLiangMS Please help cherry-pick this PR to `202305`. Thanks!

PR #2897, opened 2023-07-07
[Techsupport] Update the message seen during the lock acquisition failure

vivekrnv:
#### What I did

When a second techsupport instance starts while one is running, the message thrown before exiting is not very user friendly. Thus updating the message for it to make more sense.


dgsudharsan (approved):

liat-grozovik:
@vivekrnv  please update the PR description and the requested branches

qiluo-msft (approved):

StormLiangMS:
not bug fix, but no harm to have for cherry-pick.

PR #2899, opened 2023-07-11
[dualtor][route_check] filter out `soc_ipv6` 

zjswhhh:
#### What I did
Skip `soc_ipv6` when comparing appl_db and asic_db route entries on `active-active` dualtor. Because it's creating false alarms when tunnel route is added for soc ips. The route is meant to be programmed in asic only. 

sign-off: Jing Zhang zhangjing@microsoft.com 
#### How to verify it
Run the updated script on active-active dualtor devices.

### DIFF
--- a/scripts/route_check.py
+++ b/scripts/route_check.py
@@ -585,6 +585,9 @@ def get_soc_ips(config_db):
         if mux_entry.get("cable_type", "") == "active-active" and "soc_ipv4" in mux_entry:
             soc_ips.append(mux_entry["soc_ipv4"])
 
+            if "soc_ipv6" in mux_entry and mux_entry["soc_ipv6"]:
+                soc_ips.append(mux_entry["soc_ipv6"])
+
     return soc_ips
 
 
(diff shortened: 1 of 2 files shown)


prsunny (dismissed):

prsunny:
Please add a test for the same

scripts/route_check.py
```diff
@@ -585,6 +585,9 @@ def get_soc_ips(config_db):
         if mux_entry.get("cable_type", "") == "active-active" and "soc_ipv4" in mux_entry:
             soc_ips.append(mux_entry["soc_ipv4"])
 
+            if "soc_ipv6" in mux_entry and mux_entry["soc_ipv6"]:
```
lolyu, line 588: Placing the check here will make `soc_ipv6` dependent on `soc_ipv4` presence, better to check `soc_ipv4` and `soc_ipv6` separately.

zjswhhh: > Placing the check here will make `soc_ipv6` dependent on `soc_ipv4` presence, better to check `soc_ipv4` and `soc_ipv6` separately.

Addressed.

lolyu (approved):

PR #2918, opened 2023-07-19
Refine [override config table] for corner cases

wenyiz2021:
`sudo config override-config-table` can be used in 2 flows:
1. from ndm process
2. used by user from device to quickly change config partially/fully

in the 2nd case, user can input wrong format config, e.g. not accommodate for ndm standard. we should abort CLI in this case.
another example is user tries to override with empty config, in this case nothing will be overridden.

fixes: https://github.com/sonic-net/sonic-utilities/issues/2916
#### What I did
handle the corner cases in CLI

#### How to verify it

1. empty golden config:
```
admin@str3-8800-lc2-1:/etc/sonic$ sudo config override-config-table golden_config_db.json
Removing configDB overriden table first ...
Overriding input config to configDB ...
Overriding completed. No service is restarted.
Removing configDB overriden table first ...
Overriding input config to configDB ...
Overriding completed. No service is restarted.
Removing configDB overriden table first ...
Overriding input config to configDB ...
Overriding completed. No service is restarted.
```
command will pass, nothing is overwritten from output of `sonic-cfggen --print-data -d`

2. config input is wrong format
```
{
    "NEW_FEATURE_TABLE": {
        "entry": {
            "field": "value",
            "state": "disabled"
        }
    }
}
```

```
admin@str3-8800-lc2-1:/etc/sonic$ sudo config override-config-table golden_config_db.json
Wrong config format! 'localhost' not found in host config! cannot override.. abort
```
command will abort


tests/config_override_test.py
```diff
@@ -381,6 +381,27 @@ def read_json_file_side_effect(filename):
...
+
+        with mock.patch('config.main.read_json_file',
+                        mock.MagicMock(side_effect=read_json_file_side_effect)):
+            runner = CliRunner()
+            result = runner.invoke(config.config.commands["override-config-table"],
+                                   ['golden_config_db.json'], obj=db)
+            # make sure program aborted with return code 1
+            assert result.exit_code == 1
+
+        for ns, config_db in cfgdb_clients.items():
```
wen587, line 400: suggest change to
assert "'localhost' not found in host config" in result.output

It exit when the localhost is not found.

wenyiz2021: I have the assertion on line 404

wen587 (approved):

StormLiangMS:
@wenyiz2021 do we have ADO for this? Looks like an enhancement other than bug fix?

wenyiz2021:
> @wenyiz2021 do we have ADO for this? Looks like an enhancement other than bug fix?

hi @StormLiangMS, without this it was failing https://github.com/sonic-net/sonic-mgmt/blob/ffac1463a768c3dee0f2894c6c67834f2387d5df/tests/override_config_table/test_override_config_table.py#L201 so it should be bug fix. 

but I also made test_override_config_table.py to skip for now, we need NDM to push new config changes for multi-asic.

gechiang:
@StormLiangMS , @yxieca , This is a bug fix that needs to be backported to 202205, 202211 and 202305.
We are seeing this issue being exposed recently under 202205.  Appreciate if you can cherry-pick this PR to the above requested branches.
Thanks!

PR #2925, opened 2023-07-24
Input check for timeout in generate_dump

ycoheNvidia:
#### What I did
Added input check for argument "-t" in generate_dump script
#### How I did it
Made sure only integer values can be received for this argument
#### How to verify it
Call generate_dump -t with non-integer values


ycoheNvidia:
@qiluo-msft can we get a reviewer for this fix please?

qiluo-msft (approved):

PR #2929, opened 2023-07-27
[build] Fix dependency issue between responses and urllib3 package. (#2928)

liushilongbuaa:
#### What I did
Cherry pick PR:
* [build] Fix dependency issue between responses and urllib3 package.


StormLiangMS (approved):
LGTM

PR #2934, opened 2023-07-31
revert unit test tests/test_clear_tag

maipbui:
#### What I did
Revert test case due to frequent failure in azp introduced in PR https://github.com/sonic-net/sonic-utilities/pull/2849


qiluo-msft (approved):

maipbui:
Will investigate the failure, re-write/ improve the unit test and add unit test back in the future.

PR #2936, opened 2023-08-02
[GCU] Add PORT table StateDB Validator

isabelmsft:
#### What I did
Add PORT table StateDB Validator
#### How I did it
Add new validating function for PORT table field and fec fields
#### How to verify it
Unit tests, updated sonic-mgmt tests: https://github.com/sonic-net/sonic-mgmt/pull/9403/files


generic_config_updater/field_operation_validators.py
```diff
@@ -126,3 +126,71 @@ def _get_fields_in_patch():
                 return False
 
     return True
+
+
+def read_statedb_entry(table, field):
+    state_db = SonicV2Connector(host="127.0.0.1")
```
qiluo-msft, line 132: >SonicV2Connector [](http://example.com/codeflow?start=15&length=16)

If you use Table class, you do not need to concat table name and key.

generic_config_updater/field_operation_validators.py
```diff
@@ -126,3 +126,71 @@ def _get_fields_in_patch():
...
+    state_db.connect(state_db.STATE_DB)
+    return state_db.get(state_db.STATE_DB, table, field)
+
+
+def port_config_update_validator(patch_element):
+
+    def _validate_key(key, port, value):
+        if key == "fec":
```
qiluo-msft, line 140: >key [](http://example.com/codeflow?start=11&length=3)

key -> field

generic_config_updater/field_operation_validators.py
```diff
@@ -126,3 +126,71 @@ def _get_fields_in_patch():
...
+            supported_fecs_str = read_statedb_entry('{}|{}'.format("PORT_TABLE", port), "supported_fecs")
+            if supported_fecs_str:
+                if supported_fecs_str != 'N/A':
+                    supported_fecs_list = [element.strip() for element in supported_fecs_str.split(',')]
+                else:
+                    supported_fecs_list = []
+            else:
+                supported_fecs_list = [ 'rs', 'fc', 'none']
```
qiluo-msft, line 148: >supported_fecs_list [](http://example.com/codeflow?start=16&length=19)

Duplicate list of constant. To prevent future mistake when we add one more element, let's define it in a centralized place and reuse.

generic_config_updater/field_operation_validators.py
```diff
@@ -126,3 +126,71 @@ def _get_fields_in_patch():
...
+            if value.strip() not in supported_fecs_list:
+                return False
+            return True
+        if key == "speed":
+            supported_speeds_str = read_statedb_entry('{}|{}'.format("PORT_TABLE", port), "supported_speeds") or ''
+            supported_speeds = [int(s) for s in supported_speeds_str.split(',') if s]
+            if supported_speeds and int(value) not in supported_speeds:
```
qiluo-msft, line 155: >int [](http://example.com/codeflow?start=36&length=3)

may throw during casting

generic_config_updater/field_operation_validators.py
```diff
@@ -126,3 +126,71 @@ def _get_fields_in_patch():
...
+                return False
+            return True
```
qiluo-msft, line 157: >return [](http://example.com/codeflow?start=12&length=6)

Could you add a final `return false` at the end of function? The default behavior is `return None`, which is working but not reader friendly.

generic_config_updater/field_operation_validators.py
```diff
@@ -126,3 +126,71 @@ def _get_fields_in_patch():
...
+        return None
+    
+    if patch_element["op"] == "remove":
+        return True
+    
+    # for PORT speed and fec configs, need to ensure value is allowed based on StateDB
+    patch_element_str = json.dumps(patch_element)
+    path = patch_element["path"]
+    value = patch_element.get("value")
+    keys = ['fec', 'speed']
```
qiluo-msft, line 173: >keys [](http://example.com/codeflow?start=4&length=4)

-> fields

generic_config_updater/gcu_field_operation_validators.conf.json
```diff
@@ -145,6 +145,9 @@
                     }
                 }
             }
+        },
+	"PORT": {
```
qiluo-msft, line 149: >	 [](http://example.com/codeflow?start=0&length=1)

mixing tab and spaces.

tests/generic_config_updater/field_operation_validator_test.py
```diff
@@ -14,6 +14,80 @@
...
+        patch_element = {"path": "/PORT/Ethernet3/fec", "op": "add", "value": "rs"}
+        assert generic_config_updater.field_operation_validators.port_config_update_validator(patch_element) == True
+
+    @patch("generic_config_updater.field_operation_validators.read_statedb_entry", mock.Mock(return_value=""))
+    def test_port_config_update_validator_valid_fec_no_state_db(self):
+        patch_element = {"path": "/PORT/Ethernet3", "op": "add", "value": {"fec": "rs"}}
+        assert generic_config_updater.field_operation_validators.port_config_update_validator(patch_element) == True
+    
+    @patch("generic_config_updater.field_operation_validators.read_statedb_entry", mock.Mock(return_value=""))
+    def test_port_config_update_validator_invalid_fec_no_state_db(self):
```
qiluo-msft, line 87: >test_port_config_update_validator_invalid_fec_no_state_db [](http://example.com/codeflow?start=8&length=57)

Could you add unit test or sonic-mgmt test for config CLI with yang validation?

generic_config_updater/field_operation_validators.py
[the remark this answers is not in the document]
isabelmsft: Changed to use use_socket_path

isabelmsft: I see.. is Table class recommended over SonicV2Connector use_unix_socket_path? In the future, this function may be extended to include tables other than PORT

generic_config_updater/field_operation_validators.py
isabelmsft: Thanks, I defined the default supported_fecs_list as a constant

generic_config_updater/field_operation_validators.py
isabelmsft: yes, I changed to `return False`

tests/generic_config_updater/field_operation_validator_test.py
isabelmsft: Yes, I can add sonic-mgmt test to test config CLI with yang validation. 

Test for success scenario: 
1. Enable YANG ConfigDB validation feature (via CLI)
2. Use config CLI to configure speed/fec to valid value based on YANG validation and StateDB
3. Ensure that modification goes through 

Test for failure scenario: 
1. Enable YANG ConfigDB validation feature (via CLI)
2. Use config CLI to configure speed/fec to invalid value based on StateDB (but valid value based on YANG)
3. Ensure that output throws error "Modification of PORT table is illegal- validating function generic_config_updater.field_operation_validators.port_config_update_validator returned False"

generic_config_updater/field_operation_validators.py
isabelmsft: Changed to field

generic_config_updater/field_operation_validators.py
isabelmsft: Thank you, I added exception handling

tests/generic_config_updater/field_operation_validator_test.py
```diff
@@ -14,6 +14,85 @@
...
 
+    @patch("generic_config_updater.field_operation_validators.read_statedb_entry", mock.Mock(return_value=""))
+    def test_port_config_update_validator_valid_speed_no_state_db(self):
+        patch_element = {"path": "/PORT/Ethernet3", "op": "add", "value": {"speed": "234"}}
+        assert generic_config_updater.field_operation_validators.port_config_update_validator(patch_element) == True
+    
+    @patch("generic_config_updater.field_operation_validators.read_statedb_entry", mock.Mock(return_value=""))
+    def test_port_config_update_validator_invalid_speed_no_state_db(self):
+        patch_element = {"path": "/PORT/Ethernet3", "op": "add", "value": {"speed": "xyz"}}
+        assert generic_config_updater.field_operation_validators.port_config_update_validator(patch_element) == False
```
wen587, line 25: Looks like it hit true in unit test.

isabelmsft: Yes thanks, I changed this unit test to return existing speeds to validate against in StateDB. So now, False is returned because in this case, we attempt to typecast the value to an int. 

Previously, the unit test did not return any values in StateDB to validate against, so we never reached the typecast error being thrown. In this case, the invalid speed is caught by YANG validation so it was correct to return True before I changed this unit test.

generic_config_updater/field_operation_validators.py
qiluo-msft: I see the the same use case in scripts/portconfig. Can we have centralized constant, and reuse in this repo?

generic_config_updater/field_operation_validators.py
qiluo-msft: As I said, the benefit is no need to concat table name/separator/key. You can create multiple Table from one DBConnector, and they are not heavy.

qiluo-msft (approved):

isabelmsft:
Tested on 202305 branch, confirmed no cherry-pick conflicts

PR #2948, opened 2023-08-18
[db_migrator.py] Fix issue while upgrading from 202205 to 202211 via fast reboot

Junchao-Mellanox:
#### What I did

Fix issue while upgrading from 202205 to 202211 via fast reboot. This issue is caused by a mismatch version handling for fast reboot in db_migrator. 

In 202205, the db migrator for fast reboot flag handling is 3_0_5:

https://github.com/sonic-net/sonic-utilities/blob/56a1ae24dc8f4e2dd2ad8929506ecdc836d3dd7c/scripts/db_migrator.py#L920

However, in master(202211), the db migrator for fast reboot flag handling is 4_0_0:

https://github.com/sonic-net/sonic-utilities/blob/7435b1ca539382e644f29d424c551f6258fd2920/scripts/db_migrator.py#L993

This mismatch causes an incorrect sequence like this:

```
1. User issue fast-reboot under 202205
2. fast-reboot script set fast reboot flag by command "sonic-db-cli STATE_DB HSET "FAST_RESTART_ENABLE_TABLE|system" "enable" "true" &>/dev/null"
3. system boot to 202211
4. db_migrator found the database version is 3_0_6, it will run 4_0_0, however, it found FAST_REBOOT|system does not exist, and it set  FAST_RESTART_ENABLE_TABLE|system enable to false
5. system incorrectly performs cold reboot
```

#### How I did it

in db migrator if we see there is FAST_RESTART_ENABLE_TABLE already, we should skip fast reboot flag migration

#### How to verify it

unit test
manual test

### DIFF
--- a/scripts/db_migrator.py
+++ b/scripts/db_migrator.py
@@ -658,7 +658,7 @@ class DBMigrator():
         # overwrite the routing-config-mode as per minigraph parser
         # Criteria for update:
         # if config mode is missing in base OS or if base and target modes are not same
-        #  Eg. in 201811 mode is "unified", and in newer branches mode is "separated" 
+        #  Eg. in 201811 mode is "unified", and in newer branches mode is "separated"
         if ('docker_routing_config_mode' not in device_metadata_old and 'docker_routing_config_mode' in device_metadata_new) or \
         (device_metadata_old.get('docker_routing_config_mode') != device_metadata_new.get('docker_routing_config_mode')):
             device_metadata_old['docker_routing_config_mode'] = device_metadata_new.get('docker_routing_config_mode')
@@ -1000,12 +1000,14 @@ class DBMigrator():
         # reading FAST_REBOOT table can't be done with stateDB.get as it uses hget behind the scenes and the table structure is
         # not using hash and won't work.
         # FAST_REBOOT table exists only if fast-reboot was triggered.
-        keys = self.stateDB.keys(self.stateDB.STATE_DB, "FAST_REBOOT|system")
-        if keys:
-            enable_state = 'true'
-        else:
-            enable_state = 'false'
-        self.stateDB.set(self.stateDB.STATE_DB, 'FAST_RESTART_ENABLE_TABLE|system', 'enable', enable_state)
+        keys = self.stateDB.keys(self.stateDB.STATE_DB, "FAST_RESTART_ENABLE_TABLE|system")
+        if not keys:
+            keys = self.stateDB.keys(self.stateDB.STATE_DB, "FAST_REBOOT|system")
+            if keys:
+                enable_state = 'true'
+            else:
+                enable_state = 'false'
+            self.stateDB.set(self.stateDB.STATE_DB, 'FAST_RESTART_ENABLE_TABLE|system', 'enable', enable_state)
         self.set_version('version_4_0_1')
         return 'version_4_0_1'
 
@@ -1024,8 +1026,8 @@ class DBMigrator():
         Version 4_0_2.
         """
         log.log_info('Handling version_4_0_2')
-
-        if self.stateDB.keys(self.stateDB.STATE_DB, "FAST_REBOOT|system"):
+        enable_state = self.stateDB.get(self.stateDB.STATE_DB, 'FAST_RESTART_ENABLE_TABLE|system', 'enable')
+        if self.stateDB.keys(self.stateDB.STATE_DB, "FAST_REBOOT|system") or enable_state == 'true':
             self.migrate_config_db_flex_counter_delay_status()
 
         self.set_version('version_4_0_3')
--- a/scripts/db_migrator.py
+++ b/scripts/db_migrator.py
@@ -1026,8 +1026,7 @@ class DBMigrator():
         Version 4_0_2.
         """
         log.log_info('Handling version_4_0_2')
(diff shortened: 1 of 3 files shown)


stepanblyschak (approved):

scripts/db_migrator.py
```diff
@@ -1024,8 +1026,8 @@ def version_4_0_2(self):
         Version 4_0_2.
         """
         log.log_info('Handling version_4_0_2')
-
-        if self.stateDB.keys(self.stateDB.STATE_DB, "FAST_REBOOT|system"):
+        enable_state = self.stateDB.get(self.stateDB.STATE_DB, 'FAST_RESTART_ENABLE_TABLE|system', 'enable')
+        if self.stateDB.keys(self.stateDB.STATE_DB, "FAST_REBOOT|system") or enable_state == 'true':
```
vaibhavhd, line 1030: Why is this change needed? The check here was added (https://github.com/sonic-net/sonic-utilities/pull/2839) for old systems (like 201911) where new flag was not present. The newer branches (where `FAST_RESTART_ENABLE_TABLE` flag is present use below changes to delay flex counters:
https://github.com/sonic-net/sonic-swss/pull/1877
https://github.com/sonic-net/sonic-utilities/pull/1768

@shlomibitton can you please review if the above analysis is correct?

Junchao-Mellanox: Hi @vaibhavhd , this issue is only reproduced from 202205 to 202211. In 202205, there is already a db migrator function version_3_0_5 which changes old flag to new flag. However, in 202211, the same logic is in version_4_0_0. While upgrading from 202205 to 202211, 202205 sets new flag and we have a dump.rdb; in 202211 db_migrator just sets flag to false.

Junchao-Mellanox: Hi @stepanblyschak and @arfeigin , could you please provide your comment here?

arfeigin: @Junchao-Mellanox correct, good observation.
The suggested change for version_4_0_0 is in place and hence also this change in version_4_0_2 is needed to tie it together so it will all work as expected.

arfeigin (approved):
LGTM, good catch and solution.

scripts/db_migrator.py
arfeigin: @vaibhavhd As wrote since in version_4_0_0 in the db_migrator according to this PR we are adding new state-db entry for fast-reboot instead of the old version it will be more coherent to check for the newer version when migrating to even newer version (4_0_2) and ignore the former state-db entry.

liat-grozovik:
@vaibhavhd can you please review feedback and if possible aprrove?

scripts/db_migrator.py
vaibhavhd: I still don't understand why these two lines need to be changed. I understand the change in 4_0_0, but w/o the change in 4_0_2, what is the issue?

I am not saying that this is doing any damage, just that I don't understand why we are fixing this if this is not needed.

Junchao-Mellanox: Hi @vaibhavhd, you are right, this change is not necessary. I have updated code, could you please review again?

scripts/db_migrator.py
```diff
@@ -1000,12 +1000,14 @@ def version_4_0_0(self):
...
         # FAST_REBOOT table exists only if fast-reboot was triggered.
-        keys = self.stateDB.keys(self.stateDB.STATE_DB, "FAST_REBOOT|system")
-        if keys:
-            enable_state = 'true'
-        else:
-            enable_state = 'false'
-        self.stateDB.set(self.stateDB.STATE_DB, 'FAST_RESTART_ENABLE_TABLE|system', 'enable', enable_state)
+        keys = self.stateDB.keys(self.stateDB.STATE_DB, "FAST_RESTART_ENABLE_TABLE|system")
+        if not keys:
+            keys = self.stateDB.keys(self.stateDB.STATE_DB, "FAST_REBOOT|system")
```
vaibhavhd, line 1005: @arfeigin how do we address the conversion case where we want to set the flag, DB will not have it?

I remember we had a discussion about this in the past w/ your fastboot over warmboot infra changes. But, can't fully recollect the conclusion.

arfeigin: I think this is the discussion we had then: https://github.com/sonic-net/sonic-utilities/pull/2621#discussion_r1130519315.
I think the solution was to have the entry FAST_REBOOT_ENABLE_TABLE|system {enable, <true/false>} according to the state.
In the suggested implementation as I see:

1. If entry exists it will be preserved with either true/false.
2. If entry doesn't exist, if it had old fast-reboot entry set (fast-reboot from older version) it will be added with enable_state=true otherwise it will be added with enable_state=false.
So as I see in any case the entry will be present in state-db, either with previous values or it will be added.
In both cases it will hold accurate data as I understand.

Junchao-Mellanox: Hi @vaibhavhd , could you please check @arfeigin 's reply?

vaibhavhd (approved):

vaibhavhd (approved):

StormLiangMS:
ADO: 25048636

Junchao-Mellanox:
PR for 202211: https://github.com/sonic-net/sonic-utilities/pull/2962

PR #2952, opened 2023-08-23
Expanding the manifest capabilities 

azmy98:
Expand docker manifest capabilities

#### What I did
Expanding the manifest.json file to include new network capabilities including the
ports forwarding and docker network type configuration: host, bridge, etc. 

#### How I did it
updating the manifest py file as well as the service_creatore py file.

#### How to verify it
In order to verify this feature, we need to have the complementary part in the sonic-buildimage repo
and then need to install a docker image file with manifest that includes ports in it and network type as well using sonic-package-manager and validate that they are as expected using docker commands such as: 'docker ps' and 'docker inspect' commands


sonic_package_manager/manifest.py
```diff
@@ -213,6 +213,8 @@ def unmarshal(self, value):
         ManifestRoot('container', [
             ManifestField('privileged', DefaultMarshaller(bool), False),
             ManifestArray('volumes', DefaultMarshaller(str)),
+            ManifestField('network', DefaultMarshaller(str), ''),
+            ManifestArray('ports', DefaultMarshaller(str)),
```
stepanblyschak, line 217: Could you please add new fields (with usage examples preferably) to https://github.com/sonic-net/SONiC/blob/master/doc/sonic-application-extension/sonic-application-extension-guide.md?

sonic_package_manager/manifest.py
```diff
@@ -213,6 +213,8 @@ def unmarshal(self, value):
         ManifestRoot('container', [
             ManifestField('privileged', DefaultMarshaller(bool), False),
             ManifestArray('volumes', DefaultMarshaller(str)),
+            ManifestField('network', DefaultMarshaller(str), ''),
```
stepanblyschak, line 216: I assume the command ```sonic-package-manager show package manifest``` will display empty string ```""``` as a value of ```network``` field. Do you want the default value to be ```'host'``` rather then empty string?

stepanblyschak (changes requested):

sonic_package_manager/service_creator/creator.py
```diff
@@ -231,6 +232,18 @@ def generate_container_mgmt(self, package: Package):
 
         run_opt.append('-t')
 
+        if not is_asic_service:
+            if container_spec['network']:
+                docker_network_type = container_spec['network']
+                run_opt.append(f'--net={docker_network_type}')
+
+        if is_asic_service:
```
stepanblyschak, line 240: This is an else statement for the previous if

sonic_package_manager/service_creator/creator.py
```diff
@@ -231,6 +232,18 @@ def generate_container_mgmt(self, package: Package):
...
+            if container_spec['network']:
```
stepanblyschak, line 241: Isn't the 'network' set to 'host' by default? Think this will always fail if asic-service is True

stepanblyschak (approved):

lguohan:
this is adding new manifest for 3rd container management, can you update the doc? and provide link to the doc update pr?

azmy98:
@lguohan, I will add a documentation once the another PR is also approved: https://github.com/sonic-net/sonic-buildimage/pull/16248

azmy98:
I added the documentation PR: https://github.com/sonic-net/SONiC/pull/1601

sonic_package_manager/service_creator/creator.py
```diff
@@ -231,6 +232,17 @@ def generate_container_mgmt(self, package: Package):
 
         run_opt.append('-t')
 
+        if not is_asic_service:
+            if container_spec['network']:
+                docker_network_type = container_spec['network']
+                run_opt.append(f'--net={docker_network_type}')
+        else:
+            if container_spec['network']:
+                raise ServiceCreatorError(f"Invalid Configuration, asic-service must not contain network type")
```
qiluo-msft, line 241: >not contain network type [](http://example.com/codeflow?start=85&length=24)

empty value means default value. It is not proper to enforce empty value only.

azmy98: But ASIC services should not have network specified. This comment was given by the NVIDIA SONiC team. I am missing something?

azmy98: what do you suggest? to enforce the `none` as the docker_network_type like this
`run_opt.append(f'--net={docker_network_type}')`

qiluo-msft: Sorry, let me clarify:
1. container_spec['network'] should have default value `bridge`. ie. if it is None, treat it as `bridge`.
2. I do not propose to append run_opt, always respect user config
3. if is_asic_service and container_spec['network'] != none, you can raise an exception.

azmy98: I see, but this is what I do in this if and else that i added. 
for point 3, if the container_spec['network'] is None I raise exception, am I missing something?

azmy98: Are you suggesting to do explicitly `container_spec['network'] != 'none'`?

qiluo-msft: yes

qiluo-msft:
Please resolve conflict

PR #2953, opened 2023-08-24
[sfputil] Retrieve error-status for transceiver using sfp API

mihirpat1:
#### What I did
The mismatch in 0 based or 1 based port numbering between sfp.index and physical_port number is causing the below issues while executing the "sfputil show error-status -hw" command

1.	“sfputil show error-status -hw” command doesn’t show error-status for Ethernet0 or the last port for some platforms and the remaining ports show error-status of the next or previous port
2.	“sfputil show error-status -hw -p <port>” doesn’t display error-status
3. platform_tests/sfp/test_sfputil.py::test_check_sfputil_error_status sonic-mgmt testcase fails for some platforms

Issue details with an example of sfp.index being 1 based while physical_port being 0 based:
The sfp.index https://github.com/sonic-net/sonic-utilities/blob/7435b1ca539382e644f29d424c551f6258fd2920/sfputil/main.py#L904 is 1 based whereas the physical port numbering is 0 based (retried from port_config.ini file).
Hence, for Ethernet0 (sfp.index = 1), there is no key-value defined in output_dict (dictionary of <sfp.index>:<error_status>) for physical_port = 0.
https://github.com/sonic-net/sonic-utilities/blob/7435b1ca539382e644f29d424c551f6258fd2920/sfputil/main.py#L940

#### How I did it
I am now directly calling get_error_description using sfp object to retrieve the error description. This removes the dependency with sfp.index to retrieve error description for a port.

#### How to verify it
Ran platform_tests/sfp/test_sfputil.py::test_check_sfputil_error_status sonic-mgmt testcase and ensured it passes successfully.

No change in CLI o/p. Also, platforms which didn't work with this CLI earlier, will now show a valid CLI output.

### DIFF
--- a/sfputil/main.py
+++ b/sfputil/main.py
@@ -889,15 +889,15 @@ def fetch_error_status_from_platform_api(port):
     for logical_port_name in logical_port_list:
         physical_port = logical_port_to_physical_port_index(logical_port_name)
 
-        try:
-            error_description = platform_chassis.get_sfp(physical_port).get_error_description()
-            if is_port_type_rj45(logical_port_name):
-                output.append([logical_port_name, "N/A"])
-            else:
+        if is_port_type_rj45(logical_port_name):
+            output.append([logical_port_name, "N/A"])
+        else:
+            try:
+                error_description = platform_chassis.get_sfp(physical_port).get_error_description()
                 output.append([logical_port_name, error_description])
-        except NotImplementedError:
-            click.echo("get_error_description NOT implemented for port {}".format(logical_port_name))
-            sys.exit(ERROR_NOT_IMPLEMENTED)
+            except NotImplementedError:
+                click.echo("get_error_description NOT implemented for port {}".format(logical_port_name))
+                sys.exit(ERROR_NOT_IMPLEMENTED)
 
     return output
 
--- a/sfputil/main.py
+++ b/sfputil/main.py
@@ -882,62 +882,22 @@ def fetch_error_status_from_platform_api(port):
     """
     if port is None:
         logical_port_list = natsort.natsorted(platform_sfputil.logical)
-        # Create a list containing the logical port names of all ports we're interested in
-        generate_sfp_list_code = \
-            "sfp_list = chassis.get_all_sfps()\n"
     else:
-        physical_port_list = logical_port_name_to_physical_port_list(port)
         logical_port_list = [port]
-        # Create a list containing the logical port names of all ports we're interested in
-        generate_sfp_list_code = \
-            "sfp_list = [chassis.get_sfp(x) for x in {}]\n".format(physical_port_list)
-
-    # Code to initialize chassis object
-    init_chassis_code = \
-        "import sonic_platform.platform\n" \
-        "platform = sonic_platform.platform.Platform()\n" \
-        "chassis = platform.get_chassis()\n"
-
-    # Code to fetch the error status
-    get_error_status_code = \
-        "try:\n"\
-        "    errors=['{}:{}'.format(sfp.index, sfp.get_error_description()) for sfp in sfp_list]\n" \
-        "except NotImplementedError as e:\n"\
-        "    errors=['{}:{}'.format(sfp.index, 'OK (Not implemented)') for sfp in sfp_list]\n" \
-        "print(errors)\n"
-
-    get_error_status_command = ["docker", "exec", "pmon", "python3", "-c", "{}{}{}".format(
-        init_chassis_code, generate_sfp_list_code, get_error_status_code)]
-    # Fetch error status from pmon docker
-    try:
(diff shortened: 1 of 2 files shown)


prgeor (dismissed):

sfputil/main.py
```diff
@@ -882,62 +882,22 @@ def fetch_error_status_from_platform_api(port):
     """
     if port is None:
         logical_port_list = natsort.natsorted(platform_sfputil.logical)
-        # Create a list containing the logical port names of all ports we're interested in
```
stephenxs, line 885: SFP platform API can be invoked form the pmon docker only but it is invoked from master, which causes the exception you mentioned below.
Regarding the starting index, I think spf.index should start from 1 which is assumed also in xcvrd. if it doesn’t mismatch with physical index on some platforms, I think it should be fixed in vendor’s platform API.

stephenxs (changes requested):
As commented

sfputil/main.py
aravindmani-1: I don't think we need to use the SFP index as it is internal to vendor platforms. SFP API can be invoked from the host too like it was done for presence and eeprom API in the same file. All the API's implemented in sfputil/main.py are using host files only.

stephenxs: ok. you are probably correct according to @Junchao-Mellanox 's explaination.

aravindmani-1 (approved):

aravindmani-1 (approved):

sfputil/main.py
Junchao-Mellanox: If this PR is going to be cherry-picked to 202211/202205, then, this will be an issue.

aravindmani-1: @Junchao-Mellanox I just verified the fix in 202205 branch in Dell platforms and it worked fine. Could you please elaborate more on the issue that you expect?.

prgeor:
@Junchao-Mellanox @keboliu @stephenxs please approve if no concern

sfputil/main.py
Junchao-Mellanox: In 202205/202211, SFP platform API can be invoked from the pmon container only. So, it will be an issue.

aravindmani-1: If SFP platform API can be invoked from PMON container only, then existing CLI's like presence, eeprom would have failed. 
From the earlier implementation, error description uses PMON container for invoking platform API which deviates from other CLI's implemenation.
From this PR, error description API uses the same method as other CLI's like presence, EEPROM where it executes from the host. 

https://github.com/sonic-net/sonic-utilities/blob/d69432d1a999a1ebc702517aaf42d83765ea66c9/sfputil/main.py#L518

SFP platform API can be invoked from host too in 202205/202211. 
I confirmed the same in Dell platforms.
        root@sonic:~# show ver
        
        SONiC Software Version: SONiC.202205.345706-611449dc8
        SONiC OS Version: 11
        Distribution: Debian 11.7
        Kernel: 5.10.0-18-2-amd64
        Build commit: 611449dc8
        Build date: Thu Aug 24 13:38:00 UTC 2023
        Built by: AzDevOps@vmss-soni001UN2
        
        Platform: x86_64-dell_s6100_c2538-r0
        HwSKU: Force10-S6100
        ASIC: broadcom
        ASIC Count: 1
        ASIC Count: 1
        Serial Number: CFXYZP2
        Model Number: 0F6N2R
        Hardware Revision: A03
        Uptime: 03:07:14 up 2 days, 20:54,  1 user,  load average: 1.25, 1.62, 1.58
        Date: Mon 28 Aug 2023 03:07:14
        root@sonic:~# python3
        Python 3.9.2 (default, Feb 28 2021, 17:03:44)
        [GCC 10.2.1 20210110] on linux
        >>>
        >>> from sonic_platform.chassis import Chassis
        >>> c=Chassis()
        >>> c.get_sfp(32).get_presence()
        True
        >>> c.get_sfp(32).get_error_description()
        'OK'
        >>>

Junchao-Mellanox: This is a vendor specific behavior. Not all SFP API, but some of them can only be triggered in PMON container. That's why we have the logic to run "docker exec" there. It is for compatible with all vendors.

prgeor (approved):

mihirpat1:
@StormLiangMS - Can you please help with cherry-pick for this to 202305?
ADO - 24867010

ZhaohuiS:
@prgeor can I ask why you removed Request for 202205 branch? Nightly test is based on 202205 image.

prgeor:
> @prgeor can I ask why you removed Request for 202205 branch? Nightly test is based on 202205 image.

@ZhaohuiS this will break 202205. This issue is fixed differently in 202205

PR #2955, opened 2023-08-29
[GCU] Add BGPSentinel to CreateOnlyFilter

guangyao6:
#### What I did
BGPSentinel is the same as BGP_PEER_RANGE (create only). In order to support GCU for BGPSentinel, add BGPSentinel to GCU create only filter.
#### How I did it
Add BGPSentinel to CreateOnlyFilter and add related testcases.
#### How to verify it
Unit Test. And corresponding sonic-mgmt case: https://github.com/sonic-net/sonic-mgmt/pull/9735


wen587 (approved):
lgtm

PR #2959, opened 2023-08-29
Remove the CONFIG_DB_INIT flag dependency on db_migrator

vivekrnv:
#### What I did

db_migrator should not depend on CONFIG_DB_INITIALIZED flag. get_hwsku api call busy waits on that flag

#### How I did it

Replace get_hwsku call with get_localhost_info call which takes in a custom config_db object

### DIFF
--- a/scripts/db_migrator.py
+++ b/scripts/db_migrator.py
@@ -92,7 +92,6 @@ class DBMigrator():
         if not self.asic_type:
             log.log_error("ASIC type information not obtained. DB migration will not be reliable")
 
-        # TODO: Update get_hwsku API to take in a custom configDB object
         self.hwsku = device_info.get_localhost_info('hwsku', self.configDB)
         if not self.hwsku:
             log.log_error("HWSKU information not obtained. DB migration will not be reliable")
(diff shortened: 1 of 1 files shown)


scripts/db_migrator.py
```diff
@@ -91,7 +91,9 @@ def __init__(self, namespace, socket=None):
         self.asic_type = version_info.get('asic_type')
         if not self.asic_type:
             log.log_error("ASIC type information not obtained. DB migration will not be reliable")
-        self.hwsku = device_info.get_hwsku()
+
+        # TODO: Update get_hwsku API to take in a custom configDB object
```
vaibhavhd, line 95: Why is this TODO left? How will custom configDB help unblock the busy wait of get_hwsku?

vivekrnv: When get_hwsku() is used, it calls get_localhost_info and it calls ConfigDBConnector.connect() which busy waits. 

I'll remove the TODO comment

dgsudharsan (approved):

vaibhavhd (approved):

vaibhavhd:
@vivekrnv please update the description with how this change was validated?

vaibhavhd:
MSFT ADO: 25357782

StormLiangMS:
@vivekrnv have you test with 202305?

vivekrnv:
With this PR https://github.com/sonic-net/sonic-buildimage/pull/16116, this change is a good to have but not a must

PR #2966, opened 2023-09-02
[201911] [multi_asic] Script to monitor errors on internal links

tjchadaga:
#### What I did
Added script to monitor SAI_PORT_STAT_IF_IN_ERRORS & SAI_PORT_STAT_IF_OUT_ERRORS on internal (backend) ports of multi-asic device. 

#### How I did it
Script is added to monit to regularly scan and check for errors on internal ports. If error count is above configured threshold, and the internal port-channel has more than min_links active links, then the port is shut down and syslog is generated.

#### How to verify it
UT cases are added to test the error detection.


scripts/internal_links_monitor.py
```diff
@@ -0,0 +1,243 @@
...
+
+import syslog
+
+
+from swsssdk import  SonicV2Connector
+from sonic_py_common import multi_asic
+
+# Global variables
+MIN_SCAN_INTERVAL = 60      # Every minute
+MAX_SCAN_INTERVAL = 3600    # An hour
```
abdosi, line 12: how they are getting used ?

scripts/internal_links_monitor.py
```diff
@@ -0,0 +1,243 @@
...
+        for namespace in self.namespaces:
+            self.appdb[namespace] = self.appl_db(namespace)
+            self.configdb[namespace] = self.config_db(namespace)
+            self.link_monitor[namespace] = InternalLinkErrMontoring(
+                namespace=namespace, port_name_map='COUNTERS_PORT_NAME_MAP',
+                filter_for_backend_ports=self.filter_for_backend_ports)
+        self.error_counter_names = ['SAI_PORT_STAT_IF_IN_ERRORS',
+                                    'SAI_PORT_STAT_IF_OUT_ERRORS']
+        # Using fixed values for 201911
+        self.threshold = 0
```
abdosi, line 110: is comment applicable ? Is threshold configurable ?

scripts/internal_links_monitor.py
```diff
@@ -0,0 +1,243 @@
...
+        return ''
+
+    def get_min_links_for_lag(self, namespace, lag_name):
+        ''' Returns min_links configuration of the given portchannel '''
+        lag_info = self.configdb[namespace].get_all('CONFIG_DB', 'PORTCHANNEL|{}'.format(lag_name))        
+        return int(lag_info['min_links'])        
+
+    def get_active_lag_member_count(self, namespace, lag_name):
+        ''' Returns number of member ports that are operationally up in the given portchannel '''
+        lag_members = self.appdb[namespace].keys('APPL_DB', 'LAG_MEMBER_TABLE:{}:*'.format(lag_name))
```
abdosi, line 170: can we use config db instead of appl db ?

scripts/internal_links_monitor.py
tjchadaga: It is fixed for 201911, but would be read from configDB for master and newer branches

scripts/internal_links_monitor.py
```diff
@@ -0,0 +1,243 @@
...
+        return len(all_error_ports)
+
+def main():
+    if multi_asic.is_multi_asic():
+        link_monitor = PacketChassisInternalLinkMontoring()
+        num_of_err_ports =  link_monitor.monitor()
+        if num_of_err_ports:
+            syslog.syslog(syslog.LOG_CRIT, '{} internal ports have errors above threshold'.format(num_of_err_ports))
+            return -1
+        return 0
```
arlakshm, line 238: `return 0` for single asic as well?

scripts/internal_links_monitor.py
tjchadaga: since we are looking for active (operationally up) lag members, does it matter if we check in appl db?

scripts/internal_links_monitor.py
tjchadaga: not used in 201911, it was added for master PR. I can remove it from here.

scripts/internal_links_monitor.py
```diff
@@ -0,0 +1,243 @@
...
+        ''' Attempt to isolate given list of ports 
+            1. Check if port is already down, then skip mitigation
+            2. Check if number of active links in the portchannel where the port is a member is greater
+                than min_links. 
+                a. If active links is greater, shutdown the port
+                b. If not, return after generating syslog
+        '''
+
+        for port_name in err_port_per_ns:            
+            if self.get_port_status(namespace, port_name) == 'down':
```
arlakshm, line 197: this can happen if the internal links are flapping, because we collect only the `up` ports in the port_map?, should we proceed with shutdown on the ports iso of ignoring the port?

arlakshm (commented):
as comments

PR #2971, opened 2023-09-05
[201911] [multi_asic] Script to monitor errors on internal links 

tjchadaga:
#### What I did
Added script to monitor SAI_PORT_STAT_IF_IN_ERRORS & SAI_PORT_STAT_IF_OUT_ERRORS on internal (backend) ports of multi-asic device. 

Duplicate of https://github.com/sonic-net/sonic-utilities/pull/2966. Addressed comments provided there

#### How I did it
Script is added to monit to regularly scan and check for errors on internal ports. If error count is above configured threshold, and the internal port-channel has more than min_links active links, then the port is shut down and syslog is generated.

#### How to verify it
UT cases are added to test the error detection.


arlakshm (approved):
lgtm

abdosi (approved):

PR #2973, opened 2023-09-07
[sonic-package-manager] Increate timeout for sonic-package-manager migrate

yaqiangz:
#### What I did
When we migrate package via sonic-package-manager in some low-performance device, getting docker image from old image via docker socket may timeout by default timeout setting (60s) [client.py](https://github.com/docker/docker-py/blob/main/docker/api/client.py#L106). This PR is to increase timeout to 120s.

#### How I did it
Increase docker client timeout from 60s to 120s.

#### How to verify it
1. ut passed.
2. Build image with INCLUDE_MACSEC, and install it, macsec image was not include. And the build image with this PR, install it. Migrate package successfully.

### DIFF
--- a/sonic_package_manager/manager.py
+++ b/sonic_package_manager/manager.py
@@ -660,6 +660,10 @@ class PackageManager:
                                  skip_host_plugins=skip_host_plugins)
 
     @under_lock
+    def get_docker_client(self, dockerd_sock:str):
+        return docker.DockerClient(base_url=f'unix://{dockerd_sock}', timeout=120)
+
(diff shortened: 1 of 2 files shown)


Blueve (approved):

yaqiangz:
@yxieca Could you pls help to backport this PR to 202205? MSFT work item number is 24905913

yaqiangz:
@StormLiangMS Could you pls help to backport this bug fix PR to 202305? MSFT work item number is 24905913

PR #2990, opened 2023-09-20
Fix for GCU_DHCP_Relay KVM test Failure in HEAD PR

ridahanif96:
#### What I did

KVM Elastic test cases are failing due to missing mode attribute in configuration Updated GCU function by making it compatible with HEAD PR so that it can pass KVM Elastic test cases

#### How I did it
Modified generic_updater.py. Added a logic in  **def replace function** such that it will not fail in rollback. In this way it will pass KVM Elastic (GCU_Dhcp_relay.py test)


generic_config_updater/generic_updater.py
```diff
@@ -105,9 +105,12 @@ def replace(self, target_config):
         self.logger.log_notice("Config replacement starting.")
         self.logger.log_notice(f"Target config length: {len(json.dumps(target_config))}.")
 
-        self.logger.log_notice("Getting current config db.")
+        self.logger.log_notice(".")
         old_config = self.config_wrapper.get_config_db_as_json()
-
+        
+        old_config = switchport_mode_remove(old_config)
```
qiluo-msft, line 111: >switchport_mode_remove [](http://example.com/codeflow?start=21&length=22)

Do you want to use `self.switchport_mode_remove` ?

Please check test failure:
```
target_config = {'ACL_TABLE': {'DATAACL': {'policy_desc': 'DATAACL', 'ports': ['Ethernet4'], 'stage': 'ingress', 'type': 'L3'}, 'EVERF...}, 'VLAN': {'Vlan1000': {'dhcp_servers': ['192.0.0.1', '192.0.0.2', '192.0.0.3', '192.0.0.4'], 'vlanid': '1000'}}, ...}

    def replace(self, target_config):
        self.logger.log_notice("Config replacement starting.")
        self.logger.log_notice(f"Target config length: {len(json.dumps(target_config))}.")
    
        self.logger.log_notice(".")
        old_config = self.config_wrapper.get_config_db_as_json()
    
>       old_config = switchport_mode_remove(old_config)
E       NameError: name 'switchport_mode_remove' is not defined
```

ridahanif96: @qiluo-msft , Yes i want to remove "mode" attribute for this created this **self.switchport_mode_remove** function. All these test failures has been resolved and updated in new PR:  https://github.com/sonic-net/sonic-utilities/pull/2993. Please consider this new one.

PR #2999, opened 2023-09-27
[CMIS] Use the block size advertised by the xSFP to download the image if it is less than the default block size

stephenxs:
#### What I did

Use the block size advertised by the xSFP to download the image if it is less than the default block size

#### How I did it

It took the block size advertised by the xSFP module only when EPL was used.
In fact, it should be taken in both EPL and LPL.

#### How to verify it

Unit test

### DIFF
--- a/sfputil/main.py
+++ b/sfputil/main.py
@@ -1332,8 +1332,9 @@ def download_firmware(port_name, filepath):
 
     with click.progressbar(length=file_size, label="Downloading ...") as bar:
         address = 0
-        BLOCK_SIZE = MAX_LPL_FIRMWARE_BLOCK_SIZE if lplonly_flag else maxblocksize
-        if maxblocksize < BLOCK_SIZE:
+        if lplonly_flag:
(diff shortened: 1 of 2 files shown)


sfputil/main.py
```diff
@@ -1333,6 +1333,8 @@ def download_firmware(port_name, filepath):
     with click.progressbar(length=file_size, label="Downloading ...") as bar:
         address = 0
         BLOCK_SIZE = MAX_LPL_FIRMWARE_BLOCK_SIZE if lplonly_flag else maxblocksize
+        if maxblocksize < BLOCK_SIZE:
+            BLOCK_SIZE = maxblocksize
```
prgeor, line 1337: @stephenxs can you revise the code as follows: 
```
if (lpl page) {
  // At max the driver allows 116 Bytes write for LPL
   block size = min (maxblocksize , MAX_LPL_SIZE)
} else {
   block_size = maxblocksize
}
 ```

prgeor: @stephenxs its python, so hope you understand :)

stephenxs: @prgeor 
LGTM. Done.

prgeor (approved):

PR #3012, opened 2023-10-14
Update GCU rsyslog validator

wen587:
ADO: 25468726
#### What I did
In the privous PR https://github.com/sonic-net/sonic-utilities/pull/2991, we removed the rsyslog validator because the background daemon can now react(reset-failed & restart) the rsyslog service when the configuration changes. 

However, this change introduced an issue found in nightly tests. The problem is that there is a chance that the rsyslog service is not refreshed promptly after modification. 

We have two potential solutions: we can either modify the sonic-mgmt syslog test to wait more time for the completion of the rsyslog update or reintroduce the syslog validator with enhancements. As the update for rsyslog happens after the completion of the GCU apply-patch, I choose to modify in GCU to ensure that it accurately reflects the completion of the rsyslog change when the apply-patch process is finished.

Now the behavior in this pull request aligns with the SONiC CLI when updating rsyslog settings.

#### How I did it
Add back and update GCU rsyslog validator and align it with syslog CLI.
#### How to verify it
Unit test and E2E test

### DIFF
--- a/generic_config_updater/services_validator.py
+++ b/generic_config_updater/services_validator.py
@@ -48,6 +48,18 @@ def _service_restart(svc_name):
     return rc == 0
 
 
+def rsyslog_validator(old_config, upd_config, keys):
+    old_syslog = old_config.get("SYSLOG_SERVER", {})
+    upd_syslog = upd_config.get("SYSLOG_SERVER", {})
+
+    if old_syslog != upd_syslog:
+        os.system("systemctl reset-failed rsyslog-config rsyslog")
+        rc = os.system("systemctl restart rsyslog-config")
+        if rc != 0:
+            return False
+    return True
+
+
 def dhcp_validator(old_config, upd_config, keys):
     return _service_restart("dhcp_relay")
 
--- a/generic_config_updater/gcu_services_validator.conf.json
+++ b/generic_config_updater/gcu_services_validator.conf.json
@@ -31,6 +31,9 @@
         "PORT": {
             "services_to_validate": [ "port_service" ]
         },
+        "SYSLOG_SERVER":{
+            "services_to_validate": [ "rsyslog" ]
+        },
         "DHCP_RELAY": {
             "services_to_validate": [ "dhcp-relay" ]
         },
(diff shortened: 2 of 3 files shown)


generic_config_updater/services_validator.py
```diff
@@ -48,6 +48,18 @@ def _service_restart(svc_name):
     return rc == 0
 
 
+def rsyslog_validator(old_config, upd_config, keys):
+    old_syslog = old_config.get("SYSLOG_SERVER", {})
+    upd_syslog = upd_config.get("SYSLOG_SERVER", {})
+
+    if old_syslog != upd_syslog:
+        os.system("systemctl reset-failed rsyslog-config rsyslog")
```
qiluo-msft, line 56: >system [](http://example.com/codeflow?start=11&length=6)

@maipbui to review: why this line does not fail semgrep checker?

maipbui: it's static value, not dynamic value so the checker will not fail

maipbui: However, we can consider use `subprocess.call` here as it's better option, ref: https://stackoverflow.com/questions/44730935/advantages-of-subprocess-over-os-system, https://docs.python.org/3/library/os.html#os.system

generic_config_updater/services_validator.py
```diff
@@ -48,6 +48,18 @@ def _service_restart(svc_name):
     return rc == 0
 
 
+def rsyslog_validator(old_config, upd_config, keys):
+    old_syslog = old_config.get("SYSLOG_SERVER", {})
+    upd_syslog = upd_config.get("SYSLOG_SERVER", {})
+
+    if old_syslog != upd_syslog:
```
qiluo-msft, line 55: >!= [](http://example.com/codeflow?start=18&length=2)

Use the native recursive comparison is working in this case, but is considered dangerous.

By json protocol, the order in a list matters, but by ConfigDB schema, the order in a list does not matter. For example /EVERFLOW/ports is a list.

Could you use a general ConfigDB aware json recurisve comparison? Or in this case, just use a ConfigDB/SYSLOG_SERVER specific comparison. check example in `vlan_validator()`.

ganglyu (approved):

StormLiangMS:
@wen587  do you have test with 202305 image?

wen587:
Hi @StormLiangMS , I've tested in 202305 image.

PR #3014, opened 2023-10-16
[db_migrator] Fix the broken version chain

vivekrnv:
#### Why I did

Warm Reboot from 202205 -> latest labels is failing because of an extra version that was added to 202205 and that was missing in later branches.

#### How I did it

Add a placeholder for version 3_0_7 in latest labels

#### How to verify it

WR from 202205 to 202305 and try running db_migrator

```
root@r-anaconda-simx-139:/home/admin# db_migrator.py -o migrate
  File "/usr/local/bin/db_migrator.py", line 1133, in main
    result = getattr(dbmgtr, operation)()
  File "/usr/local/bin/db_migrator.py", line 1085, in migrate
    next_version = getattr(self, version)()
AttributeError: 'DBMigrator' object has no attribute 'version_3_0_7'
'DBMigrator' object has no attribute 'version_3_0_7'
usage: db_migrator.py [-h] [-o operation migrate, set_version, get_version] [-s unix socket] [-n asic namespace]

optional arguments:
  -h, --help            show this help message and exit
  -o operation (migrate, set_version, get_version)
                        operation to perform [default: get_version]
  -s unix socket        the unix socket that the desired database listens on
  -n asic namespace     The asic namespace whose DB instance we need to connect
```

With this change:
```

root@r-anaconda-simx-139:/home/admin# db_migrator.py -o migrate
root@r-anaconda-simx-139:/home/admin# redis-cli -n 4 hgetall "VERSIONS|DATABASE"
1) "VERSION"
2) "version_4_0_3"
```

### DIFF
--- a/scripts/db_migrator.py
+++ b/scripts/db_migrator.py
@@ -984,10 +984,19 @@ class DBMigrator():
     def version_3_0_6(self):
         """
         Version 3_0_6
-        This is the latest version for 202211 branch
         """
 
         log.log_info('Handling version_3_0_6')
+        self.set_version('version_3_0_7')
+        return 'version_3_0_7'
+
+    def version_3_0_7(self):
+        """
+        Version 3_0_7
+        This is the latest version for 202205 branch
+        """
+
+        log.log_info('Handling version_3_0_7')
         self.set_version('version_4_0_0')
         return 'version_4_0_0'
 


dgsudharsan (approved):

scripts/db_migrator.py
```diff
@@ -984,10 +984,19 @@ def version_3_0_5(self):
     def version_3_0_6(self):
         """
         Version 3_0_6
-        This is the latest version for 202211 branch
         """
 
         log.log_info('Handling version_3_0_6')
+        self.set_version('version_3_0_7')
```
vaibhavhd, line 990: I think the change here should reflect the real state of version 3_0_6. In other words, the PR that added 3_0_6 to 202205 should also have updated master branch to avoid running into this issue in the first place.

IMO the migration logic added in the PR https://github.com/sonic-net/sonic-utilities/pull/2870 should be updated here.

dgsudharsan: This is already handled as part of version 4_0_2 in master and 202305 through this PR https://github.com/sonic-net/sonic-utilities/pull/2839. Hence we may just want to add a dummy version of 3_0_6  to avoid db_migrator failure.

vadymhlushko-mlnx (approved):

vaibhavhd (approved):

StormLiangMS:
ADO:25592749

PR #3025, opened 2023-10-24
Run yang validation in unit test

ganglyu:
#### What I did
Add unit test for db_migrator.py: ConfigDB passing yang validation.
Microsoft ADO: 24657445
Pending on https://github.com/sonic-net/sonic-buildimage/pull/16974

#### How I did it
Add yang validation for unit test, and fix test data to pass yang validation.

#### How to verify it
Run sonic-utilities end to end test.

### DIFF
--- a/scripts/db_migrator.py
+++ b/scripts/db_migrator.py
@@ -7,6 +7,7 @@ import sys
 import traceback
 import re
 import sonic_yang
+import syslog
 
 from sonic_py_common import device_info, logger
 from swsscommon.swsscommon import SonicV2Connector, ConfigDBConnector, SonicDBConfig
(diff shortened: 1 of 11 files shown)


tests/db_migrator_input/config_db/cross_branch_upgrade_to_4_0_3_expected.json
```diff
@@ -3,17 +3,17 @@
         "VERSION": "version_4_0_3"
     },
     "FLEX_COUNTER_TABLE|ACL": {
-        "FLEX_COUNTER_STATUS": "true",
+        "FLEX_COUNTER_STATUS": "enable",
```
qiluo-msft, line 6: >enable [](http://example.com/codeflow?start=32&length=6)

Just wondering, should we fix yang model, or fix the test data?

scripts/db_migrator.py
```diff
@@ -1122,6 +1124,27 @@ def migrate(self):
             version = next_version
         # Perform common migration ops
         self.common_migration_ops()
+        if os.environ["UTILITIES_UNIT_TESTING"] == "2":
```
qiluo-msft, line 1127: >UTILITIES_UNIT_TESTING [](http://example.com/codeflow?start=23&length=22)

I am thinking this detection is useful in production runtime as background task, not just unit test. It will help detect bugs like:
1. db_migrator generates a new config but not yang validated
2. load_minigraph generates a new config but not yang validated

Not sure if there are other places the yang validation could help in runtime. But let's only detect and generate CRIT syslog, not to block the existing workflow.

tests/db_migrator_input/config_db/cross_branch_upgrade_to_4_0_3_expected.json
ganglyu: I will check with owner.

scripts/db_migrator.py
ganglyu: https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-py-common/sonic_py_common/logger.py#L101-L114
I can't find CRIT in logger.py

ganglyu: Updated

tests/db_migrator_input/config_db/cross_branch_upgrade_to_4_0_3_expected.json
ganglyu: @vadymhlushko-mlnx has confirmed this is a mistake.

wen587:
I saw this PR is pending on https://github.com/sonic-net/sonic-buildimage/pull/16974/files.
Just a reminder in another ADO #25626334, we are trying to use just string instead of pattern as there are many device type are not covered in this yang model: sonic-device_neighbor_metadata.yang

wen587 (approved):
LGTM

PR #3027, opened 2023-10-25
TACACSPLUS_PASSKEY_ENCRYPTION support Part - I

nmoray:
- What I did
Added a support of TACACS passkey encryption feature. 
Ref. : [HLD](https://github.com/sonic-net/SONiC/pull/1471)
This PR comprises the encryption logic.

- How I did it
Implemented the feature by following [HLD](https://github.com/sonic-net/SONiC/pull/1471)

- How to verify it
`1. Configured TACACS passkey:
root@sonic:/# config tacacs passkey <pass>

2. Verified whether passkey is encrypted:
root@sonic:/# show runningconfiguration all  | grep passkey
            "passkey": "U2FsdGVkX19kFwDeP3IhgqbLJeed3pXtazJ73FtmD3I="

3. Verified /etc/pam.d/common-auth-sonic file to validate if the passkey is decrypted correctly [Referred while ssh'ing into the device]
root@sonic:~# cat /etc/pam.d/common-auth-sonic | grep secret
auth    [success=done new_authtok_reqd=done default=ignore auth_err=die]    pam_tacplus.so server=<ip>:49 secret=<pass_in_plaintext> login=login timeout=5   try_first_pass
auth    [success=done new_authtok_reqd=done default=ignore auth_err=die]    pam_tacplus.so server=<ip>:49 secret=<pass_in_plaintext> login=login timeout=5   try_first_pass

4. Verified passkey is hidden in show tacacs output
root@sonic:~# show tacacs 
TACPLUS global auth_type pap (default)
TACPLUS global timeout 5 (default)
TACPLUS global passkey configured Yes

5. Verified user able to login into device with TACACS credentials`

- Related PR(s)
[PR#81](https://github.com/sonic-net/sonic-host-services/pull/81)


config/aaa.py
```diff
@@ -11,6 +12,41 @@
 RADIUS_MAXSERVERS = 8
 RADIUS_PASSKEY_MAX_LEN = 65
 VALID_CHARS_MSG = "Valid chars are ASCII printable except SPACE, '#', and ','"
+TACACS_PASSKEY_MAX_LEN = 65
+TACACS_SECRET_SALT = "2e6593364d369fba925092e0c1c51466c276faa127f20d18cc5ed8ae52bedbcd"
+
+def get_salt():
+    file_path = "/etc/shadow"
+    target_username = "admin"
```
madhupalu, line 20: this variable not used, can be cleaned up

madhupalu:
@nmoray please take care of unit test coverage for PR to be approved.

config/aaa.py
nmoray: Sure.

config/aaa.py
```diff
@@ -11,6 +12,41 @@
...
+                    if len(parts) == 4:
+                        salt = parts[2]
+                        break
+
+    except FileNotFoundError:
+        click.echo('File not found: ' % file_path)
+    except Exception as e:
+        click.echo('An error occurred: ' % str(e))
+
+    if salt == None:
```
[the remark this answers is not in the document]
nmoray: When there is no admin password configured.

config/aaa.py
```diff
@@ -11,6 +12,41 @@
 RADIUS_MAXSERVERS = 8
 RADIUS_PASSKEY_MAX_LEN = 65
 VALID_CHARS_MSG = "Valid chars are ASCII printable except SPACE, '#', and ','"
+TACACS_PASSKEY_MAX_LEN = 65
+TACACS_SECRET_SALT = "2e6593364d369fba925092e0c1c51466c276faa127f20d18cc5ed8ae52bedbcd"
```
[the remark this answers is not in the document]
nmoray: It is random hash and will be fixed on all the devices unless and util an explicit admin password is configured. This a fail-safe mechanism to ensure that the encryption / decryption succeeds even if there is no admin password configured on a device. 
If configured, the code will use the salt from a shadow file (from admin user credentials).

config/aaa.py
nmoray: Addressed.

config/aaa.py
```diff
@@ -240,7 +274,20 @@ def passkey(ctx, secret):
...
+            click.echo('Maximum of %d chars can be configured' % TACACS_PASSKEY_MAX_LEN)
+            return
+        elif not is_secret(secret):
+            click.echo(VALID_CHARS_MSG)
+            return
+        config_db = ConfigDBConnector()
+        config_db.connect()
+        outsecret, errs = encrypt_passkey(secret)
+        if not errs: 
+            add_table_kv('TACPLUS', 'global', 'passkey', outsecret)
```
liuh-80, line 287: Do you test if this binary will break /etc/sonic/config_db.json?

that file generated by sonic-cfggen, and when device reboot content in redis will save/load from it.

Because binary may contain any characters, I suggest convert binary to base64 string.

nmoray: HI Liuh, the hash returned by encrypt_passkey() API is in string format only. Thus there are no issues while adding it into config_db. Please refer to the below snippet.

<SNIP>
>>> out, err = encrypt_passkey("TEST")
>>> print(out)
U2FsdGVkX19yDSbDSlgRXhtBzTThlMFo8YmvfioH/qo=
>>> print(type(out))
<class 'str'>
<SNIP>

liuh-80: Thanks, this the result is a base64 format string.
So as my comments in the HLD PR, the yang model need change for longer key value, as you can see the original pass key is 'TEST' and encrypted key almost 10 times longer than original key.

liuh-80:
@nmoray , please fix coverage issue by add new test case:

@azure-pipelines-wrapper
coverage.Azure.sonic-utilities.Python3 — Pull Request Coverage

Total: 44 lines
**Missing: 39 lines**
Coverage: 11%
Threshold: 80%

config/aaa.py
nmoray: @liuh-80 one question related to sonic-system-tacacs.yang model. Is it being used for pushing the TACACS configs or pulling the telemetry data related to TACACS or both?

And in case of increasing the length of passkey leaf, do I need to make any other backward compatibility changes or nothing?

nmoray:
> @nmoray , please fix coverage issue by add new test case:
> 
> @azure-pipelines-wrapper coverage.Azure.sonic-utilities.Python3 — Pull Request Coverage
> 
> Total: 44 lines **Missing: 39 lines** Coverage: 11% Threshold: 80%

@liuh-80 can you please give me some pointers for better understanding of this AUT infra. Like, how to write a new testcase.

liuh-80:
> > @nmoray , please fix coverage issue by add new test case:
> > @azure-pipelines-wrapper coverage.Azure.sonic-utilities.Python3 — Pull Request Coverage
> > Total: 44 lines **Missing: 39 lines** Coverage: 11% Threshold: 80%
> 
> @liuh-80 can you please give me some pointers for better understanding of this AUT infra. Like, how to write a new testcase.

You can find existing test case and add new test case here: https://github.com/sonic-net/sonic-utilities/blob/master/tests/aaa_test.py

zhangyanzhao:
Reviewers, if you are ok with this PR, please help to approve it. Thanks.

madhupalu (approved):
I am good with the changes

qiluo-msft:
Please add testcases to cover new code, otherwise coverage checker will block this PR.

nmoray:
> Please add testcases to cover new code, otherwise coverage checker will block this PR.

@qiluo-msft Sure I will add the testcases. But In my opinion, if https://github.com/sonic-net/sonic-buildimage/pull/17201 PR gets merged first, it will be easy to focus on this dependent PR.

config/aaa.py
```diff
@@ -234,13 +239,37 @@ def authtype(ctx, type):
...
+        if len(secret) > TACACS_PASSKEY_MAX_LEN:
+            click.echo('Maximum of %d chars can be configured' % TACACS_PASSKEY_MAX_LEN)
+            return
+        elif not is_secret(secret):
+            click.echo(VALID_CHARS_MSG)
+            return
+
+        if encrypt:
+            try:
+                passwd = getpass.getpass()
```
brholmes1, line 258: Need to import getpass module in order for this to work

config/aaa.py
```diff
@@ -256,7 +285,8 @@ def passkey(ctx, secret):
 @click.option('-o', '--port', help='TCP port range is 1 to 65535, default 49', type=click.IntRange(1, 65535), default=49)
 @click.option('-p', '--pri', help="Priority, default 1", type=click.IntRange(1, 64), default=1)
 @click.option('-m', '--use-mgmt-vrf', help="Management vrf, default is no vrf", is_flag=True)
-def add(address, timeout, key, auth_type, port, pri, use_mgmt_vrf):
+@click.option('-e', '--encrypt', help='Enable passkey encryption feature', is_flag=True)
```
qiluo-msft, line 288: >encrypt [](http://example.com/codeflow?start=23&length=7)

This command line extra option is confusing, why you want to change TACPLUS/global/key_encrypt in 2 command lines? do you want to encrypt other passkey if the 2 commands are both used but with different encrypt option?

anders-nexthop (commented):
@nmoray this PR has many open review comments, can you please update the PR and resolve the comments?

config/aaa.py
```diff
@@ -1,16 +1,21 @@
...
 from jsonpatch import JsonPatchConflict
 from jsonpointer import JsonPointerException
 import utilities_common.cli as clicommon
+from sonic_py_common.security_cipher import master_key_mgr 
 
 ADHOC_VALIDATION = True
 RADIUS_MAXSERVERS = 8
 RADIUS_PASSKEY_MAX_LEN = 65
 VALID_CHARS_MSG = "Valid chars are ASCII printable except SPACE, '#', and ','"
+TACACS_PASSKEY_MAX_LEN = 65
```
anders-nexthop, line 16: https://github.com/sonic-net/sonic-buildimage/pull/17201 changes this limit to 256, do you want to change it here as well?

config/aaa.py
```diff
@@ -234,13 +239,37 @@ def authtype(ctx, type):
 
 @click.command()
 @click.argument('secret', metavar='<secret_string>', required=False)
+@click.option('-e', '--encrypt', help='Enable passkey encryption feature', is_flag=True)
 @click.pass_context
-def passkey(ctx, secret):
+def passkey(ctx, secret, encrypt):
     """Specify TACACS+ server global passkey <STRING>"""
     if ctx.obj == 'default':
         del_table_key('TACPLUS', 'global', 'passkey')
```
anders-nexthop: do we need to reset key_encrypt here? On the one hand I don't think it will matter, since we always set key_encrypt correctly when a new passkey is added. But on the other, we don't really want to leave stale configuration hanging around without cause.

config/aaa.py
```diff
@@ -278,7 +308,23 @@ def add(address, timeout, key, auth_type, port, pri, use_mgmt_vrf):
...
         if key is not None:
-            data['passkey'] = key
+            if encrypt:
+                try:
+                    passwd = getpass.getpass()
+                except Exception as e:
+                    click.echo('getpass aborted' % e)
+                    return
+                add_table_kv('TACPLUS', 'global', 'key_encrypt', True)
+                outsecret, errs = secure_cipher.encrypt_passkey('TACPLUS', key, passwd)
```
anders-nexthop: This won't work. Each TACPLUS server can have its own passkey, but we are only setting one global encryption password. If I configure two servers, the password for the second will override the first, and the first passkey will not be usable. Further, we are using the single global key_encrypt var for every server, so if a server is configured without it we will brick the others.

I can see two ways to resolve this. Either we can add a separate passkey for each server (which means extending the passkey manager to handle more than one TACACS key, and extend the yang model to have a key_encrypt value for each individual server), or we can enhance the logic of the cipherkey manager somewhat to re-use the same passkey for every TACACS server (we could make the 'passwd' arg of `encrypt_passkey` optional, and just use the existing passkey if it exists. we would then configure that only through the global cli command). 

I guess a third option would be to make the passkey generation an explicit step, separate from the encryption. so instead of passing a passwd to `encrypt_passkey()` we would have a separate function `add_cipher_password()` which adds a password for a given feature, and then `encrypt_passkey()` would just use that configured password. We could add logic to call `add_cipher_password()` automatically the first time `encrypt_passkey()` is called for a feature, or maybe call it only when the global option changes.

config/aaa.py
nmoray: agreed!

config/aaa.py
nmoray: If the TACPLUS table itself is deleted, no need to worry about key_encrypt as it is part of the same table only.

config/aaa.py
```diff
@@ -234,13 +239,37 @@ def authtype(ctx, type):
...
+            click.echo('Maximum of %d chars can be configured' % TACACS_PASSKEY_MAX_LEN)
+            return
+        elif not is_secret(secret):
+            click.echo(VALID_CHARS_MSG)
+            return
+
+        if encrypt:
+            try:
+                passwd = getpass.getpass()
+            except Exception as e:
```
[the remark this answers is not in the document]
nmoray: Is it okay to have following?

```
except (EOFError, KeyboardInterrupt):
    print("\nInput cancelled.")
    passwd = None
except Exception as e:
    print(f"Unexpected error: {e}")
    passwd = None
```

config/aaa.py
```diff
@@ -234,13 +239,37 @@ def authtype(ctx, type):
...
+                click.echo('getpass aborted' % e)
```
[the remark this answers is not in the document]
nmoray: Agreed

config/aaa.py
```diff
@@ -234,13 +239,37 @@ def authtype(ctx, type):
...
+                return
+            add_table_kv('TACPLUS', 'global', 'key_encrypt', True)
```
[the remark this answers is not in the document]
nmoray: Are you saying it should be (t)rue? I just referred below.
`add_table_kv('AAA', 'authentication', 'failthrough', True)`

config/aaa.py
```diff
@@ -234,13 +239,37 @@ def authtype(ctx, type):
...
+                passwd = getpass.getpass()
+            except Exception as e:
+                click.echo('getpass aborted' % e)
+                return
+            add_table_kv('TACPLUS', 'global', 'key_encrypt', True) 
+            outsecret, errs = secure_cipher.encrypt_passkey('TACPLUS', secret, passwd)
+            if not errs:
+                add_table_kv('TACPLUS', 'global', 'passkey', outsecret)
+            else:
+                click.echo('Passkey configuration failed' % errs)
```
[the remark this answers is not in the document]
nmoray: Agreed!

config/aaa.py
```diff
@@ -234,13 +239,37 @@ def authtype(ctx, type):
...
+                return
+        else:
+            add_table_kv('TACPLUS', 'global', 'key_encrypt', False)
+            add_table_kv('TACPLUS', 'global', 'passkey', secret)
+            secure_cipher.del_cipher_pass()
```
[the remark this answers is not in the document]
nmoray: yeah.

config/aaa.py
nmoray: Is there any real production usecase where the user need different passkeys for different TACACS servers? IMO, better to have a single passkey across the servers. If so, it can easily be handled by making use of is_key_encrypt_enabled() API while configuring the passkey. 

If the passkey is already encrypted while configuring the global TACACS configurations, we will check if it is already configured via is_key_encrypt_enabled() in add() function while configuring TACACS server level configs and vice versa.

This way, we can have a single FEATURE id for TACACS, RADIUS and so on. Let me know your thoughts on the same.

anders-nexthop: I'm not sure how widespread the use case is, but since the ability exists to have different passkeys for different servers I don't think we would want to get rid of that. However, I do think we could share the same _cipherpass_ between servers, which is I think what you are suggesting here. That would work, and i see that `is_key_encrypt_enabled()` returns the cipherpass, so we can use that to pass to `encrypt_passkey()` (we might want to rename `is_key_encrypt_enabled()` func at some point to make it clear that it returns the cipherpass). Would it be better to move the `getpass()` logic to security_cipher code itself? and only prompt the user for the encryption password the first time the FEATURE id is used? That probably makes more sense if we are allowing a FEATURE id to be used by more than one caller, and it would be re-usable in that case as well.

Also, what about removing the cipherpass? Right now we always remove it whenever the feature is disabled, but that won't work if more than one caller is using the same cipherpass. We could keep a reference count, or check in the caller for any server/global tables and only remove the cipherpass if none are found (or at least none that have `key_encrypt` enabled). That seems a bit messy but is maybe the right approach. How else would we ever know to remove the cipherpass? Do we even need to get rid of a cipherpass for a feature once it's added? We do want to be able to rotate them if needed. what do you thinK?

config/aaa.py
nmoray: This is the length set for actual secret which is in plaintext. IMO, 65 chars are good enough.

config/aaa.py
nmoray: I am planning to have following design changes, let me know your thoughts on the same.
- Every module who wants to use security Cipher application has to register with it first. As a part of registration, it has to specify the Feature type and a callback function which will further be used at the time of password rotation.  This way, the security Cipher module will be having list of callbacks that needs to be called when a user wants to rotate the password. It's individual module's / application's  job to implement that callback. The implementation would be, just re-encrypt and update the encrypted key in the respective CONFIG_DB tables. This way, we can have only one password per feature type and multiple modules can make use of same password for generating their unique encrypted key.
- New API to set the password which will add a new password only if it is not already set. 
- Rotate API which will only be allowed to modify the existing password as it is having the ability to update the passkey (with re-encryption) in all the registered modules
- Change the cipher_pass text file to a json file
- Extend the cipher_pass to hold the list of callbacks per feature along with a password
Something like,
```
{
  "RADIUS": {
    "callbacks": [
      "rotate_radius_server1_passkey",
      "rotate_radius_server2_passkey"
    ],
    "password": "TEST1"
  },
  "TACPLUS": {
    "callbacks": [
      "rotate_tacplus_global_passkey"
    ],
    "password": "TEST2"
  }
}
```

config/aaa.py
```diff
@@ -278,7 +308,23 @@ def add(address, timeout, key, auth_type, port, pri, use_mgmt_vrf):
...
+                outsecret, errs = secure_cipher.encrypt_passkey('TACPLUS', key, passwd)
+                if not errs:
+                    data['passkey'] = outsecret
+                else:
+                    click.echo('Passkey configuration failed' % errs)
+                    return
+            else:
+                add_table_kv('TACPLUS', 'global', 'key_encrypt', False)
+                data['passkey'] = key
+                secure_cipher.del_cipher_pass()
```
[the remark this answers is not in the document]
nmoray: As per the updated design, the application module has to call deregister(callback) which will eventually remove the respective callback from the list in json file. And if the registered callback list is empty thenthe respective password will also be removed.

config/aaa.py
nmoray: @anders-nexthop Please refer to https://github.com/sonic-net/sonic-buildimage/pull/22711 PR. Here I have updated the existing design to support both rotate feature as well as the support of reusability of same the password.

nmoray:
This PR has been closed and created a new one https://github.com/sonic-net/sonic-utilities/pull/3936.

PR #3031, opened 2023-10-31
[config] config reload should generate sysinfo if missing 

wen587:
ADO:17746072
#### What I did
Missing platform and mac in CONFIG_DB will result in container failure. We should make the config reload generate those info if missing.
#### How I did it
Add missing sys info if config_db.json doesn't contain it.
#### How to verify it
Unit test

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1540,6 +1540,19 @@ def reload(db, filename, yes, load_sysinfo, no_service_restart, force, file_form
             click.echo("The config file {} doesn't exist".format(file))
             continue
 
+        if file_format == 'config_db':
+            file_input = read_json_file(file)
+
+            platform = file_input.get("DEVICE_METADATA", {}).\
+                get("localhost", {}).get("platform")
+            mac = file_input.get("DEVICE_METADATA", {}).\
+                get("localhost", {}).get("mac")
+
+            if not platform or not mac:
+                log.log_warning("Input file does't have platform or mac. platform: {}, mac: {}"
+                    .format(None if platform is None else platform, None if mac is None else mac))
+                load_sysinfo = True
+
         if load_sysinfo:
             try:
                 command = [SONIC_CFGGEN_PATH, "-j", file, '-v', "DEVICE_METADATA.localhost.hwsku"]
--- a/config/main.py
+++ b/config/main.py
@@ -148,12 +148,8 @@ def write_json_file(json_input, fileName):
     try:
         with open(fileName, 'w') as f:
             json.dump(json_input, f, indent=4)
-    except FileNotFoundError:
-        click.echo("{}".format(str(e)), err=True)
-        raise click.Abort()
     except Exception as e:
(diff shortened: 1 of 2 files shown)


config/main.py
```diff
@@ -1525,6 +1534,12 @@ def reload(db, filename, yes, load_sysinfo, no_service_restart, force, file_form
         # Get the file from user input, else take the default file /etc/sonic/config_db{NS_id}.json
         if cfg_files:
             file = cfg_files[inst+1]
+            # Save to tmpfile in case of stdin input which can only be read once
+            if file.endswith("stdin"):
```
ganglyu, line 1538: The file endswith "stdin" is stdin?

tests/config_test.py
```diff
@@ -261,6 +262,34 @@ def test_config_reload(self, get_cmd_module, setup_single_broadcom_asic):
 
             assert "\n".join([l.rstrip() for l in result.output.split('\n')][:1]) == reload_config_with_sys_info_command_output
 
+    def test_config_reload_stdin(self, get_cmd_module, setup_single_broadcom_asic):
+        with mock.patch("utilities_common.cli.run_command", mock.MagicMock(side_effect=mock_run_command_side_effect)) as mock_run_command:
+            (config, show) = get_cmd_module
+
+            dev_stdin = os.path.join(mock_db_path, "dev_stdin")
```
ganglyu, line 269: this input is not stdin?

ganglyu (approved):

config/main.py
```diff
@@ -142,6 +143,14 @@ def read_json_file(fileName):
         raise Exception(str(e))
     return result
 
+# write given JSON file
+def write_json_file(json_input, fileName):
+    try:
+        with open(fileName, 'w') as f:
+            json.dump(json_input, f, indent=4)
+    except Exception as e:
+        raise Exception(str(e))
```
qiluo-msft, line 152: except and raise immediately does not add value. It is the same as no except at all.

config/main.py
```diff
@@ -1525,6 +1534,12 @@ def reload(db, filename, yes, load_sysinfo, no_service_restart, force, file_form
         # Get the file from user input, else take the default file /etc/sonic/config_db{NS_id}.json
         if cfg_files:
             file = cfg_files[inst+1]
+            # Save to tmpfile in case of stdin input which can only be read once
+            if file == "/dev/stdin":
+                file_input = read_json_file(file)
+                (_, tmpfname) = tempfile.mkstemp(dir="/tmp", suffix="_configReloadStdin")
```
qiluo-msft, line 1540: Look around the login inside this function, it seems one time reading of file will be enough. Let's take further effort to reorg the existing code and optimize the logic.

wen587: [sonic-cfggen -j](https://github.com/sonic-net/sonic-utilities/blob/master/config/main.py#L1573) read it first. Then [sonic-cfggen](https://github.com/sonic-net/sonic-utilities/blob/master/config/main.py#L1624) read it second.

PR #3041, opened 2023-11-15
Fix test execution on Bookworm

saiarcot895:
#### What I did

Running tests with `python3 -m pytest` instead of `python setup.py test` causes some breakages on Bookworm. Some of this has to do with what module searc paths get added, while others appear to be related to changes in the `mock` module between Python 3.9 and Python 3.11.

Fix this and make sure it works on both Bullseye and Bookworm.

#### How I did it

Fix mock.patch entries so that the correct object reference gets patched, and remove duplicate test functions in `TestMuxcable`.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,2 +1,3 @@
 [pytest]
 addopts = --cov-config=.coveragerc --cov --cov-report html --cov-report term --cov-report xml --junitxml=test-results.xml -vv
+pythonpath = .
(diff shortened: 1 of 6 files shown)


saiarcot895:
For `tests/muxcable_test.py` specifically: number of tests executed went from 96 to 97 (which is exepcted, as one test for JSON output had the same name as a different test case, and has been renamed in this PR so that it is executed).

vdahiya12 (approved):

PR #3044, opened 2023-11-16
route_check - get ROUTE_TABLE content fail - patch

inspurSDN:
Root Cause:
        1. Not all /128 routes can be removed and should be treated as special

    Solution:
        1. only remove /128 when tbl.get(k)[0] == False

### DIFF
--- a/scripts/route_check.py
+++ b/scripts/route_check.py
@@ -356,6 +356,12 @@ def filter_out_local_interfaces(keys):
     tbl = swsscommon.Table(db, 'ROUTE_TABLE')
 
     for k in keys:
+        if tbl.get(k)[0] == False:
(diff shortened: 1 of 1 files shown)


scripts/route_check.py
```diff
@@ -356,6 +356,12 @@ def filter_out_local_interfaces(keys):
     tbl = swsscommon.Table(db, 'ROUTE_TABLE')
 
     for k in keys:
+        if tbl.get(k)[0] == False:
+            # ISU2023101259067, if kernel route prefix is xxxx::yy/128, fpmsyncd will write ROUTE_TABLE:xxxx::yy without /128 to apll db
```
prsunny, line 360: What is the ISUXXX, please remove it from comment. Also, can you add a unit-test?

PR #3047, opened 2023-11-20
Fix added to restart DHCP Relay service automatically in the background

saravanan-i-CLS:
… address is changed

#### What I did
When VLAN ip address is modified (either added or deleted), DHCP relay is not restarted automatically which intern causes interface ip address not being inserted in DHCP relay packets
#### How I did it
Restart DHCP Relay service in background when click cli tries to modify IP address of VLAN which has DHCP IPV4 helper or DHCP IPV6 Destination address configured already.
#### How to verify it
Add or remove Vlan Ip address which has DHCP helper ipv4 address configured


config/main.py
jasper-hou-micas, line 1: When multi ips are configured on the vlan , can we choose one to be the giaddr ?  As if giaddr  is changed and dhcp server is lack of related ip pool，dhcp relay will fail.

PR #3048, opened 2023-11-20
[wol] Implement wol command line utility

lizhijianrd:
#### What I did

Implement `wol` command-line utility based on [Wake-on-LAN-HLD.md](https://github.com/sonic-net/SONiC/blob/master/doc/wol/Wake-on-LAN-HLD.md).

#### How to verify it

Add unittest to verify code.

### DIFF
--- a/wol/main.py
+++ b/wol/main.py
@@ -1,0 +1,184 @@
+#!/usr/bin/env python3
+
+"""
+use wol to generate and send Wake-On-LAN (WOL) "Magic Packet" to specific interface
+
+Usage: wol_click [OPTIONS] INTERFACE TARGET_MAC
+
+  Generate and send Wake-On-LAN (WOL) "Magic Packet" to specific interface
+
+Options:
+  -b           Use broadcast MAC address instead of target device's MAC
+               address as Destination MAC Address in Ethernet Frame Header.
+               [default: False]
+  -p password  An optional 4 or 6 byte password, in ethernet hex format or
+               quad-dotted decimal  [default: ]
+  -c count     For each target MAC address, the count of magic packets to
+               send. count must between 1 and 5. This param must use with -i.
+               [default: 1]
+  -i interval  Wait interval milliseconds between sending each magic packet.
+               interval must between 0 and 2000. This param must use with -c.
+               [default: 0]
+  -v           Verbose output  [default: False]
+  -h, --help   Show this message and exit.
+
+Examples:
+  wol Ethernet10 00:11:22:33:44:55
+  wol Ethernet10 00:11:22:33:44:55 -b
+  wol Vlan1000 00:11:22:33:44:55,11:33:55:77:99:bb -p 00:22:44:66:88:aa
+  wol Vlan1000 00:11:22:33:44:55,11:33:55:77:99:bb -p 192.168.1.1 -c 3 -i 2000
+"""
+
+import binascii
+import click
+import copy
+import netifaces
+import os
+import socket
+import time
+
+CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
+EPILOG = """\b
+Examples:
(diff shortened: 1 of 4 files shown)


wol/main.py
```diff
@@ -0,0 +1,184 @@
...
+            password = socket.inet_aton(value)
+        except OSError:
+            raise click.BadParameter("invalid password format")
+    else:  # The length of a valid MAC address is 17.
+        try:
+            password = binascii.unhexlify(value.replace(':', '').replace('-', ''))
+        except binascii.Error:
+            raise click.BadParameter("invalid password format")
+    if len(password) not in [4, 6]:
+        raise click.BadParameter("password must be 4 or 6 bytes")
```
Blueve, line 140: this statement is conflicting with line 128

wol/main.py
```diff
@@ -0,0 +1,184 @@
...
+    return password
```
Blueve, line 141: The function is unlikely just `validate` the password, it parsed the password.
Suggest change name to `parse_password`

wol/main.py
```diff
@@ -0,0 +1,184 @@
...
+
+def validate_interface(ctx, param, value):
+    if value not in netifaces.interfaces():
+        raise click.BadParameter("invalid SONiC interface name {}".format(value))
+    if get_interface_operstate(value) != 'up':
+        raise click.BadParameter("interface {} is not up".format(value))
+    return value
+
+
+def validate_target_mac(ctx, param, value):
```
Blueve, line 113: `parse_target_mac`

wol/main.py
```diff
@@ -0,0 +1,184 @@
...
+    mac_list = []
+    for mac_s in value.split(','):
+        try:
+            mac_b = binascii.unhexlify(mac_s.replace(':', '').replace('-', ''))
```
Blueve, line 117: what is the `_b` means?

wol/main.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+def get_interface_operstate(interface):
+    with open('/sys/class/net/{}/operstate'.format(interface), 'r') as f:
+        return f.read().strip().lower()
+
+
+def get_interface_mac(interface):
+    mac = netifaces.ifaddresses(interface)[netifaces.AF_LINK][0].get('addr')
+    return MacAddress(binascii.unhexlify(mac.replace(':', '').replace('-', '')))
```
Blueve, line 80: There are many places you use this pattern:
```
binascii.unhexlify(mac.replace(':', '').replace('-', '')
```

can you extract a method for it?

Blueve (changes requested):

wol/main.py
lizhijianrd: line 128 `return` directly, will not hit this line.

wol/main.py
lizhijianrd: Good catch! updated the function name.

wol/main.py
lizhijianrd: Updated.

wol/main.py
lizhijianrd: I used this suffix to describe the variable type is bytes. I've refactored the code of class `MacAddress`. No longer need this suffix now.

wol/main.py
lizhijianrd: Thanks for pointing out. I've refactor the `MacAddress` class to integrate this logic.

wol/main.py
Blueve: I mean the wording. You said that *must be* 4 or 6, but 0 is also acceptable based on your implementation.

lizhijianrd: I see. Updated.

Blueve (approved):

lizhijianrd:
Link sonic-net/SONiC#1508

PR #3052, opened 2023-11-24
Reduce generate_dump mem usage for cores

davidm-arista:
Add the core files to the tarball while they are been processed, this ensures that only one core file at a time will be consuming flash space inside the tarpath and the tarball.

### DIFF
--- a/scripts/generate_dump
+++ b/scripts/generate_dump
@@ -1611,9 +1611,9 @@ save_crash_files() {
         for file in $(find_files "/var/core/"); do
             # don't gzip already-gzipped log files :)
             if [ -z "${file##*.gz}" ]; then
-                save_file $file core false
+                save_file $file core false true
             else
-                save_file $file core true
+                save_file $file core true true
             fi
         done
     fi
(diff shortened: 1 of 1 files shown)


dgsudharsan:
@vivekrnv  Please review

vivekrnv (commented):
So the assumption is saving into tarball would be efficient than saving to flash path.  But we are not compressing the file while appending it to tarball so i'm not very sure how efficient this is than the current approach. 

Can you provide date to signify this?

StormLiangMS:
@vivekrnv @dgsudharsan could you help to comment if this is good to merge.

vivekrnv (approved):

dgsudharsan (approved):

StormLiangMS (approved):

StormLiangMS:
ADO:26118419

StormLiangMS:
@davidm-arista there is a cherry pick failure, could you help to create a separate PR to 202305?

liushilongbuaa:
Sorry. Maybe it is a bug in automation.
checking.

PR #3053, opened 2023-11-27
Migrate GNMI table

ganglyu:
#### What I did
We will use GNMI container to replace TELEMETRY container in public repo.
GNMI is using GNMI table for configuration, so we need to migrate configuration in CONFIG_DB.
For L2 mode, there's no GNMI configuration in minigraph, and there's no TELEMETRY table in CONFIG_DB, so we will not generate GNMI table.

#### How I did it
If there's no minigraph and golden config, copy GNMI configuration from TELEMETRY table.
If there're minigraph and golden config, copy GNMI configuration from minigraph or golden config.

#### How to verify it
Run db migrator unit test, test with minigraph and test without minigraph.


scripts/db_migrator.py
```diff
@@ -567,6 +567,25 @@ def migrate_telemetry(self):
         if not certs:
             self.configDB.set_entry("TELEMETRY", "certs", telemetry_data.get("certs"))
 
+    def migrate_gnmi(self):
+        # GNMI - add missing key
+        if not self.minigraph_data or 'GNMI' not in self.minigraph_data:
+            # If there's no GNMI in minigraph, copy configuration from CONFIG_DB TELEMETRY table
```
qiluo-msft, line 573: >If there's no GNMI in minigraph, copy configuration from CONFIG_DB TELEMETRY table [](http://example.com/codeflow?start=14&length=82)

Could you double check if this works with L2 switch mode? [https://github.com/sonic-net/SONiC/wiki/L2-Switch-mode](https://github.com/sonic-net/SONiC/wiki/L2-Switch-mode)

qiluo-msft: We previously has a bug that mistakenly generating config for telemetry in L2 mode.

scripts/db_migrator.py
```diff
@@ -1090,7 +1109,8 @@ def version_4_0_4(self):
         Version 4_0_4.
         """
         log.log_info('Handling version_4_0_4')
-
+        # Updating GNMI table
+        self.migrate_gnmi()
```
yxieca, line 1113: Adding to an existing version is not safe. What if this operation was executed on a device running an older sonic image which already has database version reached 4.0.4?

scripts/db_migrator.py
```diff
@@ -567,6 +567,25 @@ def migrate_telemetry(self):
         if not certs:
             self.configDB.set_entry("TELEMETRY", "certs", telemetry_data.get("certs"))
 
+    def migrate_gnmi(self):
+        # GNMI - add missing key
+        if not self.minigraph_data or 'GNMI' not in self.minigraph_data:
+            # If there's no GNMI in minigraph, copy configuration from CONFIG_DB TELEMETRY table
+            gnmi = self.configDB.get_entry('TELEMETRY', 'gnmi')
```
qiluo-msft, line 574: >get_entry [](http://example.com/codeflow?start=33&length=9)

Will this line still work if `TELEMETRY` table is completely missing in ConfigDB?

scripts/db_migrator.py
ganglyu: Fixed

scripts/db_migrator.py
ganglyu: yes, if TELEMETRY table is missing, db_migrator will not generate GNMI table.

scripts/db_migrator.py
```diff
@@ -618,6 +618,25 @@ def migrate_telemetry(self):
...
             self.configDB.set_entry("TELEMETRY", "certs", telemetry_data.get("certs"))
 
+    def migrate_gnmi(self):
+        if self.config_src_data:
+            if 'GNMI' in self.config_src_data:
+                # If there's GNMI in minigraph or golden config, copy configuration from config_src_data
+                gnmi_data = self.config_src_data['GNMI']
+                log.log_notice('Migrate GNMI configuration')
+                if 'gnmi' in gnmi_data:
+                    self.configDB.set_entry("GNMI", "gnmi", gnmi_data.get('gnmi'))
```
qiluo-msft, line 628: >self.configDB [](http://example.com/codeflow?start=20&length=13)

There are upgrade/downgrade use cases. If self.configDB['GNMI'] already exists, you should not modify, right?

ganglyu: I think database version is used to detect upgrade and downgrade, and maybe db_migrator will not invoke migrate_gnmi for downgrade?

qiluo-msft (approved):
LGTM, please check with other reviewers.

PR #3076, opened 2023-12-12
Support golden config in db migrator

ganglyu:
#### What I did
Need to support golden config in db migrator.

#### How I did it
If there's golden config json, read from golden config instead of minigraph.
And db migrator will use golden config data to generate new configuration.

#### How to verify it
Run unit test.


wen587 (dismissed):

scripts/db_migrator.py
```diff
@@ -53,11 +55,21 @@ def __init__(self, namespace, socket=None):
...
         # load config data from minigraph to get the default/hardcoded values from minigraph.py
         # this is to avoid duplicating the hardcoded these values in db_migrator
         self.minigraph_data = None
         try:
-            if os.path.isfile(MINIGRAPH_FILE):
+            if os.path.isfile(GOLDEN_CFG_FILE):
+                with open(GOLDEN_CFG_FILE) as f:
+                    golden_data = json.load(f)
+                    if namespace is None:
+                        self.minigraph_data = golden_data
```
qiluo-msft, line 69: >minigraph_data [](http://example.com/codeflow?start=29&length=14)

It's confusing to change the meaning of "minigraph_data". Suggest make it immutable, and create new var for new purpose.

If both GOLDEN_CFG_FILE and MINIGRAPH_FILE exists, the table inside GOLDEN_CFG_FILE will override table from MINIGRAPH_FILE. Please check with @wen587 and try reuse this override logic.

qiluo-msft (changes requested):
As comment

wen587 (approved):

wen587 (dismissed):

wen587 (approved):

utilities_common/helper.py
```diff
@@ -64,3 +65,21 @@ def get_port_pbh_binding(db_wrap, port, ns):
     ret = m_engine.fetch(req)
     pbh_tables, _ = get_matched_keys(ret)
     return pbh_tables
+
+
+def update_config(current_config, config_input):
+    """
+    Override current config with golden config
```
qiluo-msft, line 72: >Override [](http://example.com/codeflow?start=4&length=8)

Thanks for improving the code comments!

scripts/db_migrator.py
```diff
@@ -53,15 +57,41 @@ def __init__(self, namespace, socket=None):
...
+                self.__minigraph_data = parse_xml(MINIGRAPH_FILE)
         except Exception as e:
             log.log_error('Caught exception while trying to parse minigraph: ' + str(e))
             pass
+        # When both golden config and minigraph exists, override minigraph config with golden config
+        # config_src_data is the source of truth for config data
+        # this is to avoid duplicating the hardcoded these values in db_migrator
+        self.config_src_data = None
+        if self.__minigraph_data:
+            self.config_src_data = self.__minigraph_data
```
qiluo-msft, line 90: >__minigraph_data [](http://example.com/codeflow?start=40&length=16)

This is shallow copy. So any future change on right side will also appear on left side.
I understand it is not efficient to make `minigraph_data` immutable. So instead, you can call it `config_src_data`, and override it.

`__xxx` convention does not help your case, it is just make the member super private.

ganglyu: Fixed

ganglyu: For reused function, add an argument to update_config, and update_config will use deepcopy by default, but db migrator will use shallow copy.
For db migrator, add a method to generate config_src_data. golden_config_data and minigraph_data are local variables, so it's impossible to alter them and then we can prevent unintended change.

wen587 (approved):

qiluo-msft (approved):

yxieca (approved):

PR #3100, opened 2023-12-25
Fix database initialization for db_migrator

ganglyu:
#### What I did
db_migrator failed to initialize SonicDBConfig, and I fix this issue.

#### How I did it
If SonicDBConfig is already initialized, do not invoke initialize() again.

#### How to verify it
Run unit test, and verified on DUT.


FengPan-Frank (approved):
LGTM

PR #3102, opened 2023-12-27
Run yang validation in db migrator

ganglyu:
#### What I did
Add unit test/runtime alert for db_migrator.py/load_minigraph: ConfigDB passing yang validation

#### How I did it
Use subprocess to run yang validation, and then warm-reboot performance will not change.
Check /etc/sonic/mgmt_test_mark, this mark_file means this device is under end to end test.

* For unit test, db_migrator will wait for validation result and raise exception.
* For end to end test, db_migrator will wait for validation result and raise exception.

#### How to verify it
Run unit test and run end to end test


scripts/config_validator.py
```diff
@@ -0,0 +1,44 @@
...
+    # Run yang validation
+    yang_parser = sonic_yang.SonicYang(YANG_MODELS_DIR)
+    yang_parser.loadYangModel()
+    try:
+        yang_parser.loadData(configdbJson=config)
+        yang_parser.validate_data_tree()
+    except sonic_yang.SonicYangException as e:
+        log.log_error("Yang validation failed: " + str(e))
+        raise
+    if len(yang_parser.tablesWithOutYang):
```
wen587, line 38: We could make it 100% in our StarLab DUT.
However, I suspect if we could cover it 100% in PROD devices. It will break the PROD workflow if not fully coverred in PROD.
I prefer we make the check as log_warning and not raise exception right now.

wen587 (approved):

PR #3111, opened 2024-01-05
Backup STATE_DB PORT_TABLE|Ethernet during warm-reboot

mihirpat1:
#### What I did
Currently, entire PORT_TABLE in STATE_DB is being deleted during warm-reboot. Due to this, `host_tx_ready` changes to false after warm-reboot which causes the link to remain down.

#### How I did it
Backing up `host_tx_ready`, `NPU_SI_SETTINGS_SYNC_STATUS` and `CMIS_REINIT_REQUIRED` fields from `STATE_DB PORT_TABLE* during warm-reboot now.

#### How to verify it
Verified that host_tx_ready in STATE_DB PORT_TABLE is retained after warm-reboot and the link remains up. Also, ensured that the keys CMIS_REINIT_REQUIRED and NPU_SI_SETTINGS_SYNC_STATUS are retained after warm-reboot.
Before warm-reboot
```
root@sonic:/home/admin# redis-cli -n 6 hgetall "PORT_TABLE|Ethernet0"
 1) "state"
 2) "ok"
 3) "netdev_oper_status"
 4) "up"
 5) "admin_status"
 6) "up"
 7) "mtu"
 8) "9100"
 9) "CMIS_REINIT_REQUIRED"
10) "false"
11) "NPU_SI_SETTINGS_SYNC_STATUS"
12) "NPU_SI_SETTINGS_DEFAULT"
13) "supported_speeds"
14) "40000,100000"
15) "supported_fecs"
16) "none,rs"
17) "host_tx_ready"
18) "true"
19) "speed"
20) "100000"
21) "fec"
22) "N/A"
root@sonic:/home/admin# 
```

After warm-reboot script backs up PORT_TABLE and deletes unwanted fields
```
root@sonic:/home/admin# redis-cli -n 6 hgetall "PORT_TABLE|Ethernet0"
1) "CMIS_REINIT_REQUIRED"
2) "false"
3) "NPU_SI_SETTINGS_SYNC_STATUS"
4) "NPU_SI_SETTINGS_DEFAULT"
5) "host_tx_ready"
6) "true"
root@sonic:/home/admin# 
```

After switch boot-up post warm-reboot
```
root@sonic:/home/admin# redis-cli -n 6 hgetall "PORT_TABLE|Ethernet0"
 1) "state"
 2) "ok"
 3) "netdev_oper_status"
 4) "up"
 5) "admin_status"
 6) "up"
 7) "mtu"
 8) "9100"
 9) "supported_speeds"
10) "40000,100000"
11) "supported_fecs"
12) "none,rs"
13) "CMIS_REINIT_REQUIRED"
14) "false"
15) "NPU_SI_SETTINGS_SYNC_STATUS"
16) "NPU_SI_SETTINGS_DEFAULT"
17) "host_tx_ready"
18) "true"
19) "speed"
20) "100000"
21) "fec"
22) "N/A"
root@sonic:/home/admin# 
```

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -253,6 +253,7 @@ function backup_database()
                                           and not string.match(k, 'WARM_RESTART_ENABLE_TABLE|') \
                                           and not string.match(k, 'VXLAN_TUNNEL_TABLE|') \
                                           and not string.match(k, 'BUFFER_MAX_PARAM_TABLE|') \
+                                          and not string.match(k, 'PORT_TABLE|') \
                                           and not string.match(k, 'FAST_RESTART_ENABLE_TABLE|') then
                 redis.call('del', k)
             end
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -247,13 +247,22 @@ function backup_database()
     # Delete keys in stateDB except FDB_TABLE|*, MIRROR_SESSION_TABLE|*, WARM_RESTART_ENABLE_TABLE|*, FG_ROUTE_TABLE|*
     sonic-db-cli STATE_DB eval "
         for _, k in ipairs(redis.call('keys', '*')) do
-            if not string.match(k, 'FDB_TABLE|') and not string.match(k, 'WARM_RESTART_TABLE|') \
+            if string.match(k, 'PORT_TABLE|Ethernet') then
+                for i, f in ipairs(redis.call('hgetall', k)) do
+                    if i % 2 == 1 then
+                        if not string.match(f, 'host_tx_ready') \
+                            and not string.match(f, 'NPU_SI_SETTINGS_SYNC_STATUS') \
+                            and not string.match(f, 'CMIS_REINIT_REQUIRED') then
+                            redis.call('hdel', k, f)
+                        end
+                    end
+                end
+            elseif not string.match(k, 'FDB_TABLE|') and not string.match(k, 'WARM_RESTART_TABLE|') \
                                           and not string.match(k, 'MIRROR_SESSION_TABLE|') \
                                           and not string.match(k, 'FG_ROUTE_TABLE|') \
                                           and not string.match(k, 'WARM_RESTART_ENABLE_TABLE|') \
                                           and not string.match(k, 'VXLAN_TUNNEL_TABLE|') \
                                           and not string.match(k, 'BUFFER_MAX_PARAM_TABLE|') \
-                                          and not string.match(k, 'PORT_TABLE|') \
                                           and not string.match(k, 'FAST_RESTART_ENABLE_TABLE|') then
                 redis.call('del', k)
             end


scripts/fast-reboot
```diff
@@ -247,7 +247,17 @@ function backup_database()
     # Delete keys in stateDB except FDB_TABLE|*, MIRROR_SESSION_TABLE|*, WARM_RESTART_ENABLE_TABLE|*, FG_ROUTE_TABLE|*
     sonic-db-cli STATE_DB eval "
         for _, k in ipairs(redis.call('keys', '*')) do
-            if not string.match(k, 'FDB_TABLE|') and not string.match(k, 'WARM_RESTART_TABLE|') \
+            if string.match(k, 'PORT_TABLE|Ethernet') then
+                for i, f in ipairs(redis.call('hgetall', k)) do
+                    if i % 2 == 1 then
```
vaibhavhd, line 252: I am not clear on what this check looking for - ` if i % 2 == 1`

scripts/fast-reboot
```diff
@@ -247,7 +247,17 @@ function backup_database()
     # Delete keys in stateDB except FDB_TABLE|*, MIRROR_SESSION_TABLE|*, WARM_RESTART_ENABLE_TABLE|*, FG_ROUTE_TABLE|*
     sonic-db-cli STATE_DB eval "
         for _, k in ipairs(redis.call('keys', '*')) do
-            if not string.match(k, 'FDB_TABLE|') and not string.match(k, 'WARM_RESTART_TABLE|') \
+            if string.match(k, 'PORT_TABLE|Ethernet') then
```
vaibhavhd, line 250: @zjswhhh do you think keeping port_table intact in state-db will cause any change in mux init logic for dualtor warmboot?

scripts/fast-reboot
```diff
@@ -247,7 +247,17 @@ function backup_database()
     # Delete keys in stateDB except FDB_TABLE|*, MIRROR_SESSION_TABLE|*, WARM_RESTART_ENABLE_TABLE|*, FG_ROUTE_TABLE|*
```
vaibhavhd, line 247: Please update the comment accordingly.

scripts/fast-reboot
```diff
@@ -247,7 +247,17 @@ function backup_database()
...
     sonic-db-cli STATE_DB eval "
         for _, k in ipairs(redis.call('keys', '*')) do
-            if not string.match(k, 'FDB_TABLE|') and not string.match(k, 'WARM_RESTART_TABLE|') \
+            if string.match(k, 'PORT_TABLE|Ethernet') then
+                for i, f in ipairs(redis.call('hgetall', k)) do
+                    if i % 2 == 1 then
+                        if not string.match(f, 'host_tx_ready') \
+                            and not string.match(f, 'NPU_SI_SETTINGS_SYNC_STATUS') \
+                            and not string.match(f, 'CMIS_REINIT_REQUIRED') then
```
vaibhavhd, line 255: 1. From the PR's description it sounds like you need only `host_tx_ready` field. What is the reason for also persisting NPU and CMIS fields? Please add the reasoning as a comment here too.
2. Have you considered cross branch and in-branch warm-reboots where the 3 fields might get modified (deleted, name change, etc?) in the target image? How will the target image handle a scenario when state db has some filed which is not supported. This question arises from the fact that you are not preserving entire table, but just 3 fields from it. Existing cases preserved entire table.

To prevent us from getting into any unlikely scenarios in # 2 above, this handling can be done by the target image's portorch. In other words, when the device boots into target image, the portsorch initializes these fields afresh after checking that system is undergoing warmboot by checking system warm-restart flag. The downside in this approach is that portorch will have no idea what these fields were set to in the base image.

The argument is same for other fields such as OPER state (netdev_oper_status) and speeds. In the recovery path of warmboot these fields are set afresh based on SAI get calls (I think)

longhuan-cisco:
Just curious, what was the reason to not back up the entire table? Is it because some of the fields (e.g. `netdev_oper_status `) should be re-populated after warm-reboot?

mihirpat1:
> Just curious, what was the reason to not back up the entire table? Is it because some of the fields (e.g. `netdev_oper_status `) should be re-populated after warm-reboot?

@longhuan-cisco - Yes, you are correct. Hence, we decided to preserve selected fields which xcvrd/OA cares about and delete other fields from STATE_DB.

scripts/fast-reboot
prgeor: @vaibhavhd you may want to review this https://github.com/sonic-net/sonic-utilities/pull/3240/files

prgeor (approved):

prgeor:
@StormLiangMS @yxieca @bingwang-ms please cherry pick this to 202311. Need for warm reboot support for platforms using CMIS optics

prgeor:
@bingwang-ms we need this in 202405

bingwang-ms:
@prgeor Seems there is cherry-pick conflict. Please double check

mihirpat1:
> @prgeor Seems there is cherry-pick conflict. Please double check

@bingwang-ms I have removed the 202405 tags since this is already part of 202405.

PR #3114, opened 2024-01-10
[Mellanox buffer migrator] Do not touch the buffer model on generic SKUs if the buffer configuration is empty

stephenxs:
#### What I did

Do not touch the buffer model on generic SKUs if the buffer configuration is empty.

#### How I did it

Set the buffer model to traditional on generic SKUs in Mellanox db migrator only if the buffer configuration is not default and not empty.

#### How to verify it

Manually and mock test.

#### Details ####
Buffer configuration contains two parts:
1. the buffer model in `DEVICE_METADATA|localhost` which is from `init_cfg.json` and can be updated by Mellanox buffer migrator
2. the buffer pools, profiles, PGs, and queues which are renderred from the buffer templates in `config qos reload`

There was a logic to update the buffer model in Mellanox buffer migrator: if the buffer configuration is not default, the buffer model is set to traditional. However, if a device is installed from ONIE, the buffer configuration is also empty. As a result, the traditional buffer manager starts after the device is installed from ONIE, and it requires to restart the buffer manager to switch to the dynamic model. This can be done only by `config reload`.
It didn't matter since it was required to execute `config qos reload` to complete buffer configuration which required `config save` and `config reload` in any case due to issue https://github.com/sonic-net/sonic-buildimage/issues/9088.
Now that the issue has been fixed and `config reload` isn't required anymore to complete `config qos reload`, we should avoid setting the buffer model to traditional in such case, otherwise `config reload` is still required to switch the buffer model.

Verified the following scenarios:
1. non-default configuration generic SKU upgrade from 202305: warm/cold boot: expected: traditional model
2. default configuration generic SKU upgrade from 201911/202305: warm/cold boot: expected: dynamic model
3. install from ONIE: expected: dynamic model
4. MSFT SKU upgrade from 201911 by cold boot/ from 202012 by warm boot: expected: traditional model

### DIFF
--- a/scripts/mellanox_buffer_migrator.py
+++ b/scripts/mellanox_buffer_migrator.py
@@ -117,6 +117,7 @@ class MellanoxBufferMigrator():
         self.default_speed_list = ['1000', '10000', '25000', '40000', '50000', '100000', '200000', '400000']
         self.default_cable_len_list = ['5m', '40m', '300m']
         self.is_buffer_config_default = True
+        self.is_buffer_config_empty = False
 
     mellanox_default_parameter = {
         "version_1_0_2": {
@@ -645,7 +646,10 @@ class MellanoxBufferMigrator():
         # If some buffer pool is not default ones, don't need migrate
         for buffer_pool in default_buffer_pool_list_old:
             if buffer_pool not in configdb_buffer_pool_names and buffer_pool != 'ingress_lossy_pool':
+                # Check whether it's empty buffer configuration
                 log.log_notice("Default pool {} isn't in CONFIG_DB, skip buffer pool migration".format(buffer_pool))
+                if not self.configDB.get_keys('BUFFER_*'):
+                    self.is_buffer_config_empty = True
                 return True
 
         default_buffer_pools_old = self.mlnx_default_buffer_parameters(old_version, "buffer_pools")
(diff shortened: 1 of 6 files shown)


bingwang-ms:
Sorry I missed this review.
Can you please confirm if the change only applies to the scenario when installing SONiC from ONiE?

stephenxs:
> Sorry I missed this review. Can you please confirm if the change only applies to the scenario when installing SONiC from ONiE?

Hi
It is applied whenever the buffer configuration is empty. In most cases, it should be the case that the device is installed from ONIE.
In case a user removes all the buffer/QoS configurations and then saves the configuration, and then executes `config reload` it should also be applied.

bingwang-ms (approved):

bingwang-ms:
> > Sorry I missed this review. Can you please confirm if the change only applies to the scenario when installing SONiC from ONiE?
> 
> Hi It is applied whenever the buffer configuration is empty. In most cases, it should be the case that the device is installed from ONIE. In case a user removes all the buffer/QoS configurations and then saves the configuration, and then executes `config reload` it should also be applied.

Is this change required for other release branch?

stephenxs:
> > > Sorry I missed this review. Can you please confirm if the change only applies to the scenario when installing SONiC from ONiE?
> > 
> > 
> > Hi It is applied whenever the buffer configuration is empty. In most cases, it should be the case that the device is installed from ONIE. In case a user removes all the buffer/QoS configurations and then saves the configuration, and then executes `config reload` it should also be applied.
> 
> Is this change required for other release branch?

Yes. We suggest cherry-picking it to 202311. Tag added.

keboliu (approved):

PR #3130, opened 2024-01-23
[Master]: Additional check to skip FRR-Offloaded check if the bgp route-src was not selected as best

deepak-singhal0408:
#### What I did
Fixes https://github.com/sonic-net/sonic-buildimage/issues/17877
#### How I did it
Added additional check to skip FRR-Offloaded check if the routeSrc BGP was not selected as best.
#### How to verify it
Ran the script on multi-asic KVM device, and could confirm the route_check is passing.

### DIFF
--- a/scripts/route_check.py
+++ b/scripts/route_check.py
@@ -542,7 +542,7 @@ def check_frr_pending_routes():
             for entry in entries:
                 if entry['protocol'] != 'bgp':
                     continue
-				
+
                 # TODO: Also handle VRF routes. Currently this script does not check for VRF routes so it would be incorrect for us
                 # to assume they are installed in ASIC_DB, so we don't handle them.
                 if entry['vrfName'] != 'default':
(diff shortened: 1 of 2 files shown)


deepak-singhal0408:
@stepanblyschak for viz.

scripts/route_check.py
```diff
@@ -548,6 +548,10 @@ def check_frr_pending_routes():
                 if entry['vrfName'] != 'default':
                     continue
 
+		# skip if this bgp source prefix is not selected as best
```
stepanblyschak, line 551: Replace tabs with spaces

stepanblyschak (changes requested):
LGTM, but fix indent with spaces

stepanblyschak (approved):

deepak-singhal0408:
@qiluo-msft, @prsunny can you please help merge? Thanks!

prsunny (approved):

PR #3139, opened 2024-02-01
Fix link down issue on CMIS optics during warm-reboot

longhuan-cisco:
#### What I did

For now, STATE_DB PORT_TABLE will get cleaned during warm-reboot, causing `host_tx_ready` not preserved after warm-reboot, and CMIS optics will be brought down by cmis_mgr due to host_tx_ready=false.
This fix is to preserve STATE_DB PORT_TABLE during warm-reboot, to avoid link down on cmis optics.

#### How to verify it

Verified in warm-reboot case, 400G optics is not down anymore.

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -276,6 +276,7 @@ function backup_database()
     sonic-db-cli STATE_DB eval "
         for _, k in ipairs(redis.call('keys', '*')) do
             if not string.match(k, 'FDB_TABLE|') and not string.match(k, 'WARM_RESTART_TABLE|') \
+                                          and not string.match(k, 'PORT_TABLE|') \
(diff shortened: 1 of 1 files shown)


longhuan-cisco:
@mihirpat1 @prgeor Could you please take a look? This is the issue/fix Jagan discussed with you couple months ago.

mihirpat1:
> @mihirpat1 @prgeor Could you please take a look? This is the issue/fix Jagan discussed with you couple months ago.

@longhuan-cisco - I have currently created the below PR for fixing this issue
https://github.com/sonic-net/sonic-utilities/pull/3111/files

longhuan-cisco:
dup of https://github.com/sonic-net/sonic-utilities/pull/3111/files

PR #3155, opened 2024-02-06
Add vlan validation in config interface ip add command

matiAlfaro:
~

#### What I did
Fix for https://github.com/sonic-net/sonic-buildimage/issues/16975
Prevent configuration of illegal Vlan name in "config interface ip add" command
#### How I did it
Added Vlan name validation
##### Positive tests
config interface ip add Vlan100 1.1.1.1/24
config interface ip add Vlan4093 1.1.1.1/24
##### Negative tests
config interface ip add Vlan000000000000003 1.1.1.1/24
config interface ip add Vlan01 1.1.1.1/24
config interface ip add Vlan1.2 1.1.1.1/24
On illegal Vlan name command succeeded with no error
config interface ip add Vlan030 1.1.1.1/24
Error: Vlan030 is not a valid Vlan name

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -4513,6 +4513,13 @@ def ip(ctx):
     """Set IP interface attributes"""
     pass
 
+def validate_vlan_format(text):
+    pattern = re.compile(r'^Vlan([1-9]\d{0,2}|[1-3]\d{3}|40[0-8][0-9]|409[0-4])$')
+
+    if pattern.match(text):
+        return True
+    else:
+        return False
 #
 # 'add' subcommand
 #
@@ -4544,6 +4551,7 @@ def add(ctx, interface_name, ip_addr, gw):
     if interface_is_in_portchannel(portchannel_member_table, interface_name):
         ctx.fail("{} is configured as a member of portchannel."
                 .format(interface_name))
+    
 
     try:
         ip_address = ipaddress.ip_interface(ip_addr)
@@ -4576,6 +4584,12 @@ def add(ctx, interface_name, ip_addr, gw):
     table_name = get_interface_table_name(interface_name)
     if table_name == "":
         ctx.fail("'interface_name' is not valid. Valid names [Ethernet/PortChannel/Vlan/Loopback]")
+    
+    if table_name == "VLAN_INTERFACE":
+        if not validate_vlan_format(interface_name):
+            ctx.fail(f"{interface_name} is not a valid Vlan name")
+            return
+    
     interface_entry = config_db.get_entry(table_name, interface_name)
     if len(interface_entry) == 0:
         if table_name == "VLAN_SUB_INTERFACE":
--- a/config/main.py
+++ b/config/main.py
@@ -4516,6 +4516,11 @@ def ip(ctx):
 def validate_vlan_format(text):
     pattern = re.compile(r'^Vlan([1-9]\d{0,2}|[1-3]\d{3}|40[0-8][0-9]|409[0-4])$')
     return pattern.fullmatch(text) is not None
+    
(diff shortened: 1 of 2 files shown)


amazor (approved):

prsunny (dismissed):

config/main.py
```diff
@@ -4576,6 +4579,12 @@ def add(ctx, interface_name, ip_addr, gw):
     table_name = get_interface_table_name(interface_name)
     if table_name == "":
         ctx.fail("'interface_name' is not valid. Valid names [Ethernet/PortChannel/Vlan/Loopback]")
+    
+    if table_name == "VLAN_INTERFACE":
+        if not validate_vlan_format(interface_name):
```
dgsudharsan, line 4584: Shouldn't we be checking the presence of Vlan rather than validating the format? My recommendation is to query the VLAN_TABLE and confirm if Vlan is present before assigning an IP address

dgsudharsan (changes requested):

config/main.py
matiAlfaro: From my POV, you should be allowed to configure IP interface on a Vlan before creating it.
Once Vlan is created, configuration will apply.
This kind of behavior is also common on other switch devices (for example, Catalyst)

prsunny: I tend to agree with @dgsudharsan here. Just creating VLAN interface is not handled in backend (swss etc) until a VLAN table is present. So, why not check VLAN table entry and return an error log? But @dgsudharsan , do you think we can add this check to Vlan creation?

matiAlfaro: Added query of VLAN_TABLE as you suggest,
Also added option to ignore/skip this query, please let me know if you'd like this option removed.

config/main.py
```diff
@@ -4524,8 +4529,9 @@ def validate_vlan_format(text):
 @click.argument('interface_name', metavar='<interface_name>', required=True)
 @click.argument("ip_addr", metavar="<ip_addr>", required=True)
 @click.argument('gw', metavar='<default gateway IP address>', required=False)
+@click.option('-i','--ignore-vlan',is_flag=True, required=False, help="Allow configuration on not created Vlan")
```
prsunny, line 4532: I don't think this must be handled in this way. Lets simply throw an error if vlan does not exist asking user to create Vlan first?

matiAlfaro: done

dgsudharsan (approved):

prsunny (approved):

isabelmsft:
This validation requirement is not reflected in the VLAN yang model. Would you please improve the yang model so that it aligns with this requirement?

isabelmsft (changes requested):
This validation requirement is not reflected in the VLAN yang model: https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-vlan.yang. Would you please improve the yang model so that it aligns with this requirement?

matiAlfaro:
@isabelmsft - Please see https://github.com/sonic-net/sonic-buildimage/pull/18207

isabelmsft:
> @isabelmsft - Please see [sonic-net/sonic-buildimage#18207](https://github.com/sonic-net/sonic-buildimage/pull/18207)

Thanks @matiAlfaro, the requirement that vlan exists before IP address is configured is supported in yang model here: https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-vlan.yang#L125

So this PR LGTM

isabelmsft (approved):

dgsudharsan:
@yxieca Can you please cherry-pick this fix to 202311?

PR #3170, opened 2024-02-19
Optimize lag_keepalive by crafting the LACPDU packet ourselves

saiarcot895:
#### What I did

Instead of waiting for a LACPDU packet to be sent and capturing that (which involves waiting roughly 30 seconds), get the necessary information from teamd and craft it ourselves. This means that the 60-second wait in making sure that a LACPDU packet is captured and the keepalive script is ready can be largely eliminated (this has been reduced to 10 seconds to make sure the script has a chance to craft the packets and send some LACPDUs).

Fixes sonic-net/sonic-buildimage#17416.

#### How to verify it

Verified on KVM that the lag_keepalive script started correctly. Also verified that during a warm-reboot, the script correctly starts and send LACPDU packets every second (the delay between starting the script and the message where it said it is ready was 2 seconds).


scripts/fast-reboot
```diff
@@ -817,8 +817,8 @@ trap '' EXIT HUP INT QUIT TERM KILL ABRT ALRM
 # this is a non-blocking call, and the process will die in 300s
 debug "Starting lag_keepalive to send LACPDUs ..."
 timeout 300 python ${LAG_KEEPALIVE_SCRIPT} &
-# give the lag_keepalive script a chance to get ready (30s) and collect one lacpdu before going down (30s)
-sleep 60
+# give the lag_keepalive script a chance to send some LACPDUs
+sleep 10
```
vaibhavhd, line 821: This approach still uses a wait approach, and the assumption is that the LACPDUs will be sent.

In mostly all platforms, sending these LACPDUs has become a prerequisite. So, may be we should do a quick capture here to ensure that LACPDUs are being sent out via (all?) LAGs. If not, abort warm-reboot.

scripts/lag_keepalive.py
```diff
@@ -22,47 +29,64 @@ def log_error(msg):
...
+    return proc.communicate()[0], proc.returncode
+
+
+def get_port_channel_config(portChannelName):
+    (processStdout, _) = getCmdOutput(["teamdctl", portChannelName, "state", "dump"])
+    return json.loads(processStdout)
+
+
+def craft_lacp_packet(portChannelConfig, portName):
+    portConfig = portChannelConfig["ports"][portName]
```
vaibhavhd, line 43: What is a possibility of any of the following key accesses from the dict will fail with KeyError?

vaibhavhd: Please add key presence check wherever needed, and default values (??) where keys are missing.

scripts/lag_keepalive.py
```diff
@@ -22,47 +29,64 @@ def log_error(msg):
...
-        count=1, timeout=30)
-    lag_member_to_packet[lag_member] = sniffed_packet
+def getCmdOutput(cmd):
+    proc = subprocess.Popen(cmd, stdout=subprocess.PIPE)
+    return proc.communicate()[0], proc.returncode
+
+
+def get_port_channel_config(portChannelName):
+    (processStdout, _) = getCmdOutput(["teamdctl", portChannelName, "state", "dump"])
+    return json.loads(processStdout)
```
vaibhavhd, line 39: Exception handling?

scripts/lag_keepalive.py
saiarcot895: There's a try-except in `main()` that catches everything and retries immediately.

scripts/lag_keepalive.py
saiarcot895: Handled via `main()` try-catch.

scripts/fast-reboot
saiarcot895: So doing a capture here might make things a bit messy. Instead, I added logic in the Python script to fork into the background after collecting information about the LAGs. That way, if there's an issue at this stage, then it'll get caught.

vaibhavhd (approved):

PR #3182, opened 2024-02-28
[ipintutil]Handle exception in show ip interfaces command

dgsudharsan:
#### What I did
Handle exception in show ip interfaces command when executed during config reload. Sometimes during config reload the interfaces are removed and if show ip interfaces was executed during this time we may get the below traceback. The interface would exist when the call multi_asic_get_ip_intf_from_ns was made but would have been removed in the subsequent for loop which tries to get ip interface data for each interface

```
show ip interfaces

  File "/usr/local/bin/ipintutil", line 276, in

     main()
  File "/usr/local/bin/ipintutil", line 269, in main
    ip_intfs = get_ip_intfs(af, namespace, display)
  File "/usr/local/bin/ipintutil", line 232, in get_ip_intfs
    ip_intfs_in_ns = get_ip_intfs_in_namespace(af, namespace, display)
  File "/usr/local/bin/ipintutil", line 153, in get_ip_intfs_in_namespace
    ipaddresses = multi_asic_util.multi_asic_get_ip_intf_addr_from_ns(namespace, iface)
  File "/usr/local/lib/python3.9/dist-packages/utilities_common/multi_asic.py", line 186, in multi_asic_get_ip_intf_addr_from_ns

     ipaddresses = netifaces.ifaddresses(iface)
ValueError: You must specify a valid interface name.

```

#### How I did it
Adding try exception block so that if an interface is not present, it would be skipped.

#### How to verify it
Running show ip interface command and performing config reload in parallel

### DIFF
--- a/scripts/ipintutil
+++ b/scripts/ipintutil
@@ -150,7 +150,10 @@ def get_ip_intfs_in_namespace(af, namespace, display):
         ip_intf_attr = []
         if namespace != constants.DEFAULT_NAMESPACE and skip_ip_intf_display(iface, display):
             continue
-        ipaddresses = multi_asic_util.multi_asic_get_ip_intf_addr_from_ns(namespace, iface)
+        try:
+            ipaddresses = multi_asic_util.multi_asic_get_ip_intf_addr_from_ns(namespace, iface)
+        except ValueError:
+            continue
         if af in ipaddresses:
             ifaddresses = []
             bgp_neighs = {}
--- a/scripts/ipintutil
+++ b/scripts/ipintutil
@@ -29,6 +29,7 @@ try:
             mock_tables.dbconnector.load_namespace_config()
         else:
             import mock_tables.mock_single_asic
+            mock_tables.mock_single_asic.add_unknown_intf=True
 except KeyError:
     pass
 
--- a/tests/mock_tables/mock_single_asic.py
+++ b/tests/mock_tables/mock_single_asic.py
@@ -4,6 +4,8 @@ from unittest import mock
 from sonic_py_common import multi_asic
 from utilities_common import multi_asic as multi_asic_util
 
+add_unknown_intf=False
+
 mock_intf_table = {
     '': {
         'eth0': {
@@ -60,6 +62,8 @@ def mock_single_asic_get_ip_intf_from_ns(namespace):
     interfaces = []
     try:
         interfaces = list(mock_intf_table[namespace].keys())
+        if add_unknown_intf:
+            interfaces.append("unknownintf")
     except KeyError:
         pass
     return interfaces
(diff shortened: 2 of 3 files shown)


dgsudharsan:
@qiluo-msft  The code change is simple and just adding a try exception block. This scenario is hard to replicate through UT and I request exception for coverage for this PR

oleksandrivantsiv (approved):

scripts/ipintutil
```diff
@@ -150,7 +150,10 @@ def get_ip_intfs_in_namespace(af, namespace, display):
         ip_intf_attr = []
         if namespace != constants.DEFAULT_NAMESPACE and skip_ip_intf_display(iface, display):
             continue
-        ipaddresses = multi_asic_util.multi_asic_get_ip_intf_addr_from_ns(namespace, iface)
+        try:
+            ipaddresses = multi_asic_util.multi_asic_get_ip_intf_addr_from_ns(namespace, iface)
+        except ValueError:
+            continue
```
qiluo-msft, line 156: Is it better to print error message and exit? If you continue, `ipaddresses` is not initialized and used.

qiluo-msft:
> @qiluo-msft The code change is simple and just adding a try exception block. This scenario is hard to replicate through UT and I request exception for coverage for this PR

Recommend still add UT to cover the lines. It is possible to mock some functions to reach the code branch.

scripts/ipintutil
dgsudharsan: I don't think its correct. This will avoid other interfaces detail to not be shown. This is a very rare scenario during config reload.
Here is the scenario - While getting the interfaces there are 4 ports (docker0, eth0, lo0 and Loopback0). But while it enters the for loop in the parallel config reload flow Loopback0 gets deleted. However other interfaces exist and those information should be displayed. We can skip Loopback0 as we get valueerror meaning the interface does not exist now

dgsudharsan:
> > @qiluo-msft The code change is simple and just adding a try exception block. This scenario is hard to replicate through UT and I request exception for coverage for this PR
> 
> Recommend still add UT to cover the lines. It is possible to mock some functions to reach the code branch.

I am trying to add. Will update if I can add

dgsudharsan:
> > > @qiluo-msft The code change is simple and just adding a try exception block. This scenario is hard to replicate through UT and I request exception for coverage for this PR
> > 
> > 
> > Recommend still add UT to cover the lines. It is possible to mock some functions to reach the code branch.
> 
> I am trying to add. Will update if I can add

Added UT and coverage is passing now.

scripts/ipintutil
qiluo-msft: Sorry, there is no uninitialized variable in use.

qiluo-msft (approved):

PR #3188, opened 2024-03-04
[config] Add YANG alerting for override

wen587:
#### What I did
Add alerting for YANG validation when load_minigraph during override. This is to alert early if golden config is invalid which will breaks GCU feature. 
#### How I did it
Add alerting when `is_yang_config_validation_enabled` is not set during load_minigraph with override
#### How to verify it
Unit test


config/main.py
```diff
@@ -1968,6 +1968,17 @@ def override_config_table(db, input_config_db, dry_run):
             validate_config_by_cm(cm, ns_config_input, "config_input")
             # Validate updated whole config
             validate_config_by_cm(cm, updated_config, "updated_config")
+        else:
+            cm = None
+            try:
+                cm = ConfigMgmt(configdb=config_db)
```
qiluo-msft, line 1974: Please add comment what do you verify?

config/main.py
```diff
@@ -1968,6 +1968,17 @@ def override_config_table(db, input_config_db, dry_run):
...
+                cm.validateConfigData()
+            except Exception as ex:
+                log.log_warning("Failed to validate running config. Alerting: {}".format(ex))
+
+            # Validate updated whole config
+            if cm:
+                validate_config_by_cm_alerting(cm, updated_config, "updated_config")
```
qiluo-msft, line 1981: also what do you verified here?

wen587:
Hi @vaibhavhd , this extra YANG validation would be happen during load_minigraph. Just want to make sure this doesn't impact warm reboot.

wangxin:
Discussed with @wen587 offline about using log_err or log_warning. Considering check failure does not really mean something is broken and not working, use log_warning may be better.

qiluo-msft (approved):

StormLiangMS:
HI @wen587 this one more like an enhancement, is that must to have for 202305?

wen587:
Hi @StormLiangMS , the PR is conflict fixed in https://github.com/sonic-net/sonic-utilities/pull/3195.
I suggest to add to 202305. This will help us to find YANG issue in PROD by syslog without using GCU features.

PR #3196, opened 2024-03-07
[Techsupport]Handle SAI kv pair if present in sai common profile

dgsudharsan:
#### What I did
Handle the SAI key value pair if it is present in sai common profile rather than specific profile for mellanox platforms. The concept of common sai profile is introduced in https://github.com/sonic-net/sonic-buildimage/pull/18074 . After this the techsupport started to fail because of the absence of SAI_DUMP_STORE_PATH

#### How I did it
Check if the variable is not present in platform specific file and then read the common file. If the common file is not accessible due to syncd being down, fallback to default path which is hardcoded.

#### How to verify it
Running techsupport and ensuring it exits with code 0


vivekrnv (approved):

ganglyu (approved):

ayurkiv-nvda:
Need to cherry-pick it for 202305

PR #3203, opened 2024-03-11
[graceful reboot] Add the pre_reboot_hook script execution, add the watchdog arm before the reboot 

vadymhlushko-mlnx:
**DEPENDS ON: [[graceful reboot] Rename the platform_reboot to the pre_reboot_hook, remove the sysfs power cycle #18324](https://github.com/sonic-net/sonic-buildimage/pull/18324)**

#### What I did
Add the `pre_reboot_hook` script execution, add the watchdog arm before the reboot, in order to support the `graceful reboot` instead of the sysfs power cycle, to avoid filesystem corruption.

#### How I did it
Modify the `scripts/reboot`

#### How to verify it
1. Start watching logs by using `show log -f` and `journalctl -p debug -f`
2. Execute the `reboot` command from the switch CLI
3. Check in logs that all systemd services terminated correctly

### DIFF
--- a/scripts/reboot
+++ b/scripts/reboot
@@ -1,4 +1,10 @@
 #!/bin/bash
+
+declare -r EXIT_SUCCESS=0
+declare -r EXIT_ERROR=1
+declare -r WATCHDOG_UTIL="/usr/local/bin/watchdogutil"
+declare -r PRE_REBOOT_HOOK="pre_reboot_hook"
+
 DEVPATH="/usr/share/sonic/device"
 PLAT_REBOOT="platform_reboot"
 PLATFORM_UPDATE_REBOOT_CAUSE="platform_update_reboot_cause"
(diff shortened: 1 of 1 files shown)


scripts/reboot
```diff
@@ -242,6 +253,23 @@ if [ -x ${DEVPATH}/${PLATFORM}/${PLATFORM_UPDATE_REBOOT_CAUSE} ]; then
     ${DEVPATH}/${PLATFORM}/${PLATFORM_UPDATE_REBOOT_CAUSE}
 fi
 
+if [ -x ${DEVPATH}/${PLATFORM}/${PRE_REBOOT_HOOK} ]; then
```
prgeor, line 256: @vadymhlushko-mlnx why do we need pre_reboot_hook?

vadymhlushko-mlnx: @prgeor this PR comes in bundle [[graceful reboot] Rename the platform_reboot to the pre_reboot_hook, remove the sysfs power cycle #18324](https://github.com/sonic-net/sonic-buildimage/pull/18324).

For MLNX rename the `platform_reboot` script to `pre_reboot_hook` because we remove the `sysfs power cycle` from `platform

prgeor (approved):

PR #3075, opened 2023-12-10
route_check: Skip route checks if bgp feature is not enabled

anamehra:
If bgp is not enabled, get_frr_routes() gets empty list and route check fails and throws a traceback. Adding check to to skip route checks  bgp feature is disabled. On the Chassis supervisor, bgp may be disabled. 

Signed-off-by: Anand Mehra anamehra@cisco.com

#### What I did
Fixes https://github.com/sonic-net/sonic-buildimage/issues/17400

Fixed the route_check.py traceback error on Chassis Supervisor.

#### How I did it
Added feature bgp check before performing frr route check. This prevents failure when bgp is not enabled on Supervisor.

#### How to verify it
run route_check.py on Supervisor. No traceback should be observed.

#### Previous command output (if the output of a command-line utility has changed)
```
root@sonic:/home/cisco# route_check.py 
Traceback (most recent call last):
  File "/usr/local/bin/route_check.py", line 822, in <module>
    sys.exit(main()[0])
  File "/usr/local/bin/route_check.py", line 809, in main
    ret, res= check_routes()
  File "/usr/local/bin/route_check.py", line 757, in check_routes
    rt_frr_miss = check_frr_pending_routes()
  File "/usr/local/bin/route_check.py", line 547, in check_frr_pending_routes
    frr_routes = get_frr_routes()
  File "/usr/local/bin/route_check.py", line 343, in get_frr_routes
    output = subprocess.check_output('show ip route json', shell=True)
  File "/usr/lib/python3.9/subprocess.py", line 424, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.9/subprocess.py", line 528, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command 'show ip route json' returned non-zero exit status 1.
```
#### New command output (if the output of a command-line utility has changed)
```
root@sonic:/home/cisco# route_check.py 
root@sonic:/home/cisco# 
```

#### Which release branch to backport (provide reason below if selected)

- [x] 202305

PR #3237, opened 2024-03-22
Skip check_frr_pending_routes in Supervisor card in chassis

saksarav-nokia:
#### What I did
 Skip the frr_pending_routes check in Supervisor to avoid the python back trace since there is no ip or ipv6 route and json.loads fails 
#### How I did it
Added a check not to call frr_pending_routes 
#### How to verify it
Verified the route check and "sudo monit status" in SUP and LC

### DIFF
--- a/scripts/route_check.py
+++ b/scripts/route_check.py
@@ -50,6 +50,7 @@ import subprocess
 from ipaddress import ip_network
 from swsscommon import swsscommon
 from utilities_common import chassis
+from sonic_py_common import device_info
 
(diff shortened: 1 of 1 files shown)


saksarav-nokia:
@mlok-nokia for viz

saksarav-nokia:
As per Deepak, https://github.com/sonic-net/sonic-utilities/pull/3075 will address this issue. We will test with #3075 and close #3237.

saksarav-nokia:
This PR is not required and closing it

PR #3239, opened 2024-03-23
Add fabric monitoring commands.

jfeng-arista:
#### What I did
Add a force option for config fabric port unisolate command. 

Add a show command to display if a up link is get isolated or not . Example output is :

```
# show fabric isolation
asic0
  Local Link    Auto Isolated    Manual Isolated    Isolated
------------  ---------------  -----------------  ----------
           0                0                  0           0
           1                0                  0           0
           2                0                  0           0
....
```
Add test for the new commands. 
The test is failed now as it needs this https://github.com/sonic-net/sonic-swss/pull/3089 merged in first.

### DIFF
--- a/config/fabric.py
+++ b/config/fabric.py
@@ -2,7 +2,10 @@ import click
 import utilities_common.cli as clicommon
 import utilities_common.multi_asic as multi_asic_util
 from sonic_py_common import multi_asic
-from swsscommon.swsscommon import SonicV2Connector, ConfigDBConnector
+from swsscommon.swsscommon import SonicV2Connector, ConfigDBConnector, APP_FABRIC_PORT_TABLE_NAME
+
+FABRIC_PORT_STATUS_TABLE_PREFIX = APP_FABRIC_PORT_TABLE_NAME+"|"
+
 
 #
 # 'config fabric ...'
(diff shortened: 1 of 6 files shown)


jfeng-arista:
the test failure needs this https://github.com/sonic-net/sonic-swss/pull/3089 merged in first. 
I could comment out failure test case for now, but that will cause the coverage check fail, let me know if we should leave this as is for now. thank you

jfeng-arista:
@vmittal-msft @arlakshm

config/fabric.py
```diff
@@ -66,19 +69,13 @@ def isolate(portid, namespace):
 #
 @port.command()
 @click.argument('portid', metavar='<portid>', required=True)
+@click.option('-f', '--force', is_flag=True, help='Force to unisolate a link even if it is auto isolated.')
```
arlakshm, line 72: add default value as false?

config/fabric.py
```diff
@@ -102,6 +130,15 @@ def unisolate(portid, namespace):
...
     config_db.mod_entry("FABRIC_PORT", portName, {'isolateStatus': False})
 
+    if force:
+        forceShutCnt = int( portConfigData['forceUnisolateStatus'] )
+        forceShutCnt += 1
+        config_db.mod_entry( "FABRIC_PORT", portName,
+                             {'forceUnisolateStatus': forceShutCnt})
+
+        print("Force unisolate the link.")
+        print("It will clear all fabric link monitoring status for this link!")
```
arlakshm, line 140: please use click.echo iso print

arlakshm (commented):
as comments

config/fabric.py
jfeng-arista: > add default value as false?

when I run the test on or command on a system , it works as it has default value of false , so I did not have it before. 
I can add it

config/fabric.py
jfeng-arista: > please use click.echo iso print

done. thank you

arlakshm (approved):

PR #3108, opened 2024-01-01
Fix for switchport mode PR

sabakram:
### What I did

This PR is Fixture for PR#2419. The original code PR was reverted due to nonavailability of YANG Model. The code provided by this PR is exactly same as reviewed & approved by reveiwers
### How I did it

Code Modification added in original PR is added again with new PR
### How to verify it

New commands have been added in Command-Reference.md All the syntax and examples have been added there and they can be verified by running the specific command

PR #3243, opened 2024-03-26
Fix for Db migrator & Vlan Changes

sabakram:
### What I did
This PR is Fixture for https://github.com/sonic-net/sonic-utilities/pull/3108.  We have removed dbmigrator changes to avoid backward compatibility issues.

### How I did it
Removed Dbmigrator changes from dbmigrator.py and dbmigrator_test.py 

### How to verify it
Dbmigrator changes has been removed


sabakram:
@wen587 @qiluo-msft 
Hi, we have removed DB migrator changes and catered to backward compatibility concerns in vlan.py. Can you please review this PR and suggest any other changes required to avoid backward compatibility issues. This is in reference to PR #3108 .

config/vlan.py
```diff
@@ -316,9 +316,7 @@ def add_vlan_member(db, vid, port, untagged, multiple, except_flag):
             elif not is_port:
                 port_data = config_db.get_entry('PORTCHANNEL',port)
 
-            if "mode" not in port_data: 
-                ctx.fail("{} is in routed mode!\nUse switchport mode command to change port mode".format(port))
-            else:
+            if "mode" in port_data:
                 existing_mode = port_data["mode"]
             
             if existing_mode == "routed":
```
wen587, line 322: If "mode" not in `port_data`, the `existing_mode` should be undefined. 
Why is it not caught by build or test?

wen587:
Hi @sabakram , I tried submodule without your PR. https://github.com/sonic-net/sonic-buildimage/pull/18480 It pass sonic-mgmt test. I suggest to revert it first to unblock submodule update. 
You can have more time to improve the unit test to cover more use cases, especially the one used in [test_vlan_interface.py](https://github.com/sonic-net/sonic-mgmt/blob/master/tests/generic_config_updater/test_vlan_interface.py)

sabakram:
Hi @wen587 , ok we can revert this PR to avoid blocking for submodule head. but this revert is already done once earler. 
and this new PR was created totally after suggestions.
This case is being pending since long and every time all changes are done as per suggestions from start but still we have open PR for same case which is also causing blocking issues and compatibility issues. It would be great help if you reviewed PR and provide suggested changes again at once so we can re do all code to resolve this case.

wen587:
Hi @sabakram , sure I'm willing to help with the review

PR #3244, opened 2024-03-26
[dualtor_neighbor_check] Fix the script not exists issue

lolyu:
#### What I did
Fix https://github.com/sonic-net/sonic-buildimage/issues/17200

* Microsoft ADO (number only): 27317600

Upon advanced-reboot, the Lua script is flushed, so the EVALSHA command
might suffer from the script not-existed issue. Let's check if the
script exists before calling EVALSHA.

Signed-off-by: Longxiang Lyu <lolv@microsoft.com>

#### How I did it
As the above.

#### How to verify it
UT and verify on dualtor testbed.


yxieca (approved):

lolyu:
Hi @StormLiangMS, could you please help cherry-pick?

StormLiangMS:
hi @lolyu  do we have test with 202305 with this one?

PR #3257, opened 2024-04-03
Fix db_migrate.py show error and back trace while loading configuration on Linecard

liuh-80:
Fix db_migrate.py show error and back trace while loading configuration on Linecard

#### Why I did it
    Fix [issue @](https://github.com/sonic-net/sonic-buildimage/issues/18389)

#### How I did it
    Revert code change by https://github.com/sonic-net/sonic-utilities/pull/3100
    Check DB config initialize state and ignore when initialized.

#### How to verify it
    Pass all UT.
    Manually test.

##### Work item tracking
- Microsoft ADO **(number only)**: 27384235

    N/A

    Fix db_migrate.py show error and back trace while loading configuration on Linecard


scripts/db_migrator.py
```diff
@@ -1277,7 +1276,14 @@ def main():
         socket_path = args.socket
         namespace = args.namespace
 
-        load_db_config()
+        # Can't load global config base on the resoult of is_multi_asic(), because on multi-asic device, when db_migrate.py
```
liuh-80, line 1279: Code change here verified by this sonic-buildimage PR: https://github.com/sonic-net/sonic-buildimage/pull/18502, all test case passed.

ganglyu (dismissed):

scripts/db_migrator.py
qiluo-msft, line 1279: >resoult [](http://example.com/codeflow?start=47&length=7)

typo

scripts/db_migrator.py
```diff
@@ -1277,7 +1276,14 @@ def main():
...
+        # run on the local database, ASIC instance will have not created the /var/run/redis0/sonic-db/database-config.json
+        if args.namespace is not None:
+            if not SonicDBConfig.isGlobalInit():
+                SonicDBConfig.load_sonic_global_db_config(namespace=args.namespace)
+        else:
+            if not SonicDBConfig.isInit():
+                SonicDBConfig.load_sonic_db_config()
```
judyjoseph, line 1287: I feel it is more like a timing issue as you pointed out, Can we call db_mibrator, after /var/run/redis0/sonic-db/database-config.json is created ? 

Here, https://github.com/sonic-net/sonic-swss-common/blob/7db6ccf5fb427d4832d6c54ab1f485ddd87b6697/common/dbconnector.h#L107 -- it doesn't make a difference if we pass/don't pass the namespace parameter to load_sonic_global_db_config()

scripts/db_migrator.py
liuh-80: Fixed.

scripts/db_migrator.py
```diff
@@ -1277,7 +1276,14 @@ def main():
         socket_path = args.socket
         namespace = args.namespace
 
-        load_db_config()
```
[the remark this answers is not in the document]
liuh-80: This is not a general issue, according to the code of this script and Judy's comments, this script seems designed to be run before ASIC instance database0 and database1 generate database-config.json.
For other code using load_db_config this issue should not happen.

scripts/db_migrator.py
liuh-80: I search sonic code but not found any code use the '-n' parameter of db_migrator.py to pass the namespace parameter.
I will add comments in the https://github.com/sonic-net/sonic-buildimage/issues/18389 to ask author check if this can fix by your suggestion.

For load_sonic_global_db_config(namespace=args.namespace), I think it's better keep the parameter to keep the compatibility, incase we change the code of load_sonic_global_db_config in future.

mlok-nokia: @liuh-80 @judyjoseph  all instance database@.service have dependency "Requires=database.service
After=database.service".  db_mihgrator script in local host database will not be able to wait for all other instance ready  unless we change the dependency.

qiluo-msft (approved):
Please check with other reviewers.

ganglyu (approved):

PR #3259, opened 2024-04-04
Improve Semgrep CI

maipbui:
#### What I did
Semgrep's default ruleset (p/default) somehow lost some important rules
#### How I did it
Keep use p/default and add another rule
#### How to verify it
Added test code to this PR and Semgrep CI failed
```
diff --git a/generic_config_updater/change_applier.py b/generic_config_updater/change_applier.py
index d0818172..6e1473a4 100644
--- a/generic_config_updater/change_applier.py
+++ b/generic_config_updater/change_applier.py
@@ -169,6 +169,7 @@ class ChangeApplier:
     def _get_running_config(self):
         (_, fname) = tempfile.mkstemp(suffix="_changeApplier")
         os.system("sonic-cfggen -d --print-data > {}".format(fname))
+        os.system("sonic-cfggen -d --print-data > {}".format(fname))
         run_data = {}
         with open(fname, "r") as s:
             run_data = json.load(s)
```
Failed result: https://github.com/sonic-net/sonic-utilities/actions/runs/8559846841/job/23457508614?pr=3259

### DIFF
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -18,4 +18,4 @@ jobs:
       - uses: actions/checkout@v3
       - run: semgrep ci
         env:
-           SEMGREP_RULES: p/default
+           SEMGREP_RULES: r/all
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -18,4 +18,4 @@ jobs:
       - uses: actions/checkout@v3
       - run: semgrep ci
         env:
-           SEMGREP_RULES: r/all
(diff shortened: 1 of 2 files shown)


qiluo-msft (dismissed):

.github/workflows/semgrep.yml
```diff
@@ -18,4 +18,4 @@ jobs:
       - uses: actions/checkout@v3
       - run: semgrep ci
         env:
-           SEMGREP_RULES: p/default
+           SEMGREP_RULES: r/all
```
qiluo-msft, line 21: like https://registry.semgrep.dev/ruleset? Is it possible to add small incremental set?
Enabling all rules will be over verbose.

qiluo-msft (approved):

PR #3283, opened 2024-04-23
Update for the procedures for insertion/hot swap of Switch Fabric Module(SFM) by using "config chassis modules shutdown/startup" commands

JunhongMao:
#### Why I did it
For the Nokia SONiC chassis procedures for insertion/hot swap of Switch Fabric Module(SFM), 
the previous solution was using the below commands.
```
sudo nokia_cmd set shutdown-sfm <SFM-Num/Physical-Slot>
```
The below 4 PRs intend to add the below commands for the equivalent operations.
https://github.com/sonic-net/sonic-platform-daemons/pull/491
https://github.com/sonic-net/sonic-utilities/pull/3283
https://github.com/nokia/sonic-platform/pull/6
https://github.com/sonic-net/sonic-buildimage/pull/18938
```
sudo config chassis modules shutdown/startup <module name>
```
The HLD for Shutdown and Startup of the Fabric Module is below:
https://github.com/sonic-net/SONiC/pull/1694

The below PR was replaced.
https://github.com/sonic-net/sonic-buildimage/pull/18578

##### Work item tracking
- Microsoft ADO **(number only)**:

#### How I did it

1. When the cli command "sudo config chassis modules startup/shutdown" runs, it directly calls config/fabric_module_set_admin_status.py to do the related operations.
#### How to verify it

```
The below test was carried out on FABRIC-CARD3 module on the supervisor card.
1. Shutdown
sudo config chassis modules shutdown FABRIC-CARD3

2. Check the status to see if the FABRIC-CARD3 was down.
$ show chassis modules status
        Name             Description    Physical-Slot    Oper-Status    Admin-Status       Serial
------------  ----------------------  ---------------  -------------  --------------  -----------
...
FABRIC-CARD3             Unavailable                4          Empty            down          N/A

sudo tail -f /var/log/syslog | grep "pmon#chassisd:"
May  1 00:07:54.192037 ixre-cpm-chassis15 WARNING pmon#chassisd: Module FABRIC-CARD3 went off-line!
 ...

 
3. Start up the module
sudo config chassis modules startup FABRIC-CARD3


4. Check the status
$ show chassis modules status
        Name             Description    Physical-Slot    Oper-Status    Admin-Status       Serial
------------  ----------------------  ---------------  -------------  --------------  -----------
...
FABRIC-CARD3                    SFM4                4         Online              up  01214400362

sudo tail -f /var/log/syslog | grep "pmon#chassisd:"
May  1 00:26:29.501687 ixre-cpm-chassis15 NOTICE pmon#chassisd: Module FABRIC-CARD3 recovered on-line!


5. To test if the operation is still valid when the system reboot. For example, first shut down, 
then after saving config and reboot, the module should keep shutdown status. 
$ sudo config save
Existing files will be overwritten, continue? [y/N]: y

Then check the status to see if the FABRIC-CARD3 was down.
$ show chassis modules status
        Name             Description    Physical-Slot    Oper-Status    Admin-Status       Serial
------------  ----------------------  ---------------  -------------  --------------  -----------
...
FABRIC-CARD3             Unavailable                4          Empty            down          N/A


```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205
- [x] 202211
- [x] 202305

- [x] 202205


config/fabric_module_set_admin_status.py
```diff
@@ -0,0 +1,89 @@
...
+        # wait for service is down
+        time.sleep(2)
+
+        logger.log_info("Power off {} module ...".format(module))
+        hw_slot = sfm_hw_slot_mapping[num]
+        nokia_common._power_onoff_SFM(hw_slot,False)
+        logger.log_info("Chassis module {} shutdown completed".format(module))
+    else:
+        logger.log_info("Starting up chassis module {}".format(module))
+        hw_slot = sfm_hw_slot_mapping[num]
```
arlakshm, line 68: For Linecards the `admin state` is updated in  config_db  in the cli handler, chassisd calls the `set_admin_state` platform API for the module. Can we follow the same apporach here for shutdown the fabric card. The swss shutdown and startup can remain here.

arlakshm (commented):
please add UTs

config/fabric_module_set_admin_status.py
```diff
@@ -0,0 +1,89 @@
...
+sfm_hw_slot_mapping = {
+    0: 15,
+    1: 16,
+    2: 17,
+    3: 18,
+    4: 19,
+    5: 20,
+    6: 21,
+    7: 22
+}
```
mlok-nokia, line 22: There should not be any nokia specifiec code. Need to be moved to module.set_admin_state()

config/fabric_module_set_admin_status.py
```diff
@@ -0,0 +1,75 @@
...
+#
+# Copyright (c) 2024, Nokia
+# All rights reserved.
+
+import re
+import subprocess
+import time
+from swsscommon.swsscommon import SonicV2Connector
+from sonic_py_common.logger import Logger
+from platform_ndk import nokia_common
```
wumiaont, line 11: Should not have any vendor specific import here. sonic-utilities is sonic generic.

config/fabric_module_set_admin_status.py
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env python3
+#
+# Copyright (c) 2024, Nokia
```
wumiaont, line 3: This is sonic generic code. Should not be Nokia copyright.

config/fabric_module_set_admin_status.py
```diff
@@ -0,0 +1,75 @@
...
+                asic_list.append(asic_id)
+
+        logger.log_info("Shutting down chassis module {}".format(module))
+
+        for asic in asic_list:
+            logger.log_info("Stopping swss@{} and syncd@{} ...".format(asic, asic))
+            process = subprocess.Popen(['sudo', 'systemctl', 'stop', 'swss@{}.service'.format(asic)],
+                                        stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+            stdout, stderr = process.communicate()
+            outstr = stdout.decode('ascii')
```
wumiaont, line 47: Why you have outstr defined here and not used? Same 2 other places below.

config/chassis_modules.py
```diff
@@ -17,6 +18,28 @@ def modules():
     """Configure chassis modules"""
     pass
 
+def get_config_module_state(db, chassis_module_name):
+    config_db = db.cfgdb
+    fvs = config_db.get_entry('CHASSIS_MODULE', chassis_module_name)
+    if not fvs:
+        return None
```
mlok-nokia, line 25: Based on the design, this entry not found,  return "up"

config/chassis_modules.py
```diff
@@ -33,8 +56,14 @@ def shutdown_chassis_module(db, chassis_module_name):
        not chassis_module_name.startswith("FABRIC-CARD"):
         ctx.fail("'module_name' has to begin with 'SUPERVISOR', 'LINE-CARD' or 'FABRIC-CARD'")
 
+    #To avoid duplicate operation
+    if get_config_module_state(db, chassis_module_name) == 'down':
+        click.echo("Duplicate operation for " + chassis_module_name)
```
mlok-nokia, line 61: Should log "Module is already in down state". Not "duplication operation"

config/chassis_modules.py
```diff
@@ -45,5 +74,13 @@ def shutdown_chassis_module(db, chassis_module_name):
 def startup_chassis_module(db, chassis_module_name):
     """Chassis-module startup of module"""
     config_db = db.cfgdb
+    ctx = click.get_current_context()
+    
+    #To avoid duplicate operation
+    if get_config_module_state(db, chassis_module_name) is None:
+        click.echo("Duplicate operation for " + chassis_module_name)
+        return
```
mlok-nokia, line 82: It should not be "none".  Also, should log "Module {} is ready set to up state".

config/fabric_module_set_admin_status.py
```diff
@@ -0,0 +1,62 @@
...
+import subprocess
+import time
+from swsscommon.swsscommon import SonicV2Connector
+from sonic_py_common.logger import Logger
+import utilities_common.cli as clicommon
+
+# Name: fabric_module_set_admin_status.py, version: 1.0
+# Syntax: fabric_module_set_admin_status <module_name> <up/down>
+def fabric_module_set_admin_status(module, state):
+    logger = Logger("fabric_module_set_admin_status.py")
```
mlok-nokia, line 13: Remove this line.

config/fabric_module_set_admin_status.py
```diff
@@ -0,0 +1,62 @@
...
+    logger.set_min_log_priority_info()
```
mlok-nokia, line 14: Do we need this?

config/fabric_module_set_admin_status.py
```diff
@@ -0,0 +1,62 @@
...
+
+    if not module.startswith("FABRIC-CARD"):
+        logger.log_warning("Failed to set {} state. Admin state can only be set on Fabric module.".format(module))
```
mlok-nokia, line 17: This is wrong.  This is only applicable to Nokia platform.  May not for other vendor

config/fabric_module_set_admin_status.py
```diff
@@ -0,0 +1,62 @@
...
+        return
+
+    if (state != "up" and state != "down"):
+        logger.log_warning("Failed to set {}. Admin state can only be set to up or down.".format(state))
+        return
```
mlok-nokia, line 22: Do we need this?

config/fabric_module_set_admin_status.py
```diff
@@ -0,0 +1,62 @@
...
+
+    num = int(re.search(r"(\d+)$", module).group())
+    chassisdb = SonicV2Connector(host="127.0.0.1")
+    chassisdb.connect("CHASSIS_STATE_DB")
+
+    if state == "down":
+        asics_keys_list = chassisdb.keys("CHASSIS_STATE_DB", "CHASSIS_FABRIC_ASIC_TABLE*")
+        asic_list = []
+        for service in asics_keys_list:
```
mlok-nokia, line 31: This is not "service", this is asic_key. Should use asic_key

config/chassis_modules.py
```diff
@@ -45,5 +74,13 @@ def shutdown_chassis_module(db, chassis_module_name):
...
 
     config_db.set_entry('CHASSIS_MODULE', chassis_module_name, None)
+    if not get_config_module_state_timeout(ctx, db, chassis_module_name, None):
+        fabric_module_set_admin_status(chassis_module_name, 'up')
```
mlok-nokia, line 86: check if module is FABRIC, then do it

config/fabric_module_set_admin_status.py
```diff
@@ -0,0 +1,62 @@
...
+            logger.log_info("Stopping swss@{} and syncd@{} ...".format(asic, asic))
+            clicommon.run_command('sudo systemctl stop swss@{}.service'.format(asic))
+            # wait for service is down
+            time.sleep(5)
+            chassisdb.delete("CHASSIS_STATE_DB","CHASSIS_FABRIC_ASIC_TABLE|asic" + str(asic))
+            logger.log_info("Start swss@{} and syncd@{} ...".format(asic, asic))
+            clicommon.run_command('sudo systemctl start swss@{}.service'.format(asic))
+    else:
+        # wait SFM HW init done.
+        time.sleep(5)
```
mlok-nokia, line 50: No need to wait

config/fabric_module_set_admin_status.py
```diff
@@ -0,0 +1,62 @@
...
+            logger.log_warning("Failed to get {}'s asic list.".format(module))
+            return
+        logger.log_info("Shutting down chassis module {}".format(module))
+        for asic in asic_list:
+            logger.log_info("Stopping swss@{} and syncd@{} ...".format(asic, asic))
+            clicommon.run_command('sudo systemctl stop swss@{}.service'.format(asic))
+            # wait for service is down
+            time.sleep(5)
+            chassisdb.delete("CHASSIS_STATE_DB","CHASSIS_FABRIC_ASIC_TABLE|asic" + str(asic))
+            logger.log_info("Start swss@{} and syncd@{} ...".format(asic, asic))
```
wumiaont, line 46: It seems from the code here, the swss and syncd containers are bounced (bring down - bring back up). So in shutdown fabric module case, the corresponding swss and synced containers will keep there (with a bounce)? Is this by design?

config/chassis_modules.py
```diff
@@ -17,6 +18,71 @@ def modules():
...
+def get_config_module_state(db, chassis_module_name):
+    config_db = db.cfgdb
+    fvs = config_db.get_entry('CHASSIS_MODULE', chassis_module_name)
+    if not fvs:
+        return 'up'
+    else:
+        return fvs['admin_status']
+
+
+TIMEOUT_SECS = 10
```
mlok-nokia, line 31: Please move this define to beginning section.

mlok-nokia (approved):
They are good to me

config/chassis_modules.py
```diff
@@ -33,8 +105,17 @@ def shutdown_chassis_module(db, chassis_module_name):
...
+    # To avoid duplicate operation
+    if get_config_module_state(db, chassis_module_name) == 'down':
+        click.echo("Module {} is already in down state".format(chassis_module_name))
+        return
+
+    click.echo("Shutting down chassis module {}".format(chassis_module_name))
     fvs = {'admin_status': 'down'}
     config_db.set_entry('CHASSIS_MODULE', chassis_module_name, fvs)
+    if chassis_module_name.startswith("FABRIC-CARD"):
+        if not get_config_module_state_timeout(ctx, db, chassis_module_name, 'down'):
```
bmridul, line 117: why is this check required esp with a delay of upto 10 seconds?

config/chassis_modules.py
```diff
@@ -17,6 +21,74 @@ def modules():
...
+            clicommon.run_command('sudo systemctl stop swss@{}.service'.format(asic))
+
+        # wait for service is down
+        time.sleep(2)
+        click.echo("Delete related CAHSSIS_FABRIC_ASIC_TABLE entries")
+
+        for asic in asic_list:
+            chassisdb.delete("CHASSIS_STATE_DB", "CHASSIS_FABRIC_ASIC_TABLE|asic" + str(asic))
+
+        for asic in asic_list:
```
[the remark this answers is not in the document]
mlok-nokia: This entry needs to be deleted.  Otherwise,  if user set admin up,  this entry will trigger the syncd to initialize Fabric too early before the hardware is ready for it. This cause crash.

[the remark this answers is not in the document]
bmridul: My comment was - why we are restarting the swss services for the NPUs on the fabric card which is down. (line# 86)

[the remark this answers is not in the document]
wumiaont: Swss services are started for all fabric asics after bootup. They are staying in wait state till the fabric card is online.  Then that service changes to start state which will bring up swss container and finally changes back to wait state.
This is the services dump from one chassis where slot 3 and 6 have fabric cards.
root      764593       1  0 May08 ?        00:00:00 /bin/bash /usr/local/bin/swss.sh wait 7
root      765252       1  0 May08 ?        00:00:00 /bin/bash /usr/local/bin/swss.sh wait 0
root      765255       1  0 May08 ?        00:00:00 /bin/bash /usr/local/bin/swss.sh wait 1
root      765261       1  0 May08 ?        00:00:00 /bin/bash /usr/local/bin/swss.sh wait 12
root      765263       1  0 May08 ?        00:00:00 /bin/bash /usr/local/bin/swss.sh wait 13
root      765266       1  0 May08 ?        00:00:00 /bin/bash /usr/local/bin/swss.sh wait 14
root      765267       1  0 May08 ?        00:00:00 /bin/bash /usr/local/bin/swss.sh wait 15
root      765272       1  0 May08 ?        00:00:00 /bin/bash /usr/local/bin/swss.sh wait 2
root      765274       1  0 May08 ?        00:00:00 /bin/bash /usr/local/bin/swss.sh wait 3
root      765289       1  0 May08 ?        00:00:00 /bin/bash /usr/local/bin/swss.sh wait 6
root      765292       1  0 May08 ?        00:00:00 /bin/bash /usr/local/bin/swss.sh wait 8
root      765295       1  0 May08 ?        00:00:00 /bin/bash /usr/local/bin/swss.sh wait 9
root     1183800       1  0 May08 ?        00:00:00 /bin/bash /usr/local/bin/swss.sh wait 10
root     1183981       1  0 May08 ?        00:00:00 /bin/bash /usr/local/bin/swss.sh wait 5
root     1183982       1  0 May08 ?        00:00:00 /bin/bash /usr/local/bin/swss.sh wait 11
root     1183999       1  0 May08 ?        00:00:00 /bin/bash /usr/local/bin/swss.sh wait 4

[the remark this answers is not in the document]
judyjoseph: @mlok-nokia Another related question - I assume since the module state is set to down earlier, chassisd will not populate the "CHASSIS_FABRIC_ASIC_TABLE|asic" again for this asic's after the module update interval CHASSIS_INFO_UPDATE_PERIOD_SECS - is that right?

config/chassis_modules.py
```diff
@@ -17,6 +21,74 @@ def modules():
...
+    if len(asic_list) == 0:
+        return
+
+    if state == "down":
+        for asic in asic_list:
+            click.echo("Stop swss@{} and syncd@{} services".format(asic, asic))
+            clicommon.run_command('sudo systemctl stop swss@{}.service'.format(asic))
+
+        # wait for service is down
+        time.sleep(2)
```
anamehra, line 77: should check service state instead of sleep. A timeout may be set.

config/chassis_modules.py
```diff
@@ -17,6 +21,74 @@ def modules():
...
+        click.echo("Delete related CAHSSIS_FABRIC_ASIC_TABLE entries")
+
```
anamehra, line 79: shouldn't a platform hook be called at this stage to let platform power down the FC and remove and asic removal operations?

config/chassis_modules.py
```diff
@@ -17,6 +21,74 @@ def modules():
...
+        for asic in asic_list:
+            chassisdb.delete("CHASSIS_STATE_DB", "CHASSIS_FABRIC_ASIC_TABLE|asic" + str(asic))
+
+        for asic in asic_list:
+            click.echo("Start swss@{} and syncd@{} services".format(asic, asic))
+            clicommon.run_command('sudo systemctl reset-failed swss@{}.service'.format(asic))
```
anamehra, line 85: a stop was done earlier, wondering why a reset-failed would be required? Any issue observed?
And if required, won't it be required for all peer services?

config/chassis_modules.py
```diff
@@ -17,6 +21,74 @@ def modules():
...
+    chassisdb = db.db
+    chassisdb.connect("CHASSIS_STATE_DB")
+    asic_list = get_asic_list_from_db(chassisdb, chassis_module_name)
+
+    if len(asic_list) == 0:
+        return
+
+    if state == "down":
+        for asic in asic_list:
+            click.echo("Stop swss@{} and syncd@{} services".format(asic, asic))
```
anamehra, line 73: The swss service stop will stop more peer services. The log should reflect accordingly.

config/chassis_modules.py
```diff
@@ -17,6 +21,74 @@ def modules():
...
+
+        for asic in asic_list:
+            chassisdb.delete("CHASSIS_STATE_DB", "CHASSIS_FABRIC_ASIC_TABLE|asic" + str(asic))
+
+        for asic in asic_list:
+            click.echo("Start swss@{} and syncd@{} services".format(asic, asic))
+            clicommon.run_command('sudo systemctl reset-failed swss@{}.service'.format(asic))
+            clicommon.run_command('sudo systemctl start swss@{}.service'.format(asic))
+    else:
+        for asic in asic_list:
```
anamehra, line 88: should not explicitly assume that if not 'down' its up. better add a check.

config/chassis_modules.py
```diff
@@ -17,6 +21,74 @@ def modules():
...
+            click.echo("Start swss@{} and syncd@{} services".format(asic, asic))
+            clicommon.run_command('sudo systemctl start swss@{}.service'.format(asic))
```
anamehra, line 90: why a start is required here? swss service should already be running and waiting for asic presence notification.

config/chassis_modules.py
```diff
@@ -17,6 +21,74 @@ def modules():
...
+        for asic in asic_list:
+            chassisdb.delete("CHASSIS_STATE_DB", "CHASSIS_FABRIC_ASIC_TABLE|asic" + str(asic))
+
+        for asic in asic_list:
+            click.echo("Start swss@{} and syncd@{} services".format(asic, asic))
+            clicommon.run_command('sudo systemctl reset-failed swss@{}.service'.format(asic))
+            clicommon.run_command('sudo systemctl start swss@{}.service'.format(asic))
+    else:
+        for asic in asic_list:
+            click.echo("Start swss@{} and syncd@{} services".format(asic, asic))
```
anamehra, line 89: won't we require a platform hook to let platform power on the FC module?

tests/chassis_modules_test.py
```diff
@@ -202,6 +259,41 @@ def test_config_startup_module(self):
         result_out = " ".join((result_lines[header_lines]).split())
         assert result_out.strip('\n') == show_linecard0_startup_output.strip('\n')
 
+    def test_config_startup_module_fabric(self):
```
anamehra, line 262: I would suggest to add test cases for the following scenarios as well.
1. FC is up and doing startup
2. FC is down and doing startup
3. FC not present.

config/chassis_modules.py
mlok-nokia: By design, it should not be such case -- unless users just execute issue command "systemctl stop  swss@/syncd@" without bring down the hardware.  This handling just covers this case.  
As normal operation,  if user use CLI "config chassis module shutdown " a Fabric module. We shutdown service, delete the CHASSIS_FABRIC_ASIC_TABLE entry, we restart the service since we have the asic_list at this point.  This is the design which we discussed in the meeting that no need the Vendor provide the hook to do it.

config/chassis_modules.py
mlok-nokia: The design to restart the service here is to avoid to require a Vendor hook to start the correct asic service for a particular Fabric module.

config/chassis_modules.py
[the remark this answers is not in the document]
mlok-nokia: Correct.  It will be updated only if the MODULE admin_status is UP (or module admin_status default value which is UP )

config/chassis_modules.py
mlok-nokia: The power down/up module is existing design which has been handled by chassisd  -- ModuleConfigUpdater.    The design of these PRs is to avoid requiring any Vendor hook.

config/chassis_modules.py
mlok-nokia: By design, it should not be such case -- unless users just execute issue command "systemctl stop swss@/syncd@" without bring down the hardware. This handling just covers this case.
As normal operation, if user use CLI "config chassis module shutdown " a Fabric module. We shutdown service, delete the CHASSIS_FABRIC_ASIC_TABLE entry, we restart the service since we have the asic_list at this point. This is the design which we discussed in the meeting that no need the Vendor provide the hook to do it.

config/chassis_modules.py
anamehra: Please add that as comment in code.

config/chassis_modules.py
anamehra: Please add a comment in code.

config/chassis_modules.py
anamehra: Are you suggesting that the Vendor does not need to power down the FC? So the slot is always powered up. Once a new FC or old FC is inserted again, it will be powered up but admin down? Won't the asic detected already and inserted in db by pmon module?

mlok-nokia: > Are you suggesting that the Vendor does not need to power down the FC? So the slot is always powered up. Once a new FC or old FC is inserted again, it will be powered up but admin down? Won't the asic detected already and inserted in db by pmon module?

No, FC is power down by the  set_admin_state in module.py (Vendor specified code).  This is an exists functionality which has been done in the chassisd->ModuleConfigUpdater().  If Vendors support shutdown/startup FC, they need to modify their sonic-platform module.py -> set_admin_state(0 function to support it

config/chassis_modules.py
JunhongMao: Addressed.

config/chassis_modules.py
JunhongMao: Addressed.

config/chassis_modules.py
JunhongMao: The log already reflects the peer services, for example:
```
admin@ixre-cpm-chassis15:~$ sudo config chassis modules shutdown FABRIC-CARD3
Shutting down chassis module FABRIC-CARD3
Stop swss@6 and syncd@6 services
Stop swss@7 and syncd@7 services
Delete related CAHSSIS_FABRIC_ASIC_TABLE entries
Start swss@6 and syncd@6 services
Start swss@7 and syncd@7 services
```

JunhongMao:
@arlakshm @judyjoseph @anamehra, Please review again and approve the merge, thanks.

kenneth-arista:
@jfeng-arista for awareness

JunhongMao:
@arlakshm @judyjoseph @anamehra, please review again and approve the merge, thanks.

config/chassis_modules.py
JunhongMao: Thanks. Addressed.

judyjoseph (dismissed):
LGTM

JunhongMao:
@kenneth-arista, please review and approve it, thanks.

config/chassis_modules.py
```diff
@@ -17,6 +22,81 @@ def modules():
...
+
+        click.echo("Delete related CAHSSIS_FABRIC_ASIC_TABLE entries")
+
+        for asic in asic_list:
+            chassisdb.delete("CHASSIS_STATE_DB", "CHASSIS_FABRIC_ASIC_TABLE|asic" + str(asic))
+
+        # Start the services in case of the users just execute issue command "systemctl stop swss@/syncd@"
+        # without bring down the hardware
+        for asic in asic_list:
+            click.echo("Start swss@{} and peer services".format(asic))
```
kenneth-arista, line 91: Nitpick: the string does not coincide with the subsequent command; e.g. `sudo systemctl reset-failed`. Please revise accordingly

kenneth-arista (approved):
Looks good to me.

config/chassis_modules.py
JunhongMao: Addressed.

judyjoseph (approved):

anamehra (approved):

PR #3284, opened 2024-04-23
Migrate AAA table in db_migrator

liuh-80:
Migrate AAA table in db_migrator

#### Why I did it
    per-command AAA need enable in warm-upgrade case

#### How I did it
    Add db_migrator code to migrate AAA table

#### How to verify it
    Pass all test case.
    Add new test case.

    N/A

    Migrate AAA table in db_migrator


scripts/db_migrator.py
```diff
@@ -1200,6 +1200,55 @@ def set_version(self, version=None):
...
         self.configDB.set_entry(self.TABLE_NAME, self.TABLE_KEY, entry)
 
+    def enable_per_command_aaa(self):
+        aaa_table = self.configDB.get_table('AAA')
+        if not aaa_table:
+            log.log_info('AAA table does not exist ignore setup per-command accounting&authorization.')
+            return
+
+        authentication_config_old = self.configDB.get_entry('AAA', 'authentication')
+        # If device enabled TACACS authentication, then enable TACACS per-command accounting and authorization
```
qiluo-msft, line 1210: >enable [](http://example.com/codeflow?start=56&length=6)

The decision should also be driven by GOLDEN_CFG_FILE. GOLDEN_CFG_FILE could express the intention to enable AAA feature or disable, and db_migrator should respect the intention. ref: migrate_telemetry()

qiluo-msft: The old image's enablement does not matter. As long as the required passkey is there, and intention is to enable, then we should enable.

qiluo-msft, line 1210: >accounting [](http://example.com/codeflow?start=82&length=10)

Is it better to use separate PRs for accounting and authorization enablement?

liuh-80: Fixed, also discussion with Gang about how to handle version, because AAA table should migrate on all version >= 202205, so add migrate to common_migration_ops.

liuh-80: Fixed, this PR only contains migrate accounting part. will submit another PR later.

ganglyu (approved):

qiluo-msft (approved):

ganglyu (approved):

liuh-80:
Per-command authorization migrate code add in this PR: https://github.com/sonic-net/sonic-utilities/pull/3296

PR #3286, opened 2024-04-23
Add CLI for bmp configdb entity Enable/Disable

FengPan-Frank:
Work item tracking
Microsoft ADO (number only):27353337

#### What I did
Add CLI for bmp configdb entity Enable/Disable.

#### How I did it
Add python related code for Get/Set configdb entity, which is on https://github.com/sonic-net/sonic-buildimage/pull/18530

#### How to verify it
Unite test.


config/main.py
```diff
@@ -3916,6 +3916,172 @@ def del_user(db, user):
...
+@click.pass_context
+def enable_bgp_neighbor_table(ctx):
+    """enable bgp-neighbor-table sessions
+       In the case of Multi-Asic platform, corresponding database instance of neighbor will be operated.
+    """
+    log.log_info("'bmp enable bgp-neighbor-table' executing...")
+    config_db = ConfigDBConnector()
+    config_db.connect()
+    try:
+        bmp_table = config_db.get_table('BMP')
```
qiluo-msft, line 3949: >BMP [](http://example.com/codeflow?start=41&length=3)

This table is not inside your HLD. You mentioned `FEATURE|bmp` table there.

config/main.py
```diff
@@ -3916,6 +3916,172 @@ def del_user(db, user):
...
+        if not bmp_table:
+            bmp_table =  {'table': {'bgp_neighbor_table': 'true'}}
+        else:
+            bmp_table['table']['bgp_neighbor_table'] = 'true'
```
qiluo-msft, line 3953: >bgp_neighbor_table [](http://example.com/codeflow?start=32&length=18)

Is it possible that this field is missing, and this line will throw?

qiluo-msft, line 3953: >bmp_table['table']['bgp_neighbor_table'] = 'true' [](http://example.com/codeflow?start=12&length=49)

Make sure this line is covered in unit test and coverage result.

config/main.py
```diff
@@ -3916,6 +3916,172 @@ def del_user(db, user):
...
+
+        config_db.mod_entry('BMP', 'table', bmp_table['table'])
+    except ValueError as e:
+        ctx.fail("Invalid ConfigDB. Error: {}".format(e))
```
qiluo-msft, line 3957: >ctx.fail("Invalid ConfigDB. Error: {}".format(e)) [](http://example.com/codeflow?start=8&length=49)

Make sure this line is covered in unit test and coverage result. I understand this is challenging. Do you anticipate a valid code flow to reach here? If yes, could you write such unit testcase?

config/main.py
qiluo-msft: or you need to iterate the HLD PR.

config/main.py
```diff
@@ -3916,6 +3916,172 @@ def del_user(db, user):
             click.echo("Restart service snmp failed with error {}".format(e))
             raise click.Abort()
 
+#
+# 'bmp' group ('config bmp ...')
```
qiluo-msft, line 3920: >bmp [](http://example.com/codeflow?start=3&length=3)

Offline discussed, we need a feature level on/off. Some user may choose to disable the feature completely and do not waste any RAM or CPU.

config/main.py
FengPan-Frank: feature switch config will be in https://github.com/sonic-net/sonic-buildimage/pull/18946

config/main.py
FengPan-Frank: Added in https://github.com/sonic-net/sonic-buildimage/pull/18946/files#diff-94736fc06ef95ff9c8dc4dd4088f87f1c62ad322286b40cfd5b1b6c75bcb3764

config/main.py
FengPan-Frank: removed, there should not be such error since both entry exist and non-exist case are handled.

config/main.py
qiluo-msft: Still applicable

config/main.py
qiluo-msft: No, your link is build-time flag. I am talking about runtime enable/disable.

config/main.py
qiluo-msft: My point is that this is ConfigDB design, and your HLD does not mention the design.

FengPan-Frank: HLD contains this config cli design, refer below
image

tests/config_test.py
```diff
@@ -392,6 +395,165 @@ def teardown_class(cls):
...
+        import config.main
+        importlib.reload(config.main)
+
+    def test_enable_bgp_neighbor_table(
+            self,
+            get_cmd_module,
+            setup_single_broadcom_asic):
+        (config, show) = get_cmd_module
+        jsonfile_config = os.path.join(mock_bmp_db_path, "bmp_invalid.json")
+        from .mock_tables import dbconnector
```
liuh-80, line 412: The mock data been load to this dbconnector, but clirunner not using this.

You need load json to the config. you can check if this can work:

            # create object
            config.INIT_CFG_FILE = jsonfile_init_cfg
            config.DEFAULT_CONFIG_DB_FILE =  jsonfile_config

config/main.py
qiluo-msft: Thanks! above screenshot is the `config` CLI design. My point is that the HLD should has a ConfigDB schema design. Since you already has [https://github.com/sonic-net/sonic-buildimage/pull/18530](https://github.com/sonic-net/sonic-buildimage/pull/18530) merged. I will close this commemnt.

tests/config_test.py
FengPan-Frank: Figure out issue is related to parameter @clicommon.pass_db

config/main.py
FengPan-Frank: fixed.

FengPan-Frank: fixed.

config/main.py
qiluo-msft: also need testcase for bmp feature enable/disable

config/main.py
```diff
@@ -4191,6 +4191,142 @@ def del_user(db, user):
...
+    """BMP-related configuration"""
+
+
+#
+# 'enable' subgroup ('config bmp enable ...')
+#
+@bmp.group()
+@clicommon.pass_db
+def enable(db):
+    """Enable BMP table dump """
```
xincunli-sonic, line 4210: >"""Enable BMP table dump """ [](http://example.com/codeflow?start=4&length=28)

Should add pass?

config/main.py
```diff
@@ -4191,6 +4191,142 @@ def del_user(db, user):
...
+
+
+#
+# 'bgp-neighbor-table' command ('config bmp enable bgp-neighbor-table')
```
xincunli-sonic, line 4214: >'config bmp enable bgp-neighbor-table [](http://example.com/codeflow?start=32&length=37)

Would below cli format be better for understanding and less code?

config bmp --feature bgp-neighbor-table --state enabled
config bmp --feature bgp-neighbor-table --state disabled

Or

config bmp set bgp-neighbor-table enabled
config bmp set bgp-neighbor-table disabled

Or more straightforward

config bmp set bgp-neighbor-table=enabled bgp-rib-in-table=disabled

config/main.py
FengPan-Frank: done.

config/main.py
FengPan-Frank: This will require HLD update https://github.com/sonic-net/SONiC/blob/50cb55989649d5392b48b54bb7ca8263c883e162/doc/bmp/bmp.md#3-cli, previously I just follow bgp cli style, if no blocking issue, could we keep it first?

xincunli-sonic: Even though, can we have a common function to do db related operation, and just leave simple call in each cli part?

FengPan-Frank: done.

config/main.py
qiluo-msft: I see runtime flag. How about testcase for bmp feature enable/disable?

FengPan-Frank: Tracked via https://msazure.visualstudio.com/One/_workitems/edit/27588923

tests/config_test.py
```diff
@@ -702,6 +705,200 @@ def teardown_class(cls):
...
+class TestBMPConfig(object):
+    @classmethod
+    def setup_class(cls):
+        print("SETUP")
+        os.environ['UTILITIES_UNIT_TESTING'] = "1"
+        yield
+        print("TEARDOWN")
+        os.environ["UTILITIES_UNIT_TESTING"] = "0"
+
+    def test_disable_bgp_neighbor_table(
```
xincunli-sonic, line 717: >def test_disable_bgp_neighbor_table [](http://example.com/codeflow?start=4&length=35)

Minor suggestion: use @pytest.mark.parametrize to minimize repeating code.

FengPan-Frank: Thanks for the suggestion, updated.

xincunli-sonic (approved):

PR #3288, opened 2024-04-24
[config]Improve config save cli to save to one file for multiasic

wen587:
HLD design : https://github.com/sonic-net/SONiC/pull/1684
ADO: 27595294
#### What I did
Add support for config save to one file for multi-aisc.
#### How I did it
Extend support for one file save for multiasic using the below format:
```
{
  "localhost": {/*host config*/},
  "asic0": {/*asic0 config*/},
  ...
  "asicN": {/*asicN config*/}
}
```
#### How to verify it
Unit test and manual test on multiasic platform.
Example running multi:
```
admin@str2-8800-sup-2:~$ sudo config save -y tmp.json
Integrate each ASIC's config into a single JSON file tmp.json.
admin@str2-8800-sup-2:~$ cat tmp.json |more
{
    "localhost": {
        "ACL_TABLE": {
            "NTP_ACL": {
                "policy_desc": "NTP_ACL",
                "services": [
                    "NTP"
...
    "asic0": {
        "AUTO_TECHSUPPORT": {
            "GLOBAL": {
                "available_mem_threshold": "10.0",
```


qiluo-msft:
Do you have a HLD PR to link? Since this is a new feature, people need to read HLD to understand your big picture and goals in each code PR.

config/main.py
```diff
@@ -1237,18 +1237,29 @@ def save(filename):
     if filename is not None:
         cfg_files = filename.split(',')
 
-        if len(cfg_files) != num_cfg_file:
+        # If only one filename is provided in multi-ASIC mode,
+        # save all ASIC configurations to that single file.
+        if len(cfg_files) == 1 and multi_asic.is_multi_asic():
+            single_file_mode = True
+            filename = cfg_files[0]
+            cfg_files = [filename] * num_cfg_file
```
xincunli-sonic, line 1245: > * num_cfg_file [](http://example.com/codeflow?start=34&length=15)

No need this one

config/main.py
```diff
@@ -1271,6 +1282,16 @@ def save(filename):
         with open(file, 'w') as config_db_file:
```
qiluo-msft, line 1282: >config_db_file [](http://example.com/codeflow?start=32&length=14)

If user want to save to one file by providing expected command line argument, we should not save individual files for each asic? That is one of the reason that audience want to review HLD.

xincunli-sonic:
>                 file = "/etc/sonic/config_db{}.json".format(inst)

How about save all configurations to /etc/sonic too? Also, could we change reload from a single file in same PR which has consistent logic for save and reload?

---
Refers to: config/main.py:1271 in a30ce28. [](commit_id = a30ce28109825e9e3cdb35d422dd74006543154c, deletion_comment = False)

wen587:
> Do you have a HLD PR to link? Since this is a new feature, people need to read HLD to understand your big picture and goals in each code PR.

Will submit a HLD PR for this.

config/main.py
wen587: We don't save individual files for each asic. For multi-asic, it just save to one file finally.

config/main.py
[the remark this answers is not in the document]
wen587: The duplicate is to make code resuable for multiasic.

wen587:
> > ```
> >             file = "/etc/sonic/config_db{}.json".format(inst)
> > ```
> 
> How about save all configurations to /etc/sonic too? Also, could we change reload from a single file in same PR which has consistent logic for save and reload?
> 
> Refers to: config/main.py:1271 in [a30ce28](https://github.com/sonic-net/sonic-utilities/commit/a30ce28109825e9e3cdb35d422dd74006543154c). [](commit_id = [a30ce28](https://github.com/sonic-net/sonic-utilities/commit/a30ce28109825e9e3cdb35d422dd74006543154c), deletion_comment = False)

Hi @xincunli-sonic , what do you mean save all configuration to /etc/sonic. If we didn't specify the path, it will save default config_db.json, config_db0.json... to /etc/sonic.

config reload in single file is much more complex than config save. I prefer to split PR for this.

xincunli-sonic (approved):
LGTM

config/main.py
qiluo-msft: Then why it is in a for-loop?

config/main.py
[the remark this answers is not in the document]
qiluo-msft: but we should not waste CPU time to loop through the list.

config/main.py
```diff
@@ -1237,18 +1237,29 @@ def save(filename):
     if filename is not None:
         cfg_files = filename.split(',')
 
-        if len(cfg_files) != num_cfg_file:
+        # If only one filename is provided in multi-ASIC mode,
+        # save all ASIC configurations to that single file.
+        if len(cfg_files) == 1 and multi_asic.is_multi_asic():
```
wenyiz2021, line 1242: in this case for multi-asic, when single file name is provided to do `config save`, we are saving to 1 single file:
1. in this single file, do we have duplicate content from both host and asics?
2. what is the consequences of overwriting all config in one file? should we prevent this behavior

wenyiz2021 (changes requested):
in this case for multi-asic, when single file name is provided to do config save, we are saving to 1 single file:

in this single file, do we have duplicate content from both host and asics?
what is the consequences of overwriting all config in one file? should we prevent this behavior

config/main.py
wen587: 1. We do have duplicate content. Just like we have duplicate content in current `config_db.json,config_db0.json,config_dbN.json`.
2. This is by design. NDM golden config for multi-asic will be single file also. Please check the HLD design doc.https://github.com/sonic-net/SONiC/pull/1684

wenyiz2021: I see.
how do we interpret back the single file, into config dbs?

wen587: I will create another PR to extend `config reload` CLI to interpret it.

wenyiz2021: sounds good. please paste the other PR once ready. thanks!

wenyiz2021 (dismissed):

config/main.py
```diff
@@ -1237,7 +1254,13 @@ def save(filename):
...
+            filename = cfg_files[0]
+            save_to_single_file_multi(db, filename)
```
qiluo-msft, line 1261: >save_to_single_file_multi [](http://example.com/codeflow?start=12&length=25)

save_to_single_file_multi -> multi_asic_save_config

qiluo-msft (commented):
LGMT with minor comment on function name

xincunli-sonic (approved):

config/main.py
```diff
@@ -1155,6 +1155,22 @@ def validate_gre_type(ctx, _, value):
...
+    """
+    all_files_config = {}
+    cfgdb_clients = db.cfgdb_clients
+
+    for ns, config_db in cfgdb_clients.items():
+        current_config = config_db.get_config()
+        sonic_cfggen.FormatConverter.to_serialized(current_config)
+        asic_name = "localhost" if ns == DEFAULT_NAMESPACE else ns
+        all_files_config[asic_name] = sort_dict(current_config)
+    click.echo("Integrate each ASIC's config into a single JSON file {}.".format(filename))
```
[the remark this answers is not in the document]
wen587: Hi @judyjoseph , It is instructional message. 
It will not cause any sonic-mgmt test failure, because we didn't allow multiasic to save to single file without this PR implementation.
It is an extention for config save usage when multiasic and specify one file to save all asic config.

xincunli-sonic (approved):
LGTM

wenyiz2021 (approved):

judyjoseph (approved):
LGTM

PR #3295, opened 2024-04-28
[Graceful Reboot] Check the watchdog supporting capability before arm it in the reboot scripts

keboliu:
#### What I did
In the current implementation, the reboot script will arm the watchdog with 180s timeout periods, not all the platforms can support a such long watchdog timeout period. 

For example, Nvidia "x86_64-mlnx_msn2100-r0" and "x86_64-mlnx_msn2010-r0" platforms can only support 32s timeout periods.  When use "watchdogutil" to arm WD on them with 180s periods, the WD timeout will only be set to the max value of 32s.  Within these timeout periods, the platforms even can not finish shutdown, so each time during the reboot WD will always kick in and interrupt the graceful shutdown. it will also cause a confusing reboot cause(triggered by the watchdog). On this kind of platform, watchdog should not be armed to avoid issues.

#### How I did it

Add a list of platforms that do not support long watchdog timeout periods (180s), for these platforms the reboot command will skip the arm watchdog step.

#### How to verify it

perform reboot on all the platforms, watchdog can be armed or skipped accordingly after performing the reboot command.

### DIFF
--- a/scripts/reboot
+++ b/scripts/reboot
@@ -11,6 +11,8 @@ PLATFORM_UPDATE_REBOOT_CAUSE="platform_update_reboot_cause"
 REBOOT_CAUSE_FILE="/host/reboot-cause/reboot-cause.txt"
 PLATFORM_REBOOT_PRE_CHECK="platform_reboot_pre_check"
 REBOOT_TIME=$(date)
+PLATFORMS_NOT_SUPPORT_180S_WD=("x86_64-mlnx_msn2100-r0" "x86_64-mlnx_msn2010-r0")
+IS_PLATFORM_SUPPORT_180S_WD=1
 
 # Reboot immediately if we run the kdump capture kernel
 VMCORE_FILE=/proc/vmcore
@@ -177,6 +179,19 @@ function check_conflict_boot_in_fw_update()
     fi
 }
 
+function is_platform_support_180s_wd()
+{
+    # Test whether the platform support 180s wathdog
+    for i in "${PLATFORMS_NOT_SUPPORT_180S_WD[@]}"
+    do
+        if [ "$i" == "${PLATFORM}" ] ; then
+            debug "Platform ${PLATFORM} doesn't support 180s watchdog"
+            IS_PLATFORM_SUPPORT_180S_WD=0
+            return
+        fi
+    done
+}
+
 function parse_options()
 {
     while getopts "h?v" opt; do
@@ -265,9 +280,13 @@ if [ -x ${DEVPATH}/${PLATFORM}/${PRE_REBOOT_HOOK} ]; then
     fi
 fi
 
-if [ -x ${WATCHDOG_UTIL} ]; then
+is_platform_support_180s_wd
+
+if [[ -x ${WATCHDOG_UTIL} && "${IS_PLATFORM_SUPPORT_180S_WD}" == 1 ]]; then
     debug "Enabling the Watchdog before reboot"
     ${WATCHDOG_UTIL} arm
+else
+    debug "Skip enabling the watchdog"
 fi
 
 if [ -x ${DEVPATH}/${PLATFORM}/${PLAT_REBOOT} ]; then


dprital:
@prgeor , Can you please review ?

dprital:
@vaibhavhd , Can you please review as well ?

scripts/reboot
```diff
@@ -11,6 +11,8 @@ PLATFORM_UPDATE_REBOOT_CAUSE="platform_update_reboot_cause"
 REBOOT_CAUSE_FILE="/host/reboot-cause/reboot-cause.txt"
 PLATFORM_REBOOT_PRE_CHECK="platform_reboot_pre_check"
 REBOOT_TIME=$(date)
+PLATFORMS_NOT_SUPPORT_180S_WD=("x86_64-mlnx_msn2100-r0" "x86_64-mlnx_msn2010-r0")
```
vaibhavhd, line 14: This deny-list approach is reactive, and not proactive. Is there a way the support can be detected automatically? If you can add such a support for MLNX only platforms then is that still better that this?

scripts/reboot
```diff
@@ -11,6 +11,8 @@ PLATFORM_UPDATE_REBOOT_CAUSE="platform_update_reboot_cause"
 REBOOT_CAUSE_FILE="/host/reboot-cause/reboot-cause.txt"
```
vaibhavhd, line 11: Do you not plan to handle other kind of reboots which also arm the system with watchdog? Eg., warm-reboot?

scripts/reboot
```diff
@@ -265,9 +280,13 @@ if [ -x ${DEVPATH}/${PLATFORM}/${PRE_REBOOT_HOOK} ]; then
...
 fi
 
-if [ -x ${WATCHDOG_UTIL} ]; then
+is_platform_support_180s_wd
+
+if [[ -x ${WATCHDOG_UTIL} && "${IS_PLATFORM_SUPPORT_180S_WD}" == 1 ]]; then
     debug "Enabling the Watchdog before reboot"
     ${WATCHDOG_UTIL} arm
+else
+    debug "Skip enabling the watchdog"
```
vaibhavhd, line 289: There are downsides to blanket disable watchdog. Especially in the case of warm-reboot and such time sensitive reboots.

scripts/reboot
```diff
@@ -177,6 +179,19 @@ function check_conflict_boot_in_fw_update()
     fi
 }
 
+function is_platform_support_180s_wd()
+{
```
vaibhavhd, line 183: I think this logic is better suited to be placed in WATCHDOG_UTIL

vaibhavhd:
Is it ideal for a platform to take more than 32s to come up after last exit point (when the watchdog is armed). Isn't that a problem on its own and should be caught by watchdog as it happens today. If not, what's the reason for such long delays in these platforms.

keboliu:
> Is it ideal for a platform to take more than 32s to come up after last exit point (when the watchdog is armed). Isn't that a problem on its own and should be caught by watchdog as it happens today. If not, what's the reason for such long delays in these platforms.

Hi @vaibhavhd thanks for reviewing my PR.

Since in the code, the watchdog period is set to 180s, I suppose the hard limit of the shutdown time should be 180s (considering lots of stuff like FW upgrading could need to be done, I think this is reasonable), so the shutdown time on these systems longer than 32s shouldn't be a problem.  
This issue is caused by the hardware limitation(watchdog period limitation) of the systems. So if you think this issue shall be fixed from the platform side, I can do it inside our platform  API, to deny the 180s watchdog setting on these platforms since the setting is beyond its capability. The reboot behavior will be the same as in this PR(which means the watchdog shouldn't kick in on these platforms), but the change will be platform-specific, Does this sound good to you?

keboliu:
> > Is it ideal for a platform to take more than 32s to come up after last exit point (when the watchdog is armed). Isn't that a problem on its own and should be caught by watchdog as it happens today. If not, what's the reason for such long delays in these platforms.
> 
> Hi @vaibhavhd thanks for reviewing my PR.
> 
> Since in the code, the watchdog period is set to 180s, I suppose the hard limit of the shutdown time should be 180s (considering lots of stuff like FW upgrading could need to be done, I think this is reasonable), so the shutdown time on these systems longer than 32s shouldn't be a problem. This issue is caused by the hardware limitation(watchdog period limitation) of the systems. So if you think this issue shall be fixed from the platform side, I can do it inside our platform API, to deny the 180s watchdog setting on these platforms since the setting is beyond its capability. The reboot behavior will be the same as in this PR(which means the watchdog shouldn't kick in on these platforms), but the change will be platform-specific, Does this sound good to you?

@vaibhavhd  would you please share your thoughts?

PR #3296, opened 2024-04-29
Migrate AAA table per-command authorization in db_migrator

liuh-80:
Migrate AAA table per-command authorization in db_migrator

#### Why I did it
    per-command AAA need enable in warm-upgrade case

#### How I did it
    Add code to migrate per-command aunthorization

#### How to verify it
    Pass all test case.
    Add new test case.

    N/A

    Migrate AAA table per-command authorization in db_migrator


liuh-80:
This PR is based on per-command accounting migrate PR: https://github.com/sonic-net/sonic-utilities/pull/3284

scripts/db_migrator.py
```diff
@@ -840,6 +840,18 @@ def migrate_aaa(self):
             self.configDB.set_entry("AAA", "accounting", accounting_new)
             log.log_info('Migrate AAA accounting: {}'.format(accounting_new))
 
+        # setup per-command authorization
+        tacplus_config = self.configDB.get_entry('TACPLUS', 'global')
+        if 'passkey' in tacplus_config and '' != tacplus_config.get('passkey'):
+            authorization = self.configDB.get_entry('AAA', 'authorization')
+            if not authorization:
```
qiluo-msft, line 847: >authorization [](http://example.com/codeflow?start=19&length=13)

Could you provide a table to analyze the old_config, new_config, actions for all the situations?

I am trying to read: if old_config has empty /AAA/authorization, you will migrate. Otherwise, noop.

liuh-80: According to your comments in https://github.com/sonic-net/sonic-utilities/pull/3284:

[qiluo-msft](https://github.com/qiluo-msft) [2 weeks ago](https://github.com/sonic-net/sonic-utilities/pull/3284#discussion_r1581653675)
The old image's enablement does not matter. As long as the required passkey is there, and intention is to enable, then we should enable.


So, we only check the new config:


<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:dt="uuid:C2F41010-65B3-11d1-A29F-00AA00C14882"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=OneNote.File>
<meta name=Generator content="Microsoft OneNote 15">
</head>

<body lang=en-US style='font-family:Calibri;font-size:11.0pt'>


<div style='direction:ltr'>


New_config | Action
-- | --
Passkey does not   exist | Not enable
Passkey exist   Authorization   enable | Enable
Passkey exist   Authorization not   enable | Not enable



</div>


</body>

</html>

scripts/db_migrator.py
```diff
@@ -840,6 +840,18 @@ def migrate_aaa(self):
...
+                authorization_new = aaa_new.get("authorization")
+                self.configDB.set_entry("AAA", "authorization", authorization_new)
+                log.log_info('Migrate AAA authorization: {}'.format(authorization_new))
+        else:
+            # If no passkey, setup per-command authorization will block remote user command
```
qiluo-msft, line 852: >If no passkey [](http://example.com/codeflow?start=14&length=13)

As the designed in your table, you should disable Authorization. However, the current implement is just log and skip.

liuh-80: Fixed, disable authorization by delete entry.

qiluo-msft (approved):

PR #3297, opened 2024-04-30
Add show fabric rate command

jfeng-arista:
#### What I did

Added "show fabric counters rate mbps" command. 

The output of the command is 
```
  ASIC    Link ID    Rx Data Mbps    Tx Data Mbps
------  ---------  --------------  --------------
 asic0          0               0            19.8
 asic0          1               0            19.8
 asic0          2               0            39.8
 asic0          3               0            39.8
...

  ASIC    Link ID    Rx Data Mbps    Tx Data Mbps
------  ---------  --------------  --------------
 asic1          0               0               0
 asic1          1               0               0
 asic1          2               0               0
 asic1          3               0               0
...
```
The HLD is at here

https://github.com/sonic-net/SONiC/blob/master/doc/voq/fabric.md
Amendment is at https://github.com/sonic-net/SONiC/pull/1656

### DIFF
--- a/show/fabric.py
+++ b/show/fabric.py
@@ -75,6 +75,7 @@ def queue(namespace):
         cmd += ['-n', str(namespace)]
     clicommon.run_command(cmd)
 
+
 @counters.command()
 @multi_asic_util.multi_asic_click_option_namespace
 def rate(namespace):
--- a/show/fabric.py
+++ b/show/fabric.py
@@ -75,17 +75,10 @@ def queue(namespace):
         cmd += ['-n', str(namespace)]
     clicommon.run_command(cmd)
 
-
-@counters.group(invoke_without_command=True)
-def rate():
-    """Show fabric countes rate"""
-    pass
-
-
-@rate.command()
+@counters.command()
 @multi_asic_util.multi_asic_click_option_namespace
-def mbps(namespace):
-    """Show fabric counters rate mbps"""
+def rate(namespace):
+    """Show fabric counters rate"""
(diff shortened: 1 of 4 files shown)


kenneth-arista:
@arlakshm

show/fabric.py
```diff
@@ -74,3 +74,19 @@ def queue(namespace):
...
+
+@counters.group(invoke_without_command=True)
+def rate():
+    """Show fabric countes rate"""
+    pass
+
+
+@rate.command()
+@multi_asic_util.multi_asic_click_option_namespace
+def mbps(namespace):
```
arlakshm, line 87: why do we need this? can we have the command `show fabric counters rate` and we can have the default speed as `mbps` ?

jfeng-arista: > why do we need this? can we have the command `show fabric counters rate` and we can have the default speed as `mbps` ?

Thank you for your comments . I just followed one that we discussed in the HLD ( 2.8.3)  ,  let me know if this is ok or want me change to the new command format that you mentioned and update the HLD as well. 

thank you

jfeng-arista: > > why do we need this? can we have the command `show fabric counters rate` and we can have the default speed as `mbps` ?
> 
> Thank you for your comments . I was just following the one that we discussed in the HLD ( 2.8.3) , let me know if this is ok or want me change to the new command format that you mentioned and update the HLD as well.
> 
> thank you

revised after last Friday's discussion, and update the HLD as well.

arlakshm (approved):

PR #3236, opened 2024-03-22
Update sonic-utilities to support new SKU Mellanox-SN5600-O128

stephenxs:
#### What I did

Update sonic-utilities to support new SKU Mellanox-SN5600-O128

1. Add the SKU to the generic configuration updater
2. Simplify the logic of the buffer migrator to support the new SKU

#### How to verify it

Manual and unit tests

PR #3308, opened 2024-05-06
[202311] Update sonic-utilities to support new SKU Mellanox-SN5600-O128 (#3236)

stephenxs:
This is to backport PR #3236 to 202311 branch

Update sonic-utilities to support new SKU Mellanox-SN5600-O128

1. Add the SKU to the generic configuration updater
2. Simplify the logic of the buffer migrator to support the new SKU

Manual and unit tests


stephenxs:
No need to keep the PR as the original one can be cherry-picked without a conflict. Closing it.

PR #3333, opened 2024-05-21
Add a check for ensuring mirror session ACLs are programmed to ASIC

ryanzhu706:
Description
Add a check for ensuring mirror session ACLs are programmed to ASIC

What is the issue?
This fix is to address an issue where an ACL is added to CONFIG_DB, but before it could be programmed to ASIC, Orchagent is paused.
This leads to APPLY_VIEW failure when base image OA could not process this ACL entry and target image's OA still creates it.
The issue has an image fix available at https://github.com/sonic-net/sonic-sairedis/pull/1240
This issue is very rare, and has been caught by upgrade path tests only once in thousands of iterations.

What is this fix?
A new logic is added to check if mirror session ACLs for arp and nd are added to ASIC..
ACLs are looked into ASIC_DB and matched using SAI_ACL_ENTRY_ATTR_PRIORITY attribute.
SAI_ACL_ENTRY_ATTR_PRIORITY for arp ACL is 8888 and for nd is 8887
If one of the ACLs is found missing then warmboot is aborted.

Tested:


scripts/fast-reboot
```diff
@@ -272,13 +273,45 @@ function backup_database()
     fi
 }
 
+function check_mirror_session_acls()
+{
+    ACL_ARP="missing"
```
vaibhavhd, line 278: As discussed: 

Please improve the log handling here by adding a new debug line to indicate start of this function call.

scripts/fast-reboot
```diff
@@ -272,13 +273,45 @@ function backup_database()
...
+    ACL_ND="missing"
+    start_time=${SECONDS}
+    elapsed_time=$((${SECONDS} - ${start_time}))
+    while [[ ${elapsed_time} -lt 10 ]]; do
+        ACL_ENTRIES=( $(redis-cli -n 1 KEYS "*" | grep SAI_OBJECT_TYPE_ACL_ENTRY) )
```
vaibhavhd, line 283: As we found out in the test - this step is problematic. Please add a way to handle the case when no ACL_ENTRY objects are found, and grep fails with non-zero exit code. 

The goal is for any failure we see here, the error should be gracefully handled and exit code should be `EXIT_NO_MIRROR_SESSION_ACLS`

vaibhavhd: Wouldn't your latest change also fail when grep does not find anything?  In the current code if grep fails then you wouldn't move to next step to check `$?`, rather the code will just exit from this line.

Example code below. But please test it out.

```bash
CHECK_ACL_ENTRIES=0
ACL_OUTPUT=$(sonic-db-cli ASIC_DB KEYS "*" | grep SAI_OBJECT_TYPE_ACL_ENTRY) || CHECK_ACL_ENTRIES=$?
if [[ CHECK_ACL_ENTRIES-ne 0 ]]; then
      error "Failed to retrieve SAI_OBJECT_TYPE_ACL_ENTRY from ASIC_DB"
fi
# Convert output to list
ACL_ENTRIES=( ${ACL_OUTPUT} )
```

vaibhavhd (dismissed):
Please test the change and update the description before PR merge

saiarcot895:
You'll also need to fix up the email addresses in the commits (or squash the commits) for EasyCLA to pass.

vaibhavhd (approved):

ayurkiv-nvda:
Hello @vaibhavhd 
We found that in while loop we will fail after 1 iteration if ACL entries were not found in redis
I suggest continuing retry cycle.
Please take a look at suggested solution:
https://github.com/ayurkiv-nvda/sonic-utilities/pull/26

Thanks

PR #3344, opened 2024-05-29
[SPM] Add logic to disable the feature before stopping it and enabling it before starting

vadymhlushko-mlnx:
#### What I did
Add logic to disable the feature before stopping and enabling it before starting in order to properly clean the `systemd` symlinks to avoid issues with `delayed` attribute explained in the `How to verify it` section.

#### How I did it
Add the `systemctl disable ...` after the `systemctl stop...` and the `systemctl enable ...` before the `systemctl start ..` for some feature.

#### How to verify it
- Add repository for some `featureX`
- `sonic-package-manager repository <featureX> <URL>`
- Install `featureX` version 1.0.0 where the `delayed` flag is equal to `false` (`delayed` flag means - the feature will be started right after the system boots or after the `PortInitDone` event)
- `sonic-package-manager install featureX==1.0.0 -y`
- Enable the feature in SONiC
- `config feature state featureX enabled`
- Install `featureX` version 1.0.1 where the `delayed` flag is equal to `true`
- `sonic-package-manager install featureX==1.0.1 -y`
- Check the `manifest` file to verify the `delayed` field value
- `sonic-package-manager show package manifest featureX`
- `config save -y`
- `reboot`
- Check that the `featureX` is delayed on the system start

### DIFF
--- a/sonic_package_manager/manager.py
+++ b/sonic_package_manager/manager.py
@@ -1017,8 +1017,10 @@ class PackageManager:
 
     def _stop_feature(self, package: Package):
         self._systemctl_action(package, 'stop')
+        self._systemctl_action(package, 'disable')
 
     def _start_feature(self, package: Package):
+        self._systemctl_action(package, 'enable')
         self._systemctl_action(package, 'start')
 
     def _systemctl_action(self, package: Package, action: str):
--- a/tests/sonic_package_manager/test_manager.py
+++ b/tests/sonic_package_manager/test_manager.py
@@ -2,7 +2,7 @@
 
 import re
 import unittest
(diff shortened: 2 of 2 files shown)


stepanblyschak (approved):

liat-grozovik:
@lguohan can you suggest a reviewer for this issue?

liat-grozovik:
@lguohan kindly reminder the fix should go for both 202311 and 202405

liat-grozovik:
@qiluo-msft could you please help to review?

qiluo-msft (approved):

stepanblyschak:
@azpw run

qiluo-msft (approved):

PR #3347, opened 2024-05-30
Fix show fabric monitor capacity command when the feature is disabled.

jfeng-arista:
#### What I did

"show fabric monitor capacity" command has a index error if the feature is disabled. With this change, the command can still display information without errors. 

the issue is tracking at https://github.com/sonic-net/sonic-utilities/issues/3350


jfeng-arista:
@arlakshm @kenneth-arista @vmittal-msft

vmittal-msft (approved):

PR #3349, opened 2024-05-31
[config]Support single file reload for multiasic

wen587:
ADO: 27595279
#### What I did
Extend config reload to support single file reloading for multi-asic
#### How I did it
Add the single file reload support for mutli-asic 
#### How to verify it
Unit test and manual test on multi-asic DUT


config/main.py
```diff
@@ -1160,7 +1160,7 @@ def validate_gre_type(ctx, _, value):
         raise click.UsageError("{} is not a valid GRE type".format(value))
 
 
-def multi_asic_save_config(db, filename):
+def multiasic_save_to_singlefile(db, filename):
```
xincunli-sonic, line 1163: >multiasic_save_to_singlefile [](http://example.com/codeflow?start=4&length=28)

what is difference with "show runningconfiguration all"?

config/main.py
```diff
@@ -1227,6 +1227,91 @@ def validate_patch(patch):
     except Exception as e:
         raise GenericConfigUpdaterError(f"Validate json patch: {patch} failed due to:{e}")
 
+
+def multiasic_validate_single_file(filename):
+    ns_list = [DEFAULT_NAMESPACE, *multi_asic.get_namespace_list()]
+    file_input = read_json_file(filename)
+    file_ns_list = [DEFAULT_NAMESPACE if key == HOST_NAMESPACE else key for key in file_input]
+    if set(ns_list) != set(file_ns_list):
```
xincunli-sonic, line 1235: >set(ns_list) != set(file_ns_list) [](http://example.com/codeflow?start=7&length=33)

Can we have minus operation which will give the missing part.

config/main.py
```diff
@@ -1227,6 +1227,91 @@ def validate_patch(patch):
...
+        click.echo("Input file {} must contain all asics config".format(filename))
+        raise click.Abort()
+
+
+def load_sysinfo_if_missing(asic_config):
+    platform = asic_config.get("DEVICE_METADATA", {}).\
```
xincunli-sonic, line 1241: >DEVICE_METADATA [](http://example.com/codeflow?start=32&length=15)

Possible to add a constant?

config/main.py
```diff
@@ -1227,6 +1227,91 @@ def validate_patch(patch):
...
+                        .format(None if platform is None else platform, None if mac is None else mac))
+        return True
+    return False
+
+
+def flush_configdb(namespace=DEFAULT_NAMESPACE):
+    if namespace is DEFAULT_NAMESPACE:
+        config_db = ConfigDBConnector()
+    else:
+        config_db = ConfigDBConnector(use_unix_socket_path=True, namespace=namespace)
```
xincunli-sonic, line 1256: >namespace=namespace [](http://example.com/codeflow?start=65&length=19)

if namespace is empty, it will return host configdb, right?

config/main.py
```diff
@@ -1227,6 +1227,91 @@ def validate_patch(patch):
...
+
+
+def migrate_db_to_lastest(namespace=DEFAULT_NAMESPACE):
+    # Migrate DB contents to latest version
+    db_migrator = '/usr/local/bin/db_migrator.py'
+    if os.path.isfile(db_migrator) and os.access(db_migrator, os.X_OK):
+        if namespace is DEFAULT_NAMESPACE:
+            command = [db_migrator, '-o', 'migrate']
+        else:
+            command = [db_migrator, '-o', 'migrate', '-n', str(namespace)]
```
xincunli-sonic, line 1271: >str(namespace) [](http://example.com/codeflow?start=59&length=14)

why we need str convert?

config/main.py
```diff
@@ -1227,6 +1227,91 @@ def validate_patch(patch):
...
+
+def multiasic_write_to_db(filename, load_sysinfo):
+    file_input = read_json_file(filename)
+    for ns in [DEFAULT_NAMESPACE, *multi_asic.get_namespace_list()]:
+        asic_name = HOST_NAMESPACE if ns == DEFAULT_NAMESPACE else ns
+        asic_config = file_input[asic_name]
+
+        asic_load_sysinfo = True if load_sysinfo else False
+        if not asic_load_sysinfo:
+            asic_load_sysinfo = load_sysinfo_if_missing(asic_config)
```
xincunli-sonic, line 1283: >asic_load_sysinfo [](http://example.com/codeflow?start=12&length=17)

This name is not clear to indicate the platform or mac are missing

config/main.py
```diff
@@ -1227,6 +1227,91 @@ def validate_patch(patch):
...
+        if asic_load_sysinfo:
+            cfg_hwsku = asic_config.get("DEVICE_METADATA", {}).\
+                get(HOST_NAMESPACE, {}).get("hwsku")
+            if not cfg_hwsku:
+                click.secho("Could not get the HWSKU from config file,  Exiting!!!", fg='magenta')
+                sys.exit(1)
+
+        client, _ = flush_configdb(ns)
+
+        if asic_load_sysinfo:
```
xincunli-sonic, line 1294: >asic_load_sysinfo [](http://example.com/codeflow?start=11&length=17)

IMHO, this variable equals True that means something missed? Then I am quite confusion we write to db under missing circumstance?

config/main.py
wen587: The difference is that `show run all` will have `bgpraw` data that is not part of ConfigDB

config/main.py
wen587: Right.

config/main.py
wen587: We can leave it as it represents a general config table.

config/main.py
wen587: Copried from earlier code. Could be some historic reason. I will remove it.

config/main.py
wen587: There is `load_sysinfo` flag in `load_minigraph` which is to generate sysinfo which is (platform and mac).
The initial thought is to keep the name identical since they are doing the same thing. I'll explain more on next comment.

config/main.py
wen587: No matter the config source is minigraph or Golden Config, the config source itself could not get the mac in advance. The platform info and mac address was generated at run time.
Take single asic as example, Golden Config cannot provide such infomation. But SONiC container will not successfully get up if platform and mac are missing in CONFIGDB. That's why we need to check if sysinfo is missing and generate this info as needed.

config/main.py
wen587: added

xincunli-sonic (dismissed):
LGTM

config/main.py
wen587: This part of code was copied from the original. Seems one more difference for use_unix_socket_path. The default value is False. Not sure any historic reason on this.

config/main.py
```diff
@@ -1227,6 +1227,92 @@ def validate_patch(patch):
     except Exception as e:
         raise GenericConfigUpdaterError(f"Validate json patch: {patch} failed due to:{e}")
 
+
+def multiasic_validate_single_file(filename):
+    ns_list = [DEFAULT_NAMESPACE, *multi_asic.get_namespace_list()]
+    file_input = read_json_file(filename)
+    file_ns_list = [DEFAULT_NAMESPACE if key == HOST_NAMESPACE else key for key in file_input]
```
qiluo-msft, line 1234: >HOST_NAMESPACE [](http://example.com/codeflow?start=48&length=14)

what is the diff HOST_NAMESPACE vs DEFAULT_NAMESPACE?

config/main.py
```diff
@@ -1227,6 +1227,92 @@ def validate_patch(patch):
...
+    if set(ns_list) != set(file_ns_list):
+        diff = set(ns_list) - set(file_ns_list)
```
qiluo-msft, line 1236: >diff [](http://example.com/codeflow?start=8&length=4)

If the purpose is to print error message, just directly print them in original content, do not manipulate or combine.

config/main.py
```diff
@@ -1227,6 +1227,92 @@ def validate_patch(patch):
...
+        click.echo("Input file {} must contain all asics config. Missing {}".format(filename, diff))
+        raise click.Abort()
+
+
+def load_sysinfo_if_missing(asic_config):
+    platform = asic_config.get("DEVICE_METADATA", {}).\
```
qiluo-msft, line 1242: >asic_config.get("DEVICE_METADATA", {}) [](http://example.com/codeflow?start=15&length=38)

You can create a var just for DEVICE_METADATA value, and prevent breaking long line, which is not easy to read.

config/main.py
```diff
@@ -1227,6 +1227,92 @@ def validate_patch(patch):
...
+        get(HOST_NAMESPACE, {}).get("platform")
+    mac = asic_config.get("DEVICE_METADATA", {}).\
+        get(HOST_NAMESPACE, {}).get("mac")
+    if not platform or not mac:
+        log.log_warning("Input file does't have platform or mac. platform: {}, mac: {}"
```
qiluo-msft, line 1247: >does [](http://example.com/codeflow?start=36&length=4)

typo

config/main.py
```diff
@@ -1227,6 +1227,92 @@ def validate_patch(patch):
...
+                        .format(None if platform is None else platform, None if mac is None else mac))
```
qiluo-msft, line 1248: >if [](http://example.com/codeflow?start=37&length=2)

If you just want to print error message, keep statements simple, for example:
1. if mac is missing, warn that mac is missing
2. if platform is missing, warm that platform is missing

config/main.py
wen587: HOST_NAMESPACE  is 'localhost'
DEFAULT_NAMESPACE is ''

config/main.py
```diff
@@ -1227,6 +1227,95 @@ def validate_patch(patch):
...
+                filename, ns_list, file_ns_list)
+        )
+        raise click.Abort()
+
+
+def load_sysinfo_if_missing(asic_config):
+    device_metadata = asic_config.get('DEVICE_METADATA', {})
+    platform = device_metadata.get("localhost", {}).get("platform")
+    mac = device_metadata.get("localhost", {}).get("mac")
+    if not platform or not mac:
```
qiluo-msft, line 1247: >if not platform or not mac: [](http://example.com/codeflow?start=4&length=27)

you can remove this line, and reorg below as if-elif-else

config/main.py
```diff
@@ -1227,6 +1227,95 @@ def validate_patch(patch):
...
+            log.log_warning("platfrom is missing from Input file")
+        if not mac:
+            log.log_warning("mac is missing from Input file")
+        return True
+    return False
+
+
+def flush_configdb(namespace=DEFAULT_NAMESPACE):
+    if namespace is DEFAULT_NAMESPACE:
+        config_db = ConfigDBConnector()
```
qiluo-msft, line 1258: >ConfigDBConnector [](http://example.com/codeflow?start=20&length=17)

Is it same as the statement in `else`? if yes, no need to check namespace.

config/main.py
```diff
@@ -1227,6 +1227,95 @@ def validate_patch(patch):
...
+    else:
+        config_db = ConfigDBConnector(use_unix_socket_path=True, namespace=namespace)
+
+    config_db.connect()
+    client = config_db.get_redis_client(config_db.CONFIG_DB)
+    client.flushdb()
+    return client, config_db
+
+
+def migrate_db_to_lastest(namespace=DEFAULT_NAMESPACE):
```
qiluo-msft, line 1268: >migrate_db_to_lastest [](http://example.com/codeflow?start=4&length=21)

@vaibhavhd Could you help review?

config/main.py
```diff
@@ -1227,6 +1227,95 @@ def validate_patch(patch):
...
+    db_migrator = '/usr/local/bin/db_migrator.py'
+    if os.path.isfile(db_migrator) and os.access(db_migrator, os.X_OK):
+        if namespace is DEFAULT_NAMESPACE:
+            command = [db_migrator, '-o', 'migrate']
+        else:
+            command = [db_migrator, '-o', 'migrate', '-n', namespace]
+        clicommon.run_command(command, display_cmd=True)
+
+
+def multiasic_write_to_db(filename, load_sysinfo):
```
qiluo-msft, line 1279: >multiasic_write_to_db [](http://example.com/codeflow?start=4&length=21)

@rlhui @abdosi @judyjoseph Could you help review?

config/main.py
```diff
@@ -1227,6 +1227,95 @@ def validate_patch(patch):
...
+
+        asic_load_sysinfo = True if load_sysinfo else False
+        if not asic_load_sysinfo:
+            asic_load_sysinfo = load_sysinfo_if_missing(asic_config)
+
+        if asic_load_sysinfo:
+            cfg_hwsku = asic_config.get("DEVICE_METADATA", {}).\
+                get("localhost", {}).get("hwsku")
+            if not cfg_hwsku:
+                click.secho("Could not get the HWSKU from config file,  Exiting!!!", fg='magenta')
```
qiluo-msft, line 1293: >!!! [](http://example.com/codeflow?start=79&length=3)

Keep only one `!`

config/main.py
```diff
@@ -1227,6 +1227,95 @@ def validate_patch(patch):
...
+        else:
+            config_db = ConfigDBPipeConnector(use_unix_socket_path=True, namespace=ns)
+
+        config_db.connect(False)
+        sonic_cfggen.FormatConverter.to_deserialized(asic_config)
+        data = sonic_cfggen.FormatConverter.output_to_db(asic_config)
+        config_db.mod_config(sonic_cfggen.FormatConverter.output_to_db(data))
+        client.set(config_db.INIT_INDICATOR, 1)
+
+        migrate_db_to_lastest(ns)
```
qiluo-msft, line 1316: >migrate_db_to_lastest [](http://example.com/codeflow?start=8&length=21)

Is this multi-asic db_migrate required? Is it a new feature in this PR?

config/main.py
wen587: No. There is one more difference with argument `use_unix_socket_path`. The default value is False. This part is copied from original code.

xincunli-sonic (approved):

judyjoseph:
looks ok to me. Btw are you validating the switch booting up with a single config file ? or this is just "config reload" command changes.  I have this question, because there could be startup scripts where we explicitly check for the presence of config_db<n>.json

wen587:
> looks ok to me. Btw are you validating the switch booting up with a single config file ? or this is just "config reload" command changes. I have this question, because there could be startup scripts where we explicitly check for the presence of config_db.json

This is just config reload command and it will not impact booting up.

qiluo-msft (approved):

qiluo-msft:
@gechiang Could you help backport?

wen587:
https://github.com/sonic-net/sonic-utilities/pull/3481

PR #3375, opened 2024-06-20
fix: fix show bgp summary output typo

cyw233:
#### What I did
I fixed a typo in the output of `show [ip|ipv6] bgp summary` command: `Neighbhor` -> `Neighbor`

```text
Neighbhor      V     AS    MsgRcvd    MsgSent    TblVer    InQ    OutQ  Up/Down      State/PfxRcd  NeighborName
-----------  ---  -----  ---------  ---------  --------  -----  ------  ---------  --------------  --------------
10.0.0.1       4  65200       4632      11028         0      0       0  00:18:31             8514  ARISTA01T2
```

```text
Neighbor      V     AS    MsgRcvd    MsgSent    TblVer    InQ    OutQ  Up/Down      State/PfxRcd  NeighborName
----------  ---  -----  ---------  ---------  --------  -----  ------  ---------  --------------  --------------
10.0.0.1      4  65200       4632      11028         0      0       0  00:18:31             8514  ARISTA01T2
```


tests/bgp_commands_test.py
```diff
@@ -25,32 +25,32 @@
 Peer groups 4, using 256 bytes of memory
 
 
-Neighbhor      V     AS    MsgRcvd    MsgSent    TblVer    InQ    OutQ  Up/Down    State/PfxRcd    NeighborName
```
yejianquan, line 28: Great find, but could you please do global search 'Neighbhor' in sonic-mgmt to see whether we already uses the wrong magic string?
If so, let's do compatible support first in sonic-mgmt for the correction

cyw233: Good idea! I know a few places are using 'Neighbhor' in sonic-mgmt. Will raise PR to fix them as well.

abdosi (approved):

cyw233:
Created a PR in `sonic-mgmt` to sync with this typo fix: https://github.com/sonic-net/sonic-mgmt/pull/13395

yejianquan:
Hi @gechiang , do you think we need to do compatibility on our production scripts?
I don't know whether there's a possibility that, they just use 'Neighbhor' to check the BGP state.
In that case, if we correct the typo, maybe affect something.

gechiang:
We should not make this change.
@zhangyanzhao asked me to make this change before and I did not think it is good idea to change it after this has been out like this misspelled for so many years.  Many community users may already adopt to this miss spelled word into their scripts and if we change this at this point, it will impact many of those automation scripts.
We should revert this change and just live with the way it is...

bingwang-ms:
Based on the discussion, I removed the label for 202405 branch. Please re-add it if you think it's still required.
But it's a good catch!

PR #3389, opened 2024-06-28
[DPB]Fix return code in case of failure

dgsudharsan:
#### What I did
Fix return code when DPB command fails. Return of 0 indicates success which might be a problem to detect failures using automation

#### How I did it
Updated return code to 1

#### How to verify it
Basic manual testing


dgsudharsan:
@bingwang-ms Can you please review this? This is similar to  https://github.com/sonic-net/sonic-utilities/pull/3357

bingwang-ms (approved):

PR #3396, opened 2024-07-03
[Reboot] remove exec from the platform_reboot call when executing reboot in kdump kernel

DennisChiuEC:
#### What I did
Remove exec from the platform_reboot call to handle any hang issue during reboot in kdump kernel

#### How to verify it
Trigger kernel panic after uninstalling the platform driver and check if the device can reboot without hang.


DennisChiuEC:
reset the inactivity timer

PR #3413, opened 2024-07-11
[Draft] Add remote compatible show ip bgp and -n all for show ip bgp network

BYGX-wcr:
#### What I did

I enable the direct execution of "show ip bgp" on chassis supervisor and "-n all" option for "show ip bgp network" on multi-asic machine

#### How I did it
1. I modified the show ip bgp command to detect whether the current machine is a supervisor. If so, it will invoke rexec.cli to remotely exec the command on all linecards.
2. I add code in show ip bgp network command to pull the list of namespaces of the current device and run the command for each namespace.

#### How to verify it
On chassis supervisor, run "show ip bgp ..."
On a multi-asic device, run "show ip bgp network -n all"

### DIFF
--- a/show/bgp_frr_v4.py
+++ b/show/bgp_frr_v4.py
@@ -142,7 +142,7 @@ def network(ipaddress, info_type, namespace):
     if namespace == "all":
         if multi_asic.is_multi_asic():
             for ns in multi_asic.get_namespace_list():
-                click.echo("======== namespace {} ========".format(ns))
+                click.echo("\n======== namespace {} ========".format(ns))
                 output = bgp_util.run_bgp_show_command(command, ns)
                 click.echo(output.rstrip('\n'))
         else:
--- a/show/bgp_frr_v4.py
+++ b/show/bgp_frr_v4.py
@@ -21,7 +21,8 @@ def bgp():
     """Show IPv4 BGP (Border Gateway Protocol) information"""
     if device_info.is_supervisor():
         # if the device is a chassis, the command need to be executed by rexec
-        click.echo("Since the current device is a chassis supervisor, this command will be executed remotely on all linecards")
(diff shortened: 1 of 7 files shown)


show/main.py
```diff
@@ -1187,7 +1187,11 @@ def protocol(verbose):
     ip.add_command(bgp)
     from .bgp_frr_v6 import bgp
     ipv6.add_command(bgp)
-
+elif device_info.is_supervisor():
```
wenyiz2021, line 1190: you don't need to add supervisor for bgp setup. supervisor in chassis will not have bgps

wenyiz2021: nvm, you are adding this cmd on sup to execute for all linecards

show/bgp_frr_v4.py
```diff
@@ -102,10 +111,16 @@ def neighbors(ipaddress, info_type, namespace):
 def network(ipaddress, info_type, namespace):
     """Show IP (IPv4) BGP network"""
 
-    if multi_asic.is_multi_asic() and namespace not in multi_asic.get_namespace_list():
-        ctx = click.get_current_context()
-        ctx.fail('-n/--namespace option required. provide namespace from list {}'\
-            .format(multi_asic.get_namespace_list()))
+    namespace = namespace.strip()
+    if multi_asic.is_multi_asic():
```
wenyiz2021, line 115: I know you are just keeping original logic here, that if it is multi-asic, and namespace is not provided, error out.
but I think original logic may not be updated, can you change to: if multi asic, and namespace is not provided, print for all asics?
refer to: https://github.com/sonic-net/sonic-utilities/pull/2427
another pros for this is, you are running this cmd on sup, if one chassis has both multi/single-asic LCs, `show ip bgp sum` will output for all LCs, instead of having multi-asic LC error out, in this case we could have at least cmd working for both multi/single-asic.

PR #3424, opened 2024-07-15
[Bug Fix] Fix disk check test and drops group test

BYGX-wcr:
#### What I did

I fixed a problem that would cause disk_check_test and drops_group_test to fail when running the tests twice.

#### How I did it

tests/drop_groups_test.py: In the setup_class() of TestDropCounters class, I invoked a function to remove cache files.
tests/disk_check_test.py: In the teardown_class() of TestDiskCheck class, I run a command to remove /tmp/tmp* files.

#### How to verify it

Run the tests in the slave containers several times repeatedly and see if they passed. Before adding these fixes, the tests should fail starting from the second time.

### DIFF
--- a/tests/drops_group_test.py
+++ b/tests/drops_group_test.py
@@ -3,6 +3,7 @@ import sys
 
 import shutil
 from click.testing import CliRunner
+from utilities_common.cli import UserCache
 
 test_path = os.path.dirname(os.path.abspath(__file__))
 modules_path = os.path.dirname(test_path)
@@ -12,7 +13,6 @@ sys.path.insert(0, modules_path)
 
 import show.main as show
 import clear.main as clear
-from utilities_common.cli import UserCache
 
 expected_counter_capabilities = """\
 Counter Type           Total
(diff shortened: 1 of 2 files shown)


tests/drops_group_test.py
```diff
@@ -97,14 +98,17 @@
...
+    cache = UserCache("dropstat")
+    cache.remove_all()
 
 class TestDropCounters(object):
     @classmethod
     def setup_class(cls):
         print("SETUP")
-        if os.path.exists(dropstat_path):
-            shutil.rmtree(dropstat_path)
+        remove_tmp_dropstat_file()
```
qiluo-msft, line 111: >remove_tmp_dropstat_file [](http://example.com/codeflow?start=8&length=24)

do you still need remove dir `dropstat_path`?

BYGX-wcr: I don't think so. The dropstat_path seems to refer to a non-relevant location.

qiluo-msft (approved):

PR #3426, opened 2024-07-16
Fix the Bug of SONiC Extension Packages Migration

PeterTSW-EC:
#### What I did
* SONiC updates the format of the CLI section in the manifest.json on 17/5/2023. ([Link](https://github.com/sonic-net/SONiC/blob/master/doc/sonic-application-extension/sonic-application-extention-hld.md#manifest-path-7))
* The implementation was merged to master, 202311, or later branches. ([Link](https://github.com/sonic-net/sonic-utilities/pull/2753))
* However, the [PR #2753 Comment](https://github.com/sonic-net/sonic-utilities/pull/2753#issuecomment-1678604198) reported that an error occurred when the CLI migration.
* The root cause is occurred while a CLI field (show/config/clear) is empty ('').
* This PR corrects the manifest.json parser that leaves the empty list ([]) instead of a list with an empty string (['']) when the CLI field is empty ('').

#### How I did it
* Update the manifest.py, and add a condition that returns an empty list ([]) when the CLI field is an empty string (''); Otherwise, return a list with a single string path (['/xxx/xxx']). 

#### How to verify it
Execute the command "sudo sonic-package-manager install -y --from-tarball ./docker-test.gz", or install a new version of the SONiC image. (It executes the "sonic-package-manager migrate" during the installation. )


PeterTSW-EC:
The [PR #2753](https://github.com/sonic-net/sonic-utilities/pull/2753#issuecomment-1678604198)'s comment was mentioned about the same issue.

PR #3429, opened 2024-07-18
[202305] Input check for timeout in generate_dump (#2925)

maipbui:
Cherry-pick #2925
ADO 28548370
#### What I did
Added input check for argument "-t" in generate_dump script

#### How I did it
Made sure only integer values can be received for this argument

#### How to verify it
Call generate_dump -t with non-integer values


StormLiangMS (approved):
LGTM

PR #3437, opened 2024-07-20
Fix show techsupport date issue

ganglyu:
#### What I did
Show techsupport is designed to collect logs and core files since given date.
I find that some core files are missing when given date is relative, for example "5 minutes ago".
Microsoft ADO: 28737486

#### How I did it
Create the reference file at the start of the script, and don't update it in find_files.

#### How to verify it
Run end to end test: show_techsupport/test_auto_techsupport.py


wen587 (approved):
ltgm

PR #3439, opened 2024-07-23
Fix kexec_unload failure on secure boot enabled platforms

vivekrnv:
#### Why I did

kexec -u on systems with SB enabled is failing

Fixes https://github.com/sonic-net/sonic-buildimage/issues/18007

```
root@sn5600:/home/admin# mokutil --sb-state
SecureBoot enabled
root@sn5600:/home/admin# kexec -u
kexec unload failed: Permission denied

[ 3443.199576] ima: impossible to appraise a kernel image without a file descriptor; try using kexec_file_load syscall.
```

#### How to verify it

Use the -a argument with kexec 

```
 -a, --kexec-syscall-auto  Use file based syscall for kexec and fall
                      back to the compatibility syscall when file based
                      syscall is not supported or the kernel did not
                      understand the image
```

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -147,7 +147,7 @@ function clear_boot()
     # common_clear
     debug "${REBOOT_TYPE} failure ($?) cleanup ..."
 
-    /sbin/kexec -u || /bin/true
+    /sbin/kexec -u -a || /bin/true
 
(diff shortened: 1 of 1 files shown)


davidpil2002 (approved):
appoved

saiarcot895 (approved):

PR #3455, opened 2024-07-29
Adding j2c+ platforms for supporting NOKIA devices in GCU validator

okaravasi:
Currently file generic_config_updater/gcu_field_operation_validators.conf.json is used from test cases to verify if the platform under test allowes updates to config paths:

> /PFC_WD 
> /BUFFER_POOL 
> /BUFFER_PROFILE 
> /WRED_PROFILE 

Currently pltaforms for j2c+ chips are not included with as a result the tests report erroneously status for the relevant test cases that verify these paths.
The current PR adds these platforms to support testing on this paths also as expected.
More specifically a new entry for j2c+ chips was added under broadcom entry. For this new entry, 2 NOKIA hwskus are added.

#### How to verify it
Ran existing GCU tests in newly added platforms and verified that the test pass. Ran also in differnet platform than the added and verified there are no unexpecetd failures.


okaravasi:
Changes for this PR have been integrated into [PR #3495](https://github.com/sonic-net/sonic-utilities/pull/3495) and will be handled there. This ticket can be closed.

PR #3459, opened 2024-08-01
Auditd Enhancement CLI

maipbui:
HLD https://github.com/sonic-net/SONiC/pull/1713
Microsoft ADO **(number only)**: 27951427
#### What I did
Add Auditd CLI design

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -7035,27 +7035,27 @@ def audit():
 @audit.command('enable')
 @click.pass_context
 def enable(ctx):
-    """Enbale the audit enhancement feature """
+    """Enable the audit enhancement feature"""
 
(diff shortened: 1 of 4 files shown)


config/main.py
```diff
@@ -7019,6 +7019,49 @@ def del_ntp_server(ctx, ntp_ip_address):
...
+#
+@audit.command('enable')
+@click.pass_context
+def enable(ctx):
+    """Enable the security auditing feature"""
+
+    config_db = ValidatedConfigDBConnector(ConfigDBConnector())
+    config_db.connect()
+    try:
+        config_db.mod_entry("AUDIT", "config", {"enabled": "true"})
```
FengPan-Frank, line 7043: do we need to handle config node AUDIT not existing case?

PR #3476, opened 2024-08-06
[SNMP:Contact] Fix for config replace failure with SNMP Contact issue

Verma-Anukul:
==> Issue:
• Config replace/yang-validation fails for SNMP contact configuration.
• In SNMP contact CLI, email field is mandatory
Related to https://github.com/sonic-net/sonic-buildimage/issues/12620

==> Root Cause:
• sonic-snmp.yang defines leaf Contact, but sonic-utilities (CLI backend) writes to ConfigDB with key {contact_name}. So YANG validation fails when the key is {contact_name} and not the string "Contact"
• CLI current implementation is expecting two mandatory arguments

==> Fix:
• sonic-utilities (CLI backend) code is updated to use Yang leaf name "Contact" string as key and user passed input as "value"
• CLI implementation (sonic-utilities) is changed to expect only a single string, where user can pass any string of choice (name, email, phone, address or combination of these). Accordingly, "show" CLIs will display only "contact" instead of "name" & "email"

Migration script is updated to take care of config_db migration
Tests:

Test CLIs manually
Test config replace
Test config migration

PRs :
sonic-buildimage - https://github.com/sonic-net/sonic-buildimage/pull/19834
sonic-utilities -  https://github.com/sonic-net/sonic-utilities/pull/3476
SONiC (docs) -  https://github.com/sonic-net/SONiC/pull/1774

Signed-off-by: Anukul Verma <anukverm@cisco.com>


Verma-Anukul:
@lguohan @qiluo-msft @venkatmahalingam @dgsudharsan @SuvarnaMeenakshi @jleveque @liushilongbuaa @xincunli-sonic

Please help in reviewing and merging this change

config/main.py
```diff
@@ -4053,41 +4053,27 @@ def replace_community(db, current_community, new_community):
...
-    """ Add snmp contact name and email """
-    snmp = db.cfgdb.get_table("SNMP")
+def add_contact(db, contact):
+    """ Add snmp contact"""
+    config_db = ValidatedConfigDBConnector(db.cfgdb)
+    if isinstance(contact, tuple):
+        contact = " ".join(contact)
+    snmp = config_db.get_table("SNMP")
     try:
         if snmp['CONTACT']:
```
qiluo-msft, line 4066: >if snmp['CONTACT']: [](http://example.com/codeflow?start=8&length=19)

Could you just check? this way you do not need to catch the KeyError exception.
```
if "CONTACT" not in snmp:
```

config/main.py
```diff
@@ -4098,67 +4084,59 @@ def add_contact(db, contact, contact_email):
...
+        contact = " ".join(contact)
     snmp = db.cfgdb.get_table("SNMP")
     try:
-        if not contact in (list(snmp['CONTACT'].keys()))[0]:
-            click.echo("SNMP contact {} is not configured".format(contact))
-            sys.exit(1)
-        else:
-            db.cfgdb.set_entry('SNMP', 'CONTACT', None)
-            click.echo("SNMP contact {} removed from configuration".format(contact))
+        if contact == snmp['CONTACT']['Contact']:
```
qiluo-msft, line 4096: >if contact == snmp['CONTACT']['Contact']: [](http://example.com/codeflow?start=8&length=41)

Could you just check below? you do not need to catch KeyError exception.

```
if "CONTACT" in snmp and 'Contact' in snmp['CONTACT'] and contact == snmp['CONTACT']['Contact']:
``` #Closed

File: config/main.py [hunk 3]
```diff
@@ -4098,67 +4084,59 @@ def add_contact(db, contact, contact_email):
...
                 clicommon.run_command(['systemctl', 'reset-failed', 'snmp.service'], display_cmd=False)
                 clicommon.run_command(['systemctl', 'restart', 'snmp.service'], display_cmd=False)
             except SystemExit as e:
                 click.echo("Restart service snmp failed with error {}".format(e))
                 raise click.Abort()
+        else:
+            click.echo("SNMP Contact {} does not exist.".format(contact))
+            sys.exit(1)
     except KeyError:
         if "CONTACT" not in snmp.keys():
```
qiluo-msft, line 4114: >if "CONTACT" not in snmp.keys(): [](http://example.com/codeflow?start=8&length=32)

another possibility is that 'Contact' is not in snmp['CONTACT']. If this happen, the exception is silently swallowed. #Closed

config/main.py
```diff
@@ -4098,67 +4084,59 @@ def add_contact(db, contact, contact_email):
...
+    snmp = config_db.get_table("SNMP")
     try:
-        current_snmp_contact_name = (list(snmp['CONTACT'].keys()))[0]
-        if current_snmp_contact_name == contact:
-            current_snmp_contact_email = snmp['CONTACT'][contact]
-        else:
-            current_snmp_contact_email = ''
-        if contact == current_snmp_contact_name and contact_email == current_snmp_contact_email:
-            click.echo("SNMP contact {} {} already exists".format(contact, contact_email))
+        snmp_contact = snmp['CONTACT']['Contact']
```
qiluo-msft, line 4129: >snmp['CONTACT']['Contact'] [](http://example.com/codeflow?start=23&length=26)

Could you just check below? you do not need to catch KeyError exception.

```
if "CONTACT" in snmp and 'Contact' in snmp['CONTACT'] and contact == snmp['CONTACT']['Contact']:
``` #Closed

File: config/main.py [hunk 5]
```diff
@@ -4098,67 +4084,59 @@ def add_contact(db, contact, contact_email):
...
+        if contact in snmp_contact:
```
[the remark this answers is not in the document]
qiluo-msft: this is a valid comment/bug. Please fix it.

config/main.py
```diff
@@ -4167,9 +4145,8 @@ def modify_contact(db, contact, contact_email):
                 click.echo("Restart service snmp failed with error {}".format(e))
                 raise click.Abort()
     except KeyError:
-        if "CONTACT" not in snmp.keys():
-            click.echo("Contact name {} is not configured".format(contact))
-            sys.exit(3)
+        click.echo("Cannot modify SNMP Contact.  You must use 'config snmp contact add' command.'")
```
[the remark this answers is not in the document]
Verma-Anukul: Done

config/main.py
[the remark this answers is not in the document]
Verma-Anukul: Done

config/main.py
Verma-Anukul: We need to fold check here
1. If Data exists , else return error to add first
2. If data is same as new one, else return error tht already same config exists

config/main.py
Verma-Anukul: This is delete case, if Contact field is not having any value then yes we are not returning any error

config/main.py
Verma-Anukul: Two fold check is done to return different errors in these two scenarios

1. Contact is not configured and user is trying to delete
2. Different contact is configured compare to what user is deleting

config/main.py
Verma-Anukul: Thanks for the suggestion. In this case, seems it is using try-except intentionally since accessing the key directly (snmp['CONTACT']) reflects the assumption that the key should generally exist, and we treat its absence as an exceptional condition. This also avoids an extra lookup that if would require.

Also it was existing code, so i didn't change it and same way of handling is followed for other clis

Let me know if there's a specific reason to prefer the in check here — happy to revisit if needed.

Verma-Anukul:
@lguohan @qiluo-msft @venkatmahalingam @dgsudharsan @SuvarnaMeenakshi @jleveque @liushilongbuaa @xincunli-sonic @vaibhavhd 

Please help in reviewing and merging this change

tests/db_migrator_input/config_db/snmp_contact_input.json
```diff
@@ -0,0 +1,8 @@
+{
+    "SNMP|CONTACT": {
+        "test-name": "test-email@test.com"
+    },
+    "VERSIONS|DATABASE": {
+            "VERSION": "version_202505_01"
```
qiluo-msft, line 6: >             [](http://example.com/codeflow?start=0&length=12)

could you use 8 spaces indentation this line?

config/main.py
qiluo-msft: My proposal is to simplify the code, because there is a check `if "CONTACT" not in snmp.keys()` inside except anyway.

try-except inside a except block is weird.

scripts/db_migrator.py
```diff
@@ -900,6 +900,20 @@ def migrate_aaa(self):
             if keys:
                 self.configDB.delete(self.configDB.CONFIG_DB, authorization_key)
 
+    def migrate_config_db_snmp_contact(self):
+        """
+        Migrate SNMP": { "CONTACT": { "joe": "joe@email.com"}} to
```
qiluo-msft, line 905: >Migrate [](http://example.com/codeflow?start=8&length=7)

Will you also change the file dockers/docker-snmp/snmpd.conf.j2?

qiluo-msft:
> **config snmp contact add/del/modify**

I think we need to step back and rethink about command line interfaces first.

If the system only allow 0 or 1 contact, we just need `config snmp contact set/del`.

1. `config snmp contact set <contact>` will create a new contact if non contact before, or overwrite contact if contact exist.
2. `config snmp contact del` do not need existing contact as command line argument.

---
Refers to: doc/Command-Reference.md:10689 in 7eb7fcb. [](commit_id = 7eb7fcb9e07f37ff036ac71ff96c522ab99ec0f2, deletion_comment = False)

PR #3484, opened 2024-08-09
Add timeout for rexec's get_password

BYGX-wcr:
#### What I did

I added a timeout setting for the get_password function in rexec module so that automatic pipelines that does not expect password input will not get blocked. The current timeout setting is 10 sec.

#### How I did it

Add a SIGALRM signal before waiting for password input. 

#### How to verify it

Run a "show ip bgp summary" on SONiC Chassis Supversior and does not input password until it times out.

Before adding such a mechanism, if you do not input password when run "rexec -c <cmd>" you will be blocked at the following output:
```
Since the current device is a chassis supervisor, this command will be executed remotely on all linecards
Password for username 'XXX':
```

After adding such a mechanism, if you do not input password when run "rexec -c <cmd>" you will see a timeout message after 10 seconds.
```
Since the current device is a chassis supervisor, this command will be executed remotely on all linecards
Password for username 'XXX': 
Aborted! Timeout when waiting for password input.
```

### DIFF
--- a/rcli/utils.py
+++ b/rcli/utils.py
@@ -1,5 +1,5 @@
 import click
-from getpass import getpass
+import getpass
 import os
 import signal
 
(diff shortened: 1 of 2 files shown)


gechiang (approved):
LGTM.

gechiang:
@BYGX-wcr , please do get additional sign off from @judyjoseph,  @arlakshm and @qiluo-msft before merge this PR.
Thanks!

PR #3486, opened 2024-08-11
Move from bootctl to mokutil when checking for Secure Boot status

DavidZagury:
#### What I did
Moved to use mokutil instead of bootctl as bootctl is no longer available in Bookworm.
This affected reboot scripts, and upgrade scenario.

#### How I did it
Change calls to _bootctl status_ with _mokutil --sb-state_

#### How to verify it
After fixing the scripts to check reboot:

root@sn5600:/home/admin# soft-reboot

SECURE_UPGRADE_ENABLED=1
[[ CHECK_SECURE_UPGRADE_ENABLED -ne 0 ]]
load_kernel_secure
invoke_kexec -s
packet_write_wait: port 22: Broken pipe
admin@sn5600:~$ show reboot-cause
User issued 'soft-reboot' command [User: admin, Time: Tue Jul 23 11:06:43 PM UTC 2024]

### DIFF
--- a/scripts/soft-reboot
+++ b/scripts/soft-reboot
@@ -93,7 +93,7 @@ function clear_lingering_reboot_config()
     if [[ -f ${WARM_DIR}/${REDIS_FILE} ]]; then
         mv -f ${WARM_DIR}/${REDIS_FILE} ${WARM_DIR}/${REDIS_FILE}.${TIMESTAMP} || /bin/true
     fi
-    /sbin/kexec -u || /bin/true
+    /sbin/kexec -u -a || /bin/true
 }
 
 SCRIPT=$0
@@ -147,9 +147,17 @@ function setup_reboot_variables()
     fi
 }
 
+function invoke_kexec() {
+    /sbin/kexec -l "$KERNEL_IMAGE" --initrd="$INITRD" --append="$BOOT_OPTIONS" $@
+}
+
 function load_kernel() {
     # Load kernel into the memory
-    /sbin/kexec -l "$KERNEL_IMAGE" --initrd="$INITRD" --append="$BOOT_OPTIONS"
+    invoke_kexec -a
+}
+
+function load_kernel_secure() {
+    invoke_kexec -s
 }
 
 function reboot_pre_check()
@@ -215,7 +223,14 @@ stop_sonic_services
 
 clear_lingering_reboot_config
 
-load_kernel
+# check if secure boot is enabled 
+CHECK_SECURE_UPGRADE_ENABLED=0
+SECURE_UPGRADE_ENABLED=$(mokutil --sb-state 2>/dev/null | grep -c "enabled") || CHECK_SECURE_UPGRADE_ENABLED=$?
+if [[ CHECK_SECURE_UPGRADE_ENABLED -ne 0 ]]; then
+    load_kernel
+else
+    load_kernel_secure
+fi
 
 # Update the reboot cause file to reflect that user issued 'reboot' command
 # Upon next boot, the contents of this file will be used to determine the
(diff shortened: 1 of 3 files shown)


scripts/soft-reboot
```diff
@@ -93,7 +93,7 @@ function clear_lingering_reboot_config()
     if [[ -f ${WARM_DIR}/${REDIS_FILE} ]]; then
         mv -f ${WARM_DIR}/${REDIS_FILE} ${WARM_DIR}/${REDIS_FILE}.${TIMESTAMP} || /bin/true
     fi
-    /sbin/kexec -u || /bin/true
```
Yarden-Z, line 96: Why is this change required?

sonic_installer/bootloader/grub.py
```diff
@@ -164,7 +164,8 @@ def is_secure_upgrade_image_verification_supported(self):
             if ! [ -n "$(ls -A /sys/firmware/efi/efivars 2>/dev/null)" ]; then
                 mount -t efivarfs none /sys/firmware/efi/efivars 2>/dev/null
             fi
-            SECURE_UPGRADE_ENABLED=$(bootctl status 2>/dev/null | grep -c "Secure Boot: enabled")
+            SECURE_UPGRADE_ENABLED=$(mokutil --sb-state 2>/dev/null | \
+                grep -c "enabled") || CHECK_SECURE_UPGRADE_ENABLED=$?
```
Yarden-Z, line 168: From where are you getting the "CHECK_SECURE_UPGRADE_ENABLED" variable?

scripts/soft-reboot
```diff
@@ -93,7 +93,7 @@ function clear_lingering_reboot_config()
...
+    /sbin/kexec -u -a || /bin/true
```
Yarden-Z, line 96: Why is this change required?

sonic_installer/bootloader/grub.py
DavidZagury: @Yarden-Z  Remove as unneeded

qiluo-msft:
Could you link this PR to HLD PR?

scripts/soft-reboot
qiluo-msft: I have the same question. If this is valid and general bug fix, we need to evaluate the backport needs. Could you separate it into a standalone PR?

scripts/soft-reboot
```diff
@@ -147,9 +147,17 @@ function setup_reboot_variables()
...
 }
 
+function invoke_kexec() {
+    /sbin/kexec -l "$KERNEL_IMAGE" --initrd="$INITRD" --append="$BOOT_OPTIONS" $@
+}
+
 function load_kernel() {
     # Load kernel into the memory
-    /sbin/kexec -l "$KERNEL_IMAGE" --initrd="$INITRD" --append="$BOOT_OPTIONS"
+    invoke_kexec -a
```
qiluo-msft, line 156: >-a [](http://example.com/codeflow?start=17&length=2)

why changing? If this is valid and general bug fix, we need to evaluate the backport needs. Could you separate it into a standalone PR?

scripts/soft-reboot
vivekrnv: kexec -u on systems with SB enabled is failing. Already fixed for WR here: https://github.com/sonic-net/sonic-utilities/pull/3439. 

Since we are updating soft-reboot to work on Secure boot systems. i've also added this change

scripts/soft-reboot
vivekrnv: Similar to above comment.

ycoheNvidia (approved):
Looks fine to me

ycoheNvidia (approved):

vivekrnv (approved):

liat-grozovik:
@qiluo-msft can you please review comment feedback?

qiluo-msft (approved):

PR #3490, opened 2024-08-15
Exit early if YANG validation fails in Golden Config

wen587:
ADO: 29254166
#### What I did
Exit early if golden config fails YANG validation
#### How I did it
Check before stop service in load_minigraph
#### How to verify it
Unit test


ganglyu (approved):

config/main.py
```diff
@@ -1983,8 +1983,23 @@ def load_minigraph(db, no_service_restart, traffic_shift_away, override_config,
...
+        config_db = cfgdb_clients[DEFAULT_NAMESPACE]
+
+        cm = None
+        try:
+            cm = ConfigMgmt(configdb=config_db)
+            cm.validateConfigData()
+        except Exception as ex:
+            log.log_warning("The existing config fails YANG validation. Alerting: {}".format(ex))
+
         config_to_check = read_json_file(golden_config_path)
```
xincunli-sonic, line 1997: >golden_config_path [](http://example.com/codeflow?start=41&length=18)

what if golden config is multiasic flavor? Like 

[
  {
    "localhost": {
      ...
    }
  },
  {
    "asic0": {
      ...
    }
  },
  {
    "asic1": {
      ...
    }
  }
]

xincunli-sonic: Recommend validate function without cfgdb_clients.

if multi_asic.is_multi_asic():
            for asic in multi_asic.get_namespace_list():
                target_config = config_to_check.pop(asic)
                if not SonicYangCfgDbGenerator().validate_config_db_json(target_config):
                    return False

xincunli-sonic:
>         table_hard_dependency_check(host_config)

This part also needs add other asics' config check?

---
Refers to: config/main.py:2007 in 16481df. [](commit_id = 16481df500df03bb503feb67396d81007f512b07, deletion_comment = False)

wen587:
> > ```
> >     table_hard_dependency_check(host_config)
> > ```
> 
> This part also needs add other asics' config check?
> 
> Refers to: config/main.py:2007 in [16481df](https://github.com/sonic-net/sonic-utilities/commit/16481df500df03bb503feb67396d81007f512b07). [](commit_id = [16481df](https://github.com/sonic-net/sonic-utilities/commit/16481df500df03bb503feb67396d81007f512b07), deletion_comment = False)

Currently it only has dependency check about AAA which is only exist in host. It is weakened version of YANG validation based on PROD issue. I think after yang validation enabled, we can remove it or just add parts that YANG could not covered.

config/main.py
wen587: I want to skip Golden Config for mutliaisc right now. It will not report error if multiasic flavor.
The reason is that we have 100% table covered in single asic, and I have monitored for a while and single asic Golden Config is always YANG validated. So we have confidence it will not exit early which block the PROD flow.
But we don't have such confidence in multiasic.

config/main.py
```diff
@@ -1983,8 +1983,23 @@ def load_minigraph(db, no_service_restart, traffic_shift_away, override_config,
...
+        cfgdb_clients = db.cfgdb_clients
+        # Pick the host config in case of multi-asic
+        config_db = cfgdb_clients[DEFAULT_NAMESPACE]
+
+        cm = None
+        try:
+            cm = ConfigMgmt(configdb=config_db)
+            cm.validateConfigData()
+        except Exception as ex:
+            log.log_warning("The existing config fails YANG validation. Alerting: {}".format(ex))
```
FengPan-Frank, line 1995: Do we need exit for this exception? or is there real case that validateConfigData() gets failed and validate_config_by_cm() gets succeed?

config/main.py
xincunli-sonic: Do we have ADO track the multiasic change?

config/main.py
```diff
@@ -1983,8 +1983,23 @@ def load_minigraph(db, no_service_restart, traffic_shift_away, override_config,
...
-        # Dependency check golden config json
+        cfgdb_clients = db.cfgdb_clients
+        # Pick the host config in case of multi-asic
+        config_db = cfgdb_clients[DEFAULT_NAMESPACE]
+
+        cm = None
+        try:
+            cm = ConfigMgmt(configdb=config_db)
+            cm.validateConfigData()
+        except Exception as ex:
```
qiluo-msft, line 1994: >except [](http://example.com/codeflow?start=8&length=6)

I see there are several options or design decisions. How about we move the options into a conf file, and then we can provide a default one. The benefit is possibility of runtime override if any urgency happen, or any potential phased rollout (even we do not plan for it).
1. yang_invalid_final_config_throw: yes/no
2. yang_invalid_final_config_logging: yes/no
3. yang_invalid_final_config_loglevel: WARN/ERROR/etc
4. namespaces: default/all
5. etc

qiluo-msft, line 1994: >Exception [](http://example.com/codeflow?start=15&length=9)

Could you use less-general exception type?

config/main.py
wen587: [Task 29182008](https://msazure.visualstudio.com/One/_workitems/edit/29182008): Fail the Golden Config validation early in T2 scenario

config/main.py
wen587: validateConfigData will validate current running config. validate_config_by_cm here is to validate golden config.
We cannot assume the running config is already yang validated. We only make sure it is yang validted after load_minigraph+override by golden config.

config/main.py
wen587: I think we want to block the process since currently Golden Config is already yang validated based on recent deployement. If it fail, there must be some modification or corner case. And we can use this failure to push the fix from feature owner. 
If we let it pass, we will leave bad config into PROD, which will take long time fix. Is that what we want?

config/main.py
wen587: Changed to sonic_yang to avoid CONFIG_MGMT init failure

xincunli-sonic (approved):
LGTM

ganglyu (approved):

FengPan-Frank (approved):

PR #3513, opened 2024-09-02
Upgrade pyroute2 and improve cli response time

vivekrnv:
#### What I did

- Older pyroute2 depends on distutils. Thus upgrade the version to latest to improve import time. A similar issue for natsort is reported here https://github.com/sonic-net/sonic-buildimage/issues/17246. However pyroute2 import is still heavy in bookworm and thus every CLI command is slow compared to bullseye. 

```
<Bullseye>
root@msn2700:/home/admin# time python3 -c "import pyroute2"
real	0m0.378s
user	0m0.308s
sys	0m0.060s

<Before upgrade>
root@msn2700:/home/admin# time python3 -c "import pyroute2"
real        0m0.707s
user        0m0.425s
sys        0m0.097s

<After upgrade>
root@msn2700:/home/admin#  time python3 -c "import pyroute2"
real	0m0.511s
user	0m0.433s
sys	0m0.075s
```

- To fix this, i've delayed the pyroute2 import into the method where it is actually used, this has an improvement of 0.4 sec for all the CLI commands on slower CPU devices

```
root@msn2700:/home/admin#  time python3 -c "import utilities_common.cli as clicommon"
real	0m0.693s
user	0m0.579s
sys	0m0.109s

root@msn2700/home/admin#  time python3 -c "import utilities_common.cli as clicommon"
real	0m0.363s
user	0m0.271s
sys	0m0.072s
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -249,7 +249,7 @@ setup(
         'pexpect>=4.8.0',
         'semantic-version>=2.8.5',
         'prettyprinter>=0.18.0',
-        'pyroute2>=0.5.14, <0.6.1',
+        'pyroute2==0.7.12',
         'requests>=2.25.0, <=2.31.0',
         'tabulate==0.9.0',
         'toposort==1.6',
--- a/utilities_common/multi_asic.py
+++ b/utilities_common/multi_asic.py
@@ -3,7 +3,6 @@ import functools
 
 import click
 import netifaces
-import pyroute2
 from natsort import natsorted
 from sonic_py_common import multi_asic, device_info
 from utilities_common import constants
@@ -170,6 +169,7 @@ def multi_asic_args(parser=None):
     return parser
 
 def multi_asic_get_ip_intf_from_ns(namespace):
+    import pyroute2
     if namespace != constants.DEFAULT_NAMESPACE:
         pyroute2.netns.pushns(namespace)
     interfaces = natsorted(netifaces.interfaces())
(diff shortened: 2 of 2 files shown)


saiarcot895 (dismissed):

vivekrnv:
@qiluo-msft Requesting waiver to ignore the coverage failure. 

I tried adding an UT to cover the multi_asic_get_ip_intf_from_ns and multi_asic_get_ip_intf_addr_from_ns methods (https://github.com/sonic-net/sonic-utilities/commit/50e311ade00e8f78275f31b8e1e4fda3d3482d2a) .  

But these are monkey patched here https://github.com/sonic-net/sonic-utilities/blob/master/tests/mock_tables/mock_multi_asic.py#L93. 

So i couldn't call the actual method to capture it under the coverage scan. Besides, my changes just change the location of import and so no major functionality change that warrants a UT

keboliu (approved):

saiarcot895 (approved):

qiluo-msft (approved):

qiluo-msft:
Thanks for the UT effort! Could you further explore if you can just disable multi_asic_get_ip_intf_from_ns monkey patch in your specific testcase? We should not globally enable one function monkey patch, because it will prevent us to UT the the function itself.

---
In reply to: [2330382035](https://github.com/sonic-net/sonic-utilities/pull/3513#issuecomment-2330382035) [](http://example.com/codeflow?ancestors=2330382035)

vivekrnv:
> Thanks for the UT effort! Could you further explore if you can just disable multi_asic_get_ip_intf_from_ns monkey patch in your specific testcase? We should not globally enable one function monkey patch, because it will prevent us to UT the the function itself.
> 
> In reply to: [2330382035](https://github.com/sonic-net/sonic-utilities/pull/3513#issuecomment-2330382035)

Hi, i checked but monkey patching is global. once the file is imported and loaded, the original api is lost. 

I suppose this can be updated but changing this would also probably require some other Multi Asic UT's to also be updated and so i think this should be taken as a separate item

bingwang-ms:
@vivekrnv Can you raise a PR to do the cherry-pick to 202405 since automation is not working?

PR #3515, opened 2024-09-03
Revert "[wol] Implement wol command line utility"

w1nda:
we use compiled wol cli in and after 202411: https://github.com/sonic-net/sonic-buildimage/pull/19206

So, we reverts PR sonic-net/sonic-utilities#3048


doc/Command-Reference.md
```diff
@@ -222,8 +222,6 @@
 * [Static DNS Commands](#static-dns-commands)
   * [Static DNS config command](#static-dns-config-command)
   * [Static DNS show command](#static-dns-show-command)
-* [Wake-on-LAN Commands](#wake-on-lan-commands)
```
lizhijianrd, line 225: We should keep these doc.
Only revert the code.

w1nda: OK, will save the doc
https://github.com/sonic-net/sonic-utilities/pull/3516

lizhijianrd (dismissed):

lizhijianrd (approved):

PR #3566, opened 2024-10-03
[SmartSwitch] Extend reboot script for rebooting SmartSwitch

vvolam:
#### What I did
Extended reboot script for SmartSwitch cases to reboot entire SmartSwitch or a specific DPU

#### How I did it
Implemented changes according to [SmartSwitch Reboot flow HLD](https://github.com/sonic-net/SONiC/blob/605c3a56ac2717dbbb638433e7bb13054fc05a31/doc/smart-switch/reboot/reboot-hld.md)

#### How to verify it
- Verified the script on non-smart switch and didn't find any regressions. Also, script throws errors if any new smart switch related parameters are given by user.
- Verified on NVIDIA smart switch.


scripts/reboot
```diff
@@ -11,6 +11,8 @@ PLATFORM_UPDATE_REBOOT_CAUSE="platform_update_reboot_cause"
 REBOOT_CAUSE_FILE="/host/reboot-cause/reboot-cause.txt"
 PLATFORM_REBOOT_PRE_CHECK="platform_reboot_pre_check"
 REBOOT_TIME=$(date)
+PLATFORM_JSON_FILE="platform.json"
+PLATFORM_JSON_PATH="${DEVPATH}/${PLATFORM}/${PLATFORM_JSON_FILE}"
```
gpunathilell, line 15: The `${PLATFORM}` definition comes after the variable is being used [in this line ](https://github.com/sonic-net/sonic-utilities/blob/7792b3e9eb341e91b3255d189c98bfa71eb43587/scripts/reboot#L33). This has to be moved

scripts/reboot
```diff
@@ -176,9 +185,128 @@ function check_conflict_boot_in_fw_update()
     fi
 }
 
+# Function to retrieve DPU IP from CONFIG_DB
+function get_dpu_ip()
+{
+    local DPU_NAME=$1
+    dpu_ip=$(sonic-db-cli CONFIG_DB HGET "DHCP_SERVER_IPV4_PORT|bridge-midplane|${DPU_NAME}" "ips")
```
gpunathilell, line 192:
```suggestion
    dpu_ip=$(sonic-db-cli CONFIG_DB HGET "DHCP_SERVER_IPV4_PORT|bridge-midplane|${DPU_NAME}" "ips@")
```

scripts/reboot
```diff
@@ -176,9 +185,128 @@ function check_conflict_boot_in_fw_update()
...
+    fi
+    debug "$DPU_NAME GNMI port:$port"
+}
+
+# Function to get reboot status from DPU
+function get_reboot_status()
+{
+    local dpu_ip=$1
+    local port=$2
+    reboot_status=$(gnoi_client -target ${dpu_ip}:${port} -logtostderr -insecure -rpc RebootStatus)
```
gpunathilell, line 216: Would these `gnoi_client` commands be available outside the gnmi docker ? If not we need to use the docker exec command instead

scripts/reboot
```diff
@@ -192,6 +320,13 @@ function parse_options()
             f )
                 REBOOT_FLAGS+=" -f"
                 ;;
+            d )
+                REBOOT_DPU="yes"
+                DPU_MODULE_NAME="$optarg"
```
gpunathilell, line 325:
```suggestion
                DPU_MODULE_NAME="$OPTARG"
```

scripts/reboot
vvolam: that's a good catch. When testing, I did right thing and forgot to capture in code. Thanks!

scripts/reboot
```diff
@@ -215,6 +350,55 @@ function linecard_reboot_notify_supervisor()
...
+    for (( i=0; i<"$NUM_DPU"; i++ )); do
+        echo "Rebooting DPU module dpu$i"
+        reboot_dpu_module "dpu$i" &
+    done
+    wait
+}
+
+# Function to handle scenarios on smart switch
+function handle_smart_switch() {
+    if [ -f "$PLATFORM_JSON_PATH" ]; then
```
gpunathilell, line 366: This would set `SMART_SWITCH="yes"` only on the NPU and not on the DPUs

scripts/reboot_helper.py
```diff
@@ -0,0 +1,135 @@
...
+        log_err("Module {module_name} not found")
+        return False
+
+    except Exception as e:  # noqa: F841
+        log_err("Error occurred while rebooting module {module_name}: {repr(e)}")
+        return False
+
+
+def is_dpu():
+    """Check if script is running on DPU module"""
```
gpunathilell, line 96: This would return `True` only for NPU and not for DPUs, (to check if the current platform is a DPU platform I think we can use the same logic used in [docker_image_ctl.j2](https://github.com/sonic-net/sonic-buildimage/blob/a5a9dacad7fadd7fdd064858f2a586424ad4f766/files/build_templates/docker_image_ctl.j2#L375)

scripts/reboot
```diff
@@ -176,9 +185,128 @@ function check_conflict_boot_in_fw_update()
...
+
+function reboot_dpu_module()
+{
+    local DPU_NAME=$1
+    local DPU_INDEX=${DPU_NAME//[!0-9]/}
+
+    debug "User requested rebooting device ${DPU_NAME} ..."
+
+    # Retrieve DPU IP and GNMI port
+    dpu_ip=$(get_dpu_ip "${DPU_NAME}")
```
gpunathilell, line 253: Both `get_dpu_ip` and `get_gnmi_port` do not return anything

scripts/reboot_helper.py
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env python3
```
gpunathilell, line 1: Not sure if `reboot_helper.py` script is not copied to the target image on build

scripts/reboot
```diff
@@ -215,6 +350,55 @@ function linecard_reboot_notify_supervisor()
...
+            echo "Invalid, '-p' option not specified for a DPU"
+            exit ${EXIT_ERROR}
+        elif [[ "$is_dpu" != "True" && "$PRE_SHUTDOWN" == "yes" ]]; then
+            echo "Invalid '-p' option specified for a non-DPU"
+            exit ${EXIT_ERROR}
+        fi
+
+        # If not a DPU, reboot all DPUs in parallel
+        if [[ "$is_dpu" != "True" ]]; then
+            reboot_all_dpus "$NUM_DPU"
```
gpunathilell, line 397: If NPUs + DPUs need reboot we need to pass the `MODULE_REBOOT_SMARTSWITCH` to the module reboot function, if individual DPUs need reboot we use the `MODULE_REBOOT_DPU` as per reboot hld:

reboot(self, reboot_type):
```
Define new reboot_type as MODULE_REBOOT_DPU for DPU only reboot and MODULE_REBOOT_SMARTSWITCH for entire switch reboot.

```

scripts/reboot
```diff
@@ -215,6 +350,55 @@ function linecard_reboot_notify_supervisor()
...
+            echo "User requested to reboot the device ${DPU_MODULE_NAME}"
+            reboot_dpu_module "$DPU_MODULE_NAME"
+        else
+            echo "Invalid '-d' option specified for a non-smart switch"
+            exit ${EXIT_ERROR}
+        fi
+    fi
+
+    if [[ "$SMART_SWITCH" == "yes" ]]; then
+        is_dpu=$(python3 -c "import reboot_helper; reboot_helper.is_dpu()")
```
gpunathilell, line 384: The output of `is_dpu` should be printed

scripts/reboot
vvolam: This could be because the code might not be present for this in sonic image. I will check on the possible ETA for the changes in that case. Thank you!

scripts/reboot
```diff
@@ -176,9 +185,128 @@ function check_conflict_boot_in_fw_update()
...
+        fi
+    done
+
+    # Check if DPU exists and retrieve bus info
+    DPU_BUS_INFO=$(get_dpu_bus_info "${DPU_NAME}")
+
+    # Update STATE_DB and handle PCIe removal and rescan
+    sonic-db-cli state_db set "PCIE_DETACH_INFO|${DPU_NAME}" '{"dpu_id": "'${DPU_INDEX}'", "dpu_state": "detaching", "bus_info": "'${DPU_BUS_INFO}'"}'
+
+    echo 1 > /sys/bus/pci/devices/${DPU_BUS_INFO}/remove
```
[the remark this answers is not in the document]
vvolam: Yes, I will update on the pci_detach function but HLD to updated to not use pcie_reattach but use kernel sysfs mechanism to use for rescan. I have to remove the code come platform-common to reflect the same. Does vendor code do anything special in the pci_reattach() other than rescan?

scripts/reboot_helper.py
vvolam: Completed. Thank you!

scripts/reboot_helper.py
vvolam: updated. Thank you!

scripts/reboot
vvolam: got it.

scripts/reboot
vvolam: I will fix the HLD, it was not updated. Thank you!

gpunathilell: The reboot script should execute the reboot of DPUS in parallel. I think the differentiation between `MODULE_REBOOT_DPU` and `MODULE_REBOOT_SMARTSWITCH` is used by platforms to decide whether to wait for the dpus to finish rebooting or not (if we are rebooting individual DPUs -> `MODULE_REBOOT_DPU` -> wait for DPU to finish rebooting. If we are rebooting all DPUS + switch -> `MODULE_REBOOT_SMARTSWITCH`-> NPU will just issue reboot command to DPUs and not wait for the result as it is undergoing reboot itself ) Correct me if I am wrong

scripts/reboot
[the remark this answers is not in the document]
gpunathilell: Yes, in our case we have additional code to be executed along with kernel sysfs mechanism, is it okay to call the function if it is implemented, otherwise to use the sysfs calls directly?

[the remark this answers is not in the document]
vvolam: Understood. updated it.

scripts/reboot_helper.py
```diff
@@ -0,0 +1,224 @@
...
+        log.log_info("Detach command sent for module {}".format(module_name))
+        return True
+    except NotImplementedError:
+        raise NotImplementedError("PCI detach not implemented on this platform: {type(e).__name__}")
+    except Exception as e:
+        log.log_error("Unexpected error occurred while detaching module {}: {}".format(module_name, e))
+        return False
+
+
+def main():
```
prgeor, line 192: @vvolam main() function in helper?

scripts/reboot_helper.py
```diff
@@ -0,0 +1,224 @@
...
+    except NotImplementedError:
+        raise NotImplementedError("Reboot not implemented on this platform: {type(e).__name__}")
+    except Exception as e:
+        log.log_error("Unexpected error occurred while rebooting module {}: {}".format(module_name, e))
+        return False
+
+    return True
+
+
+def is_dpu():
```
prgeor, line 130: @vvolam can we add this https://github.com/sonic-net/sonic-utilities/blob/7cbcfda5e517e5c8a8cfe7140468774b107d6280/counterpoll/main.py#L20 to some common library in sonic-utilities and use that?

scripts/reboot
```diff
@@ -176,9 +185,138 @@ function check_conflict_boot_in_fw_update()
...
+        fi
+
+        sleep "$poll_interval"
+        waited_time=$((waited_time + poll_interval))
+	    if [ $waited_time -ge $dpu_halt_services_timeout ]; then
+            echo "Error: Timeout waiting for DPU ${DPU_NAME} to finish rebooting"
+            exit ${EXIT_ERROR}
+        fi
+    done
+
```
prgeor, line 303: @vvolam may be put into a function `wait_for_dpu_reboot_status()` ?

scripts/reboot_helper.py
```diff
@@ -0,0 +1,224 @@
...
+
+    log.log_info("Detaching module {}...".format(module_name))
+    try:
+        platform_chassis.pci_detach(module_name)
+        log.log_info("Detach command sent for module {}".format(module_name))
+        return True
+    except NotImplementedError:
+        raise NotImplementedError("PCI detach not implemented on this platform: {type(e).__name__}")
+    except Exception as e:
+        log.log_error("Unexpected error occurred while detaching module {}: {}".format(module_name, e))
```
prgeor, line 188: @vvolam we agreed that pci detach/attach need not be platform specific?

scripts/reboot
```diff
@@ -215,6 +360,56 @@ function linecard_reboot_notify_supervisor()
...
+        if [[ "$SMART_SWITCH" == "yes" ]]; then
+            echo "User requested to reboot the device ${DPU_MODULE_NAME}"
+            reboot_dpu_module "$DPU_MODULE_NAME"
+        else
+            echo "Invalid '-d' option specified for a non-smart switch"
+            exit ${EXIT_ERROR}
+        fi
+    fi
+
+    is_dpu=$(python3 -c "import reboot_helper; reboot_helper.is_dpu()")
```
prgeor, line 393: @vvolam as mentioned can we move this to a common lib inside sonic-utitilies for future reuse?

scripts/reboot_helper.py
vvolam: @prgeor This is to write unit tests for main function as well for maximum code coverage, else I couldn't call in unit tests.

scripts/reboot_helper.py
vvolam: @prgeor yes, I am not using attach method but we planned to support detach method in future. But Gagan from NVIDIA mentioned they are doing something special in this PCI detach method and provided comment to modify accordingly.

@gpunathilell @oleksandrivantsiv FYI.

scripts/reboot
vvolam: @gpunathilell gnmi port changes are expected to be available by 11/15.

gpunathilell: what I meant was the function call is not returning anything as we have no echo statements from the function(even if the gnmi ports are added manually the function doesn't return anything, and the value in `dpu_ip` is being overwritten with empty string)

scripts/reboot
oleksandrivantsiv: @vvolam we agreed that `MODULE_REBOOT_SMARTSWITCH` would be used to initiate DPU reboots during a system reboot, without waiting for the DPUs to fully boot, to reduce overall reboot time. What has changed? Why is there a need to update the HLD?

scripts/reboot_helper.py
oleksandrivantsiv: @vvolam, @prgeor in the HLD, we defined an API for attaching and detaching the DPU PCI, which was added to the base class in [PR #504](https://github.com/sonic-net/sonic-platform-common/pull/504) of the sonic-platform-common repository. Why aren’t we using it in the implementation? Our agreement was that if the vendor implemented the API, we would use it; otherwise, we would fall back on the default attach/detach implementation.

scripts/reboot_helper.py
```diff
@@ -0,0 +1,224 @@
...
+    platform = platform_info.get('platform')
+    if not platform:
+        log.log_error("Platform does not exist in platform_info")
+        return False
+
+    config_data = load_platform_json(platform)
+    if config_data is None:
+        return False
+
+    return any(re.search(r'\.DPU$', key) for key in config_data.keys())
```
oleksandrivantsiv, line 150: `return 'DPU' in config_data` will do the same, but will be more efficient.

scripts/reboot_helper.py
oleksandrivantsiv, line 130: Why don't we add this to the device_info.py file (in the same way as is_smartswitch(): https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-py-common/sonic_py_common/device_info.py#L579) to prevent code duplication. It will allow to reuse it in many places.

vvolam: Good point. I will do this change first and then refactor the code. Thanks!

vvolam: created a PR https://github.com/sonic-net/sonic-buildimage/pull/20724 for these smartswitch common DPU helpers.

scripts/reboot
vvolam: Was the config DB set properly on the server? could you please check "sonic-db-cli CONFIG_DB HGET "DHCP_SERVER_IPV4_PORT|bridge-midplane|DPU<#>" "ips@""" is returning valid value?

scripts/reboot_helper.py
```diff
@@ -0,0 +1,224 @@
...
+    Args:
+        module_name (str): The name of the module to reboot.
+
+    Returns:
+        bool: True if the reboot command was successfully sent, False otherwise.
+    """
+    # Load the platform chassis if not already loaded
+    load_platform_chassis()
+
+    if not is_smartswitch():
```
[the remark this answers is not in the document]
vvolam: @prgeor modified. Thank you!

scripts/reboot_helper.py
vvolam: Yes, created a https://github.com/sonic-net/sonic-buildimage/pull/20724 for add DPU common libraries.

scripts/reboot
```diff
@@ -176,9 +185,138 @@ function check_conflict_boot_in_fw_update()
...
+    sonic-db-cli state_db set "PCIE_DETACH_INFO|${DPU_NAME}" '{"dpu_id": "'${DPU_INDEX}'", "dpu_state": "detaching", "bus_info": "'${DPU_BUS_INFO}'"}'
+
+    pci_detach_module "${DPU_NAME}" "${DPU_BUS_INFO}"
+    reboot_platform_module "${DPU_NAME}"
+    echo 1 > /sys/bus/pci/rescan
+
+    sonic-db-cli state_db del "PCIE_DETACH_INFO|${DPU_NAME}"
+}
+
 function parse_options()
```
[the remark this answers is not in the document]
vvolam: @prgeor Moved all smartswitch functions to reboot_smartswitch_helper script.

scripts/reboot_helper.py
vvolam: Updated reboot HLD, also updated PR using PCI detach/reattach APIs.

scripts/reboot
vvolam: @prgeor yes moved it. Thank you!

scripts/reboot_smartswitch_helper
```diff
@@ -0,0 +1,293 @@
...
+{
+    local dpu_ip=$1
+    local port=$2
+
+    if [[ -z "$PLATFORM_JSON_PATH" ]]; then
+        log_message "Error: PLATFORM_JSON_PATH is not defined"
+        exit $EXIT_ERROR
+    fi
+
+    local dpu_halt_services_timeou
```
oleksandrivantsiv, line 108: dpu_halt_services_timeout?

oleksandrivantsiv: it would be better to merge line 108 and 109:
```
local dpu_halt_services_timeout=$(jq -r '.dpu_halt_services_timeout' "$PLATFORM_JSON_PATH" 2>/dev/null)
```

scripts/reboot_smartswitch_helper
```diff
@@ -0,0 +1,293 @@
...
+    local REBOOT_TYPE=$2
+    local DPU_INDEX=${DPU_NAME//[!0-9]/}
+
+    debug "User requested rebooting device ${DPU_NAME} ..."
+
+    # Send reboot command to DPU
+    gnmi_reboot_dpu "${DPU_NAME}"
+    if [ $? -ne 0 ]; then
+        log_message "Error: Failed to send gnoi command to reboot DPU ${DPU_NAME}"
+        return ${EXIT_ERROR}
```
oleksandrivantsiv, line 172: We should not interrupt the reboot if the pre-shutdown fails. 
If the pre-shutdown fails we should log error message and continue to the platform reboot

scripts/reboot_smartswitch_helper
```diff
@@ -0,0 +1,293 @@
...
+    fi
+
+    DPU_BUS_INFO=$(jq -r --arg DPU_NAME "${DPU_NAME}" '.DPUS[] | select(has($DPU_NAME)) | .[$DPU_NAME].bus_info' "$PLATFORM_JSON_PATH")
```
gpunathilell, line 175:
```suggestion
    DPU_BUS_INFO=$(jq -r --arg DPU_NAME "$DPU_NAME" '.DPUS[$DPU_NAME].bus_info' "$PLATFORM_JSON_PATH")
```

scripts/reboot_smartswitch_helper
```diff
@@ -0,0 +1,293 @@
...
+    if [ -z "$DPU_BUS_INFO" ]; then
+        log_message "Error: Failed to retrieve bus info for DPU ${DPU_NAME}"
+        return ${EXIT_ERROR}
+    fi
+
+    # Update STATE_DB and handle PCIe removal and rescan
+    sonic-db-cli STATE_DB set "PCIE_DETACH_INFO|${DPU_NAME}" '{"dpu_id": "'${DPU_INDEX}'", "dpu_state": "detaching", "bus_info": "'${DPU_BUS_INFO}'"}'
+
+    pci_detach_module "${DPU_NAME} ${DPU_BUS_INFO}"
```
gpunathilell, line 184:
```suggestion
    pci_detach_module ${DPU_NAME} ${DPU_BUS_INFO}
```

scripts/reboot_smartswitch_helper
```diff
@@ -0,0 +1,293 @@
...
+    fi
+    return ${EXIT_ERROR}
+}
+
+# Function to detach PCI module
+function pci_detach_module()
+{
+    local DPU_NAME=$1
+    local DPU_BUS_INFO=$2
+    python3 -c "from reboot_helper import pci_detach_module; pci_detach_module('${DPU_NAME}')"
```
gpunathilell, line 70: reboot_helper.py not in import path, import fails

gpunathilell: Could be due to missing PR, need to double check

scripts/reboot_helper.py
```diff
@@ -0,0 +1,161 @@
...
+        reboot_type (str): The type of reboot requested for the module.
+
+    Returns:
+        bool: True if the reboot command was successfully sent, False otherwise.
+    """
+    if not load_and_verify(module_name):
+        return False
+
+    # Attempt to reboot the module
+    if not hasattr(platform_chassis, 'reboot'):
```
gpunathilell, line 88: the module reboot is not part of the chassis attribute, it is for each module, we need to obtain modules using `get_all_modules` and then perform the reboot

scripts/reboot_helper.py
```diff
@@ -0,0 +1,161 @@
...
+SYSLOG_IDENTIFIER = "reboot_helper"
+
+EXIT_FAIL = -1
+EXIT_SUCCESS = 0
+
+# Global logger instance
+log = logger.Logger(SYSLOG_IDENTIFIER)
+
+# Global variable for platform chassis
+platform_chassis = None
```
prgeor, line 20: Do the following at once during init and don't need to keep checking the module loaded or not

```
  import sonic_platform.platform
   platform_chassis = sonic_platform.platform.Platform().get_chassis()
```

scripts/reboot_helper.py
```diff
@@ -0,0 +1,161 @@
...
+        if not status:
+            log.log_error("Reboot status for module {}: {}".format(module_name, status))
+            return False
+        return True
+    except Exception as e:
+        log.log_error("Unexpected error occurred while rebooting module {}: {}".format(module_name, e))
+        return False
+
+
+def pci_detach_module(module_name):
```
prgeor, line 104: @vvolam can this go to utilities-common/chassis.py as it can be reused?

scripts/reboot_helper.py
```diff
@@ -0,0 +1,161 @@
...
+        if not status:
+            log.log_error("PCI detach status for module {}: {}".format(module_name, status))
+            return False
+        return True
+    except Exception as e:
+        log.log_error("Unexpected error occurred while detaching module {}: {}".format(module_name, e))
+        return False
+
+
+def pci_reattach_module(module_name):
```
prgeor, line 134: @vvolam can this go to utilities-common/chassis.py as it can be reused?

scripts/reboot_smartswitch_helper
vvolam: Fixed it. Thank you!

scripts/reboot_smartswitch_helper
vvolam: Got it. Thanks!

scripts/reboot_smartswitch_helper
vvolam: Modified, Thank you!

scripts/reboot_smartswitch_helper
vvolam: I have modified the helper to common utilities. Please check updated code. Thank you!

scripts/reboot_helper.py
vvolam: I have obtained module_index and calling accordingly.

scripts/reboot_helper.py
vvolam: @prgeor I have updated code to use the load_platform_chassis() from utilities_common/util_base.py.

scripts/reboot_helper.py
vvolam: @prgeor Since these are module_base API's. I have moved all the code to utilities_common/module_base.py and written corresponding unit tests.

scripts/reboot_smartswitch_helper
```diff
@@ -0,0 +1,291 @@
...
+        echo 1 > /sys/bus/pci/rescan
+    fi
+}
+
+# Function to reboot DPU
+function reboot_dpu_platform()
+{
+    local DPU_NAME=$1
+    local REBOOT_TYPE=$2
+    python3 -c "rom utilties_common.module_base import ModuleHelper; ModuleHelper.reboot_module('${DPU_NAME}', '${REBOOT_TYPE}')"
```
gpunathilell, line 94:
```suggestion
    python3 -c "from utilties_common.module_base import ModuleHelper; ModuleHelper.reboot_module('${DPU_NAME}', '${REBOOT_TYPE}')"
```

scripts/reboot_smartswitch_helper
```diff
@@ -0,0 +1,291 @@
...
+    fi
+    return ${EXIT_ERROR}
+}
+
+# Function to detach PCI module
+function pci_detach_module()
+{
+    local DPU_NAME=$1
+    local DPU_BUS_INFO=$2
+    python3 -c "from utilities_common.module_base import ModuleHelper; ModuleHelper.pci_detach_module('${DPU_NAME}')"
```
gpunathilell, line 70: ModuleHelper functions being called without being instantiated. would not work

utilities_common/module_base.py
```diff
@@ -0,0 +1,108 @@
...
+        Reboot the specified module by invoking the platform API.
+
+        Args:
+            module_name (str): The name of the module to reboot.
+            reboot_type (str): The type of reboot requested for the module.
+
+        Returns:
+            bool: True if the reboot command was successfully sent, False otherwise.
+        """
+        module_index = util.try_get(self.platform_chassis.get_module_index, module_name, default=INVALID_MODULE_INDEX)
```
gpunathilell, line 39: The current util_base.py does not accept function arguments, may beed to be modified

scripts/reboot_smartswitch_helper
```diff
@@ -0,0 +1,291 @@
...
+        echo 1 > /sys/bus/pci/devices/${DPU_BUS_INFO}/remove
+    fi
+}
+
+# Function to rescan PCI module
+function pci_reattach_module()
+{
+    local DPU_NAME=$1
+    local DPU_BUS_INFO=$2
+    python3 -c "from utilties_common.module_base import ModuleHelper; ModuleHelper.pci_reattach_module('${DPU_NAME}')"
```
gpunathilell, line 82:
```suggestion
    python3 -c "from utilities_common.module_base import ModuleHelper; ModuleHelper.pci_reattach_module('${DPU_NAME}')"
```

scripts/reboot
```diff
@@ -192,6 +203,13 @@ function parse_options()
             f )
                 REBOOT_FLAGS+=" -f"
                 ;;
+            d )
+                REBOOT_DPU="yes"
+                DPU_MODULE_NAME="$OPTARG"
+                ;;
+            p )
+                PRE_SHUTDOWN="yes"
```
prgeor, line 211: @vvolam why pre-shutdown is an option and not mandatory for smartwitch?

scripts/reboot_smartswitch_helper
prgeor, line 1: @vvolam any specific reason to use a bash script instead of a python script? We could have designed a SmartSwtichRebootMgr() class in python and you can always call pythons cript from reboot bash script.

gpunathilell:
For reboot, is it also possible to check the CONFIG_DB entry for the admin state for the DPUs? if the CONFIG_DB entry is missing or set to down, then we do not need to reboot the DPUs at all

scripts/reboot
vvolam: @prgeor It is only required on DPU and invalid on NPU.

scripts/reboot_smartswitch_helper
vvolam: @prgeor As there are many commands that needs to issued, I preferred bash script to keep it simple.

vvolam: If it is python script, we need to wrap bash commands. It is mix of both python and bash commands.

scripts/reboot_smartswitch_helper
vvolam: Thank you for this! I have fixed it.

utilities_common/module_base.py
vvolam: Thank you, I have added one more function try_get_args which takes arguments also as input.

scripts/reboot_smartswitch_helper
```diff
@@ -0,0 +1,295 @@
...
+    fi
+}
+
+# Function to reboot DPU
+function reboot_dpu_platform()
+{
+    local DPU_NAME=$1
+    local REBOOT_TYPE=$2
+    python3 -c "
+    from utilities_common.module import ModuleHelper
```
gpunathilell, line 95: Remove indentation, as this causes failure in python execution

utilities_common/module.py
```diff
@@ -0,0 +1,126 @@
...
+        if module_index < 0:
+            log.log_error("Unable to get module-index for {}". format(module_name))
+            return False
+
+        if not hasattr(self.platform_chassis.get_module(module_index), 'reboot'):
+            log.log_error("Reboot method not found in platform chassis")
+            return False
+
+        log.log_info("Rebooting module {} with reboot_type {}...".format(module_name, reboot_type))
+        status = self.try_get_args(self.platform_chassis.get_module(module_index).reboot, reboot_type, False)
```
gpunathilell, line 67:
```suggestion
        status = self.try_get_args(self.platform_chassis.get_module(module_index).reboot, reboot_type, default=False)
```

gpunathilell: If default is not specified the arguments are passed onto the callback from try_get_args

scripts/reboot_smartswitch_helper
```diff
@@ -0,0 +1,295 @@
...
+    # Update STATE_DB and handle PCIe removal and rescan
+    sonic-db-cli STATE_DB set "PCIE_DETACH_INFO|${DPU_NAME}" '{"dpu_id": "'${DPU_INDEX}'", "dpu_state": "detaching", "bus_info": "'${DPU_BUS_INFO}'"}'
+
+    pci_detach_module ${DPU_NAME} ${DPU_BUS_INFO}
+    if [ $? -ne 0 ]; then
+        log_message "Error: Failed to detach PCI module for DPU ${DPU_NAME}"
+        return ${EXIT_ERROR}
+    fi
+
+    reboot_dpu_platform ${DPU_NAME}
```
gpunathilell, line 192:
```suggestion
    reboot_dpu_platform ${DPU_NAME} ${REBOOT_TYPE}
```

gpunathilell:
General question, When we perform `reboot -d DPU0` - request reboot for DPU0, do we need to exit after this call in reboot script? `handle_smart_switch "$REBOOT_DPU" "$PRE_SHUTDOWN" "$DPU_MODULE_NAME"` as we are requesting the reboot of a specific DPU and not requesting the reboot of NPU as well?

scripts/reboot_smartswitch_helper
vvolam: Removed. Thank you!

utilities_common/module.py
vvolam: Yes, Thank you!

scripts/reboot_smartswitch_helper
vvolam: Fixed it. Thank you!

vvolam:
> General question, When we perform `reboot -d DPU0` - request reboot for DPU0, do we need to exit after this call in reboot script? `handle_smart_switch "$REBOOT_DPU" "$PRE_SHUTDOWN" "$DPU_MODULE_NAME"` as we are requesting the reboot of a specific DPU and not requesting the reboot of NPU as well?

@gpunathilell, Yes, this is a good catch. we should not proceed with other steps, if only a single DPU reboot is requested. Modified the code. Thank you!

utilities_common/module.py
```diff
@@ -0,0 +1,126 @@
+#!/usr/bin/env python3
+#
+# module_base.py
+#
+# Utility helper for modules within SONiC
+
+from sonic_py_common import logger
```
prgeor, line 7: @vvolam can you use SysLogger? https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-py-common/sonic_py_common/syslogger.py as logger has inherent limitation of polluting the global SYSLOG_IDENTIFIER

utilities_common/chassis.py
```diff
@@ -16,3 +16,19 @@ def get_chassis_local_interfaces():
...
+def is_smartswitch():
+    return hasattr(device_info, 'is_smartswitch') and device_info.is_smartswitch()
+
+
+def is_dpu():
+    return hasattr(device_info, 'is_dpu') and device_info.is_dpu()
+
+
+def get_num_dpus():
+    return hasattr(device_info, 'get_num_dpus') and device_info.get_num_dpus()
```
prgeor, line 30: @vvolam not sure why you want to return boolean value for this ?

utilities_common/chassis.py
```diff
@@ -16,3 +16,19 @@ def get_chassis_local_interfaces():
...
+
+
+def get_dpu_list():
+    return hasattr(device_info, 'get_dpu_list') and device_info.get_dpu_list()
```
prgeor, line 34: @vvolam are you trying to return boolean value?

utilities_common/module.py
vvolam: @prgeor Modified it. But everywhere in the code, logger is used.

utilities_common/chassis.py
vvolam: Fixed it. Thank you!

utilities_common/chassis.py
```diff
@@ -16,3 +16,23 @@ def get_chassis_local_interfaces():
...
+    return hasattr(device_info, 'is_dpu') and device_info.is_dpu()
+
+
+def get_num_dpus():
+    if hasattr(device_info, 'get_num_dpus'):
+        return device_info.get_num_dpus()
+    return 0
+
+
+def get_dpu_list():
```
rameshraghupathy, line 35: @vvolam This may have interactions with the PR https://github.com/sonic-net/sonic-utilities/pull/3271.  Can you please resolve them? Currently do you have any consumer for this?

vvolam: @rameshraghupathy Sure, once PR is merged, I can fix the conflicts. With latest modified commits, I don't have in this current PR.

vvolam:
@prgeor please help review and merge, if no other changes needed.

scripts/reboot
```diff
@@ -225,6 +243,19 @@ fi
...
+if [[ $smart_switch_result -ne 0 ]]; then
+    exit $smart_switch_result
+fi
+
+# On a smartswitch, complete the DPU reboot and exit
+is_smartswitch=$(is_smartswitch)
+if [ "$is_smartswitch" == "True" ] && [ "$REBOOT_DPU" == "yes" ]; then
+    exit $smart_switch_result
+fi
+fi
```
gpunathilell, line 257:
```suggestion
```

scripts/reboot_smartswitch_helper
```diff
@@ -0,0 +1,291 @@
...
+        return ${EXIT_SUCCESS}
+    fi
+    return ${EXIT_ERROR}
+}
+
+# Function to detach PCI module
+function pci_detach_module()
+{
+    local DPU_NAME=$1
+    local DPU_BUS_INFO=$2
```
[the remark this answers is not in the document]
vvolam: @rameshraghupathy We don't need bus_info for rescan. We just need bus_info for removal. Also, this is as per HLD. If we want to use entire path. We need to modify HLD to make sure everyone is on same page. Please suggest?

scripts/reboot_smartswitch_helper
```diff
@@ -0,0 +1,291 @@
...
+        return ${EXIT_ERROR}
+    fi
+
+    reboot_dpu_platform ${DPU_NAME} ${REBOOT_TYPE}
+    if [ $? -ne 0 ]; then
+        log_message "Error: Failed to send platform command to reboot DPU ${DPU_NAME}"
+        return ${EXIT_ERROR}
+    fi
+
+    pci_reattach_module ${DPU_NAME} ${DPU_BUS_INFO}
```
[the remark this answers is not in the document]
vvolam: @rameshraghupathy reboot_dpu_platform() returns only when reboot is completed successfully either success or failure.

utilities_common/module.py
```diff
@@ -0,0 +1,129 @@
...
+
+        log.log_info("Rebooting module {} with reboot_type {}...".format(module_name, reboot_type))
+        status = self.try_get_args(self.platform_chassis.get_module(module_index).reboot, reboot_type, default=False)
+        if not status:
+            log.log_error("Reboot status for module {}: {}".format(module_name, status))
+            return False
+
+        return True
+
+    def pci_detach_module(self, module_name):
```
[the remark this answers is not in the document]
vvolam: @rameshraghupathy Added a note to cover the same. Thank you!

scripts/reboot_smartswitch_helper
```diff
@@ -0,0 +1,291 @@
...
+
+# Function to rescan PCI module
+function pci_reattach_module()
+{
+    local DPU_NAME=$1
+    local DPU_BUS_INFO=$2
+    python3 -c "from utilities_common.module import ModuleHelper; helper = ModuleHelper(); helper.pci_reattach_module('${DPU_NAME}')"
+    if [ $? -ne 0 ]; then
+        log_message "Error: PCI reattach vendor API is not available"
+        echo 1 > /sys/bus/pci/rescan
```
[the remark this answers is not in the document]
vvolam: @rameshraghupathy You are right, I have fixed it to call for same specific device which is detached previously!

gpunathilell:
@vvolam There are some issues with the commits, changes seen in files not relevant to reboot HL. Please check

vvolam:
> @vvolam There are some issues with the commits, changes seen in files not relevant to reboot HLD. Please check

@gpunathilell Thank you Gagan! Merge of latest public and rebase went crazy! Fixed it. Could you please review now?

scripts/reboot_smartswitch_helper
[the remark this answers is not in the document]
rameshraghupathy: @vvolam you are using "${DPU_BUS_INFO}" for both remove and rescan.  This may not work and it should be slightly different for rescan and remove for MtFuji platform.  Not sure about other platforms. Following is an example and how should it be:
DPU_PCI_REMOVE_PATH:  '/sys/bus/pci/devices/0000:1a:00.0/remove'
DPU_PCI_RESCAN_PATH: '/sys/bus/pci/devices/0000:15:03.0/pci_bus/0000:1a/rescan'
So the best option would be to leave these two to the users to define them in the platform.json file instead of '.DPUS[$DPU_NAME].bus_info'.
Example: '.DPUS[$DPU_NAME].pcie_remove_path' and '.DPUS[$DPU_NAME].pcie_rescan_path'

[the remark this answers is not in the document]
vvolam: @rameshraghupathy As discussed in the meeting, if there is any vendor specific rescan, we will use platform API. Only if vendor has not implemented, we will go-ahead with OS generic rescan.

KrisNey-MSFT:
hi @gpunathilell - may I ask the next steps required for this one please?

gpunathilell:
> hi @gpunathilell - may I ask the next steps required for this one please?

HI @KrisNey-MSFT Testing is not done yet, as we have to integrate both PMON changes (some PRs are still pending merge and undergoing changes) and the Reboot HLD related changes (some PRs have merge conflicts with the PRs related to PMON changes)so takes time to create a test image and do some testing

scripts/reboot_smartswitch_helper
```diff
@@ -0,0 +1,291 @@
...
+    sonic-db-cli CONFIG_DB HGET "DPU_PORT|$DPU_NAME" "gnmi"
+}
+
+# Function to get reboot status from DPU
+function get_reboot_status()
+{
+    local dpu_ip=$1
+    local port=$2
+    local reboot_status
+    reboot_status=$(docker exec -i gnmi gnoi_client -target "${dpu_ip}:${port}" -logtostderr -insecure -notls -rpc RebootStatus 2>/dev/null)
```
gpunathilell, line 51: After the changes regarding -notls, the communication is fixed, but the command shows a different error
This command yields the following output:
```System Reboot
panic: rpc error: code = Unknown desc = systemd: Invalid reboot method: 3

goroutine 1 [running]:
main.systemReboot({0xb2a918, 0xc0001e1da0}, {0xb273d8, 0xc000113ac0})```

**Comment by KrisNey-MSFT** [comment]
Thank you!

File: utilities_common/module.py [hunk 47]
```diff
@@ -0,0 +1,133 @@
...
+        Reboot the specified module by invoking the platform API.
+
+        Args:
+            module_name (str): The name of the module to reboot.
+            reboot_type (str): The type of reboot requested for the module.
+
+        Returns:
+            bool: True if the reboot command was successfully sent, False otherwise.
+        """
+        module_index = self.try_get_args(self.platform_chassis.get_module_index, module_name,
```
rameshraghupathy, line 57: The module_name for this API has to be uppercase.  Can you add "module_name = module_name.upper()" before calling the API?

utilities_common/module.py
```diff
@@ -0,0 +1,133 @@
...
+        Note: Caller to make sure this method is not invoked concurrently with
+        pci_detach_module for the same module.
+
+        Args:
+            module_name (str): The name of the module to rescan.
+
+        Returns:
+            bool: True if the rescan command was successfully sent, False otherwise.
+        """
+        module_index = self.try_get_args(self.platform_chassis.get_module_index, module_name,
```
rameshraghupathy, line 117: The module_name for this API has to be uppercase. Can you add "module_name = module_name.upper()" before calling the API?

utilities_common/module.py
```diff
@@ -0,0 +1,133 @@
...
+        Note: Caller to make sure this method is not invoked concurrently with
+        pci_reattach_module for the same module.
+
+        Args:
+            module_name (str): The name of the module to detach.
+
+        Returns:
+            bool: True if the detach command was successfully sent, False otherwise.
+        """
+        module_index = self.try_get_args(self.platform_chassis.get_module_index, module_name,
```
rameshraghupathy, line 87: The module_name for this API has to be uppercase. Can you add "module_name = module_name.upper()" before calling the API?

utilities_common/chassis.py
```diff
@@ -16,3 +16,23 @@ def get_chassis_local_interfaces():
                 lst = data[1].split(",")
                 return lst
     return lst
+
```
rameshraghupathy, line 19: Needs to be merged with "https://github.com/sonic-net/sonic-utilities/pull/3271"  There are missing APIs
1. common_utilities/chassis.py should use get_dpu_list() instead of get_all_dpus()

2. config/chassis_modules.py should use get_dpu_list() instead of get_all_dpus

vvolam: @rameshraghupathy I will merge the "https://github.com/sonic-net/sonic-utilities/pull/3271" changes once the PR is merge and update this file accordingly. Thank you!

utilities_common/module.py
vvolam: Added it. Thank you!

utilities_common/module.py
vvolam: Added it. Thank you!

utilities_common/module.py
vvolam: Added it. Thank you!

gpunathilell:
```
2025-01-28 21:11:44 - Error: Failed to send reboot status command to DPU dpu1
/usr/local/bin/reboot_smartswitch_helper: line 126: [: null: integer expression expected
```
Error seen with latest changes

gpunathilell:
Also there are two issues which need to be addressed, the provisioning for the DPU and the switch GNMI container configuration (as these configurations are required to make sure that the gnoi_client command can be executed from the switch)
GNMI command execution fails with EOF error (due to GNMI container being shut down during pre shutdown) This issue has to be handled as per HLD (make sure GNMI container is not shutdown on DPU)

KrisNey-MSFT:
Dependent upon PMON via Cisco, and a few Issues filed.

gpunathilell:
For the pcie related changes which are being done, we need to have differentiation in the way DPU reboot is handled in single DPU reboot and system reboot scenarios (`ModuleBase.MODULE_REBOOT_DPU` vs `ModuleBase.MODULE_REBOOT_SMARTSWITCH`) for `MODULE_REBOOT_SMARTSWITCH` we just schedule the DPU reboot and exit from the reboot call i.e. pcie is not back up after the reboot is executed, so we can not remove the STATE_DB entry in case of `ModuleBase.MODULE_REBOOT_SMARTSWITCH` (or perform rescan) so the entry deletion has to be skippped in case of system reboot

scripts/reboot_smartswitch_helper
vvolam: This output is not from RebootStatus RPC but from Reboot RPC, I have fixed it. Thank you!

vvolam:
> ```
> 2025-01-28 21:11:44 - Error: Failed to send reboot status command to DPU dpu1
> /usr/local/bin/reboot_smartswitch_helper: line 126: [: null: integer expression expected
> ```
> 
> Error seen with latest changes

This is because of dpu_halt_services_timeout not present in platform.json. I have updated script to have a default timeout in such cse. Thank you!

scripts/reboot_smartswitch_helper
```diff
@@ -0,0 +1,292 @@
...
+        port=$GNMI_PORT # Default GNMI port
+    fi
+    log_message "GNMI port ${DPU_NAME}: $port"
+
+    if [ -z "$dpu_ip" ]; then
+        log_message "Error: Failed to retrieve DPU IP for ${DPU_NAME}"
+        return ${EXIT_ERROR}
+    fi
+
+    docker exec -i gnmi gnoi_client -target ${dpu_ip}:${port} -logtostderr -insecure -notls -rpc Reboot -jsonin '{"method":HALT}' 2>/dev/null
```
ganglyu, line 151: Is the GNMI/GNOI server running in insecure mode on the DPU?

vvolam: @ganglyu No, insecure is not required, removed it. Thank you!

scripts/reboot_smartswitch_helper
```diff
@@ -0,0 +1,293 @@
...
+    local DPU_NAME=$1
+    sonic-db-cli CONFIG_DB HGET "DPU_PORT|$DPU_NAME" "gnmi_port"
+}
+
+# Function to get reboot status from DPU
+function get_reboot_status()
+{
+    local dpu_ip=$1
+    local port=$2
+    $(docker exec gnmi gnoi_client -target "${dpu_ip}:${port}" -logtostderr -notls -module System -rpc RebootStatus &>/dev/null)
```
hdwhdw, line 50: We need some integration test here to prevent CLI for gnoi_client accidentally changes. I recommend writing our own python grpc client, example: https://github.com/sonic-net/sonic-mgmt/pull/17068, but that can be a long shot.

scripts/reboot_smartswitch_helper
```diff
@@ -0,0 +1,293 @@
...
+        log_message "ERROR: Failed to send gnoi command to reboot ${DPU_NAME}"
+    fi
+
+    local DPU_BUS_INFO=$(jq -r --arg DPU_NAME "$DPU_NAME" '.DPUS[$DPU_NAME].bus_info' "$PLATFORM_JSON_PATH")
+    if [ -z "$DPU_BUS_INFO" ] || [ "$DPU_BUS_INFO" = "null" ]; then
+        log_message "ERROR: Failed to retrieve bus info for ${DPU_NAME}"
+        return ${EXIT_ERROR}
+    fi
+
+    if [[ "$REBOOT_TYPE" != $MODULE_REBOOT_SMARTSWITCH ]]; then
```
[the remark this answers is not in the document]
vvolam: > Is there a check to see if reboot is being done on a DPU which is currently powered off and skipping reboot for that specific DPU?(Admin State=Down Operational State=Down)

We rely on platform API return code to determine this outcome.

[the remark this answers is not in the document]
vvolam: > Even in the case of `MODULE_REBOOT_SMARTSWITCH` we still need to suppresss the pcied warnings

With the discussions that are in process, redis-cli requests will be done inside PCI attach/detach. Once that is finalized, I will modify it here.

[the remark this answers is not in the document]
gpunathilell: > > Is there a check to see if reboot is being done on a DPU which is currently powered off and skipping reboot for that specific DPU?(Admin State=Down Operational State=Down)

Does this mean the reboot should be skipped assuming the operational state for the specific DPU is down in the platform specific API?

scripts/reboot
```diff
@@ -225,6 +246,18 @@ fi
...
+smart_switch_result=$?
+if [[ $smart_switch_result -ne 0 ]]; then
+    exit $smart_switch_result
+fi
+
+# On a smartswitch, complete the DPU reboot and exit
+is_smartswitch=$(python3 -c 'from utilities_common.chassis import is_smartswitch; print(is_smartswitch())')
+if [ "$is_smartswitch" == "True" ] && [ "$REBOOT_DPU" == "yes" ]; then
+    exit $smart_switch_result
+fi
```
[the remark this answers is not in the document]
vvolam: I made the changes. Thank you!

ganglyu (dismissed):

scripts/reboot_smartswitch_helper
```diff
@@ -0,0 +1,303 @@
...
+    local DPU_INDEX=${DPU_NAME//[!0-9]/}
+
+    debug "User requested rebooting device ${DPU_NAME} ..."
+
+    # Check if the DPU operation status is online before rebooting
+    local oper_status
+    oper_status=$(show chassis modules status "${DPU_NAME^^}" | sed -n '/^ *DPU/ s/.*\s\+\(Online\|Offline\)\s\+.*/\1/p')
+    if [ $? -ne 0 ]; then
+        log_message "ERROR: Failed to retrieve DPU status."
+    elif
```
gpunathilell, line 175:
```suggestion
    else
```

gpunathilell (approved):

scripts/reboot_smartswitch_helper
gpunathilell: Otherwise LGTM

vvolam: Fixed. Thank you very much for all your help

gpunathilell (approved):

ganglyu (approved):

PR #3573, opened 2024-10-11
Fix  slash in path.

xincunli-sonic:
#### What I did

Addressing issue [#20377](https://github.com/sonic-net/sonic-buildimage/issues/20377). The issue caused by unescape in JsonPointer implementation which followed [RFC 6901](https://www.rfc-editor.org/rfc/rfc6901)

```python
pointer = jsonpointer.JsonPointer(path)

...

class JsonPointer(object):
    """A JSON Pointer that can reference parts of a JSON document"""

    # Array indices must not contain:
    # leading zeros, signs, spaces, decimals, etc
    _RE_ARRAY_INDEX = re.compile('0|[1-9][0-9]*$')
    _RE_INVALID_ESCAPE = re.compile('(~[^01]|~$)')

    def __init__(self, pointer):

        # validate escapes
        invalid_escape = self._RE_INVALID_ESCAPE.search(pointer)
        if invalid_escape:
            raise JsonPointerException('Found invalid escape {}'.format(
                invalid_escape.group()))

        parts = pointer.split('/')
        if parts.pop(0) != '':
            raise JsonPointerException('Location must start with /')

        parts = [unescape(part) for part in parts]
        self.parts = parts
```

#### How I did it

Re-escape `/` to `~1` to match the real key in json, and [JsonPatch](https://www.rfc-editor.org/rfc/rfc6902#appendix-A.14) will handle it correctly.

#### How to verify it

```shell
admin@str2-7250-lc1-2:~$ cat link.json 
[
    {
        "op": "add",
        "path": "/asic1/PORTCHANNEL_INTERFACE/PortChannel106|10.0.0.6~131",
        "value": {}
    }
]
admin@str2-7250-lc1-2:~$ sudo config apply-patch link.json 
sonic_yang(6):Note: Below table(s) have no YANG models: DHCP_SERVER
sonic_yang(6):Note: Below table(s) have no YANG models: LOGGER
sonic_yang(6):Note: Below table(s) have no YANG models: LOGGER
Patch Applier: asic1: Patch application starting.
Patch Applier: asic1: Patch: [{"op": "add", "path": "/PORTCHANNEL_INTERFACE/PortChannel106|10.0.0.6~131", "value": {}}]
Patch Applier: asic1 getting current config db.
Patch Applier: asic1: simulating the target full config after applying the patch.
Patch Applier: asic1: validating all JsonPatch operations are permitted on the specified fields
Patch Applier: asic1: validating target config does not have empty tables,
Patch Applier: asic1: sorting patch updates.
Patch Applier: The asic1 patch was converted into 0 changes.
Patch Applier: asic1: applying 0 changes in order.
Patch Applier: asic1: verifying patch updates are reflected on ConfigDB.
Patch Applier: asic1 patch application completed.
Patch applied successfully.
```

### DIFF
--- a/generic_config_updater/generic_updater.py
+++ b/generic_config_updater/generic_updater.py
@@ -19,7 +19,7 @@ def extract_scope(path):
     if not path:
         raise Exception("Wrong patch with empty path.")
     pointer = jsonpointer.JsonPointer(path)
-    
+
     # Re-escapes `/` in path parts by converting `/` back to `~1`.
     parts = [part.replace("/", "~1") for part in pointer.parts]
     if not parts:
--- a/generic_config_updater/generic_updater.py
+++ b/generic_config_updater/generic_updater.py
@@ -17,11 +17,11 @@ CHECKPOINT_EXT = ".cp.json"
 
 def extract_scope(path):
     if not path:
-        raise Exception("Wrong patch with empty path.")
+        raise GenericConfigUpdaterError("Wrong patch with empty path.")
     pointer = jsonpointer.JsonPointer(path)
 
-    # Re-escapes `/` in path parts by converting `/` back to `~1`.
-    parts = [part.replace("/", "~1") for part in pointer.parts]
+    # Re-escapes
+    parts = [jsonpointer.escape(part) for part in pointer.parts]
     if not parts:
         raise GenericConfigUpdaterError("Wrong patch with empty path.")
     if parts[0].startswith("asic"):
--- a/generic_config_updater/generic_updater.py
+++ b/generic_config_updater/generic_updater.py
@@ -19,7 +19,9 @@ def extract_scope(path):
     if not path:
         raise Exception("Wrong patch with empty path.")
     pointer = jsonpointer.JsonPointer(path)
-    parts = pointer.parts
+    
+    # Re-escapes `/` in path parts by converting `/` back to `~1`.
+    parts = [part.replace("/", "~1") for part in pointer.parts]
     if not parts:
         raise GenericConfigUpdaterError("Wrong patch with empty path.")
     if parts[0].startswith("asic"):
--- a/tests/generic_config_updater/multiasic_change_applier_test.py
+++ b/tests/generic_config_updater/multiasic_change_applier_test.py
@@ -25,6 +25,12 @@ class TestMultiAsicChangeApplier(unittest.TestCase):
             "/asic0123456789/PORTCHANNEL/PortChannel102/admin_status": (
                 True, "asic0123456789", "/PORTCHANNEL/PortChannel102/admin_status"
             ),
+            "/asic1/PORTCHANNEL_INTERFACE/PortChannel106|10.0.0.6/31": (
+                True, "asic1", "/PORTCHANNEL_INTERFACE/PortChannel106|10.0.0.6/31"
+            ),
+            "/asic1/PORTCHANNEL_INTERFACE/PortChannel106|10.0.0.6~131": (
+                True, "asic1", "/PORTCHANNEL_INTERFACE/PortChannel106|10.0.0.6~131"
+            ),
             "/localhost/BGP_DEVICE_GLOBAL/STATE/tsa_enabled": (
                 True, "localhost", "/BGP_DEVICE_GLOBAL/STATE/tsa_enabled"
             ),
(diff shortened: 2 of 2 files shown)


generic_config_updater/generic_updater.py
```diff
@@ -19,7 +19,9 @@ def extract_scope(path):
     if not path:
         raise Exception("Wrong patch with empty path.")
     pointer = jsonpointer.JsonPointer(path)
-    parts = pointer.parts
+
+    # Re-escapes `/` in path parts by converting `/` back to `~1`.
+    parts = [part.replace("/", "~1") for part in pointer.parts]
```
wen587, line 24: Is it a issue for multi-asic only? 
I remember in single asic, the jsonpointer has built in feature to parse / to ~1

xincunli-sonic: It was introduced by multi asic change, we have to extract the namespace from the patch, then JsonPointer will do escape `~1` back to `/`

wen587: Sorry. I don't get it. 
By the code here, the extract_scope doesn't differentiate single or multi. How do we guarantee single asic is not broken? https://github.com/sonic-net/sonic-utilities/blob/319f58d5d5dbe86e2f0aeece67b8274dcd3f3ce8/config/main.py#L1599-L1600

xincunli-sonic: For single asic, the portchannel interface will not having slash in port, but multi asic does have, hence, we could do escape slash to ~1 in jsonpatch.

xincunli-sonic: Another reason is single asic logic is taking path as input parameter directly, but multi asic, we have to do combine with parts by Jsonpointer result..

wen587 (approved):

qiluo-msft:
>             except Exception:

Can we use specific (not too general) exception type?

---
Refers to: tests/generic_config_updater/multiasic_change_applier_test.py:80 in 26ee2a4. [](commit_id = 26ee2a4e99bbfb1adff460621950355c5ea69646, deletion_comment = False)

generic_config_updater/generic_updater.py
qiluo-msft, line 24: >replace [](http://example.com/codeflow?start=18&length=7)

this replace is not enough. Could you use jsonpointer.escape(s) : str->str

xincunli-sonic: Discussed offline, we should reuse escape function of jsonpointer to do full escape.

qiluo-msft:
>             except Exception:

still applicable.

---
In reply to: [2442871334](https://github.com/sonic-net/sonic-utilities/pull/3573#issuecomment-2442871334) [](http://example.com/codeflow?ancestors=2442871334)

---
Refers to: tests/generic_config_updater/multiasic_change_applier_test.py:80 in 26ee2a4. [](commit_id = 26ee2a4e99bbfb1adff460621950355c5ea69646, deletion_comment = False)

xincunli-sonic:
>             except Exception:

This is test path, and main purpose is catch any extract failure, not focusing on specific error.

---
In reply to: [2445514861](https://github.com/sonic-net/sonic-utilities/pull/3573#issuecomment-2445514861) [](http://example.com/codeflow?ancestors=2445514861,2442871334)

---
Refers to: tests/generic_config_updater/multiasic_change_applier_test.py:80 in 26ee2a4. [](commit_id = 26ee2a4e99bbfb1adff460621950355c5ea69646, deletion_comment = False)

qiluo-msft (approved):

mlok-nokia:
@xincunli-sonic This PR has cherry-pick conflict. Please help to address it so that can cherry-pick to 202405. Thanks

xincunli-sonic:
This requires the previous [PR](https://github.com/sonic-net/sonic-utilities/pull/3630) be cherry picked into 202405, after it done, I will reuquest the current PR backport again.

mlok-nokia:
> This requires the previous [PR](https://github.com/sonic-net/sonic-utilities/pull/3630) be cherry picked into 202405, after it done, I will reuquest the current PR backport again.

Thansk.

bingwang-ms:
Included in 202405 by PR https://github.com/sonic-net/sonic-utilities/pull/3649

PR #3576, opened 2024-10-15
Add YANG validation for config reload if file is given

wen587:
#### What I did
Add constraint for YANG when config reload, which is already enabled in load_minigraph
#### How I did it
Check YANG vaidation for the config
#### How to verify it
Unit test

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1810,6 +1810,22 @@ def reload(db, filename, yes, load_sysinfo, no_service_restart, force, file_form
             click.echo("Input {} config file(s) separated by comma for multiple files ".format(num_cfg_file))
             return
 
+    if filename is not None:
+        if multi_asic.is_multi_asic():
+            # Multiasic has not 100% fully validated. Thus pass here.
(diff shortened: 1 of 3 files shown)


xincunli-sonic (approved):

config/main.py
```diff
@@ -1372,6 +1372,19 @@ def multiasic_write_to_db(filename, load_sysinfo):
         migrate_db_to_lastest(ns)
 
 
+def config_file_yang_validation(filename):
+    config_to_check = read_json_file(filename)
+    sy = sonic_yang.SonicYang(YANG_DIR)
+    sy.loadYangModel()
+    try:
+        sy.loadData(configdbJson=config_to_check)
+        sy.validate_data_tree()
```
ganglyu, line 1381: Do you want to check table without yang models?

wen587: offline synced. will init another PR fo load)minigraph

ganglyu (approved):

PR #3583, opened 2024-10-22
Record and warn tables which not covered by YANG

wen587:
ADO: 29917681
#### What I did
Sent a warn syslog if there are tables not covered by yang.
#### How I did it
Sent a warn syslog if there are tables not covered by yang.
#### How to verify it
Manual test

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -2323,7 +2323,7 @@ def validate_config_by_cm_alerting(cm, config_json, jname):
     except Exception as ex:
         log.log_warning("Failed to validate {}. Alerting: {}".format(jname, ex))
 
-    if len(cm.tablesWithOutYang):
+    if len(cm.tablesWithOutYang()):
         log.log_warning("YANG failed to cover tables: {}.".format(str(cm.tablesWithOutYang)))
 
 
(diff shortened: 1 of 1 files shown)


ganglyu (approved):

config/main.py
```diff
@@ -2323,6 +2323,9 @@ def validate_config_by_cm_alerting(cm, config_json, jname):
     except Exception as ex:
         log.log_warning("Failed to validate {}. Alerting: {}".format(jname, ex))
 
+    if len(cm.tablesWithOutYang()):
+        log.log_warning("YANG failed to cover tables: {}.".format(str(cm.tablesWithOutYang)))
```
qiluo-msft, line 2327: log_error?

qiluo-msft (approved):

bingwang-ms:
Is a warning log good enough? Do we need to raise an exception?

wen587:
> Is a warning log good enough? Do we need to raise an exception?

Hi @bingwang-ms , no need to raise an exception. This is to help us discover non-YANG tables. We don't want to pause the load_minigraph process

PR #3596, opened 2024-11-05
 Fix for integer overflow of counter value if its too large

harjotsinghpawra:
#### What I did

Changed the data type casting to float if count coming is too large or in scientific notation .

#### How I did it
made it float then type cast to int for display purpose

#### How to verify it
Manually passed big value and see the count output if it breaks or works


wsycqyz (approved):
lğtm

StormLiangMS (approved):
LGTM

PR #3605, opened 2024-11-08
Optimize techsupport reducing number of vtysh calls in scale sceario

dgsudharsan:
#### What I did
Optimize techsupport collecting neighbor information having single call instead of multiple calls which increases techsupport time

#### How I did it
By combining all vtysh calls into one inside neighbor loop

#### How to verify it
Running techsupport with 256 neighbors. This reduces by more than 1.5 minutes

Before the fix
image

After the fix
image


dgsudharsan:
Updated techsupport sonic-mgmt test here https://github.com/sonic-net/sonic-mgmt/pull/15466

scripts/generate_dump
```diff
@@ -523,24 +523,37 @@ save_bgp_neighbor() {
     local ns=$(get_vtysh_namespace $asic_id)
 
     neighbor_list_v4=$(${timeout_cmd} bash -c "vtysh $ns -c 'show ip bgp neighbors' | grep 'BGP neighbor is' | awk -F '[, ]' '{print \$4}' | awk /\\\./")
-    for word in $neighbor_list_v4; do
-        save_cmd "vtysh $ns -c \"show ip bgp neighbors $word advertised-routes\"" "ip.bgp.neighbor.$word.adv$asic_id"
-        save_cmd "vtysh $ns -c \"show ip bgp neighbors $word routes\"" "ip.bgp.neighbor.$word.rcv$asic_id"
-    done
+    if [ ! -z "$neighbor_list_v4" ]; then
+        v4_cmd="vtysh "
+        for word in $neighbor_list_v4; do
```
FengPan-Frank, line 528: Great optimization!
Is multiAsic case missed here?

dgsudharsan: I think this API operates per ns. I would recommend to extend this once merged if we want to combine all ns into one.

r12f: hi Sud, I think part of what FengPan might be thinking is that - do you mind trying this on the multi-asic case and make sure nothing regresses?

dgsudharsan: @r12f  We don't have a multiasic platform. However we ran the sonic-mgmt show techsupport test and it passes with the changes in the PR https://github.com/sonic-net/sonic-mgmt/pull/15466

scripts/generate_dump
```diff
@@ -524,27 +524,42 @@ save_bgp_neighbor() {
...
+            v4_cmd="${v4_cmd} $ns -Ec 'show bgp ip neighbors $word advertised-routes' "
```
qiluo-msft, line 530: >$word [](http://example.com/codeflow?start=61&length=5)

It is easy to inject malicious command line by. Could you validate the `word` context before using it?

scripts/generate_dump
```diff
@@ -524,27 +524,42 @@ save_bgp_neighbor() {
...
-    for vrf in $vrf_list; do
-        neighbor_list=`${timeout_cmd} bash -c "vtysh $ns -c 'show ip bgp vrf $vrf neighbors' | grep 'BGP neighbor is' | awk -F '[, ]' '{print \$4}'"`
-        for word in $neighbor_list; do
-            save_cmd "vtysh $ns -c \"show ip bgp vrf $vrf neighbors $word advertised-routes\"" "ip.bgp.neighbor.$vrf.$word.adv$asic_id"
-            save_cmd "vtysh $ns -c \"show ip bgp vrf $vrf neighbors $word routes\"" "ip.bgp.neighbor.$vrf.$word.rcv$asic_id"
+
+    if [ ! -z "$vrf_list" ]; then
+        vrf_cmd="vtysh "
+        for vrf in $vrf_list; do
+            neighbor_list=`${timeout_cmd} bash -c "vtysh $ns -c 'show ip bgp vrf $vrf neighbors' | grep 'BGP neighbor is' | awk -F '[, ]' '{print \$4}'"`
```
qiluo-msft, line 555: >$vrf [](http://example.com/codeflow?start=81&length=4)

Could you validate the `vrf` context before using it?

scripts/generate_dump
maipbui, line 530: >bgp ip [](http://example.com/codeflow?start=44&length=6)

original code is `show ip bgp neighbors`, is this expected?

scripts/generate_dump
```diff
@@ -524,27 +524,42 @@ save_bgp_neighbor() {
...
+            for word in $neighbor_list; do
+                vrf_cmd="${vrf_cmd} $nc -Ec 'show ip bgp vrf $vrf neighbors $word advertised-routes' "
```
maipbui, line 557: >$nc [](http://example.com/codeflow?start=36&length=3)

is this variable newl, should it be $ns?

scripts/generate_dump
dgsudharsan: I modified it to show bgp ipv4 neighbors now. It is to be aligned with the ipv6 command below as well. Both show ip bgp neighbor and show bgp ipv4 neighbor yield same results

scripts/generate_dump
dgsudharsan: This is obtained from the show vrf command output and loop through the values. Don't think we require one more validation

scripts/generate_dump
dgsudharsan: How can we inject $word? This is a local variable used to loop through the neighbor list obtained from previous command. Don't see a risk here?

liat-grozovik:
@qiluo-msft  @FengPan-Frank any further comments?

FengPan-Frank (approved):
LGTM

xwjiang-ms:
@dgsudharsan this PR is causing PR test failure, could you help to review the revert PR?
https://github.com/sonic-net/sonic-utilities/pull/3694

PR #3192, opened 2024-03-05
[warm-reboot] Add support to pass extra kernel cmdline parameters

pavannaregundi:
#### What I did
Add support to optionally pass extra kernel command line arguments during warm-reboot kexec load. 

#### How I did it
Extra command line arguments can be supplied from variable 'KEXEC_LOAD_EXTRA_CMDLINE_LINUX' in "installer.conf" present in device hwsku folder.

```
# cat installer.conf 
KEXEC_LOAD_EXTRA_CMDLINE_LINUX="<extra command line arguments>"
```

#### How to verify it
Triggered warm-reboot on Marvell AC5XRD platform and checked for successful warmboot.

PR #3613, opened 2024-11-12
[ARM64][Warm-Reboot]: Get boot arguments from uboot environment

marvellparthiv:
Fetch boot arguments from the next boot image by reading it from uboot environment variables.

#### What I did
In current scenario, during warm-reboot for arm64, boot arguments are fetched from device tree of the current image. This causes an issue during upgrade/downgrade with warm-reboot.

#### How I did it
With the change, boot arguments for arm64 is fetched from uboot environment variables. The logic relies on creation of
"sonic_bootargs" environment variable to fetch the right boot argument.

#### How to verify it
Tested by sonic to sonic install followed by warm-reboot.


pavannaregundi:
@saiarcot895 Can you please review this. PR has been rebased to latest code.

saiarcot895 (approved):

pavannaregundi:
@saiarcot895 before this is cherry picked, we need to cherry pick PR #3192 to resolve the conflict.

PR #3618, opened 2024-11-15
Fix counterpoll show for ENI Counters

vivekrnv:
#### What I did

Fix the object name and because of which, counterpoll show for ENI is not working


oleksandrivantsiv (approved):

prsunny (approved):

prsunny:
@qiluo-msft , please merge

PR #3621, opened 2024-11-19
[broadcom] Add broadcom ASIC info to "show version" click command

wdoekes:
#### What I did

When the ASIC platform is broadcom, expand the "show version" output with additional versions. This helps keeping track of different versions when building/testing.

#### How I did it

Added _get_broadcom_info() function that calls 'bcmcmd' with 'bsv' and 'show unit'.

#### How to verify it

Run `show version` on a device with broadcom ASIC and observe how it now lists this:

```diff
 # show version

 SONiC Software Version: SONiC.master.0-123456789
 SONiC OS Version: 12
 Distribution: Debian 12.8
 Kernel: 6.1.0-22-2-amd64
 Build commit: 123456789
 Build date: Tue Nov 12 15:28:30 UTC 2024
 Built by: sonic-builder

 Platform: x86_64-accton_as9716_32d-r0
 HwSKU: Accton-AS9716-32D
 ASIC: broadcom
 ASIC Count: 1
+ASIC API BRCM SAI ver: [10.1.42.0]
+ASIC API OCP SAI ver: [1.13.2]
+ASIC API SDK ver: [sdk-6.5.29]
+ASIC Model: Unit 0 chip BCM56980_B0 (current)
 Serial Number: ...
```

(see diff above)

(see diff above)

#### Additional comments

I am aware that:

1. This is broadcom specific. I find it useful. Especially since the trouble I've been having with different versions working/not-working depending on libsaibcm.
2. The code might not be in the best location. I'll gladly move parts to a submodule/helper file. Please give me feedback on where you'd rather see it.
3. If syncd is down, the `show version` command takes slightly longer because of the timeouts. I think this is acceptable.

In my case, I could not get `master` from November 2024 to run on the Accton-AS7326-56X or Accton-AS9716-32D, except when I reverted ffb9bc021 and 433d039a4; reverting the BRCM SAI/SDK versions. While the BRCM/SAI versions will not get fetched if syncd is down (they will be in syslog), these clues are helpful when comparing output.

Another output:
```diff
 # show version

 SONiC Software Version: SONiC.202405.0-123456789
 SONiC OS Version: 12
 Distribution: Debian 12.8
 Kernel: 6.1.0-22-2-amd64
 Build commit: 123456789
 Build date: Mon Nov 18 15:56:07 UTC 2024
 Built by: sonic-builder

 Platform: x86_64-accton_as7326_56x-r0
 HwSKU: Accton-AS7326-56X
 ASIC: broadcom
 ASIC Count: 1
+ASIC API BRCM SAI ver: [11.2.13.1]
+ASIC API OCP SAI ver: [1.14.0]
+ASIC API SDK ver: [sdk-6.5.30-SP4]
+ASIC API CANCUN ver: [06.04.01]
+ASIC Model: Unit 0 chip BCM56873_A0 (current)
 Serial Number: ...
```

### DIFF
--- a/show/main.py
+++ b/show/main.py
@@ -1480,6 +1480,18 @@ def version(verbose):
     platform_info = device_info.get_platform_info()
     chassis_info = platform.get_chassis_info()
 
+    if platform_info['asic_type'] == 'broadcom':
+        asic_info = _get_broadcom_info()
+        asic_info_str = []
+        asic_info_str.extend(
+            # NOTE: No ':' because that is in the output as well.
+            ["ASIC API {}".format(line) for line in asic_info['versions']])
+        asic_info_str.extend(
+            ["ASIC Model: {}".format(line) for line in asic_info['units']])
+        asic_info_str = '\n'.join(asic_info_str)
+    else:
+        asic_info = None
+
     sys_uptime_cmd = ["uptime"]
     sys_uptime = subprocess.Popen(sys_uptime_cmd, text=True, stdout=subprocess.PIPE)
 
@@ -1496,6 +1508,8 @@ def version(verbose):
     click.echo("HwSKU: {}".format(platform_info['hwsku']))
     click.echo("ASIC: {}".format(platform_info['asic_type']))
     click.echo("ASIC Count: {}".format(platform_info['asic_count']))
+    if asic_info:
+        click.echo(asic_info_str)
     click.echo("Serial Number: {}".format(chassis_info['serial']))
     click.echo("Model Number: {}".format(chassis_info['model']))
     click.echo("Hardware Revision: {}".format(chassis_info['revision']))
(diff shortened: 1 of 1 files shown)


show/main.py
```diff
@@ -1496,6 +1508,8 @@ def version(verbose):
     click.echo("HwSKU: {}".format(platform_info['hwsku']))
     click.echo("ASIC: {}".format(platform_info['asic_type']))
     click.echo("ASIC Count: {}".format(platform_info['asic_count']))
+    if asic_info:
```
FengPan-Frank, line 1511: would this be disruptive for some test cases if adding output in the middle of show version cli?

wdoekes: Legit question.

If I look at `tests/show_platform_test.py`, it should not change (for now) because it has a mellanox ASIC_TYPE and not a broadcom one.

My reason for placing it in the middle was this: _the top info can then be copy-pasted into bug reports without  including machine-specific serial numbers._

If you want, I can add a test there for the broadcom model with this output included.

wdoekes: B.t.w. I'd rather fetch the info once at initialization from the ASIC and get it from a state(?) db during `show version`, but that's probably more invasive.

FengPan-Frank: my thought is better to keep output backward compatibility, since we are not sure if there's other customize scripts would be broken.

wdoekes: Ok ok.

So, after the second paragraph and before the docker image list then?

PR #3626, opened 2024-11-22
Fix typo in route_check_script

deepak-singhal0408:
#### What I did
Fixed a typo in route_check script. If result is not empty, script might fail due to undefined variable.

#### How I did it
Corrected the typo. 

#### How to verify it
NA


deepak-singhal0408:
got introduced as part of PR https://github.com/sonic-net/sonic-utilities/pull/3604

deepak-singhal0408:
MSFT ADO: 30327574

PR #3632, opened 2024-11-29
 [copp] Add CoPP trap for DHCP L2 broadcasts

SapphirusBeryl:
#### What I did
This PR adds support for the DHCP_L2 trap types.

#### How I did it
It appears that these traps are needed on Innovium SAI in order for DHCP broadcast packets to be trapped.
Out of the box, the DHCP Relay doesn't do anything (since DHCP broadcasts don't appear on the CPU), but with these extra traps applied it works.

#### How to verify it
 
- Set up test switch using Innovium SAI (e.g. Supermicro T7132S)
- Run tcpdump on switch to watch for DHCP packets
- Run DHCP client on a connected server
- Observe that no packets are received
- Apply patch, enable L2 traps
- Re-run tcpdump and DHCP client
- Observe that packets are received and relayed, and that the client receives an answer

#### Related

https://github.com/sonic-net/sonic-swss/pull/2987
https://github.com/sonic-net/sonic-buildimage/pull/20970


jasper-hou-micas:
Is it seen only on  Innovium SAI  or  is that common on other SAI like Broadcom?

PR #3654, opened 2024-12-05
Revert config reload with YANG validation

wen587:
#### What I did
The sonic vm image has yang validation issue with default config setup. Limit YANG check to golden config only before fix default factory config.
#### How I did it
Restrict YANG validation to golden config only.
#### How to verify it
Unit test and integration test.


xwjiang-ms (approved):

bingwang-ms:
@wen587 Please sync code with master branch to address PR checker issue

PR #3656, opened 2024-12-05
[config] Restrict YANG validation to Golden Config

wen587:
#### What I did
The sonic vm image has yang validation issue with default config setup. Limit YANG check to golden config only before fix default factory config.
#### How I did it
Restrict YANG validation to golden config only.
#### How to verify it
Unit test and integration test.


bingwang-ms:
Is it good enough to decide whether to do YANG validation by file name? It's very easy to be bypassed.

wen587:
> Member

Created ADO(30441433) to track this. Currently vm has config init issue without this fix. Will fix vm config then fix it back.

PR #3668, opened 2024-12-09
[GCU] Add data acl table and rule check

wen587:
ADO: 30051071
#### What I did
Add a json patch check to avoid race condition issue when sonic consume ACL_TABLE and ACL_RULE together
#### How I did it
Add a special dataacl check for specific change where ACL_TABLE type change with ACL_RULE update
#### How to verify it
unit test

### DIFF
--- a/generic_config_updater/gu_common.py
+++ b/generic_config_updater/gu_common.py
@@ -215,40 +215,47 @@ class ConfigWrapper:
                     raise IllegalPatchOperationError("Modification of {} table is illegal- validating function {} returned False".format(table, function))
 
     def illegal_dataacl_check(self, old_config, upd_config):
-        # Block data ACL changes when patch includes:
-        # 1. table "type" being replaced
-        # 2. rule update on tables with table "type" replaced
-        # This will cause rase condition when swss consume the change
-        # and make acl rule inactive
-        old_acltable = old_config.get("ACL_TABLE", {})
-        upd_acltable = upd_config.get("ACL_TABLE", {})
+        '''
+        Block data ACL changes when patch includes:
+        1. table "type" being replaced
+        2. rule update on tables with table "type" replaced
+        This will cause race condition when swss consume the change of
+        acl table and acl rule and make the changed acl rule inactive
+        '''
+        old_acl_table = old_config.get("ACL_TABLE", {})
+        upd_acl_table = upd_config.get("ACL_TABLE", {})
 
         # Pick data acl table with "type" field
-        old_dacltable = [table for table, fields in old_acltable.items()
+        old_dacl_table = [table for table, fields in old_acl_table.items()
                          if fields.get("type") and fields["type"] != "CTRLPLANE"]
-        upd_dacltable = [table for table, fields in upd_acltable.items()
+        upd_dacl_table = [table for table, fields in upd_acl_table.items()
                          if fields.get("type") and fields["type"] != "CTRLPLANE"]
 
-        # Pick intersect tables that "type" being replaced
-        common_dacltable = set(old_dacltable).intersection(set(upd_dacltable))
-        modified_common_dacltable = [
-            table for table in common_dacltable
-            if old_acltable[table]["type"] != upd_acltable[table]["type"]
+        # Pick intersect common tables that "type" being replaced
+        common_dacl_table = set(old_dacl_table).intersection(set(upd_dacl_table))
+        # Identify tables from the intersection where the "type" field differs
+        modified_common_dacl_table = [
+            table for table in common_dacl_table
+            if old_acl_table[table]["type"] != upd_acl_table[table]["type"]
         ]
 
-        old_aclrule = old_config.get("ACL_RULE", {})
-        upd_aclrule = upd_config.get("ACL_RULE", {})
+        old_acl_rule = old_config.get("ACL_RULE", {})
+        upd_acl_rule = upd_config.get("ACL_RULE", {})
 
         # Pick rules with its dependent table which has "type" replaced
-        old_daclrule = [rule for rule in old_aclrule
-                        if rule.split("|")[0] in modified_common_dacltable]
-        upd_daclrule = [rule for rule in upd_aclrule
-                        if rule.split("|")[0] in modified_common_dacltable]
+        old_dacl_rule = [rule for rule in old_acl_rule
+                        if rule.split("|")[0] in modified_common_dacl_table]
+        upd_dacl_rule = [rule for rule in upd_acl_rule
+                        if rule.split("|")[0] in modified_common_dacl_table]
 
         # Block changes if acl rule change on tables with "type" replaced
-        for key in set(old_daclrule).union(set(upd_daclrule)):
-            if (old_aclrule.get(key, {}) != upd_aclrule.get(key, {})):
-                raise IllegalPatchOperationError("Modification of dataacl table is illegal")
+        for key in set(old_dacl_rule).union(set(upd_dacl_rule)):
+            if (old_acl_rule.get(key, {}) != upd_acl_rule.get(key, {})):
+                raise IllegalPatchOperationError(
+                    "Modification of dataacl rule {} is illegal: \
+                        acl table type changed in {}".format(
+                            key, modified_common_dacl_table
+                    ))
 
     def validate_lanes(self, config_db):
         if "PORT" not in config_db:
--- a/generic_config_updater/gu_common.py
+++ b/generic_config_updater/gu_common.py
@@ -181,6 +181,8 @@ class ConfigWrapper:
                 if any(op['op'] == operation and field == op['path'] for op in patch):
                     raise IllegalPatchOperationError("Given patch operation is invalid. Operation: {} is illegal on field: {}".format(operation, field))
 
+        self.illegal_dataacl_check(old_config, target_config)
+
         def _invoke_validating_function(cmd, jsonpatch_element):
             # cmd is in the format as <package/module name>.<method name>
             method_name = cmd.split(".")[-1]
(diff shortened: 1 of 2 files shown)


generic_config_updater/gu_common.py
```diff
@@ -212,6 +214,41 @@ def _invoke_validating_function(cmd, jsonpatch_element):
                 if not _invoke_validating_function(function, element):
                     raise IllegalPatchOperationError("Modification of {} table is illegal- validating function {} returned False".format(table, function))
 
+    def illegal_dataacl_check(self, old_config, upd_config):
+        # Block data ACL changes when patch includes:
+        # 1. table "type" being replaced
+        # 2. rule update on tables with table "type" replaced
+        # This will cause rase condition when swss consume the change
```
xincunli-sonic, line 221: >rase [](http://example.com/codeflow?start=26&length=4)

typo?

generic_config_updater/gu_common.py
```diff
@@ -212,6 +214,41 @@ def _invoke_validating_function(cmd, jsonpatch_element):
                 if not _invoke_validating_function(function, element):
                     raise IllegalPatchOperationError("Modification of {} table is illegal- validating function {} returned False".format(table, function))
 
+    def illegal_dataacl_check(self, old_config, upd_config):
+        # Block data ACL changes when patch includes:
```
xincunli-sonic, line 218: ># Block [](http://example.com/codeflow?start=8&length=7)

Suggest to use docstring

generic_config_updater/gu_common.py
```diff
@@ -212,6 +214,41 @@ def _invoke_validating_function(cmd, jsonpatch_element):
...
+        upd_acltable = upd_config.get("ACL_TABLE", {})
+
+        # Pick data acl table with "type" field
+        old_dacltable = [table for table, fields in old_acltable.items()
+                         if fields.get("type") and fields["type"] != "CTRLPLANE"]
+        upd_dacltable = [table for table, fields in upd_acltable.items()
+                         if fields.get("type") and fields["type"] != "CTRLPLANE"]
+
+        # Pick intersect tables that "type" being replaced
+        common_dacltable = set(old_dacltable).intersection(set(upd_dacltable))
```
xincunli-sonic, line 233: >intersection [](http://example.com/codeflow?start=46&length=12)

 Could you add a comment explaining why intersecting old and updated tables is necessary.

generic_config_updater/gu_common.py
```diff
@@ -212,6 +214,41 @@ def _invoke_validating_function(cmd, jsonpatch_element):
...
+        # Pick rules with its dependent table which has "type" replaced
+        old_daclrule = [rule for rule in old_aclrule
+                        if rule.split("|")[0] in modified_common_dacltable]
+        upd_daclrule = [rule for rule in upd_aclrule
+                        if rule.split("|")[0] in modified_common_dacltable]
+
+        # Block changes if acl rule change on tables with "type" replaced
+        for key in set(old_daclrule).union(set(upd_daclrule)):
+            if (old_aclrule.get(key, {}) != upd_aclrule.get(key, {})):
+                raise IllegalPatchOperationError("Modification of dataacl table is illegal")
```
xincunli-sonic, line 251: >IllegalPatchOperationError [](http://example.com/codeflow?start=22&length=26)

The IllegalPatchOperationError message could be improved by including information about the specific table or rule that violates the constraints.

generic_config_updater/gu_common.py
```diff
@@ -212,6 +214,41 @@ def _invoke_validating_function(cmd, jsonpatch_element):
...
+
+        # Pick data acl table with "type" field
+        old_dacltable = [table for table, fields in old_acltable.items()
+                         if fields.get("type") and fields["type"] != "CTRLPLANE"]
+        upd_dacltable = [table for table, fields in upd_acltable.items()
+                         if fields.get("type") and fields["type"] != "CTRLPLANE"]
+
+        # Pick intersect tables that "type" being replaced
+        common_dacltable = set(old_dacltable).intersection(set(upd_dacltable))
+        modified_common_dacltable = [
```
xincunli-sonic, line 234: >modified_common_dacltable [](http://example.com/codeflow?start=8&length=25)

Maybe add a comment for `modified_common_dacltable` readability:

# Identify tables from the intersection where the "type" field differs between configurations

generic_config_updater/gu_common.py
```diff
@@ -212,6 +214,41 @@ def _invoke_validating_function(cmd, jsonpatch_element):
...
+            table for table in common_dacltable
+            if old_acltable[table]["type"] != upd_acltable[table]["type"]
+        ]
+
+        old_aclrule = old_config.get("ACL_RULE", {})
+        upd_aclrule = upd_config.get("ACL_RULE", {})
```
xincunli-sonic, line 240: Could be redundant, I am wondering have they been changed?

generic_config_updater/gu_common.py
```diff
@@ -212,6 +214,41 @@ def _invoke_validating_function(cmd, jsonpatch_element):
...
+        # 1. table "type" being replaced
+        # 2. rule update on tables with table "type" replaced
+        # This will cause rase condition when swss consume the change
+        # and make acl rule inactive
+        old_acltable = old_config.get("ACL_TABLE", {})
+        upd_acltable = upd_config.get("ACL_TABLE", {})
+
+        # Pick data acl table with "type" field
+        old_dacltable = [table for table, fields in old_acltable.items()
+                         if fields.get("type") and fields["type"] != "CTRLPLANE"]
```
xincunli-sonic, line 228: >fields.get("type") [](http://example.com/codeflow?start=28&length=18)

This assumes that fields["type"] always exists if fields.get("type") is true. If not, suggest to use

if fields.get("type", None) and fields["type"] != "CTRLPLANE"]

generic_config_updater/gu_common.py
xincunli-sonic, line 221: >rase condition [](http://example.com/codeflow?start=26&length=14)

Also, can you help add more comments when will happen race condition?

generic_config_updater/gu_common.py
```diff
@@ -212,6 +214,41 @@ def _invoke_validating_function(cmd, jsonpatch_element):
                 if not _invoke_validating_function(function, element):
                     raise IllegalPatchOperationError("Modification of {} table is illegal- validating function {} returned False".format(table, function))
 
+    def illegal_dataacl_check(self, old_config, upd_config):
```
hdwhdw, line 217: Suggest make the function name more descriptive. Illegal Data ACL is a very broad term. Maybe consider something like `check_acl_table_type_changed`. Going forward if there are more illegal changes we can create a check_illegal function and have it call this, which make it more structural and readable.

generic_config_updater/gu_common.py
```diff
@@ -212,6 +214,41 @@ def _invoke_validating_function(cmd, jsonpatch_element):
...
+        # Block data ACL changes when patch includes:
+        # 1. table "type" being replaced
+        # 2. rule update on tables with table "type" replaced
+        # This will cause rase condition when swss consume the change
+        # and make acl rule inactive
+        old_acltable = old_config.get("ACL_TABLE", {})
```
hdwhdw, line 223: Here as well, prefers `old_acl_table`, and other symbols as well.

generic_config_updater/gu_common.py
```diff
@@ -212,6 +214,41 @@ def _invoke_validating_function(cmd, jsonpatch_element):
...
+        upd_acltable = upd_config.get("ACL_TABLE", {})
```
hdwhdw, line 224: `upd` for `updated` can easily confuse people. Suggest always spelling out the full word. With modern formatting tools it is better to have a long variable name than having a variable name whose meaning is unclear to readers.

https://semirise.com/variable-naming-conventions-and-best-practices

tests/generic_config_updater/field_operation_validator_test.py
```diff
@@ -188,6 +188,92 @@ def test_validate_field_operation_illegal__rm_loopback0(self):
...
+            "ACL_RULE": {
+                "OPENAI_INGRESS|RULE_1": {
+                    "IP_TYPE": "IPV6ANY"
+                }
+            }
+        }
+        config_wrapper = gu_common.ConfigWrapper()
+        config_wrapper.validate_field_operation(old_config, target_config)
+
+    def test_validate_field_operation_legal__dataacl_rule_update(self):
```
hdwhdw, line 248: The three `legal__dataacl` (additional underscore) are typos here?

tests/generic_config_updater/field_operation_validator_test.py
```diff
@@ -188,6 +188,92 @@ def test_validate_field_operation_illegal__rm_loopback0(self):
         config_wrapper = gu_common.ConfigWrapper()
         self.assertRaises(gu_common.IllegalPatchOperationError, config_wrapper.validate_field_operation, old_config, target_config)
 
+    def test_validate_field_operation_illegal__dataacl_table_type_and_rule(self):
```
hdwhdw, line 191: Suggest unit test name should contain:
1. expectation (success or fail)
2. why the expectation (failure reason)

Hence, maybe consider a name like `test_validate_field_operation_fail_illegal_type`, `test_validate_field_operation_fail_type_changed`, etc.

hdwhdw (commented):
LGTM. Mostly some nits about naming.

generic_config_updater/gu_common.py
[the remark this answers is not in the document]
wen587: Yes. This is used for Control plane which don't have such issue as data plane. it is defined in yang model: https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-templates/sonic-types.yang.j2#L97

wen587: If not exists, fields.get("type") will return None. It is same as if fields.get("type", None)

generic_config_updater/gu_common.py
wen587: This is because we only care "type" changed tables. So it will be the intersection between old and updated table.

generic_config_updater/gu_common.py
wen587: Previously it is old acl table , here is old acl rule.

tests/generic_config_updater/field_operation_validator_test.py
wen587: Here is the convention in this file. 
legal indicates success, illegal indicates fail.
Reason is the description after __

tests/generic_config_updater/field_operation_validator_test.py
wen587: This is the descriptive convention in this unit test file.

generic_config_updater/gu_common.py
wen587: Will update acltable to acl_table as suggested.
Here is the only case found in data acl. So I use this general definition. But if there are more, I will restructure the code here.

generic_config_updater/gu_common.py
wen587: Will keep upd as it was applied all over the code in GCU.

xincunli-sonic (approved):
LGTM

hdwhdw (approved):
LGTM

tests/generic_config_updater/field_operation_validator_test.py
hdwhdw: Thanks!

tests/generic_config_updater/field_operation_validator_test.py
hdwhdw: Got it. Thanks.

kperumalbfn:
@wen587 Any sonic-mgmt tests for this? Please let me know so that we can add them to 202411 as well

wen587:
> @wen587 Any sonic-mgmt tests for this? Please let me know so that we can add them to 202411 as well

Yes.sonic-mgmt already have test for this: https://github.com/sonic-net/sonic-mgmt/blob/202411/tests/generic_config_updater/test_cacl.py#L691
Please go ahead to add to 202411

PR #3671, opened 2024-12-10
Fixed the issues with sonic-clear queuecounter for egress queue and voq

saksarav-nokia:
#### What I did
 Fixed the sonic-clear queuecounter issues reported in https://github.com/sonic-net/sonic-buildimage/issues/20913
Fixed the following 2 issues
   1) For the egress queue counters, after the queue counter is cleared and show is issued with --nonzero option, the code takes the elif path at line 323 (old code) even if the diff between cached counter and current counter is 0, prints the current counter values from counter_db . This issue was for both egress queue counter and voq counter. 
2)  When the sonic-clear queuecounter is issued , the queuestat is called first without --voq option and this gets the port names and queue ids for each port in each asic , reads the egress queue counters and cache the values in /tmp/cache/queuestat/.  Then queuestat is called with --voq option and this gets all the system ports names and voq id's for each asic , reads the voq counters and cache the values in /tmp/cache/queuestat. Since each asic has the all the system ports and all the voqs, and since the cache file name is queuestat+system_port_name with out asic namespace, caching asic1's voq counters overwrites the asic0's voq counters . 

#### How I did it
1)Corrected the logic mistake for issue 1. Since cnstat_diff_print is called only if cache file is present and this should print only if non-zero & valid diff or  (not non-zero )
2) Added asic namespace with the cache file name for vow counters.
show command output is not changed.
show command output is not changed.

### DIFF
--- a/scripts/queuestat
+++ b/scripts/queuestat
@@ -294,38 +294,27 @@ class Queuestat(object):
                 old_cntr = cnstat_old_dict.get(key)
             if old_cntr is not None:
                 if self.voq:
-                   if not non_zero or ns_diff(cntr['totalpacket'], old_cntr['totalpacket']) != '0' or \
+                    if not non_zero or ns_diff(cntr['totalpacket'], old_cntr['totalpacket']) != '0' or \
                                    ns_diff(cntr['totalbytes'], old_cntr['totalbytes']) != '0' or \
                                    ns_diff(cntr['droppacket'], old_cntr['droppacket']) != '0' or \
                                    ns_diff(cntr['dropbytes'], old_cntr['dropbytes']) != '0' or \
                                    ns_diff(cntr['creditWDpkts'], old_cntr['creditWDpkts']) != '0':
-                       table.append((port, cntr['queuetype'] + str(cntr['queueindex']),
+                        table.append((port, cntr['queuetype'] + str(cntr['queueindex']),
                                    ns_diff(cntr['totalpacket'], old_cntr['totalpacket']),
                                    ns_diff(cntr['totalbytes'], old_cntr['totalbytes']),
                                    ns_diff(cntr['droppacket'], old_cntr['droppacket']),
                                    ns_diff(cntr['dropbytes'], old_cntr['dropbytes']),
                                    ns_diff(cntr['creditWDpkts'], old_cntr['creditWDpkts'])))
-                   elif not non_zero or cntr['totalpacket'] != '0' or cntr['totalbytes'] != '0' or \
-                                  cntr['droppacket'] != '0' or cntr['dropbytes'] != '0' or cntr['creditWDpkts'] != '0':
-                       table.append((port, cntr['queuetype'] + str(cntr['queueindex']),
-                                  cntr['totalpacket'], cntr['totalbytes'],
-                                  cntr['droppacket'], cntr['dropbytes'], cntr['creditWDpkts']))
                 else:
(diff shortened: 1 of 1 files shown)


saksarav-nokia:
@rlhui @vmittal-msft @arista-nwolfe for viz

scripts/queuestat
```diff
@@ -346,8 +335,10 @@ class Queuestat(object):
         for port in natsorted(self.counter_port_name_map):
             json_output[port] = {}
             cnstat_dict = self.get_cnstat(self.port_queues_map[port])
-
-            cnstat_fqn_file_name = cnstat_fqn_file + port
+            cache_ns = ''
+            if self.voq and self.namespace is not None:
```
arista-hpandya, line 339: We should also handle cases when namespace arg is provided but the voq flag is absent

scripts/queuestat
```diff
@@ -294,38 +294,27 @@ class Queuestat(object):
...
-                                  ns_diff(cntr['totalbytes'], old_cntr['totalbytes']),
-                                  ns_diff(cntr['droppacket'], old_cntr['droppacket']),
-                                  ns_diff(cntr['dropbytes'], old_cntr['dropbytes'])))
-                   elif not non_zero or cntr['totalpacket'] != '0' or cntr['totalbytes'] != '0' or \
-                                     cntr['droppacket'] != '0' or cntr['dropbytes'] != '0':
-                      table.append((port, cntr['queuetype'] + str(cntr['queueindex']),
-                               cntr['totalpacket'], cntr['totalbytes'],
-                               cntr['droppacket'], cntr['dropbytes']))
-
+                       table.append((port, cntr['queuetype'] + str(cntr['queueindex']),
```
arista-hpandya, line 313: Could you add a unit test to verify the logic change here?

arista-hpandya (changes requested):

saksarav-nokia:
> @saksarav-nokia Can you elaborate on the usecase of namespace/asic specific clear commands? I recall when I was working on adding multi-asic support to queuestat we decided to go forward with the goal that a clear would "clear all namespaces".

@arista-hpandya , queuestat is called for the following 2 scenarios for sonic-clear queuecounter.
  1) without --voq:  - In this case, the egress queue corresponding to front panel ports in the respective asic is cached when QueueStatWrapper calls QueueStat for each asic. In this case asic0 has only asic0's front panel ports and asic1 has only asic1's front panel ports

2) with --voq:  - In this case, the voq counters corresponding to system ports in the respective asic is cached when QueueStatWrapper calls QueueStat for each asic. But as you know, each asic has all the system ports of entire chassis (For J2C+, per asic in each IMM, there are system port for 18 front panel ports, 1 cpu port, 2 recycle ports) . That means asic0 and asic1 has the same set of system ports . So when QueueStat is called for asic0, the cache files are created for all system ports and when the  QueueStat is called for asic1, the cache files are created with exact same name which overwrites the cache files created for asic0. So when you show after clear, the show queue counter --voq doesn't work as expected.

scripts/queuestat
saksarav-nokia: > Could you please add a unit test to verify the logic change here?

There are tests already in https://github.com/sonic-net/sonic-utilities/blob/master/tests/queue_counter_test.py which covers these changes.

arlakshm (approved):
lgtm

bingwang-ms:
Readding label to trigger auto-cherry-pick

PR #3672, opened 2024-12-10
add sonic_asic_platform in the render context when rendering docker_image_ctl.j2

BYGX-wcr:
#### Why I did

Now, when the service-creator in sonic-package-manager renders the docker_image_ctl.j2 template, it does not include the sonic_asic_platform info in the render context. The sonic_asic_platform variable is crucial for the manipulation of template rendering, so I added the code to supplement it in the render context.

#### What I did

I added code to acquire sonic_asic_platform info and added it in the render context of docker_image_ctl.j2.

Microsoft ADO: 30477799

#### How I did it

Add additional code in the service creator of sonic package manager.

#### How to verify it

Use the updated sonic-utilities code to build a new VS image.

### DIFF
--- a/sonic_package_manager/service_creator/creator.py
+++ b/sonic_package_manager/service_creator/creator.py
@@ -29,6 +29,8 @@ from sonic_package_manager.service_creator.feature import FeatureRegistry
 from sonic_package_manager.service_creator.sonic_db import SonicDB
 from sonic_package_manager.service_creator.utils import in_chroot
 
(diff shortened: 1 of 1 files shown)


judyjoseph (approved):
LGTM , requesting @stepanblyschak to review too - to make sure this is the right approach to pass the build variables .. here eg: aonic_asic_platform.

stepanblyschak (approved):

PR #3680, opened 2024-12-15
Delete TRSNCEIVER tables while config reload 

noaOrMlnx:
#### What I did
Removed TRANSCEIVER tables while config reload / load minigraph is executed.
This code adds the option to reserve TRANSCEIVER_INFO table if xcvrd is being restarted. - https://github.com/sonic-net/sonic-platform-daemons/blob/0cb36447ebfda390d999df6516beac4c17315a7b/sonic-xcvrd/xcvrd/xcvrd.py#L2199  

The fix introduced here comes to fix the issue that the table remains also after config reload / load minigraph, which may cause config DB and TRANSCEIVER_INFO table have different ports keys. 
 
#### How I did it
Update config reload command to delete TRANSCEIVER tables: ["TRANSCEIVER_INFO", "TRANSCEIVER_STATUS", "TRANSCEIVER_PM", "TRANSCEIVER_FIRMWARE_INFO", "TRANSCEIVER_DOM_SENSOR", "TRANSCEIVER_DOM_THRESHOLD"]
#### How to verify it
Execute config reload and make sure TRANSCEIVER tables are deleted

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1322,7 +1322,8 @@ def flush_configdb(namespace=DEFAULT_NAMESPACE):
 
 
 def delete_transceiver_tables():
-    tables = ["TRANSCEIVER_INFO", "TRANSCEIVER_STATUS", "TRANSCEIVER_PM", "TRANSCEIVER_FIRMWARE_INFO", "TRANSCEIVER_DOM_SENSOR", "TRANSCEIVER_DOM_THRESHOLD"]
+    tables = ["TRANSCEIVER_INFO", "TRANSCEIVER_STATUS", "TRANSCEIVER_PM",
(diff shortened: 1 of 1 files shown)


dprital:
@prgeor , Can you please review ?

mihirpat1 (approved):

dprital (approved):

bingwang-ms:
@noaOrMlnx Will skip this PR for 202405 branch as it's considered as not a must-have fix. Please let me know for any concern. Thanks

noaOrMlnx:
@kperumalbfn This change is required to any system which doesn't contain all ports. 
Most cases, it will happen because of split to 8 on certain systems. This split demands deleting the neighbor port.
If we delete the neighbor port and then config reload, some tables can remain in Config DB, which will later cause xcvrd (inside pmon) to crash searching for non-existing tables. 
The above changes solve the issue

PR #3682, opened 2024-12-16
Disable routeCheck monit as part of config reload/minigraph stop service and enable it back of service start.

abdosi:
What I did:

 For Config reload/minigraph stop and re-enable routeCheck because
 with large route scale of 70K+ routes this can log monit error
 Transiently which can also fail in sonic-mgmt.

Why I  did:

Because of this transient issue  monit ERR log can get generated and this can result failure of sonic-mgmt test case.

How I verify:

Manual Verification via `sudo monit status routeCheck`


abdosi:
@anamehra @yejianquan for viz.

config/main.py
```diff
@@ -949,17 +950,18 @@ def _restart_services():
...
     wait_service_restart_finish('networking', last_networking_timestamp)
 
+    # Reload Monit configuration to pick up new hostname in case it changed
+    click.echo("Reloading Monit configuration ...")
+    clicommon.run_command(['sudo', 'monit', 'reload'])
+
     try:
         subprocess.check_call(['sudo', 'monit', 'status'], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
-        click.echo("Enabling container monitoring ...")
+        click.echo("Enabling container and routeCheck monitoring ...")
```
anamehra, line 959: After the route check is enabled at this point, as the links and sessions are still coming up, it may still cause same error if a route check happens, is that correct?

abdosi: @anamehra once we have start of monit probability of hitting ERR is less since it takes 5 min to initialize and then we have  15 min (3 cycle)  for any error to come. System should get stabilize by 15-20 min post monit restart.

yejianquan (approved):
LGTM

abdosi:
@yejianquan : Help with cherry-pick

PR #3685, opened 2024-12-18
Enable multi asic golden config YANG validation.

xincunli-sonic:
ADO:[30529412]
#### What I did
Add constraint for YANG when config reload, which is already enabled in load_minigraph
#### How I did it
Check YANG vaidation for the config
#### How to verify it
Unit test

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -2023,7 +2023,7 @@ def load_minigraph(db, no_service_restart, traffic_shift_away, override_config, 
             click.secho("Cannot find '{}'!".format(golden_config_path),
                         fg='magenta')
             raise click.Abort()
-        
+
         config_file_yang_validation(golden_config_path)
 
         config_to_check = read_json_file(golden_config_path)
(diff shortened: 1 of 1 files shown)


wen587 (approved):
lgtm

abdosi (approved):

abdosi:
@xincunli-sonic : should we add UT for this ?  Also do we need this in 202405 ?

FengPan-Frank (approved):
LGTM

xincunli-sonic:
> @xincunli-sonic : should we add UT for this ? Also do we need this in 202405 ?

Will add UTs for both single and multi asic in next PR, ADO: 30677081. Also, tagged for 202405 branch.

bingwang-ms:
Hi @xincunli-sonic, I'm not going to backport this change to 202405 if it's for multi-ASIC. We have a separated branch for multi-ASIC now. And we do see some problem caused by the YANG validator in 202411. Please let me know if you think this is a must-have fix for 202405.

PR #3565, opened 2024-10-03
Fix handling of optional fields in get_bgp_peer function

matofeder:
#### What I did

This commit fixes an issue in get_bgp_peer function where `local_addr` and `name` were treated as mandatory. However, `local_addr` and `name` fields are not required in BGP neighbor configuration.

#### How I did it

The local_addr and name fields are no longer mandatory in the get_bgp_peer function.

#### How to verify it

Apply BGP neighbor configuration like below (utilizing BGP_PEER_GROUP), where fields such as local_addr or name are omitted:

```json
{ 
    "BGP_GLOBALS": {
        "default": {
            "local_asn": "65001",
            "log_nbr_state_changes": "true",
            "router_id": "10.0.0.1"
        }
    },
   "BGP_PEER_GROUP": {
        "default|core": {
          "peer_group_name": "core",
          "admin_status": "true",
          "asn": "65002",
          "peer_type": "external"
        }
     },
    "BGP_NEIGHBOR": {
        "default|Ethernet4": {
            "peer_group_name": "core"
        }
    },
    "BGP_NEIGHBOR_AF": {
        "default|core|ipv4_unicast": {
            "admin_status": "true",
            "route_map_in": [
                "ALLOW"
            ],
            "route_map_out": [
                "ALLOW"
            ]
        }
    },
    "ROUTE_MAP": {
        "ALLOW|1": {
            "route_operation": "permit"
        }
    }
}
```

#### Previous command output (if the output of a command-line utility has changed)

```bash
root@sw1:~# show ip interfaces 
Traceback (most recent call last):
  File "/usr/local/bin/ipintutil", line 280, in <module>
    main()
  File "/usr/local/bin/ipintutil", line 273, in main
    ip_intfs = get_ip_intfs(af, namespace, display)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/bin/ipintutil", line 236, in get_ip_intfs
    ip_intfs_in_ns = get_ip_intfs_in_namespace(af, namespace, display)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/bin/ipintutil", line 149, in get_ip_intfs_in_namespace
    bgp_peer = get_bgp_peer()
               ^^^^^^^^^^^^^^
  File "/usr/local/bin/ipintutil", line 51, in get_bgp_peer
    local_addr = data[neighbor_ip]['local_addr']
[...]

PR #3695, opened 2024-12-24
show ip interfaces: fix exception with BGP unnumbered

bradh352:
#### What I did
Without this patch an exception is thrown when running `show ip interfaces` when BGP Unnumbered is configured:

```
root@sw1:~# show ip interfaces
  File "/usr/local/bin/ipintutil", line 280, in <module>
    main()
  File "/usr/local/bin/ipintutil", line 273, in main
    ip_intfs = get_ip_intfs(af, namespace, display)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/bin/ipintutil", line 236, in get_ip_intfs
    ip_intfs_in_ns = get_ip_intfs_in_namespace(af, namespace, display)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/bin/ipintutil", line 149, in get_ip_intfs_in_namespace
    bgp_peer = get_bgp_peer()
               ^^^^^^^^^^^^^^
  File "/usr/local/bin/ipintutil", line 51, in get_bgp_peer
    local_addr = data[neighbor_ip]['local_addr']
                 ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
KeyError: 'local_addr'
```

This patch will allow the command to complete successfully.  It shouldn't be necessary to actually query FRR for neighbor details, the prior version didn't, it just echo'd back config details.

This is a slightly different patch to what #3565 does.  It does not add blank values to the output parameter, it skips due to exception handling.

#### How I did it

Added try / except logic block around optional key/values.

#### How to verify it

Configure BGP Unnumbered via FRR Management framework and observe 'show ip interfaces' fails with above error.  Apply patch and observe expected output.

Fixes #3565
Fixes https://github.com/sonic-net/sonic-buildimage/issues/19930
Signed-off-by: Brad House (@bradh352)

### DIFF
--- a/scripts/ipintutil
+++ b/scripts/ipintutil
@@ -48,9 +48,14 @@ def get_bgp_peer():
     data = config_db.get_table('BGP_NEIGHBOR')
 
     for neighbor_ip in data.keys():
-        local_addr = data[neighbor_ip]['local_addr']
-        neighbor_name = data[neighbor_ip]['name']
-        bgp_peer.setdefault(local_addr, [neighbor_name, neighbor_ip])
+        # The data collected here will only work for manually defined neighbors
+        # so we need to ignore errors when using BGP Unnumbered.
+        try:
+            local_addr = data[neighbor_ip]['local_addr']
+            neighbor_name = data[neighbor_ip]['name']
+            bgp_peer.setdefault(local_addr, [neighbor_name, neighbor_ip])
+        except KeyError:
+            pass
     return bgp_peer
 
 
--- a/tests/mock_tables/config_db.json
+++ b/tests/mock_tables/config_db.json
@@ -2059,6 +2059,14 @@
         "asn": "65200",
         "keepalive": "3"
     },
+    "BGP_NEIGHBOR|Vlan100": {
+        "rrclient": "0",
+        "peer_type": "external",
+        "nhopself": "0",
+        "admin_status": "up",
+        "holdtime": "10",
+        "keepalive": "3"
+    },
     "SCHEDULER|scheduler.0": {
         "type": "DWRR",
         "weight": "14"


FengPan-Frank (approved):
LGTM

FengPan-Frank (commented):
pls help to add exceptional test case, thanks

bradh352:
@FengPan-Frank I'm not very familiar with the way the test infrastructure is designed, I can see the existing tests for `show ip interfaces` in `show_ip_int_test.py`, however its not clear how the running config that is used is loaded.  

From some grepping for the IPs the existing test cases use it appears to be loaded via mock_tables/mock_single_asic.py  ... 

I don't immediately see how to augment the config db, I see some tests do things like:
```
from .mock_tables import dbconnector
jsonfile_config = os.path.join(mock_db_path, "config_db")
dbconnector.dedicated_dbs['CONFIG_DB'] = jsonfile_config
db = Db()
```
But I'm pretty sure that ends up replacing the entire config_db instead of augmenting it, is that true?

I guess I could do a bunch of `runner.infoke(config.config.commands["bgp"].commands[...` to configure it

FengPan-Frank:
> show_ip_int_test

we usually use mock config to do this, refer https://github.com/search?q=repo%3Asonic-net%2Fsonic-utilities%20int_ip_input&type=code

bradh352:
> > show_ip_int_test
> 
> we usually use mock config to do this, refer https://github.com/search?q=repo%3Asonic-net%2Fsonic-utilities%20int_ip_input&type=code

Ok, that's the way i mentioned I wasn't sure about.  Does that replace the entire CONFIG_DB, or does it merge into existing configuration?

I do see it sets `os.environ['UTILITIES_UNIT_TESTING'] = "1"` in the `int_ip_input` which is used to `config` tests instead of `os.environ['UTILITIES_UNIT_TESTING'] = "2"` which is used in `show` tests but I don't see any docs on what that actually means.  Its not clear where to look in the code either for what it means.

FengPan-Frank (approved):

PR #3698, opened 2024-12-26
Fix monit connection failure when doing routeCheck/contianerCheck  monitor enable after config reload

abdosi:
Fix:
https://github.com/sonic-net/sonic-buildimage/issues/21268

How I did:
Reorder the sequence of doing enabling container_cheek and routeCheck before doing monit reload to avoid transient issue of monit sock error.

Also I add a sleep of 1 sec to make sure monitor enable configuration takes effect before we do reload.

How I verify:

Ran 20 iteration of config reload and it pass successfully

 ```
#!/bin/bash

    i=0
    while [ $i -lt 20 ]; do
        echo "iteration $i"
        sudo config reload -y -f > /dev/null
        sleep 120
        if [ $? -ne 0 ]; then
           echo "Failed iteration $i for config reload"
           break
        fi
        i=$((i+1))
        sudo monit status routeCheck
        sudo monit status container_checker
        echo "Completed $i iteration sucessfully"
    done
```


abdosi:
Hi @bingwang-ms : help with merge of this .

yaqiangz (approved):
lgtm

xwjiang-ms:
@abdosi should we cherry-pick to 202411?

PR #3700, opened 2024-12-27
Improve SONiC disk checker to handle disk full case and mount overlay fs to allow remote user login.

liuh-80:
Improve SONiC disk checker to handle disk full case and mount overlay fs to allow remote user login.

This PR depends on DB schema change: https://github.com/sonic-net/sonic-buildimage/pull/21351

#### What I did
    Currently disk checker only handle RO disk case, but when disk no free space, remote user also can't login.

#### How I did it
    Check disk free space and mount overlay fs to allow TACACS login.

#### How to verify it
    Create big file on device, make device no free space. then login with remote user should success.

##### Work item tracking
- Microsoft ADO: 30608100

### DIFF
--- a/scripts/disk_check.py
+++ b/scripts/disk_check.py
@@ -40,6 +40,7 @@ UPPER_DIR = "/run/mount/upper"
 WORK_DIR = "/run/mount/work"
 MOUNTS_FILE = "/proc/mounts"
 
+# Threshold of free block counts: On most file systems, the block size is 4096 bytes.
 FREE_SPACE_THRESHOLD = 1024
 
 EVENTS_PUBLISHER_SOURCE = "sonic-events-host"
@@ -66,9 +67,9 @@ def log_debug(m):
     _log_msg(syslog.LOG_DEBUG, "Debug", m)
 
 
-def event_pub():
+def event_pub(event):
     param_dict = FieldValueMap()
-    param_dict["fail_type"] = "read_only"
+    param_dict["fail_type"] = event
     event_publish(events_handle, EVENTS_PUBLISHER_TAG, param_dict)
 
 
@@ -77,14 +78,14 @@ def test_writable(dirs):
         rw = os.access(d, os.W_OK)
         if not rw:
             log_err("{} is not read-write".format(d))
-            event_pub()
+            event_pub("read_only")
             return False
         else:
             log_debug("{} is Read-Write".format(d))
             space = os.statvfs(d)
             if space.f_bavail < FREE_SPACE_THRESHOLD:
-                log_err("{} is no free space".format(d))
-                event_pub()
+                log_err("{} has no free disk space".format(d))
+                event_pub("disk_full")
                 return False
             else:
                 log_debug("{} has enough disk space".format(d))
--- a/scripts/disk_check.py
+++ b/scripts/disk_check.py
@@ -77,7 +77,7 @@ def event_pub(event):
     event_publish(events_handle, EVENTS_PUBLISHER_TAG, param_dict)
 
 
-def test_disk_full(dirs): 
+def test_disk_full(dirs):
     for d in dirs:
         space = os.statvfs(d)
         if space.f_bavail < FREE_SPACE_THRESHOLD:
@@ -141,7 +141,7 @@ def do_mnt(dirs, overlay_prefix):
         os.mkdir(d_upper)
         os.mkdir(d_work)
 
-        ret = run_cmd(["mount", "-t", "overlay", "{}_{}".format(overlay_prefix, d_name),\
+        ret = run_cmd(["mount", "-t", "overlay", "{}_{}".format(overlay_prefix, d_name),
                 "-o", "lowerdir={},upperdir={},workdir={}".format(d, d_upper, d_work), d])
         if ret:
             break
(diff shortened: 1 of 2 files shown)


scripts/disk_check.py
```diff
@@ -79,6 +81,14 @@ def test_writable(dirs):
             return False
         else:
             log_debug("{} is Read-Write".format(d))
+            space = os.statvfs(d)
+            if space.f_bavail < FREE_SPACE_THRESHOLD:
+                log_err("{} is no free space".format(d))
+                event_pub()
```
qiluo-msft, line 87: >event_pub [](http://example.com/codeflow?start=16&length=9)

event_pub()  will send a read_only event, but this is RW branch. Is it conflicting event?

scripts/disk_check.py
```diff
@@ -40,6 +40,8 @@
 WORK_DIR = "/run/mount/work"
 MOUNTS_FILE = "/proc/mounts"
 
+FREE_SPACE_THRESHOLD = 1024
```
qiluo-msft, line 43: >FREE_SPACE_THRESHOLD [](http://example.com/codeflow?start=0&length=20)

what is the unit? could you add code comment?

scripts/disk_check.py
```diff
@@ -79,6 +81,14 @@ def test_writable(dirs):
             return False
         else:
             log_debug("{} is Read-Write".format(d))
+            space = os.statvfs(d)
+            if space.f_bavail < FREE_SPACE_THRESHOLD:
+                log_err("{} is no free space".format(d))
```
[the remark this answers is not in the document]
liuh-80: Fixed

scripts/disk_check.py
liuh-80: Fixed, this needs a schema change: https://github.com/sonic-net/sonic-buildimage/pull/21351

scripts/disk_check.py
liuh-80: Fixed, unit is file system block, on ext4 this means 1024*4096 = 4MB

tests/disk_check_test.py
```diff
@@ -177,6 +180,72 @@ def test_readonly(self, mock_proc, mock_log):
...
+        with patch('sys.argv', ["", "-d", "/tmpx"]):
+            disk_check.main()
+
+    @patch("disk_check.syslog.syslog")
+    @patch("disk_check.subprocess.run")
+    @patch('shutil.rmtree')
+    @patch('os.access', return_value=True)
+    @patch('os.statvfs', return_value=os.statvfs_result((4096, 4096, 1909350, 1491513, 4096,
+                                                         971520, 883302, 883302, 4096, 255)))
+    def test_unmount_disk_full(self, mock_os_statvfs, mock_os_access, mock_rmtree, mock_proc, mock_log):
```
qiluo-msft, line 208: >test_unmount_disk_full [](http://example.com/codeflow?start=8&length=22)

As discussed in internal meeting, disk-full is special compared with disk-readonly is that it can recover to normal (delete large files). Does this testcase cover the recovery?

qiluo-msft:
>     There have been cases, where disk turns Read-only due to kernel bug.

The file level comment (seems a design doc) does not complete anymore, please update.

---
Refers to: scripts/disk_check.py:6 in 52a715e. [](commit_id = 52a715e5b71c2608abdccee9b375af436f6a102a, deletion_comment = False)

tests/disk_check_test.py
liuh-80: Yes, this test will cover recovery case.
I also will add a sonic-mgmt test to cover the disk-full scenario.

scripts/disk_check.py
```diff
@@ -132,13 +153,36 @@ def do_mnt(dirs):
     return ret
 
 
-def is_mounted(dirs):
+def do_unmnt(dirs, overlay_prefix):
+    for d in dirs:
+        d_name = get_dname(d)
+
+        ret = run_cmd(["umount", "-l", "{}_{}".format(overlay_prefix, d_name)])
```
saiarcot895, line 160: It's worth noting that if there's some process in `/etc` or `/home` (including any user's shell being in their home directory), then this will most likely fail. I guess since `disk_check.py` gets executed frequently, it'll eventually work, at least.

saiarcot895 (approved):

scripts/disk_check.py
liuh-80: This will not fail because unmount with -l, lazy mode will wait and unmount in background.

qiluo-msft (commented):
Minor comment on file header comment

liuh-80:
> > ```
> > There have been cases, where disk turns Read-only due to kernel bug.
> > ```
> 
> The file level comment (seems a design doc) does not complete anymore, please update.
> 
> Refers to: scripts/disk_check.py:6 in [52a715e](https://github.com/sonic-net/sonic-utilities/commit/52a715e5b71c2608abdccee9b375af436f6a102a). [](commit_id = [52a715e](https://github.com/sonic-net/sonic-utilities/commit/52a715e5b71c2608abdccee9b375af436f6a102a), deletion_comment = False)

Fixed.

qiluo-msft (approved):

PR #3703, opened 2025-01-03
Fixed show interface counters on supervisor for packet chassis

abdosi:
What I did:

On Packet chassis for supervisor do the following:

`show interfaces counter` gets sats from LC

`show infeface counters -n <> -d all` gets internal port stats

`show infeface counters  -d all` gets internal port stats


abdosi:
@yejianquan for viz.

utilities_common/portstat.py
```diff
@@ -405,7 +408,9 @@ def cnstat_print(self, cnstat_dict, ratestat_dict, intf_list, use_json, print_al
                 print(table_as_json(table, header))
             else:
                 print(tabulate(table, header, tablefmt='simple', stralign='right'))
-        if (multi_asic.is_multi_asic() or device_info.is_chassis()) and not use_json:
+        if device_info.is_voq_chassis():
```
yejianquan, line 411: Does it mean for voq chassis, we don't have internal links, so run with '-d all' makes no differences?

abdosi: @yejianquan : yes for voq chassis we have fabric links [not ethernet links] and we have separate command to get counters for those.

abdosi:
@bingwang-ms

PR #3708, opened 2025-01-09
[QOS] Skip showing unnecessary warning message

vivekrnv:
#### What I did

Don't show an unnecessary warning message (`Operation not completed successfully, please save and reload configuration.`) during config qos reload

**To repro:**
```
root@sonic:/home/admin# sonic-cfggen -H -k Mellanox-SN4700-O8V48 --preset t1 > /tmp/default_config.json

root@sonic:/home/admin# config qos reload
Operation not completed successfully, please save and reload configuration.

Running command: /usr/local/bin/sonic-cfggen -d -t /usr/share/sonic/device/x86_64-mlnx_msn4700-r0/Mellanox-SN4700-O8V48/buffers_dynamic.json.j2,/tmp/cfg_buffer.json -t /usr/share/sonic/device/x86_64-mlnx_msn4700-r0/Mellanox-SN4700-O8V48/qos.json.j2,/tmp/cfg_qos.json -y /etc/sonic/sonic_version.yml
Running command: /usr/local/bin/sonic-cfggen -j /tmp/cfg_buffer.json -j /tmp/cfg_qos.json --write-to-db

Buffer calculation model updated, restarting swss is required to take effect
```

#### How I did it

When in traditional buffer mode, qos reload will not wait for tables to be cleared, so it will set the timeout to 0 in the _wait_until_clear call. But today, this prints an error log and exits.

Thus fixed to not print the error log when the timeout is zero.

#### How to verify it

UT and manual testing

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -793,7 +793,7 @@ def _wait_until_clear(tables, interval=0.5, timeout=30, verbose=False):
                     click.echo("Some entries matching {} still exist: {}".format(table, keys[0]))
                 time.sleep(interval)
         empty = (non_empty_table_count == 0)
-    if not empty:
+    if not empty and timeout:
         click.echo("Operation not completed successfully, please save and reload configuration.")
     return empty
 
--- a/config/main.py
+++ b/config/main.py
@@ -793,7 +793,9 @@ def _wait_until_clear(tables, interval=0.5, timeout=30, verbose=False):
                     click.echo("Some entries matching {} still exist: {}".format(table, keys[0]))
                 time.sleep(interval)
         empty = (non_empty_table_count == 0)
-    if not empty and timeout:
+    if timeout == 0:
+        empty = True
+    if not empty:
         click.echo("Operation not completed successfully, please save and reload configuration.")
     return empty
 
--- a/config/main.py
+++ b/config/main.py
@@ -778,6 +778,8 @@ def storm_control_delete_entry(port_name, storm_type):
 
 
 def _wait_until_clear(tables, interval=0.5, timeout=30, verbose=False):
+    if timeout == 0:
+        return True
     start = time.time()
     empty = False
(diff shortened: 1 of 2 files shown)


stephenxs (approved):

config/main.py
```diff
@@ -793,7 +793,7 @@ def _wait_until_clear(tables, interval=0.5, timeout=30, verbose=False):
                     click.echo("Some entries matching {} still exist: {}".format(table, keys[0]))
                 time.sleep(interval)
         empty = (non_empty_table_count == 0)
-    if not empty:
+    if not empty and timeout:
```
bingwang-ms, line 796: Suggest checking `timeout != 0`

vivekrnv: I've slightly updated the logic to also return True in case timeout is 0. 

This is to be compatible with the following PR: https://github.com/sonic-net/sonic-utilities/pull/3710

liat-grozovik:
@bingwang-ms @kperumalbfn
can you please review following comments handling?

config/main.py
```diff
@@ -793,6 +793,8 @@ def _wait_until_clear(tables, interval=0.5, timeout=30, verbose=False):
                     click.echo("Some entries matching {} still exist: {}".format(table, keys[0]))
                 time.sleep(interval)
         empty = (non_empty_table_count == 0)
+    if timeout == 0:
+        empty = True
```
bingwang-ms, line 797: This code looks a little wired. How about checking the value of `timeout` and return `True` directly at the beginning of the function of ` _wait_until_clear`? Or just skip calling this function `_wait_until_clear` when `timeout == 0`?

vivekrnv: Updated, Please check

vivekrnv: @bingwang-ms Can we signoff this PR?

bingwang-ms (approved):

PR #3711, opened 2025-01-11
Add recover asic_id config if load golden config.

xincunli-sonic:
#### What I did

Restore hardware related configuration: `asic_id` from old config or local file (if not exist) since the golden config does not have them.

#### How I did it

Add a function to backfill the field `asic_id` to config db by old config or read local files (if not exist) afterload golden config.

#### How to verify it

Using lab device to download golden config which does not have these fields, and apply through load_minigraph with overridde golden config option, then check configdb dump if asic_id get populated correctly.


config/main.py
```diff
@@ -2187,6 +2194,68 @@ def override_config_by(golden_config_path):
     return
 
 
+def recover_hw_config(namespace_list=None):
```
wen587, line 2197: mac and platform is already supported. Can generate asic id here: https://github.com/sonic-net/sonic-utilities/pull/3711

config/main.py
```diff
@@ -2187,6 +2194,68 @@ def override_config_by(golden_config_path):
...
+            device_metadata["platform"] = platform
+
+            hostname = parse_hostname(MINIGRAPH_FILE)
+            asic_role = parse_asic_sub_role(MINIGRAPH_FILE, scope)
+            switch_type = parse_asic_switch_type(MINIGRAPH_FILE, scope, hostname)
+
+            if (switch_type and switch_type.lower() == "chassis-packet") or \
+               (asic_role and asic_role.lower() == "backend") or \
+               (platform == device_info.VS_PLATFORM):
+                mac = device_info.get_system_mac(scope=scope, hostname=hostname)
```
qiluo-msft, line 2228: >mac [](http://example.com/codeflow?start=16&length=3)

Can you reuse load_sysinfo_if_missing?

config/main.py
```diff
@@ -2187,6 +2194,68 @@ def override_config_by(golden_config_path):
...
+                else ConfigDBConnector(use_unix_socket_path=True, scope=scope)
+            config_db.connect()
+
+            device_metadata = config_db.get_entry("DEVICE_METADATA", "localhost") or {}
+            platform = device_info.get_platform()
+            device_metadata["platform"] = platform
+
+            hostname = parse_hostname(MINIGRAPH_FILE)
+            asic_role = parse_asic_sub_role(MINIGRAPH_FILE, scope)
+            switch_type = parse_asic_switch_type(MINIGRAPH_FILE, scope, hostname)
```
arlakshm, line 2223: why do need to get this information from minigraph.xml. If the `DEVICE_METADATA` table is supported then this info is available in config_db?

xincunli-sonic: Refactored.

config/main.py
xincunli-sonic: Refactored.

config/main.py
```diff
@@ -2199,12 +2199,15 @@ def generate_sysinfo(cur_config, config_input, ns=None):
 
     mac = None
     platform = None
+    asic_id = None
     cur_device_metadata = cur_config.get('DEVICE_METADATA')
 
     # Reuse current config's mac and platform. Generate if absent
```
qiluo-msft, line 2205: >platform [](http://example.com/codeflow?start=37&length=8)

minor comment: this code comment should be updated as code, or make it more generic.

qiluo-msft (approved):

PR #3712, opened 2025-01-13
Remove partially installer image when image install failed.

liuh-80:
Remove partially installer image when image install failed.

#### What I did
When install image failed, partially installed image not removed, which may cause disk full issue on small disk device.

#### How I did it
Handle install image SystemExit exception and uninstall image.

#### How to verify it
Manually verify.
Pass all test case

##### Work item tracking
- Microsoft ADO: 30611724

### DIFF
--- a/sonic_installer/main.py
+++ b/sonic_installer/main.py
@@ -591,7 +591,14 @@ def install(url, force, skip_platform_check=False, skip_migration=False, skip_pa
 
         echo_and_log("Installing image {} and setting it as default...".format(binary_image_version))
         with SWAPAllocator(not skip_setup_swap, swap_mem_size, total_mem_threshold, available_mem_threshold):
-            bootloader.install_image(image_path)
+            try:
+                bootloader.install_image(image_path)
+            except SystemExit as e:
+                # When install image failed with SystemExit, image is partial installed
+                echo_and_log('Install image failed with exception: {}'.format(e))
+                bootloader.remove_image(binary_image_version)
+                raise e
+
         # Take a backup of current configuration
         if skip_migration:
             echo_and_log("Skipping configuration migration as requested in the command option.")
--- a/sonic_installer/main.py
+++ b/sonic_installer/main.py
@@ -593,10 +593,14 @@ def install(url, force, skip_platform_check=False, skip_migration=False, skip_pa
         with SWAPAllocator(not skip_setup_swap, swap_mem_size, total_mem_threshold, available_mem_threshold):
             try:
                 bootloader.install_image(image_path)
-            except SystemExit as e:
-                # When install image failed with SystemExit, image is partial installed
-                echo_and_log('Install image failed with exception: {}'.format(e))
-                bootloader.remove_image(binary_image_version)
+            except BaseException as e:
+                # When install image failed with exception, image is partial installed
+                # Fully installed image will update SONiC environment by update_sonic_environment
+                # For partial installed image, only need delete image folder
+                echo_and_log('Image install failed with exception: {}'.format(e))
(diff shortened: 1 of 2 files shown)


sonic_installer/main.py
```diff
@@ -591,7 +591,18 @@ def install(url, force, skip_platform_check=False, skip_migration=False, skip_pa
...
+                bootloader.install_image(image_path)
+            except BaseException as e:
+                # When install image failed with exception, image is partial installed
+                # Fully installed image will update SONiC environment by update_sonic_environment
+                # For partial installed image, only need delete image folder
+                echo_and_log('Image install failed with exception: {}'.format(e))
+                echo_and_log('Delete partial installed image: {}'.format(binary_image_version))
+                new_image_dir = bootloader.get_image_path(binary_image_version)
+                run_command(["rm", "-rf", new_image_dir])
+                raise e
```
qiluo-msft, line 604: >raise e [](http://example.com/codeflow?start=16&length=7)

just `raise` is better, it keep original stack info.

sonic_installer/main.py
```diff
@@ -591,7 +591,18 @@ def install(url, force, skip_platform_check=False, skip_migration=False, skip_pa
...
+            try:
+                bootloader.install_image(image_path)
+            except BaseException as e:
+                # When install image failed with exception, image is partial installed
+                # Fully installed image will update SONiC environment by update_sonic_environment
+                # For partial installed image, only need delete image folder
+                echo_and_log('Image install failed with exception: {}'.format(e))
+                echo_and_log('Delete partial installed image: {}'.format(binary_image_version))
+                new_image_dir = bootloader.get_image_path(binary_image_version)
+                run_command(["rm", "-rf", new_image_dir])
```
qiluo-msft, line 603: >run_command [](http://example.com/codeflow?start=16&length=11)

possible to use python function? `rm-rf` is always dangerous.

sonic_installer/main.py
```diff
@@ -591,7 +591,18 @@ def install(url, force, skip_platform_check=False, skip_migration=False, skip_pa
 
         echo_and_log("Installing image {} and setting it as default...".format(binary_image_version))
         with SWAPAllocator(not skip_setup_swap, swap_mem_size, total_mem_threshold, available_mem_threshold):
-            bootloader.install_image(image_path)
+            try:
+                bootloader.install_image(image_path)
+            except BaseException as e:
```
qiluo-msft, line 596: >BaseException [](http://example.com/codeflow?start=19&length=13)

Can we use more specific exception type relating disk issue?

sonic_installer/main.py
qiluo-msft: Should we use `bootloader.remove_image` ?

liuh-80: We can't because remove_image need check bootloader info, but install failed at here doesn't update bootloader.

The bootloader will be update by line 612:
update_sonic_environment(bootloader, binary_image_version)

A partial installed image failed here only extract files to /host/image-version folder, so we only need to delete that folder.

sonic_installer/main.py
liuh-80: Fixed

sonic_installer/main.py
liuh-80: Fixed, install image failed with throw SystemExit exception.

sonic_installer/main.py
qiluo-msft: how about `shutil.rmtree` ?

liuh-80: Fixed

qiluo-msft (approved):

PR #3717, opened 2025-01-14
Service validator vul

mohmamedali:
#### What I did
"os.system()" uses shell invocation to execute command which is dangerous. without a static string that can lead to command injection.
#### How I did it
pass use list of strings to subprocess() - use shell=False instead.
#### How to verify it
pass UT

### DIFF
--- a/generic_config_updater/services_validator.py
+++ b/generic_config_updater/services_validator.py
@@ -1,6 +1,6 @@
 import time
 from .gu_common import genericUpdaterLogging
-from subprocess import call 
+from subprocess import call
 
 logger = genericUpdaterLogging.get_logger(title="Service Validator")
 
--- a/generic_config_updater/services_validator.py
+++ b/generic_config_updater/services_validator.py
@@ -1,6 +1,6 @@
-import os
 import time
 from .gu_common import genericUpdaterLogging
+import subprocess
 
 logger = genericUpdaterLogging.get_logger(title="Service Validator")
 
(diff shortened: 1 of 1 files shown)


FengPan-Frank (dismissed):
LGTM

generic_config_updater/services_validator.py
```diff
@@ -16,17 +16,25 @@ def set_verbose(verbose=False):
...
+    try:
+        subprocess.check_call(cmd)
+    except Exception as e:
+        return e.returncode
+    return 0
+
+
 def _service_restart(svc_name):
-    rc = os.system(f"systemctl restart {svc_name}")
+    rc = run_command(['systemctl', 'restart', svc_name])
```
maipbui, line 28: >rc = run_command(['systemctl', 'restart', svc_name]) [](http://example.com/codeflow?start=4&length=52)

I think simply replace `os.system` with `subprocess.call`, as suggested in python doc [https://docs.python.org/3/library/subprocess.html#replacing-os-system](https://docs.python.org/3/library/subprocess.html#replacing-os-system), no need to introduce new function `run_command`

mohmamedali: Yes, Done!

mohmamedali: Yes, Done!

maipbui (approved):

maipbui:
LGTM

PR #3720, opened 2025-01-16
remove os.system duo to security vulnerabilities

MSaid-dev:
What I did
"os.system()" uses shell invocation to execute dangerous commands without a static string that can lead to command injection.

How I did it
pass list of strings to subprocess() - use shell=False instead.

How to verify it
pass UT


FengPan-Frank:
thanks for fixing this vulnerable issue, pls follow up with test failure

sunhorus:
All checks passed and also unit tests

FengPan-Frank (dismissed):
LGTM

config/main.py
```diff
@@ -3629,8 +3629,8 @@ def add_snmp_agent_address(ctx, agentip, port, vrf):
     config_db.set_entry('SNMP_AGENT_ADDRESS_CONFIG', key, {})
 
     #Restarting the SNMP service will regenerate snmpd.conf and rerun snmpd
-    cmd="systemctl restart snmp"
-    os.system (cmd)
+    cmd = ["systemctl", "restart", "snmp"]
+    getstatusoutput_noshell(cmd)
```
maipbui, line 3633: >getstatusoutput_noshell(cmd) [](http://example.com/codeflow?start=4&length=28)

you can replace `os.system` with `subprocess.call`, suggested in python doc [https://docs.python.org/3/library/subprocess.html#replacing-os-system](https://docs.python.org/3/library/subprocess.html#replacing-os-system)

MSaid-dev: Done

PR #3723, opened 2025-01-21
Fix call for spanning-tree commands in dump script

DavidZagury:
#### What I did
Fix the calls for spanning-tree commands in dump script.
During call to generate techsupport, we can see that the spanning tree commands fail:
```
.
Error: No such command "spanning_tree".
timeout --foreground 5m bash -c "dummy_cleanup_method ()
.
```
#### How I did it
Change from show spanning_tree to the actual command show spanning-tree
#### How to verify it
Call show techsupport

### DIFF
--- a/scripts/generate_dump
+++ b/scripts/generate_dump
@@ -2209,10 +2209,10 @@ main() {
     wait
 
  	save_cmd "stpctl all" "stp.log"
-    save_cmd "show spanning_tree" "stp.show"
-    save_cmd "show spanning_tree statistics" "stp.stats"
-    save_cmd "show spanning_tree bpdu_guard" "stp.bg"
-    save_cmd "show spanning_tree root_guard" "stp.rg"
+    save_cmd "show spanning-tree" "stp.show"
+    save_cmd "show spanning-tree statistics" "stp.stats"
+    save_cmd "show spanning-tree bpdu_guard" "stp.bg"
+    save_cmd "show spanning-tree root_guard" "stp.rg"
 
     save_cmd "ps aux" "ps.aux" &
     save_cmd "top -b -n 1" "top" &


FengPan-Frank (approved):
Seems we'll need some test for generate_dump flow.

qiluo-msft:
@DavidZagury could you add a sonic-mgmt or unit test to prevent regression like this?

liat-grozovik (approved):

liat-grozovik:
@DavidZagury i wonder what was the PRs that changed the spanning tree and didnt catch it, can you share it?
this is a degradation which should be avoided in the PR review and the PR checkers of the original one. Maybe we are missing a unit test or a test?

DavidZagury:
> @DavidZagury i wonder what was the PRs that changed the spanning tree and didnt catch it, can you share it? this is a degradation which should be avoided in the PR review and the PR checkers of the original one. Maybe we are missing a unit test or a test?
@liat-grozovik the PR the introduced this change in sonic-utilities was https://github.com/sonic-net/sonic-utilities/pull/3567 by @divyachandralekha 

There are no unit test for this area in the code.
There is a sonic-mgmt test, that touch this area, but it only check that a specific set of commands run successfully, and does not check if all commands that run during techsupport has passed

DavidZagury:
> @DavidZagury could you add a sonic-mgmt or unit test to prevent regression like this?

@qiluo-msft 
At the moment we don't have capacity to handle such request. This bug was created from a PR from broadcom ( https://github.com/sonic-net/sonic-utilities/pull/3567 ) maybe they could create a UT in this area

wajahatrazi:
Hi @DavidZagury , 

Just wondering the steps you took and why regarding this error. Can you share the process? Would be helpful in upcoming PRs.

DavidZagury:
> Hi @DavidZagury ,
> 
> Just wondering the steps you took and why regarding this error. Can you share the process? Would be helpful in upcoming PRs.

Hi @wajahatrazi , after I saw that when running show techsupport there is the error which I shared in the PR:
```
.
Error: No such command "spanning_tree".
timeout --foreground 5m bash -c "dummy_cleanup_method ()
.
```
I check the generate dump script, saw that the relevant CLIs are:
```
show spanning_tree
show spanning_tree statistics
show spanning_tree bpdu_guard
show spanning_tree root_guard
```
I tried them manually on a switch, and saw that indead this CLIs don't work.
To understand why it was added, when and by who, I have check the PR which added these lines, and saw that it also created the show command.

The show command has been defined:
https://github.com/sonic-net/sonic-utilities/blob/9d273f16250c3ee0939838f256cd23e39b72c66f/show/main.py#L1998-L2000

In SONiC we use click to create the CLIs, and I know that when click see a command with underscore as definition, the created CLI is with dash instread, as we can see in the release note from version 7.0.0:

> Subcommands that are named by the function now automatically have the underscore replaced with a dash. If you register a function named my_command it becomes my-command in the command line interface.
https://click.palletsprojects.com/en/stable/changes/#version-7-0

I have check running the CLIs with the correct name, and the error did not appear.
I have updated the generate dump script, and made sure there are no error.

PR #3750, opened 2025-02-05
display proper message with proper errno for kvm.

hdwhdw:
#### What I did
In KVM, there is no system EEPROM. decode-syseeprom should display the information and return the correct errno (ENODEV).

Currently, on KVM, decode-syseeprom will return OK (0) even if there is no system EEPROM.

#### How I did it
Check if the platform is kvm, if so, log and return.

#### How to verify it
```
root@vlab-01:~# decode-syseeprom
Platform x86_64-kvm_x86_64-r0 does not support EEPROM
root@vlab-01:~# echo $?
19
root@vlab-01:~# python3 -c 'import os, sys; print(os.strerror(int(sys.argv[1])))' 19
No such device
```

```
admin@vlab-01:~$ sudo decode-syseeprom -s
Failed to read system EEPROM info
admin@vlab-01:~$ echo $?
0
```


scripts/decode-syseeprom
```diff
@@ -238,6 +239,12 @@ def main():
     if any(re.match(p, platform) for p in platforms_without_eeprom_db):
         support_eeprom_db = False
 
+    # Currently, kvm does not support eeprom
+    platforms_without_eeprom = ['.*kvm.*']
+    if any(re.match(p, platform) for p in platforms_without_eeprom):
+        print('Platform {} does not support EEPROM'.format(platform))
```
zbud-msft, line 245: Should we log this error instead?

zbud-msft (approved):

FengPan-Frank (approved):

PR #3755, opened 2025-02-07
Issue #3754: Modify AAA config commands to use pass_db decorator

### LINKED ISSUE #3754: [bug] AAA tests do not handle mock DB connection correctly
Many test cases in [sonic-utilities/tests/aaa_test.py ](https://github.com/sonic-net/sonic-utilities/blob/master/tests/aaa_test.py) and [sonic-utilities/tests/radius_test.py](https://github.com/sonic-net/sonic-utilities/blob/master/tests/radius_test.py) are handling mock database connections incorrectly. The state configured by AAA config commands is not the same state used by the AAA show commands that are being used as validation. Instead, desired state is manually fed into AAA show commands to produce the desired results. This is a clear break from the prevalent pattern in sonic-utilities/tests in which config commands are verified with their accompanying show commands.

THE DETAILS

`ConfigDBConnector` is built so that each instance will connect to the DB separately; users can instantiate a single `ConfigDBConnector` and use it indefinitely, or create a new one every time a DB operation is required. Config commands are each run as a separate process, so each command will have a different `ConfigDBConnector`. In [sonic-utilities/config/main.py](https://github.com/sonic-net/sonic-utilities/blob/master/config/main.py), the `ConfigDBConnector` is instantiated at a broad command level and then passed down to subcommands as needed (using the [pass_db decorator](https://github.com/sonic-net/sonic-utilities/blob/7f3957c43664082ee85d0e4b334d6b838ef861a4/utilities_common/cli.py#L20)). However, in [sonic-utilities/config/aaa.py](https://github.com/sonic-net/sonic-utilities/blob/master/config/aaa.py), no `ConfigDBConnector` is passed in, and each command creates its own link to the DB. This works fine in the actual production system, since each `ConfigDBConnector` ends up connecting to the same DB.

The pytest framework used for sonic-utilities breadth tests does not have an actual DB instantiated at all. Instead, [sonic-utilities/tests/mock_files/dbconnector.py](https://github.com/sonic-net/sonic-utilities/blob/master/tests/mock_tables/dbconnector.py) is imported at the beginning of the test. That file defines a class `SwssSyncClient` which mocks `RedisDB`. Each instance of `SwssSyncClient` is unique, and will have it's own mocked versions of Sonic tables. For most config commands this works perfectly fine; because the top-level command is responsible for creating the connection, subcommands can easily be tested by creating a `ConfigDBConnector` connected to a mock `SwssSyncClient` and then passing that same `ConfigDBConnector` into each config command in turn. Show commands are used to validate the config changes, and since these show commands are passed the same `ConfigDBConnector` object, they work as expected.

In AAA tests, however, this does not work the same way. There is no way to pass a `ConfigDBConnector` object into AAA commands, so instead each command will make its own `ConfigDBConnector`, and each of those will have a distinct `SwssSyncClient`. Both of those objects will only last for the duration of the command, and no outside state will be set. The existing AAA tests have worked around this issue by running the config command, then creating a new `ConfigDBConnector` and manually adding whatever the config command should have added, then running the show command and asserting that the output matches what is expected. Here's an example from  [sonic-utilites/tests/aaa_test.py](https://github.com/sonic-net/sonic-utilities/blob/master/tests/aaa_test.py):

```
def test_config_aaa_radius(self, get_cmd_module):
    (config, show) = get_cmd_module
    runner = CliRunner()
    db = Db()
    db.cfgdb.delete_table("AAA")

    result = runner.invoke(config.config.commands["aaa"],\
                           ["authentication", "login", "radius"], obj=db)
    print(result.exit_code)
    print(result.output)
    assert result.exit_code == 0
    assert result.output == config_aaa_empty_output

    db.cfgdb.mod_entry("AAA", "authentication", {'login' : 'radius'})

    result = runner.invoke(show.cli.commands["aaa"], [], obj=db)
    assert result.exit_code == 0
    assert result.output == show_aaa_radius_output

    result = runner.invoke(config.config.commands["aaa"],\
                           ["authentication", "login", "default"], obj=db)
    print(result.exit_code)
    print(result.output)
    assert result.exit_code == 0
    assert result.output == config_aaa_empty_output

    db.cfgdb.delete_table("AAA")

    result = runner.invoke(show.cli.commands["aaa"], [], obj=db)
    assert result.exit_code == 0
    assert result.output == show_aaa_default_output
```

This follows the standard approach for sonic-utilities tests, dutifully passing in the created 'db' object. But the config command ignores that object, writes to its own short-lived mock db, and returns. The test is forced to make calls to `mod_entry` and `delete_entry` to actually set the correct table state. This means that the only things being tested for AAA config commands are:

- that the return code is 0, meaning that the command syntax is correct and the command was accepted
- that the command returned no output, which is expected for config commands

That's it. There is no validation of the actual function of the command. While it's possible that the test writers are intentionally only wanting to test the command syntax, it seems extremely unlikely given that the rest of the tests in sonic-utilities lean heavily on this pattern to test config command internals.

Either the test infrastructure must be modified to match the behavior of the underlying implementation when handling multiple mocked db connections, or the AAA config commands need to be changed to follow the same style used in other cli commands, where the database connection is passed down from the top-level config command.

Regardless of which approach is taken, the manual workaround must also be removed from all test cases where it is present, and any lurking errors resulting from the test gap will need to be addressed.

anders-nexthop:
#### What I did

Currently all AAA config commands create a new `ConfigDBConnector` object for every operation. This doesn't interact well with the pytest mock infrastructure used in the unit tests. I changed the AAA config commands to accept a `ConfigDBConnector` from the top-level command, which follows the standard pattern used by other config commands (such as those in [sonic-utilities/config/main.py](https://github.com/sonic-net/sonic-utilities/blob/master/config/main.py)).

I fixed up tests so that they actually test config functionality, instead of manually manipulating the mock DB to return valid show results.

Some expected output in tests was not in line with actual product output. I updated the expected output in these cases.

fixes #3754 

#### How I did it

Convert AAA config commands to use the pass_db decorator pattern, which is used by most other commands in sonic-utilities. This decorator allows the ConfigDBConnector to be passed in directly.

I went through [sonic-utilities/tests/aaa_test.py ](https://github.com/sonic-net/sonic-utilities/blob/master/tests/aaa_test.py) and [sonic-utilities/tests/radius_test.py](https://github.com/sonic-net/sonic-utilities/blob/master/tests/radius_test.py), and removed all instances where manual database manipulation was being done unnecessarily (such as [here](https://github.com/sonic-net/sonic-utilities/blob/a7deb8c0f9b36992d0d401d611308bd1e50458b0/tests/aaa_test.py#L137), for example).

#### How to verify it

Removing the db modification from the AAA tests causes them to fail before my changes. After my changes, all AAA tests pass.

### DIFF
--- a/config/aaa.py
+++ b/config/aaa.py
@@ -16,8 +16,8 @@ def is_secret(secret):
     return bool(re.match('^' + '[^ #,]*' + '$', secret))
 
 
-def add_table_kv(table, entry, key, val):
-    config_db = ValidatedConfigDBConnector(ConfigDBConnector())
+def add_table_kv(db, table, entry, key, val):
+    config_db = ValidatedConfigDBConnector(db.cfgdb)
     config_db.connect()
     try:
         config_db.mod_entry(table, entry, {key:val})
@@ -26,8 +26,8 @@ def add_table_kv(table, entry, key, val):
         ctx.fail("Invalid ConfigDB. Error: {}".format(e))
 
 
-def del_table_key(table, entry, key):
-    config_db = ValidatedConfigDBConnector(ConfigDBConnector())
+def del_table_key(db, table, entry, key):
+    config_db = ValidatedConfigDBConnector(db.cfgdb)
     config_db.connect()
     data = config_db.get_entry(table, entry)
     if data:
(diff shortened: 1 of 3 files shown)


anders-nexthop:
I filled in some missing test cases to pass the code coverage threshold.

anders-nexthop:
@qiluo-msft can you give this change a review when you get a chance?

anders-nexthop:
@liuh-80 can you please review this PR?

qiluo-msft:
How to verify it? Could you also run these command "config aaa *"  on a DUT, and run some sonic-mgmt test to verify the basic functionality?

qiluo-msft:
@a-barboza Could you help review?

qiluo-msft (approved):
lgtm

a-barboza:
> @a-barboza @dbarashinvd Could you help review?

@asha-behera @adyeung

anders-nexthop:
> How to verify it? Could you also run these command "config aaa *" on a DUT, and run some sonic-mgmt test to verify the basic functionality?

```
admin@gold108-dut:~$ sudo config aaa authentication debug enable
admin@gold108-dut:~$ sudo config aaa authentication failthrough enable
admin@gold108-dut:~$ sudo config aaa authentication fallback enable
admin@gold108-dut:~$ sudo config aaa authentication trace enable
admin@gold108-dut:~$ show aaa
AAA authentication login local (default)
AAA authentication failthrough True
AAA authentication debug True
AAA authentication fallback True
AAA authentication trace True
AAA authorization login local (default)
AAA accounting login disable (default)

admin@gold108-dut:~$ sudo config aaa authentication debug disable
admin@gold108-dut:~$ sudo config aaa authentication failthrough disable
admin@gold108-dut:~$ sudo config aaa authentication fallback disable
admin@gold108-dut:~$ sudo config aaa authentication trace disable
admin@gold108-dut:~$ show aaa
AAA authentication login local (default)
AAA authentication failthrough False
AAA authentication debug False
AAA authentication fallback False
AAA authentication trace False
AAA authorization login local (default)
AAA accounting login disable (default)

admin@gold108-dut:~$ sudo config aaa authentication debug default
admin@gold108-dut:~$ sudo config aaa authentication failthrough default
admin@gold108-dut:~$ sudo config aaa authentication fallback default
admin@gold108-dut:~$ sudo config aaa authentication trace default
admin@gold108-dut:~$ show aaa
AAA authentication login local (default)
AAA authentication failthrough False (default)
AAA authorization login local (default)
AAA accounting login disable (default)

```
We have been running sonic-mgmt tests locally with these changes for a couple of months now and have not noticed any issues. Are there any specific tests you would like me to run?

PR #3766, opened 2025-02-13
sonic-utilities: port to libyang3

bradh352:
#### What I did

remove direct dependencies on libyang, and instead rely on functionality provided by sonic-yang-mgmt.

This has dependencies on other PRs in other repositories, see primary tracking https://github.com/sonic-net/sonic-buildimage/issues/22385

Fixes https://github.com/sonic-net/sonic-buildimage/issues/22385

#### How I did it

Modifications to sonic-yang-mgmt in https://github.com/sonic-net/sonic-buildimage/pull/21716 added public functions and enhancements to allow us to not need to call into libyang, thus making this code not reliant on a given version of libyang (however the new sonic-yang-mgmt code does depend on libyang3, so this is not compatible with libyang1, the goal is to limit the number of packages directly using libyang to make future upgrading easier).
The find_data_dependencies() logic has been simplified due to enhancements in sonic-yang-mgmt in specifying schema or partial paths rather than data node paths able to return all nodes with leafrefs to the specified path.  This is not only more simple, but a lot faster.

#### How to verify it

Pull all PRs from https://github.com/sonic-net/sonic-buildimage/pull/21679 into a local build system and build and test

### DIFF
--- a/generic_config_updater/gu_common.py
+++ b/generic_config_updater/gu_common.py
@@ -144,9 +144,7 @@ class ConfigWrapper:
                                         self.validate_lanes]
 
         try:
-            tmp_config_db_as_json = copy.deepcopy(config_db_as_json)
-
-            sy.loadData(tmp_config_db_as_json)
+            sy.loadData(config_db_as_json)
 
             sy.validate_data_tree()
 
@@ -319,10 +317,10 @@ class ConfigWrapper:
     def crop_tables_without_yang(self, config_db_as_json):
         sy = self.create_sonic_yang_with_loaded_models()
 
-        sy.jIn = copy.deepcopy(config_db_as_json)
-
+        # Current sonic-yang-mgmt guarantees _cropConfigDB() will deep copy if
+        # it needs to modify.
+        sy.jIn = config_db_as_json
         sy.tablesWithOutYang = dict()
-
         sy._cropConfigDB()
 
         return sy.jIn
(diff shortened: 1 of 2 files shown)


generic_config_updater/gu_common.py
```diff
@@ -615,11 +614,7 @@ def _find_leafref_paths(self, path, config):
 
         xpath = self.convert_path_to_xpath(path, config, sy)
 
-        leaf_xpaths = self._get_inner_leaf_xpaths(xpath, sy)
-
-        ref_xpaths = []
```
FengPan-Frank, line 620: why removing this line?

bradh352: previously sonic-yang-mgmt's find_data_dependencies() expected an exact node to find dependencies for ... however the *new* version allows you to specify an ancestor node in the hierarchy and it will return all nodes that reference child nodes of the provided path.   

That means that this code is no longer needing to do a complex tree recursion and never calls into libyang directly (which is what `sy.root.tree_for()` and `node.tree_dfs()` were doing, but instead only uses functionality provided by sonic-yang-mgmt.

This is orders of magnitude more efficient due to it being implemented directly in sonic-yang-mgmt, and also removes dependencies on the libyang version in use.

bradh352: See sonic-yang-mgmt PR: https://github.com/sonic-net/sonic-buildimage/pull/21716

And in particular you can see the new test case added in https://github.com/sonic-net/sonic-buildimage/pull/21716/files#diff-fe704fab7611ef139a60d6d1b0c03dfde3ba4c09d7326a9c8f10132808614a99

(under "dependencies" when using the ancestor node xpath "/test-acl:acl/ACL_TABLE" which tests this functionality)

bradh352:
closing in favor of #4118

PR #3769, opened 2025-02-18
Fix the switch level Voq counter to use switch_id Oid instead of counter name for PACKET_INTEGRITY counter

saksarav-nokia:
#### What I did
Changed the switch level Voq counter to use switch_id Oid instead of counter name for PACKET_INTEGRITY counter
#### How I did it
Changed the orchagent to use switch level Voq counter to use Switch_id Oid instead of counter name for PACKET_INTEGRITY counter. So updated the corresponding changes in sonic-utilities.
#### How to verify it
Verified with show drop count and mock test
No change in command output
No change in command output


saksarav-nokia:
@vmittal-msft

rlhui:
@vmittal-msft  please help file a msft ado

vmittal-msft:
@saksarav-nokia  changes looks ok. Did you verify sonic command to make sure switch level as well as port level drops are still being shown in output ?

vmittal-msft (approved):

saksarav-nokia:
> @saksarav-nokia changes looks ok. Did you verify sonic command to make sure switch level as well as port level drops are still being shown in output ?

Yes. I Also manaully added switch level debug counters to the COUNTER_DB and then verified the show commands.  Also the vs tests in sonic-utilities passed

SuvarnaMeenakshi (approved):

PR #3770, opened 2025-02-19
Add golden config check

xincunli-sonic:
#### What I did
Added golden config file content check in config_file_yang_validation.

#### How I did it
The function reads a JSON configuration file.
It checks if the configuration is empty or not a dictionary.
If the device is multi-ASIC, it verifies that all required namespaces exist in the configuration.

#### How to verify it

Patch change into testbed, given below cases to verify
1. Empty file
2. Not dictionary
3. Empty dictionary
4. Missing namespace if platform is multi asic.
5. Normal golden config.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1396,15 +1396,15 @@ def config_file_yang_validation(filename):
     # Check if the config is empty
     if not config:
         return
-    
+
     # Check if the config is not a dictionary
     if not isinstance(config, dict):
         return
-    
+
     # Check if the config is an empty dictionary
     if not config:
         return
-    
+
     # If the device is multi-ASIC, check if all required namespaces exist
     if multi_asic.is_multi_asic():
         required_namespaces = [HOST_NAMESPACE, *multi_asic.get_namespace_list()]
--- a/config/main.py
+++ b/config/main.py
@@ -1392,6 +1392,27 @@ def multiasic_write_to_db(filename, load_sysinfo):
 
 def config_file_yang_validation(filename):
     config = read_json_file(filename)
+
+    # Check if the config is empty
+    if not config:
+        return
+    
+    # Check if the config is not a dictionary
+    if not isinstance(config, dict):
+        return
+    
+    # Check if the config is an empty dictionary
+    if not config:
+        return
+    
+    # If the device is multi-ASIC, check if all required namespaces exist
+    if multi_asic.is_multi_asic():
+        required_namespaces = [HOST_NAMESPACE, *multi_asic.get_namespace_list()]
+        for ns in required_namespaces:
+            asic_name = HOST_NAMESPACE if ns == DEFAULT_NAMESPACE else ns
+            if asic_name not in config:
+                return
+
     sy = sonic_yang.SonicYang(YANG_DIR)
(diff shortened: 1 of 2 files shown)


config/main.py
```diff
@@ -1392,6 +1392,27 @@ def multiasic_write_to_db(filename, load_sysinfo):
 
 def config_file_yang_validation(filename):
     config = read_json_file(filename)
+
```
wen587, line 1395: This line will also fail with empty config: https://github.com/sonic-net/sonic-utilities/blob/a87d4cc74de9a23e3a5e6da388ec4fd9dd1e6f28/config/main.py#L2081

config/main.py
```diff
@@ -1392,6 +1392,27 @@ def multiasic_write_to_db(filename, load_sysinfo):
...
+    if not config:
+        return
+
+    # Check if the config is not a dictionary
+    if not isinstance(config, dict):
+        return
+
+    # Check if the config is an empty dictionary
+    if not config:
+        return
```
wen587, line 1406: duplicate with 1396

config/main.py
```diff
@@ -2048,7 +2066,10 @@ def load_minigraph(db, no_service_restart, traffic_shift_away, override_config,
                         fg='magenta')
             raise click.Abort()
 
-        config_file_yang_validation(golden_config_path)
+        if not config_file_yang_validation(golden_config_path):
+            click.secho("Invalid golden config file:'{}'!".format(golden_config_path),
+                        fg='magenta')
+            raise click.Abort()
 
         config_to_check = read_json_file(golden_config_path)
```
wen587, line 2074: {} in golden config will also fail in line 2081

xincunli-sonic: no, if we checked config_file_yang_validation return value, it will not go to line 2081.

wen587: Hi @xincunli-sonic ,  I don't think we should fial it when golden config is {}.
It is a valid input file and it means nothing to override. There is cases that we just provide {} in testbed.

config/main.py
```diff
@@ -1393,8 +1393,8 @@ def multiasic_write_to_db(filename, load_sysinfo):
 def config_file_yang_validation(filename):
     config = read_json_file(filename)
 
-    # Check if the config is empty
-    if not config or not isinstance(config, dict):
+    # Check if the config is not a dictionary
+    if not isinstance(config, dict):
```
wen587, line 1397: In line 1416 it will fail when config_to_check is None.
Also it will fail in line 2027:
https://github.com/sonic-net/sonic-utilities/blob/bf7bd4a655ee7be1feaf2d81c41336c55009b19b/config/main.py#L2077

How about add check before:
line 1416
```

   for scope in asic_list:
        config_to_check = config.get(scope) if multi_asic.is_multi_asic() else config
        if not config_to_check:
               try:
```
line 2027
```
for scope in asic_list:
       host_config = config_to_check.get(scope) if multi_asic.is_multi_asic() else config_to_check
       if not host_config:
            table_hard_dependency_check(host_config)
```

wen587 (dismissed):

wen587 (approved):

PR #3778, opened 2025-02-25
CLI addition for PFC counters --history

peterbailey-arista:
#### What I did
- Added a history option to `pfcstat` / `show pfc counters`
- Added unit tests for single and multiasic cases of the history cli

These changes were made for the PFC Historical Statistics feature: https://github.com/sonic-net/SONiC/issues/1904

#### How I did it
- Additions made to pfcstat script with new --history option
- Additions made to show CLI to take advantage of new --history option

#### How to verify it
- Run the sonic-utilities unit tests
- Run `show pfc counters --history` on a switch with these changes
 
Example output of `show pfc counters --history`

```
     Port    Priority    RX Pause Transitions    Total RX Pause Time US    Recent RX Pause Time US    Recent RX Pause Timestamp
---------  ----------  ----------------------  ------------------------  -------------------------  ---------------------------
Ethernet0        PFC0                      12                    12,000                      1,200         01/10/2008, 21:20:00
Ethernet0        PFC1                      21                    20,001                      2,001         05/18/2033, 03:33:20
Ethernet0        PFC2                      22                    20,002                      2,002         05/18/2033, 03:33:20
Ethernet0        PFC3                      23                    20,003                      2,003         05/18/2033, 03:33:20
Ethernet0        PFC4                      24                    20,004                      2,004         05/18/2033, 03:33:20
Ethernet0        PFC5                      25                    20,005                      2,005         05/18/2033, 03:33:20
Ethernet0        PFC6                      26                    20,006                      2,006         05/18/2033, 03:33:20
Ethernet0        PFC7                      27                    20,007                      2,007         05/18/2033, 03:33:20

Ethernet4        PFC0                      14                    14,000                      1,400         05/13/2014, 16:53:20
Ethernet4        PFC1                      41                    40,001                      4,001         10/02/2096, 07:06:40
Ethernet4        PFC2                      42                    40,002                      4,002         10/02/2096, 07:06:40
Ethernet4        PFC3                      43                    40,003                      4,003         10/02/2096, 07:06:40
Ethernet4        PFC4                      44                    40,004                      4,004         10/02/2096, 07:06:40
Ethernet4        PFC5                      45                    40,005                      4,005         10/02/2096, 07:06:40
Ethernet4        PFC6                      46                    40,006                      4,006         10/02/2096, 07:06:40
Ethernet4        PFC7                      47                    40,007                      4,007         10/02/2096, 07:06:40

Ethernet8        PFC0                      18                    18,000                      1,800         01/15/2027, 08:00:00
Ethernet8        PFC1                      81                    80,001                      8,001         07/06/2223, 14:13:20
Ethernet8        PFC2                      82                    80,002                      8,002         07/06/2223, 14:13:20
Ethernet8        PFC3                      83                    80,003                      8,003         07/06/2223, 14:13:20
Ethernet8        PFC4                      84                    80,004                      8,004         07/06/2223, 14:13:20
Ethernet8        PFC5                      85                    80,005                      8,005         07/06/2223, 14:13:20
Ethernet8        PFC6                      86                    80,006                      8,006         07/06/2223, 14:13:20
Ethernet8        PFC7                      87                    80,007                      8,007         07/06/2223, 14:13:20
```


scripts/pfcstat
```diff
@@ -244,55 +365,84 @@ Examples:
...
+        """
+            Print the pfc history stats
+        """
+        if os.path.isfile(hist_fqn_file):
+            try:
+                hist_cached_dict = json.load(open(hist_fqn_file, 'r'))
+                print("Last cached time was " + str(hist_cached_dict.get('time')))
+                pfcstat.history_diff_print(hist_dict, hist_cached_dict)
+            except IOError as e:
+                print(e.errno, e)
```
maipbui, line 404: >print(e.errno, e) [](http://example.com/codeflow?start=16&length=17)

consider return or raise here?

scripts/pfcstat
```diff
@@ -244,55 +365,84 @@ Examples:
...
+        """
+            Print the counters of pfc rx counter
+        """
+        if os.path.isfile(cnstat_fqn_file_rx):
+            try:
+                cnstat_cached_dict = json.load(open(cnstat_fqn_file_rx, 'r'))
+                print("Last cached time was " + str(cnstat_cached_dict.get('time')))
+                pfcstat.cnstat_diff_print(cnstat_dict_rx, cnstat_cached_dict, True)
+            except IOError as e:
+                print(e.errno, e)
```
maipbui, line 428: >print(e.errno, e) [](http://example.com/codeflow?start=16&length=17)

same

scripts/pfcstat
```diff
@@ -244,55 +365,84 @@ Examples:
...
+        """
+            Print the counters of pfc tx counter
+        """
+        if os.path.isfile(cnstat_fqn_file_tx):
+            try:
+                cnstat_cached_dict = json.load(open(cnstat_fqn_file_tx, 'r'))
+                print("Last cached time was " + str(cnstat_cached_dict.get('time')))
+                pfcstat.cnstat_diff_print(cnstat_dict_tx, cnstat_cached_dict, False)
+            except IOError as e:
+                print(e.errno, e)
```
maipbui, line 443: >print(e.errno, e) [](http://example.com/codeflow?start=16&length=17)

same

scripts/pfcstat
peterbailey-arista: Thank you for reviewing! 
I updated to address the feedback, but the design has also changed a bit since your comments.

doc/Command-Reference.md
```diff
@@ -9700,8 +9700,38 @@ This command displays the details of Rx & Tx priority-flow-control (pfc) for all
...
+  Ethernet4        PFC4                      44                    40,004                      4,004         10/02/2096, 07:06:40
+  Ethernet4        PFC5                      45                    40,005                      4,005         10/02/2096, 07:06:40
+  Ethernet4        PFC6                      46                    40,006                      4,006         10/02/2096, 07:06:40
+  Ethernet4        PFC7                      47                    40,007                      4,007         10/02/2096, 07:06:40
+  ```
+
+
+- NOTE: PFC counters (including historical stats) can be cleared by the user with the following command:
   ```
   admin@sonic:~$ sonic-clear pfccounters
```
vmittal-msft, line 9736: @peterbailey-arista this will clear pfc counters as well as history. What if user only want to clear counter and not history. Shall we give option like --history for clear counter as well ? let me know your thoughts.

scripts/pfcstat
```diff
@@ -68,6 +70,27 @@ counter_bucket_tx_dict = {
 }
 
 
+HistStats = namedtuple("HistStats", "numTransitions, totalPauseTime, recentPauseTimestamp, recentPauseTime")
+
+SAI_PREFIX = "SAI"
+EST_PREFIX = "EST"
+
+total_stat_fields = [
+    "_PORT_STAT_PFC_*_ON2OFF_RX_PKTS",
```
vmittal-msft, line 79: is this only counting on to off transitions and not off to on ?

scripts/pfcstat
```diff
@@ -197,6 +297,51 @@ class Pfcstat(object):
...
+            if key in hist_old_dict:
+                for pfc, stat in pfc_dict.items():
+                    old_stat = None
+                    if pfc in hist_old_dict.get(key):
+                        old_stat = hist_old_dict.get(key).get(pfc)
+                    # old dict [ port ] has this priority data
+                    if old_stat is not None:
+                        table.append((key,
+                                    pfc,
+                                    ns_diff(stat['numTransitions'], old_stat['numTransitions']),
```
vmittal-msft, line 319: @peterbailey-arista  Why are we only printing the difference from last history ? and not the cumulative numbers ?

scripts/pfcstat
peterbailey-arista: Yes that's correct, since off-to-on and on-to-off would be the same number +/-1 it was unnecessary to add both.

Also OFF2ON is not part of the SAI PORT attributes. We are either using the SAI or estimating the value if it is not provided by the SAI in this feature.

For ex see: https://github.com/opencomputeproject/SAI/blob/b925fc0f0fc8c620eea7d9dead7fc99e33b7b44e/inc/saiport.h#L3185

doc/Command-Reference.md
peterbailey-arista: @vmittal-msft That's correct, the history is based in part on the counters values when when they increment. It would seem to me that it is best to clear them together so that you can't have for example a very low count but extremely long history or very many transitions as that would appear confusing and contradictory.

I could look into the option but maybe this could be incremented on in the future instead of overhauling the code as it currently exists. Let me know what you think

scripts/pfcstat
peterbailey-arista: @vmittal-msft I touched on it in my other comment but the history is also meant to be a snapshot from between two time periods in the same way that I believe all the other counts are displayed in SONiC.
In this case I believe it would be confusing for the history not to follow the same pattern as the other counters especially when it is sometimes an estimated value based on those other counters.

peterbailey-arista: Thank you for reviewing please let me know if I have answered your questions or if there are others!

vmittal-msft: Thanks @peterbailey-arista. So "RX Pause Transitions" as well as "Total Paused time" will be increment between last two polling times ? Also, Can you please give me example about similar implementation for other counters. As i always assumed the counters are cumulative since the switch start and not dependent on time intervals.

peterbailey-arista: No, sorry if what I said was confusing, the sonic-clear functionality does not actually "clear" any counters, all counters do increment cumulatively from the beginning state (I'm not sure if that's switch boot-up or when exactly) but when you run sonic-clear it caches the current count and from then on the count you see with `show` is the difference between the current count and the count at the last time sonic-clear was run. 

The easiest example is the pfc pause frame counts that we currently see with `show pfc counters`, these counts are since the last clear -however the actual value stored in DB is cumulative, the value displayed by show is calculated using the difference between the value at last clear and the value now.

What I'm trying to say is that this is how the current counters work between show and clear so I have the historical stat counts work in the same way.

qiluo-msft:
This change will break sonic-mgmt testcase. @peterbailey-arista Could you fix the testcase?
```python
def extract_pfcwd_config(duthost, start_pfcwd):
    """
    Extract pfcwd info from running config

    Args:
        duthost: DUT host object

    Yields:
        pfcwd_config: dict of dicts with interface as the 1st level key and 'action', 'detect_time',
                      'restore_time' as the 2nd level keys
    """
    output = duthost.command('show pfcwd config')
    pytest_assert('Ethernet' in output['stdout'], 'No ports found in the pfcwd config')

    pfcwd_config = defaultdict()
    for line in output['stdout_lines']:
        if line.strip().startswith('Ethernet'):
            port, action, detect, restore = line.split()
            pfcwd_config.update({port: {'action': action,
                                        'detect_time': detect,
                                        'restore_time': restore}})

    yield pfcwd_config
```

```
12/09/2025 05:44:18 __init__._fixture_generator_decorator    L0088 ERROR  | 
ValueError('too many values to unpack (expected 4)')
  File "/var/src/sonic-mgmt/tests/common/plugins/log_section_start/__init__.py", line 84, in _fixture_generator_decorator
    res = next(it)
  File "/var/src/sonic-mgmt/tests/generic_config_updater/test_pfcwd_status.py", line 162, in extract_pfcwd_config
    port, action, detect, restore = line.split()
ValueError: too many values to unpack (expected 4)
```

PR #3781, opened 2025-02-25
Add namespace support for show interface neighbor expected

auspham:
#### What I did

Adding namespace support for `show interface neighbor expected`. So that we can use

```
show interface neighbor expected -n asic0
```

#### How to verify it
Verified on T2 testbed

```
admin@somehost:/var/log$ show interfaces neighbor expected -n asic1
Usage: show interfaces neighbor expected [OPTIONS] [INTERFACENAME]
Try "show interfaces neighbor expected -h" for help.

Error: no such option: -n
```

```
admin@somehost:/var/log$ show interfaces neighbor expected -n asic1
LocalPort    Neighbor    NeighborPort    NeighborLoopback    NeighborMgmt    NeighborType
-----------  ----------  --------------  ------------------  --------------  --------------
Ethernet96   ARISTAXXTX  Ethernet1       None                XXX.XX.XXX.XX   someHub
Ethernet104  ARISTAXXTX Ethernet1       None                XXX.XX.XXX.XX   otherHub
...
```

### DIFF
--- a/show/interfaces/__init__.py
+++ b/show/interfaces/__init__.py
@@ -281,16 +281,19 @@ def neighbor():
 # 'expected' subcommand ("show interface neighbor expected")
 @neighbor.command()
 @click.argument('interfacename', required=False)
+@click.option('--namespace', '-n', 'namespace', default='',
+              type=str, show_default=True, help='Namespace name or all',
+              callback=multi_asic_util.multi_asic_namespace_validation_callback)
 @clicommon.pass_db
-def expected(db, interfacename):
+def expected(db, interfacename, namespace):
     """Show expected neighbor information by interfaces"""
 
-    neighbor_dict = db.cfgdb.get_table("DEVICE_NEIGHBOR")
+    neighbor_dict = db.cfgdb_clients[namespace].get_table("DEVICE_NEIGHBOR")
     if neighbor_dict is None:
         click.echo("DEVICE_NEIGHBOR information is not present.")
         return
 
-    neighbor_metadata_dict = db.cfgdb.get_table("DEVICE_NEIGHBOR_METADATA")
+    neighbor_metadata_dict = db.cfgdb_clients[namespace].get_table("DEVICE_NEIGHBOR_METADATA")
     if neighbor_metadata_dict is None:
         click.echo("DEVICE_NEIGHBOR_METADATA information is not present.")
         return
--- a/show/interfaces/__init__.py
+++ b/show/interfaces/__init__.py
@@ -281,12 +281,13 @@ def neighbor():
 # 'expected' subcommand ("show interface neighbor expected")
 @neighbor.command()
 @click.argument('interfacename', required=False)
-@click.option('--namespace', '-n', 'namespace', default='',
-              type=str, show_default=True, help='Namespace name or all',
-              callback=multi_asic_util.multi_asic_namespace_validation_callback)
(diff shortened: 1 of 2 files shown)


show/interfaces/__init__.py
```diff
@@ -281,16 +281,19 @@ def neighbor():
 # 'expected' subcommand ("show interface neighbor expected")
 @neighbor.command()
 @click.argument('interfacename', required=False)
+@click.option('--namespace', '-n', 'namespace', default='',
```
FengPan-Frank, line 284: better to update
https://github.com/sonic-net/sonic-utilities/blob/afa3c029e3b249544766ce451f85e078f513cd21/doc/Command-Reference.md?plain=1#L5434 as well, thanks

auspham: Thanks for pointing out, updated @FengPan-Frank

sdszhang:
Will `ip netns exec asicx xxx` be enough for the requirement?

auspham:
> Will `ip netns exec asicx xxx` be enough for the requirement?

yes ip netns exec asicx will work as well. We will need to adjust the tests a bit and cannot use inherited `SonicAsic` method. This would be a nice improvement.

sdszhang (approved):
lgtm

show/interfaces/__init__.py
```diff
@@ -281,16 +281,19 @@ def neighbor():
...
+              type=str, show_default=True, help='Namespace name or all',
+              callback=multi_asic_util.multi_asic_namespace_validation_callback)
 @clicommon.pass_db
-def expected(db, interfacename):
+def expected(db, interfacename, namespace):
     """Show expected neighbor information by interfaces"""
 
-    neighbor_dict = db.cfgdb.get_table("DEVICE_NEIGHBOR")
+    neighbor_dict = db.cfgdb_clients[namespace].get_table("DEVICE_NEIGHBOR")
```
sdszhang, line 291: what's the namespace if it's single ASIC?

auspham: we have some check guard here https://github.com/sonic-net/sonic-utilities/blob/master/utilities_common/multi_asic.py#L116 that would not be able to run on single ASIC.

FengPan-Frank (approved):

show/interfaces/__init__.py
sdszhang: so `show interfaces neighbor expected` will not run on single ASIC platform?

auspham: Hi @sdszhang this is meant to be the output.

image

But good point you're bringing this up since there is some logic issue in the current logic of `multi_asic_namespace_validation_callback`.

image

Since `multi_asic.is_multi_asic` is a function, this if statement here will always be `True` regardless of the topology.

I think what the author meant to write is:

```
if not multi_asic.is_multi_asic() and value: 
        click.echo("-n/--namespace is not available for single asic")
        ctx.abort()
        return value
```

This is gonna check if our DUT executing with `-n asicX` and is multiasic or not. Now im not sure if we should modifying this since there are functions that are using this.

sdszhang (approved):

sdszhang (approved):

auspham:
hi @FengPan-Frank  could you have a review again. please. Thanks

FengPan-Frank (approved):

PR #3791, opened 2025-03-03
Fix command docker ps shows image ID instead of image name issue.

liushilongbuaa:
#### What I did
Fix docker ps command showing image ID instead of name:tag issue.
It only impacts dhcp-relay, dhcp-server and macsec.
#### How I did it
render_ctx will provide name:tag and j2 template will use it instead of ID.


yaqiangz (approved):
LGTM

wangxin (approved):

PR #3814, opened 2025-03-21
YANG: remove uses clause handling, now part of sonic-yang-mgmt

bradh352:
#### What I did

sonic-yang-mgmt now integrates uses clauses as if a uses clause didn't exist. No need to have this complex handling in sonic-utilities.

Depends on https://github.com/sonic-net/sonic-buildimage/pull/21907
Fixes https://github.com/sonic-net/sonic-buildimage/issues/22382

#### How I did it

Besides removal of the code, the only test case changes are the removal of the 'grouping' clause as it is specific to 'uses' which is not relevant and some minor reordering of expected responses.

#### How to verify it

Run `tests/cli_autogen_yang_parser_test.py` after applying both this PR and https://github.com/sonic-net/sonic-buildimage/pull/21907

N/A

N/A

Signed-off-by: Brad House (@bradh352)

### DIFF
--- a/sonic_cli_gen/yang_parser.py
+++ b/sonic_cli_gen/yang_parser.py
@@ -331,8 +331,7 @@ def on_leaf(leaf: OrderedDict,
     attr = {'name': leaf.get('@name'),
             'description': get_description(leaf),
             'is-leaf-list': is_leaf_list,
-            'is-mandatory': get_mandatory(leaf)
-           }
+            'is-mandatory': get_mandatory(leaf)}
 
     return attr
 
--- a/sonic_cli_gen/yang_parser.py
+++ b/sonic_cli_gen/yang_parser.py
@@ -36,8 +36,7 @@ class YangParser:
                             'name': 'value',
                             'description': 'value',
                             'is-leaf-list': False,
-                            'is-mandatory': False,
-                            'group': 'value'
+                            'is-mandatory': False
                         }
                         ...
                     ],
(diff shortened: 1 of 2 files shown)


kalash-nexthop (approved):

kalash-nexthop (changes requested):
Just leaving a comment here that I accidentally hit approve button. Hopefully this will be enough for "unapproving" this. Although it doesn't matter since I am not an official reviewer anyway.

sonic_cli_gen/yang_parser.py
```diff
@@ -36,8 +36,7 @@ class YangParser:
                             'name': 'value',
                             'description': 'value',
                             'is-leaf-list': False,
-                            'is-mandatory': False,
-                            'group': 'value'
+                            'is-mandatory': False
```
FengPan-Frank, line 39: for group removal, do you need to update https://github.com/sonic-net/sonic-utilities/blob/ce51df2d8c803d6f15348acd31a6f80d2d425edb/tests/cli_autogen_yang_parser_test.py#L143C9-L143C30 as well?

bhouse-nexthop: Since a uses/grouping is being done in this yang model, and sonic-yang-mgmt is actually 'compiling' it into a normalized form,  I'm keeping the test case for explicit validation here.  I don't see a reason to remove the test case since it is still stressing the schema.  I'm just removing the 'group' member of the expected output since that is no longer valid/relevant as sonic-utilities isn't responsible for handling that.

gechiang:
@adyeung , please help tag the YANG and UMF group.
Thanks!

bradh352:
> @qiluo-msft Now that [sonic-net/sonic-buildimage#21907](https://github.com/sonic-net/sonic-buildimage/pull/21907) is merged this ticket needs to be merged. I just rebased against current master to force a rebuild.

I forced the rebuild too early, the new sonic-buildimage assets aren't available yet so tests fail.  Will retry in a few hours.

bradh352:
@qiluo-msft all test pass now since the artifacts from the merged PR https://github.com/sonic-net/sonic-buildimage/pull/21907 are available.

Can this be merged?

bradh352:
FYI, all PRs are failing tests in sonic-utilities due to this not being merged!

qiluo-msft (approved):

yejianquan:
The depenent PR https://github.com/sonic-net/sonic-buildimage/pull/21907 is marked as 202505 not needed, remove

PR #3822, opened 2025-03-28
Add a check for the condition where asic_id is not present on VS platform

BYGX-wcr:
#### What I did

Add a check in generate_sysinfo function to prevent runtime error when the asic_id is none because it is not present on VS and Cisco platform.


BYGX-wcr:
The asic_id is not present on Cisco platform as well.

abdosi (approved):

BYGX-wcr:
Closed as problem has been resolved.

PR #3825, opened 2025-04-02
Enable config reload yang validation for multiasic

xincunli-sonic:
#### What I did

Remove multi asic check for reload operation since multi asis golden config has finished qualification.

### DIFF
--- a/tests/config_test.py
+++ b/tests/config_test.py
@@ -196,9 +196,12 @@ Released lock on {0}
 RELOAD_MASIC_CONFIG_DB_OUTPUT = """\
 Acquired lock on {0}
 Stopping SONiC target ...
+Running command: /usr/local/bin/sonic-cfggen -H -k  --write-to-db
 Running command: /usr/local/bin/sonic-cfggen -j /tmp/config.json --write-to-db
-Running command: /usr/local/bin/sonic-cfggen -j /tmp/config.json -n asic0 --write-to-db
-Running command: /usr/local/bin/sonic-cfggen -j /tmp/config.json -n asic1 --write-to-db
+Running command: /usr/local/bin/sonic-cfggen -H -k  -n asic0 --write-to-db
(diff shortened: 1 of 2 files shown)


tests/config_test.py
```diff
@@ -196,9 +196,12 @@
 RELOAD_MASIC_CONFIG_DB_OUTPUT = """\
 Acquired lock on {0}
 Stopping SONiC target ...
+Running command: /usr/local/bin/sonic-cfggen -H -k  --write-to-db
```
wen587, line 199: Why the ouput changes? 
I don't think config_file_yang_validation will call extra command to configure.

tests/config_test.py
```diff
@@ -1334,6 +1347,26 @@ def add_sysinfo_to_cfg_file(self):
                 }
             }
             f.write(json.dumps(device_metadata))
+        with open(self.dummy_cfg_file_asic0, 'w') as f:
```
wen587, line 1350: it is same for host and asic, why create another one?

tests/config_test.py
```diff
@@ -835,7 +838,11 @@ def read_json_file_side_effect(filename):
         with mock.patch("utilities_common.cli.run_command",
                         mock.MagicMock(side_effect=mock_run_command_side_effect)),\
             mock.patch('config.main.read_json_file',
-                       mock.MagicMock(side_effect=read_json_file_side_effect)):
+                       mock.MagicMock(side_effect=read_json_file_side_effect)),\
+            mock.patch('config.main.sonic_yang.SonicYang.loadData',
```
wen587, line 842: Why dont we call yang validation over the config, it seems we just bypass here.

tests/config_test.py
xincunli-sonic: The original test will skip calling config_file_yang_validation when it is multiasic

tests/config_test.py
xincunli-sonic: Thanks, removed since it is not required anymore.

wen587 (approved):
lgtm

PR #3826, opened 2025-04-03
VNET CLI- ADD/DEL VNET, ADD/DEL VNET ROUTE, VRF BIND/UNBIND to consider VNET + ADDITIONAL MODIFICATIONS TO SHOW VNET CLIs

kavyavanibedida:
#### What I did
- **Added 4 new VNET CLIS**(Exact command covered in the reference command doc)
  - sudo config vnet add 
  - sudo config vnet del <vnet-name> 
  - sudo config vnet add-route <vnet-name> <prefix>  <endpoint>
  - sudo config vnet del-route <vnet-name> <prefix> 
- **Modification of  existing CLIS**
   -  bind vrf/unbind vrf CLIs to consider VNET
   - show vnet interfaces to take vlan_sub_interfaces, portchannel for VNET
   - show vnet neighbors to take vlan_sub_interfaces, portchannel for VNET
- Updated the Command Line Reference Guide

#### How I did it
- Modified config/main.py for config commands
- Modified show/vnet.py to show associated interfaces for vnets

#### How to verify it
- **UT pytest coverage in vnet_test, subintf_test, ip_config_test** 
  - test_vnet_bind_unbind
  - test_vnet_add_del_route
  - test_vnet_add_del
  -  test_intf_vnet_bind_unbind
  -  test_subintf_vnet_bind_unbind


config/main.py
```diff
@@ -125,6 +125,7 @@
 QUEUE_RANGE = click.IntRange(min=0, max=255)
 GRE_TYPE_RANGE = click.IntRange(min=0, max=65535)
 ADHOC_VALIDATION = True
+VNET_NAME_MAX_LEN = 200
```
SoumyaMishra18, line 128: Lets limit the vnet name length to IFACE_NAME_MAX_LEN(15 characters).

tests/vrf_test.py
```diff
@@ -209,6 +210,164 @@ def test_vrf_bind_unbind(self):
...
+        expected_output_unbind = "Interface Po0002.101 IP disabled and address(es) removed due to unbinding VRF.\n"
+        cmds = config.config.commands["interface"].commands["vrf"].commands["unbind"]
+        result = runner.invoke(cmds, ["Po0002.101"], obj=vnet_obj)
+
+        print(result.exit_code, result.output)
+        assert result.exit_code == 0
+        assert ('vnet_name', 'Vnet_103') not in db.cfgdb.get_table('VLAN_SUB_INTERFACE')['Po0002.101']
+        assert result.output == expected_output_unbind
+
+        expected_output_bind = "Interface Ethernet0 IP disabled and address(es) removed due to binding VRF Vnet_1.\n"
```
SoumyaMishra18, line 307: After unbind, can we get the 'show vnet interfaces' to check the entries are removed?

tests/vrf_test.py
```diff
@@ -209,6 +210,164 @@ def test_vrf_bind_unbind(self):
...
+        cmds = config.config.commands["interface"].commands["vrf"].commands["bind"]
+        result = runner.invoke(cmds, ["Po0002.101", "Vnet_1"], obj=vnet_obj)
+        assert result.exit_code == 0
+        assert result.output == expected_output_bind
+        assert ('Vnet_1') in db.cfgdb.get_table('VLAN_SUB_INTERFACE')['Po0002.101']['vnet_name']
+
+        jsonfile_config = os.path.join(mock_db_path_vnet, "config_db")
+        dbconnector.dedicated_dbs['CONFIG_DB'] = jsonfile_config
+
+        expected_output = """\
```
SoumyaMishra18, line 359: this seems to be the same as expected output in line 220....can we remove this?

tests/vrf_test.py
```diff
@@ -275,3 +434,76 @@ def test_invalid_vrf_name(self):
...
+Error: 'vnet_name' must begin with 'Vnet_' .
+"""
+        # Test vnet add using length of vnet name
+        vnet_name = "Vnet_ypfbjjhyzivaythuaxlbcibgdgjkqgapedmiosjgsvddqaalformdmbwigddddddddghkbpccbzjrhefrcdeqkvgmu \
+                 bxxnkgbvjpgpnypfbjjhyzivaythuaxlbcibgdgjkqgapedmiosjgsvddqlformdmbwigghkbpccbzjrhefrcdeqkvgmubxxnkgbvj"
+        args = [vnet_name, "222", "tunnel1"]
+        result = runner.invoke(config.config.commands["vnet"].commands["add"], args, obj=vnet_obj)
+        assert result.exit_code != 0
+        assert "'vnet_name' length should not exceed 200 characters" in result.output
```
SoumyaMishra18, line 451: Use the length to be 15 characters. We will be using short vnet name and GUID will be stored as a separate entry.

tests/vrf_test.py
```diff
@@ -275,3 +434,76 @@ def test_invalid_vrf_name(self):
...
+        assert result.exit_code == 0
+
+        # Test vnet del with wrong vnet name
+        expected_output_del = "'vnet_name' must begin with 'Vnet_' "
+        result = runner.invoke(config.config.commands["vnet"].commands["del"], ["vnet_3"], obj=vnet_obj)
+        assert result.exit_code != 0
+        assert expected_output_del in result.output
+
+        # Test vnet del with long vnet name
+        expected_output_del = "'vnet_name' length should not exceed 200 characters"
```
SoumyaMishra18, line 490: length should be 15

config/main.py
```diff
@@ -5802,23 +5829,40 @@ def bind(ctx, interface_name, vrf_name):
         if interface_name is None:
             ctx.fail("'interface_name' is None!")
 
-    if not is_vrf_exists(config_db, vrf_name):
-        ctx.fail("VRF %s does not exist!"%(vrf_name))
+    isVnet = False
```
anish-n, line 5832: I feel that its confusing if we are using a ```config vrf``` command to bind to a vnet object. Even though ```config vnet``` is what one would use to create a vnet, one must then issue a config vrf to bind an interface to a vnet.  I think its better to conform to a standard model, either we use all VRF or all VNET commands so that its not split between 2 types.

anish-n (changes requested):

config/main.py
```diff
@@ -9070,6 +9124,80 @@ def motd(message):
...
+        ctx.fail("'vnet_name' length should not exceed {} characters".format(VNET_NAME_MAX_LEN))
+    if not is_vnet_exists(config_db, vnet_name):
+        ctx.fail("VNET {} does not exist!".format(vnet_name))
+    else:
+        del_interface_bind_to_vnet(config_db, vnet_name)
+        try:
+            config_db.set_entry('VNET', vnet_name, None)
+        except JsonPatchConflict as e:
+            ctx.fail("Invalid ConfigDB. Error: {}".format(e))
+        click.echo("VNET {} deleted and all associated IP addresses removed.".format(vnet_name))
```
anish-n, line 9200: missing add/delete vnet route commands?

config/main.py
```diff
@@ -5802,23 +5829,40 @@ def bind(ctx, interface_name, vrf_name):
         if interface_name is None:
             ctx.fail("'interface_name' is None!")
 
-    if not is_vrf_exists(config_db, vrf_name):
-        ctx.fail("VRF %s does not exist!"%(vrf_name))
+    isVnet = False
+    if (vrf_name.startswith('Vnet_')):
+        isVnet= True
+
+    if (not isVnet):
```
anish-n, line 5836: NIT: code will be a bit cleaner if we use ```if(isVnet)```, and else, vs ```if (not isVnet)```

anish-n (changes requested):

config/main.py
```diff
@@ -5967,23 +5995,40 @@ def bind(ctx, interface_name, vrf_name):
...
+                return
+        else:
+            if (config_db.get_entry(table_name, interface_name).get('vrf_name') == vrf_name):
+                return
+
     # Clean ip addresses if interface configured
     interface_addresses = get_interface_ipaddresses(config_db, interface_name)
-    for ipaddress in interface_addresses:
-        remove_router_interface_ip_address(config_db, interface_name, ipaddress)
+    for ip_address in interface_addresses:
```
kavyavanibedida, line 6022: Had to change to ip_address due to a pre-commit check failure

tests/vrf_test.py
kavyavanibedida: have updated to 15

tests/vrf_test.py
kavyavanibedida: have updated to 15

config/main.py
```diff
@@ -5937,7 +5991,7 @@ def enable_use_link_local_only(ctx, interface_name):
         interface_type = "PORTCHANNEL_INTERFACE"
     elif interface_name.startswith("Vlan"):
         interface_type = "VLAN_INTERFACE"
-    else:
+    else :
```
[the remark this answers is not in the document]
kavyavanibedida: Fixed

config/main.py
```diff
@@ -6090,7 +6144,7 @@ def add_vrrp_ip(ctx, interface_name, vrrp_id, ip_addr):
     if check_vrrp_ip_exist(config_db, ip_addr):
         ctx.abort()
 
-    if "/" not in ip_addr:
+    if "/" not in ip_addr :
```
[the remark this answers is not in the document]
kavyavanibedida: fixed

config/main.py
kavyavanibedida: Soumya is adding these 2 CLIs

config/main.py
kavyavanibedida: Updated as specified

tests/vrf_test.py
kavyavanibedida: yes, this checks at the end of the testcase

SoumyaMishra18 (approved):

config/main.py
```diff
@@ -515,6 +548,31 @@ def del_interface_bind_to_vrf(config_db, vrf_name):
...
+                    interface_ipaddresses = get_interface_ipaddresses(config_db, interface_name)
+                    for ip in interface_ipaddresses:
+                        remove_router_interface_ip_address(config_db, interface_name, ip)
+                    config_db.set_entry(table_name, interface_name, None)
+
+
+def del_route_bind_to_vnet(config_db, vnet_name):
+    """del interface bind to vnet
+    """
+    tables = ['VNET_ROUTE_TUNNEL']
```
anish-n, line 571: Should probably delete VNET_ROUTE here as well, in addition to the existing delete which you have for VNET_ROUTE_TUNNEL.

config/main.py
kavyavanibedida: Added these 2 CLis

tests/vrf_test.py
```diff
@@ -335,6 +495,172 @@ def test_show_vnet_guid(self):
         assert "PortChannel0002" in result.output
         assert "Loopback0" in result.output
 
+    def test_vnet_add_del(self):
```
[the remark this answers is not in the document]
anish-n: I also see that in the prior PR https://github.com/sonic-net/sonic-utilities/pull/3821/files, puts things into the same vrf_test.py, but it should indeed go into show_vnet_test.py, @SoumyaMishra18 fyi, pls fix

tests/vrf_test.py
```diff
@@ -335,6 +495,172 @@ def test_show_vnet_guid(self):
...
+        vnet_name = "Vnet_ypfbjjhyzivaythuaxlbcibgdgjkqgapedmiosjgsv"
+        args = [vnet_name, "10.10.10.10/32", "10.10.10.1"]
+        result = runner.invoke(config.config.commands["vnet"].commands["add-route"], args, obj=vnet_obj)
+        assert result.exit_code != 0
+        assert "'vnet_name' length should not exceed 15 characters" in result.output
+        assert vnet_name not in db.cfgdb.get_table('VNET_ROUTE_TUNNEL')
+
+        # Test vnet del with route deletion
+        result = runner.invoke(config.config.commands["vnet"].commands["del"], ["Vnet_4"], obj=vnet_obj)
+        assert any(key[0] != 'Vnet_4' for key in db.cfgdb.get_table('VNET_ROUTE_TUNNEL'))
```
anish-n, line 625: This should be ```all``` and not ```any```, any will only check if there is some route which exists with a key containing Vnet_4, we need to check that there are no routes containing Vnet_4.

tests/vrf_test.py
```diff
@@ -335,6 +495,172 @@ def test_show_vnet_guid(self):
...
+        assert result.exit_code != 0
+        assert expected_output_del in result.output
+
+        # Test vnet del route with long vnet name
+        expected_output_del = "'vnet_name' length should not exceed 15 characters"
+        vnet_name = "Vnet_ypfbjjhyzivaythuaxlbcibgdgjkq"
+        args = [vnet_name, "10.10.10.10/32"]
+        result = runner.invoke(config.config.commands["vnet"].commands["del-route"], args, obj=vnet_obj)
+        assert result.exit_code != 0
+        assert any(key[0] != vnet_name for key in db.cfgdb.get_table('VNET_ROUTE_TUNNEL'))
```
anish-n, line 640: This should be all as well I think

tests/vrf_test.py
```diff
@@ -335,6 +495,172 @@ def test_show_vnet_guid(self):
...
+        assert expected_output_del in result.output
+
+        # Test vnet del route
+        args = ["Vnet_3", "10.10.10.10/32"]
+        result = runner.invoke(config.config.commands["vnet"].commands["del-route"], args, obj=vnet_obj)
+        assert result.exit_code == 0
+        vnet_route_tunnel = db.cfgdb.get_table('VNET_ROUTE_TUNNEL')
+        if vnet_route_tunnel:
+            assert any(key[0] != 'Vnet_3' for key in vnet_route_tunnel)
```
anish-n, line 649: same here

anish-n (changes requested):

config/main.py
```diff
@@ -418,14 +420,45 @@ def is_vrf_exists(config_db, vrf_name):
...
+        if k[0] == vnet_name and k[1] == prefix:
+            return True
+
+    return False
+
+
+def vnet_name_is_valid(ctx, vnet_name):
+    """Check if the vnet name is valid
+    """
+    if not vnet_name.startswith("Vnet_"):
```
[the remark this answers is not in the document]
kavyavanibedida: updated as per the suggestion

tests/vrf_test.py
kavyavanibedida: added the prefix in the assert to verify the route deletion

tests/vrf_test.py
kavyavanibedida: added the prefix in the assert to verify the route deletion

config/main.py
```diff
@@ -9306,5 +9408,155 @@ def motd(message):
...
+@click.argument('guid', metavar='<guid>', type=str, required=False)
+@click.argument('scope', metavar='<scope>', required=False)
+@click.argument('advertise_prefix', metavar='<advertise_prefix>', type=bool, required=False)
+@click.argument('overlay_dmac', metavar='<overlay_dmac>', required=False)
+@click.argument('src_mac', metavar='<src_mac>', required=False)
+@click.pass_context
+def add_vnet(ctx, vnet_name, vni, vxlan_tunnel, peer_list, guid, scope, advertise_prefix, overlay_dmac, src_mac):
+    """Add Vnet"""
+    config_db = ValidatedConfigDBConnector(ctx.obj['config_db'])
+
```
anish-n, line 9439: we should be validating that vxlan_tunnel exists here before allowing vnet creation

config/main.py
```diff
@@ -9306,5 +9408,155 @@ def motd(message):
...
+@click.argument('peer_list', metavar='<peer_list>', required=False)
+@click.argument('guid', metavar='<guid>', type=str, required=False)
+@click.argument('scope', metavar='<scope>', required=False)
+@click.argument('advertise_prefix', metavar='<advertise_prefix>', type=bool, required=False)
+@click.argument('overlay_dmac', metavar='<overlay_dmac>', required=False)
+@click.argument('src_mac', metavar='<src_mac>', required=False)
+@click.pass_context
+def add_vnet(ctx, vnet_name, vni, vxlan_tunnel, peer_list, guid, scope, advertise_prefix, overlay_dmac, src_mac):
+    """Add Vnet"""
+    config_db = ValidatedConfigDBConnector(ctx.obj['config_db'])
```
anish-n, line 9438: we should be validating that the vni is a valid vni, please check the function vni_id_is_valid

config/main.py
```diff
@@ -9306,5 +9408,155 @@ def motd(message):
...
+@click.argument('profile', metavar='<profile>', required=False)
+@click.argument('primary', metavar='<primary>', required=False)
+@click.argument('monitoring', metavar='<monitoring>', required=False)
+@click.argument('adv_prefix', metavar='<adv_prefix>', type=bool, required=False)
+@click.pass_context
+def add_vnet_route(ctx, vnet_name, prefix, end_point, vni, mac_address, endpoint_monitor,
+                   profile, primary, monitoring, adv_prefix):
+    """Add VNET Route"""
+    config_db = ValidatedConfigDBConnector(ctx.obj['config_db'])
+
```
anish-n, line 9503: we should be checking for valid ip prefix

config/main.py
```diff
@@ -9306,5 +9408,155 @@ def motd(message):
...
+    vnet_name_is_valid(ctx, vnet_name)
+
```
anish-n, line 9505: we should be checking for end_point being a valid ip address

config/main.py
```diff
@@ -9306,5 +9408,155 @@ def motd(message):
...
+@click.argument('advertise_prefix', metavar='<advertise_prefix>', type=bool, required=False)
+@click.argument('overlay_dmac', metavar='<overlay_dmac>', required=False)
+@click.argument('src_mac', metavar='<src_mac>', required=False)
+@click.pass_context
+def add_vnet(ctx, vnet_name, vni, vxlan_tunnel, peer_list, guid, scope, advertise_prefix, overlay_dmac, src_mac):
+    """Add Vnet"""
+    config_db = ValidatedConfigDBConnector(ctx.obj['config_db'])
+
+    vnet_name_is_valid(ctx, vnet_name)
+
```
anish-n, line 9441: we should be validating the mac, prefix etc.

anish-n (changes requested):

config/main.py
kavyavanibedida: added the check, furthermore checks for vni and vxlan_tunnel map entry in the vxlan_tunnel_map table. @anish-n please let me know if this check is required

config/main.py
kavyavanibedida: added the check

config/main.py
kavyavanibedida: updated

config/main.py
kavyavanibedida: updated the checks. the args adhere to this 
image

config/main.py
```diff
@@ -9306,5 +9441,213 @@ def motd(message):
...
+    if is_vnet_exists(config_db, vnet_name):
+        ctx.fail("VNET {} already exists!".format(vnet_name))
```
anish-n, line 9476: Please confirm with backend swss code that this truly needs to be create only.

config/main.py
```diff
@@ -9306,5 +9441,213 @@ def motd(message):
...
+    vnet_name_is_valid(ctx, vnet_name)
+
+    if not is_vnet_exists(config_db, vnet_name):
+        ctx.fail("VNET {} doesnot exist, cannot add a route!".format(vnet_name))
+    if not clicommon.is_ipaddress_prefix(prefix):
+        ctx.fail("Invalid prefix {}".format(prefix))
+    if is_vnet_route_exists(config_db, vnet_name, prefix):
+        ctx.fail("Route already exists for the VNET {}!".format(vnet_name))
+    else:
+        endpoint_ips = endpoint.split(',')
```
anish-n, line 9573: vni, mac address, endpoint are all lists, so either follow the convention of taking just 1 entry for all 3 of these, or we allow all 3 to be passed in as a list. I am ok if we support accepting just 1 entry for all these 3 for now, not sure if @prsunny has a different perspective. 

The problem happens if we support taking this in as a list, but don't support a list for VNI, then the swss code will throw an error cause vni must be a list of equal size as that of endpoint ip.

config/main.py
kavyavanibedida: ignoring the check

config/main.py
```diff
@@ -9306,5 +9441,213 @@ def motd(message):
...
+    vnet_name_is_valid(ctx, vnet_name)
+
+    if is_vnet_exists(config_db, vnet_name):
+        ctx.fail("VNET {} already exists!".format(vnet_name))
+    else:
+        if not vni.isdigit() or clicommon.vni_id_is_valid(int(vni)) is False:
+            ctx.fail("Invalid VNI {}. Valid range [1 to 16777215].".format(vni))
+        if not is_vxlan_tunnel_exists(config_db, vxlan_tunnel):
+            ctx.fail("Vxlan tunnel {} does not exist!".format(vxlan_tunnel))
+        if not is_vni_vxlan_mapping_exists(config_db, vxlan_tunnel, vni):
```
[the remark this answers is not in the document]
kavyavanibedida: removed. Looks like the swss code creates the entry while creation https://github.com/sonic-net/sonic-swss/blob/72eafd9b2e490db0c71aab8f56e1ea58cae196d6/orchagent/vnetorch.cpp#L501

config/main.py
```diff
@@ -9306,5 +9441,213 @@ def motd(message):
...
+            ctx.fail("VNI {} is not mapped to Vxlan tunnel {}.".format(vni, vxlan_tunnel))
+
+        subvnet_dict = {}
+        subvnet_dict["vni"] = vni
+        subvnet_dict["vxlan_tunnel"] = vxlan_tunnel
+
+        if peer_list:
+            peers = peer_list.split(',')
+            for peer in peers:
+                if not is_vnet_exists(config_db, peer):
```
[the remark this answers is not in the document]
kavyavanibedida: Removed this check for peers. Seems like the SWSS code only adds the peers to the list and no verification of the vnet create objects for the peers. https://github.com/sonic-net/sonic-swss/blob/72eafd9b2e490db0c71aab8f56e1ea58cae196d6/orchagent/vnetorch.cpp#L526

config/main.py
kavyavanibedida: Thanks for pointing this. Have made the change to both create/update the vnet object as per the swss code:, addOperation https://github.com/sonic-net/sonic-swss/blob/72eafd9b2e490db0c71aab8f56e1ea58cae196d6/orchagent/vnetorch.cpp#L423

config/main.py
kavyavanibedida: Have changed it to only one entry for vni, mac, end_point, ep_monitor & primary

config/main.py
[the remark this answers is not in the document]
anish-n: For 2, orachagent will retry the creation again, assuming the peer vnet gets added later, so I think we can skip that check

config/main.py
```diff
@@ -9306,5 +9428,205 @@ def motd(message):
...
+    """Add VNET Route"""
+    config_db = ValidatedConfigDBConnector(ctx.obj['config_db'])
+
+    vnet_name_is_valid(ctx, vnet_name)
+
+    if not is_vnet_exists(config_db, vnet_name):
+        ctx.fail("VNET {} doesnot exist, cannot add a route!".format(vnet_name))
+    if not clicommon.is_ipprefix(prefix):
+        ctx.fail("Invalid prefix {}".format(prefix))
+    if is_specific_vnet_route_exists(config_db, vnet_name, prefix):
```
anish-n, line 9552: does backend support modifying this, and if so the update should be allowed.

kavyavanibedida: Removed this check. This CLI considers update. 
1. From the swss code, I see that update_route is allowed, added the testcase for update
2. Ref links https://github.com/sonic-net/sonic-swss/blob/72eafd9b2e490db0c71aab8f56e1ea58cae196d6/orchagent/vnetorch.cpp#L1182
https://github.com/sonic-net/sonic-swss/blob/72eafd9b2e490db0c71aab8f56e1ea58cae196d6/orchagent/vnetorch.cpp#L1226

config/main.py
```diff
@@ -9306,5 +9428,203 @@ def motd(message):
...
+@click.argument('vnet_name', metavar='<vnet_name>', type=str, required=True)
+@click.argument('prefix', metavar='<prefix>', required=True)
+@click.argument('endpoint', metavar='<endpoint>', required=True)
+@click.argument('vni', metavar='<vni>', required=False)
+@click.argument('mac_address', metavar='<mac_address>', required=False)
+@click.argument('endpoint_monitor', metavar='<endpoint_monitor>', required=False)
+@click.argument('profile', metavar='<profile>', type=str, required=False)
+@click.argument('primary', metavar='<primary>', required=False)
+@click.argument('monitoring', metavar='<monitoring>', type=str, required=False)
+@click.argument('adv_prefix', metavar='<adv_prefix>', required=False)
```
anish-n, line 9539: in the design we had an argument called "persistent" to take in, which defines whether the route is written to app db(non persistent) or  to config db(persistent), I don't see that here, can you please check that.

kavyavanibedida: Had used only config_db for VNET_ROUTE_TUNNEL 1. To maintain uniformity with other CLIs 2.  Have seen that VNET_ROUTE_TUNNEL moved to using config_db in SWSS

anish-n: ok, @prsunny fyi in case you have a different perspective on whether or not to include the app db route creation in the cli, currently this PR skips app db route, and only does create/delete in config_db.

config/main.py
```diff
@@ -418,14 +421,77 @@ def is_vrf_exists(config_db, vrf_name):
 
     return False
 
+
+def is_vxlan_tunnel_exists(config_db, vxlan_tunnel_name):
```
prsunny, line 425: Can you please move this to utilities? - https://github.com/sonic-net/sonic-utilities/blob/master/utilities_common/cli.py

config/main.py
```diff
@@ -418,14 +421,77 @@ def is_vrf_exists(config_db, vrf_name):
...
+    """Check if VXLAN tunnel exists
+    """
+    keys = config_db.get_keys("VXLAN_TUNNEL")
+    if keys:
+        if vxlan_tunnel_name in keys:
+            return True
+    return False
+
+
+def is_mac_address_valid(mac):
```
prsunny, line 435: This can be used for future cases as well, suggest move to common utilities

config/main.py
prsunny: I think for phase 1, this should be good

prsunny (commented):
overall looks good to me. minor comment

config/main.py
kavyavanibedida: updated

prsunny (approved):
lgtm, @qiluo-msft  to merge

PR #3831, opened 2025-04-07
Generic Configuration Updater (GCU) performance enhancements

### LINKED ISSUE #22372: Enhancement: Improve Generic Config Updater (GCU) performance
### Is it platform specific

generic

### Importance or Severity

Critical

Generic Configuration Updater is woefully slow, so much so that it is basically unusable in a lot of situations.

### Current Behavior

In a configuration I have, it takes about 15 minutes to apply.

### Proposed Behavior

It would be expected this to take no more than a minute or two.

bhouse-nexthop:
***Important***: Please review each commit (including commit message) individually.  Looking at the patch-set as a whole may cause confusion.

#### What I did

Generic Configuration Updater is extremely slow, using the python profiler it was possible to determine the worst offenders where changes could be made without affecting the overall algorithm and HLD design documentation.

Brief overview of changes:
* Prevent copy.deepcopy() calls where possible
* Don't run validation twice back to back
* Move configdb path <> xpath conversion logic to sonic-yang-mgmt where it belongs and enhance it to support schema conversion (not just data) and add caching.
* Sort table keys by the number of schema backlinks and must statements for the node to try better guess the right order of the patches to generate rather than doing it in alphabetical order which is likely to cause validation failures.
* Add ability to Group patches together in some commits where its known they will not cause issues, these are things like grouping parameter updates under the same key.
* When applying changes, do not re-read the configuration from redis twice between each applied patch (this is **extremely** slow, and actually hid a race condition).  We are mutating the configuration and a lock is held so we know the expected before and after.  There is still a final validation to ensure something didn't go sideways.

Dependencies:
 * sonic-yang-mgmt enhancements: https://github.com/sonic-net/sonic-buildimage/pull/22254
 * sonic-yang-mgmt parse uses/grouping: https://github.com/sonic-net/sonic-buildimage/pull/21907
 * sonic-utilities rely on sonic-yang-mgmt uses/grouping handling: https://github.com/sonic-net/sonic-utilities/pull/3814

Stats below ... (stats need both this and the sonic-utilities PR to be relevant)...

<ins>**Original Performance:**</ins>
Dry Run:
```
time sudo config replace -d ./config_db.json
...
real	2m51.588s
user	2m23.777s
sys	0m25.300s
```

Full:
```
time sudo config replace ./config_db.json
...
real	14m53.772s
user	12m2.376s
sys	2m8.908s
```

<ins>**With Patch**:</ins>
Dry Run:
```
time sudo config replace -d ./config_db.json
...
real	0m59.602s
user	0m56.434s
sys	0m2.110s
```

Full:
```
time sudo config replace ./config_db.json
...
real	1m54.303s
user	0m58.482s
sys	0m2.545s
```

So that's roughly 3x improvement for dry-run, and 7.5x improvement for full commit.  There is room for improvement on the full commit due to a `sleep(1)` being used between each patch because of a race condition found in the prior code (that was hidden due to a costly sanity check that has been removed).

#### How I did it

Profiling via cProfiler

#### How to verify it

Run sonic-utilities test cases, they still pass

N/A

N/A

Fixes https://github.com/sonic-net/sonic-buildimage/issues/22372


generic_config_updater/gu_common.py
```diff
@@ -145,9 +145,7 @@ def validate_config_db_config(self, config_db_as_json):
                                         self.validate_lanes]
 
         try:
-            tmp_config_db_as_json = copy.deepcopy(config_db_as_json)
```
qiluo-msft, line 148: >deepcopy [](http://example.com/codeflow?start=41&length=8)

Could you provide a link relating "With
a recent patch to sonic-yang-mgmt it is no longer mutated" ?

bhouse-nexthop: That would be this commit https://github.com/sonic-net/sonic-buildimage/pull/22254/commits/e59d0319721b4da6a884eea0e6c1deac019b1cd0 from the PR this depends on https://github.com/sonic-net/sonic-buildimage/pull/22254

generic_config_updater/gu_common.py
```diff
@@ -351,7 +349,7 @@ def create_sonic_yang_with_loaded_models(self):
             loaded_models_sy.loadYangModel() # This call takes a long time (100s of ms) because it reads files from disk
             self.sonic_yang_with_loaded_models = loaded_models_sy
 
-        return copy.copy(self.sonic_yang_with_loaded_models)
+        return self.sonic_yang_with_loaded_models
```
qiluo-msft, line 350: >sonic_yang_with_loaded_models [](http://example.com/codeflow?start=20&length=29)

Dangerous to expose a member as return value, the caller could invoke non-const member on this object.

generic_config_updater/gu_common.py
```diff
@@ -130,9 +130,8 @@ def validate_sonic_yang_config(self, sonic_yang_as_json):
         sy = self.create_sonic_yang_with_loaded_models()
 
         try:
+            # Loading data automatically does full validation
```
qiluo-msft, line 134: >Loading data automatically does full validation [](http://example.com/codeflow?start=14&length=47)

Could you add a unit testcase to ensure "Loading data automatically does full validation" ? so there will be no future regression.

generic_config_updater/gu_common.py
bhouse-nexthop: This object is actually designed to be mutated.  Each time loadData() is called internal members of this returned class are updated.  Its just really reusing the model loading that was performed earlier.

This is a pre-existing bad design, that they tried to work around with a copy.  The problem is the copy makes some assumptions which aren't valid, so the pre-existing copy doesn't provide the protection you might think so removing it actually doesn't have any negative implications.  The designed behavior of using copy was that it would preserve the loaded models but clear the data loaded. 

Really should be a separate class for loading the models, and that class object (which shouldn't be mutable) should be passed into the data parser object and created on parse.

Doing that however, is sort of out of scope for this kind of enhancement.

generic_config_updater/gu_common.py
bhouse-nexthop: the test cases that have failures are already validating this, we'll see passes where we don't expect.  Its a little hard to actually see that until the PR https://github.com/sonic-net/sonic-buildimage/pull/22254 is merged since we can't see coverage without that in this PR.

bhouse-nexthop:
I just rebased against master to force a new build.

bhouse-nexthop:
@qiluo-msft this PR is pretty well cleaned up, however it is depending on https://github.com/sonic-net/sonic-buildimage/pull/21907 and https://github.com/sonic-net/sonic-utilities/pull/3814 which haven't been merged yet.  Once those merge the tests that are failing due to the inability to resolve the paths that are imported via the uses clause will be fixed.

It would be great to get all this into 202505.

All GCU test cases pass when I run them with this PR *and* those 2 dependent PRs:
```
brad@a59a25f02d1a:/sonic/src/sonic-utilities$ pytest tests/generic_config_updater/
================================== test session starts ==================================
platform linux -- Python 3.11.2, pytest-7.2.1, pluggy-1.0.0+repack
rootdir: /sonic/src/sonic-utilities/tests, configfile: pytest.ini
plugins: pyfakefs-5.8.0, cov-4.0.0
collected 348 items

tests/generic_config_updater/change_applier_test.py ..                            [  0%]
tests/generic_config_updater/field_operation_validator_test.py .................. [  5%]
......................                                                            [ 12%]
tests/generic_config_updater/gcu_feature_patch_application_test.py ..             [ 12%]
tests/generic_config_updater/generic_updater_test.py ............................ [ 20%]
...............                                                                   [ 25%]
tests/generic_config_updater/gu_common_test.py .................................. [ 34%]
....................................                                              [ 45%]
tests/generic_config_updater/multiasic_change_applier_test.py ......              [ 46%]
tests/generic_config_updater/multiasic_generic_updater_test.py .                  [ 47%]
tests/generic_config_updater/patch_sorter_test.py ............................... [ 56%]
................................................................................. [ 79%]
......................................................................            [ 99%]
tests/generic_config_updater/service_validator_test.py ..                         [100%]

================================= 348 passed in 13.38s ==================================
```

bhouse-nexthop:
@qiluo-msft now that the dependent PRs have merged, this is passing all test cases.  Any chance you could do a final review and merge?

bhouse-nexthop:
Hi @saiarcot895, nice meeting you at the Sonic workshop today.  I mentioned this PR and you said you might be able to take a look at it.  Thanks!
-Brad

bhouse-nexthop:
@hdwhdw can you help as well?

generic_config_updater/gu_common.py
```diff
@@ -351,7 +347,7 @@ def create_sonic_yang_with_loaded_models(self):
             loaded_models_sy.loadYangModel() # This call takes a long time (100s of ms) because it reads files from disk
             self.sonic_yang_with_loaded_models = loaded_models_sy
 
-        return copy.copy(self.sonic_yang_with_loaded_models)
+        return self.sonic_yang_with_loaded_models
```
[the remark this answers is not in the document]
bhouse-nexthop: These objects are *guaranteed* to no longer be mutated with the sister PR from https://github.com/sonic-net/sonic-buildimage/pull/22254 which is why this change was made.  profiling python showed this was a large performance hit.

generic_config_updater/change_applier.py
```diff
@@ -137,25 +138,45 @@ def _report_mismatch(self, run_data, upd_data):
...
+        # This check would consume high CPU and would take a good amount of
+        # time (0.5s - 1s).
+        #
+        # The below sleep is functionally equivalent in terms of preventing the
+        # race condition (without the high CPU that might cause other control
+        # plane issues), but is of course not the proper fix.
+        #
+        # An upstream SONiC issue will be opened for the race condition, and
+        # until resolved leaving this comment in place for future reference.
+        time.sleep(1)
```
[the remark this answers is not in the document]
bhouse-nexthop: This *is* an issue, which is why https://github.com/sonic-net/sonic-buildimage/issues/22255 was filed.  This issue is pre-existing and the slowness of the prior code effectively hid the issue.  There is no way to resolve this in this patch set, it requires a large rework of the config_db reading logic.

generic_config_updater/patch_sorter.py
```diff
@@ -271,8 +281,26 @@ def from_operation(operation):
 
         return JsonMove(diff, op_type, current_config_tokens, target_config_tokens)
 
-    def apply(self, config):
-        return self.patch.apply(config)
+    def apply(self, config, in_place: bool = False):
+        if self.op_type == OperationType.REMOVE or self.op_type == OperationType.REPLACE:
+            self.orig_value = JsonMove._get_value(config, sonic_yang.SonicYang.configdb_path_split(self.path))
```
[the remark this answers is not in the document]
bhouse-nexthop: will look at this tomorrow to see if this copilot evaluation makes sense or not.  too jet lagged to look at it right now.

generic_config_updater/patch_sorter.py
```diff
@@ -90,6 +98,8 @@ def _to_jsonpatch_operation(diff, op_type, current_config_tokens, target_config_
     @staticmethod
     def _get_value(config, tokens):
         for token in tokens:
+            if isinstance(token, str) and token.isnumeric():
```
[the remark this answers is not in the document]
bhouse-nexthop: this evaluation is invalid.  The path tokens for lists in the schema cannot be negative.

bhouse-nexthop:
Hi @lguohan thanks for triggering copilot on this.  I left one open that I need to look at when I'm not jet lagged.  I'll do that tomorrow am.  Thanks!

generic_config_updater/patch_sorter.py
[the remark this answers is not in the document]
bhouse-nexthop: This is a completely invalid evaluation here.  _get_value() takes an array of values and is used to recurse down a json data structure, which can consist of dictionary keys or numeric values when used as an index within a list.

AI can't understand this.

bhouse-nexthop:
@lguohan can you review now that I've addressed?

lguohan (approved):

generic_config_updater/change_applier.py
```diff
@@ -137,25 +138,45 @@ def _report_mismatch(self, run_data, upd_data):
...
+        run_data = current_configdb
+        upd_data = prune_empty_table(change.apply(run_data, in_place=False))
         upd_keys = defaultdict(dict)
 
         for tbl in sorted(set(run_data.keys()).union(set(upd_data.keys()))):
             self._upd_data(tbl, run_data.get(tbl, {}), upd_data.get(tbl, {}), upd_keys)
 
         ret = self._services_validate(run_data, upd_data, upd_keys)
-        if not ret:
-            run_data = get_config_db_as_json(self.scope)
```
qiluo-msft, line 150: Why removing this block of code? It seems useful for a bugfix.

bhouse-nexthop: From the commit message for this change:

```
ChangeApplier: no need to re-read configdb twice per patch
During change application, the configdb was being read before applying
each patch, then again as a validation step.  Since we are holding
a lock on the configdb and we know the state as we are mutating it,
there is no reason we need to do this as it greatly slows down the
overall application of the patches to Redis.

There is also no reason at all to do the validation step, this appears
to be something left in during development to help find bugs but
not something that will ever catch any sort of issue post-development.
There is still a final validation step that will catch overall errors
that is minimal enough to not cause any performance concerns.
```

qiluo-msft: The motivation of https://github.com/sonic-net/sonic-utilities/pull/2295 is explained in PR description.

bradh352: PR #2295 is about `remove_backend_tables_from_config` not about re-reading configdb again.  When the configdb is infact read again, the `remove_backend_tables_from_config` is still called.

vmittal-msft:
@bhouse-nexthop please raise manual pr for 202511 as there are cherry-pick conflicts.

bhouse-nexthop:
@vmittal-msft manual 202511 cherry-pick raised as requested: #4820

There were merge conflicts against three GCU changes that landed in 202511 after this commit's base (#4484, #4486, and the regression test #4484 added); the resolutions are documented in that PR's description and commit message.

PR #3833, opened 2025-04-07
Fix show platform ssdhealth not showing expected output when a usb flash drive is plugged in

tirupatihemanth:
Fix `show platform ssdhealth` not showing expected output when a usb flash drive is plugged in. 

#### What I did
When 'show platform ssdhealth' is not supplied with a device name and a usb flash drive is attached we show incorrect output as we deduce incorrect device name from 'lsblk' command and pass it to ssdutil. I removed this incorrect deduction of the device name.

#### How I did it

Instead of incorrectly deducing the device using the lsblk command we can instead not pass any device name as argument to the ssdutil and it will automatically choose the device on which /host partition is mounted which is the primary storage device in our case as per the PR: https://github.com/sonic-net/sonic-utilities/pull/3399

#### How to verify it

Run `sudo show platform ssdhealth` without any arguments with a flash drive plugged into the sonic switch. 

```
root@qa-eth-vt19-1-4280:~# show platform ssdhealth
Disk Type    : NVME
Device Model : N/A
Health       : N/A
Temperature  : N/A
```

```
root@qa-eth-vt19-1-4280:/home/admin# show platform ssdhealth
Disk Type    : NVME
Device Model : Virtium VTPM24CEXI080-BM110006
Health       : 100.0%
Temperature  : 51.0C
```

In the case where the user passes the device name as argument to the command (`show platform ssdhealth /dev/nvme0n1`) output remains the same and no changes have been made.

### DIFF
--- a/show/platform.py
+++ b/show/platform.py
@@ -117,11 +117,13 @@ def ssdhealth(device, verbose, vendor):
         if platform_data:
             device = platform_data.get("chassis", {}).get("disk", {}).get("device", None)
 
-        if not device:
-            # Fallback to discovery
-            device = os.popen("lsblk -o NAME,TYPE -p | grep disk").readline().strip().split()[0]
+    # if device argument is not provided ssdutil will display the health of the disk containing
+    # the /host partition. In sonic this is the primary storage device.
+    if device:
+        cmd = ['sudo', 'ssdutil', '-d', str(device)]
+    else:
+        cmd = ['sudo', 'ssdutil']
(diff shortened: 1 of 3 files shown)


oleksandrivantsiv (approved):

judyjoseph (approved):
lgtm

judyjoseph:
Please add any branches it needs to be cherry-picked @tirupatihemanth

PR #3836, opened 2025-04-08
Fix the problematic implementation of "config route add/del" and add UTs

BYGX-wcr:
#### What I did

The original implementation of "config route add/del" was problematic. It expected either nexthop ip or outbound interface only, which does not conform with the [STATIC_ROUTE HLD](https://github.com/sonic-net/SONiC/blob/master/doc/static-route/SONiC_static_route_hdl.md). Besides, it assumed that nexthop-vrf is only associated with nexthop IP which was not correct. So, I fixed the implementation of config route add and config route del and added the missing unit tests for those commands.

#### How I did it

Change the code

#### How to verify it

* Run the command on a device loaded with the new implementation.
* Run the UTs

```
admin@xxx:~$ sudo config route add prefix 192.168.10.0/24 nexthop 172.17.0.1 dev Ethernet272
Usage: config route add [OPTIONS] prefix [vrf <vrf_name>] <A.B.C.D/M> nexthop
                        <[vrf <vrf_name>] <A.B.C.D>>|<dev <dev_name>>
Try "config route add -h" for help.

Error: interface dev doesn`t exist
```

```
admin@xxx:~$ sudo config route add prefix 192.168.10.0/24 nexthop 172.17.0.1 dev Ethernet272
admin@xxx:~$ show ip route 192.168.10.0/24
Routing entry for 192.168.10.0/24
  Known via "static", distance 1, metric 0
  Last update 00:00:13 ago
  * 172.17.0.1, via Ethernet272 inactive

```


r12f (approved):

r12f (dismissed):
lgtm and added @prsunny for review as well

prsunny:
@BYGX-wcr , could you please follow the template for description? Basically, capture the previous command and the new command.

BYGX-wcr:
> @BYGX-wcr , could you please follow the template for description? Basically, capture the previous command and the new command.

Hi Prince, this is a config command. There is no output to be read by user in command line.

prsunny:
> > @BYGX-wcr , could you please follow the template for description? Basically, capture the previous command and the new command.
> 
> Hi Prince, this is a config command. There is no output to be read by user in command line except some checking failure messages.

You can put the config command and use 'show command' to optionally provide the output.

BYGX-wcr:
> > > @BYGX-wcr , could you please follow the template for description? Basically, capture the previous command and the new command.
> > 
> > 
> > Hi Prince, this is a config command. There is no output to be read by user in command line except some checking failure messages.
> 
> You can put the config command and use 'show command' to optionally provide the output.

I have updated with an example compared with the previous command execution which gave an error message.

config/main.py
```diff
@@ -7139,53 +7155,59 @@ def route(ctx):
...
+            if i >= len(route['distance'].split(',')):
+                route['distance'] += ',0'
         else:
             route['distance'] = '0'
 
         if 'blackhole' in route:
-            route['blackhole'] += ',false'
+            if i >= len(route['blackhole'].split(',')):
+                # If the user configure with "ifname" as "null", set 'blackhole' attribute as true.
+                if route['ifname'].split(',')[i] == 'null':
```
prsunny, line 7207: Looking at this change, just want to confirm there are no backward compatibility or functional changes with this PR since the description mentioned it as a bug fix ONLY to add `ifname `option

prsunny (approved):
minor comment.

config/main.py
BYGX-wcr: Hi Prince, looking at the lines below 7211~7216, I am doing the same thing as the previous people did. However, previously the implementation did not consider the case where there are multiple nexthops being added in one command and a nexthop that is in the middle of the list has ifname = 'null'.

config/main.py
```diff
@@ -7139,53 +7155,59 @@ def route(ctx):
...
+        entry_counter = len(route['ifname'].split(','))
 
     # Alignment in case the command contains several nexthop ip
     for i in range(entry_counter):
+        # Set vrf to empty string if not defined
         if 'nexthop-vrf' in route:
-            if i > 0:
-                vrf = route['nexthop-vrf'].split(',')[0]
-                route['nexthop-vrf'] += ',' + vrf
+            if i >= len(route['nexthop-vrf'].split(',')):
```
BYGX-wcr, line 7178: @prsunny , there is one behavior change here actually. The previous implementation assumed that if the destination vrf of a nexthop is not specified, it will be the same as the destination vrf of the first nexthop. I don't think this implicit behavior aligns with the static-route HLD, which stated that the nexthop-vrf should be explicitly specified otherwise it is deemed as default vrf. So, I made the change, let me know if you want me to revert this change.

prsunny: Oh ok. I suggest revert that change as its not relevant to this PR and also we don't want to break some previous implementations using this assumption.

PR #3837, opened 2025-04-08
Add Sonic gnmi agent utility

aronovic:
#### What I did
Added a new utility that can load config templates to the DPUs on a smart switch box.

#### How I did it
The agent utility is parsing the j2 templates and creates google protocol buffer commands that are sent to the GNMI server to be executed.

#### How to verify it
First need to build the docker image that contains the utility and after it gets loaded on the smart switch and it gets invoked from within the docker.
All the instructions are in the README.md file


KrisNey-MSFT:
hi @vivekrnv - were there any review comments for this one?

KrisNey-MSFT:
hi @prsunny - could we have a reviewer for this one pls?

hdwhdw (changes requested):
Several high level questions here:
1. This is not a daemon, it is a simple CLI tool. Hence why do we need a new container? Can't we simply make it a binary in the image?
2. The main goal of this utility is render json and issue grpc. Have you consider other languages such as go which is much more efficient and much better at handling grpc/gnmi?
3. If we are committed to python, can we avoid patterns like `subprocess.run(`gnmi_set`)`, and instead directly use grpcio library in python? No brittle stdio from binary we don't control, no subprocess overhead, no disk overhead for json, no risk for shell injection.

aronovic:
> @aronovic do you have an HLD for this feature?

We do not need an HLD for this utility. Is just for smartswitch internal testing.

hdwhdw:
@aronovic Just spoke with @prsunny: Usually container code that goes into sonic-buildimage (and by extension its submodule like sonic-utility) either builds into sonic images, or get pushed to soniccr1.azurecr.io so sonic system can pulls. It looks like this container does neither of these. How are should we use this container, are we just building from source when we need it, or circulate the tarball out of band?

aronovic:
> @aronovic Just spoke with @prsunny: Usually container code that goes into sonic-buildimage (and by extension its submodule like sonic-utility) either builds into sonic images, or get pushed to soniccr1.azurecr.io so sonic system can pulls. It looks like this container does neither of these. How are should we use this container, are we just building from source when we need it, or circulate the tarball out of band?

So far, this has been circulated as a tarball and used by AMD team and Keysight to do configuration at higher scale from what is available in sonic-mgmt dash test directory. The users are also adding new j2 files for specific configurations.

hdwhdw:
What exactly do you mean by "do configuration at higher scale from what is available in sonic-mgmt dash test"? Do we have sonic-mgmt code that depends on this? The main concern here is this is some code without any automated continuous integration. So we don't have any good way to maintain it such as detecting regression.

Also if we don't need it inside sonic image, or we don't need it as any part of sonic pipeline, do we really have a good reason to upstream this into sonic-net repo? If we just need a tarball, we could build that from any personal repo, and it give you greater flexibility. @prsunny what do you think?

aronovic:
> What exactly do you mean by "do configuration at higher scale from what is available in sonic-mgmt dash test"? Do we have sonic-mgmt code that depends on this? The main concern here is this is some code without any automated continuous integration. So we don't have any good way to maintain it such as detecting regression.
> 
> Also if we don't need it inside sonic image, or we don't need it as any part of sonic pipeline, do we really have a good reason to upstream this into sonic-net repo? If we just need a tarball, we could build that from any personal repo, and it give you greater flexibility. @prsunny what do you think?

What I mean, is that utility has been used to configure a very large number of DASH objects in Keysight environment. 
The idea was to share it with larger community and to be built by whoever needs it rather that waiting for some tarball to be generated from our private repository in Cisco as it is currently the case.

hdwhdw:
@aronovic I’m concerned this sets a precedent for landing standalone, non-integrated utilities in sonic-utilities (and effectively sonic-buildimage). If we accept this as-is, it effectively opens the door to more one-off tools being added without a clear integration pipeline/story. Have you consider building this into the sonic image? or simply place it in sonic-buildimage so it has some integration and get pushed to ACR and people can pull and share?

But we cannot commit any code without continuous integration pipeline. Effectively they are dead code and we have no way of detecting regression. Based on what you say and the tarball can be generated from Cisco private repo, we should put it in sonic-buildimage and use its build system and pipeline to generate the tarball as well, instead of putting it here and we have to do it manually.

mgheorghe:
> @aronovic I’m concerned this sets a precedent for landing standalone, non-integrated utilities in sonic-utilities (and effectively sonic-buildimage). If we accept this as-is, it effectively opens the door to more one-off tools being added without a clear integration pipeline/story. Have you consider building this into the sonic image? or simply place it in sonic-buildimage so it has some integration and get pushed to ACR and people can pull and share?
> 
> But we cannot commit any code without continuous integration pipeline. Effectively they are dead code and we have no way of detecting regression. Based on what you say and the tarball can be generated from Cisco private repo, we should put it in sonic-buildimage and use its build system and pipeline to generate the tarball as well, instead of putting it here and we have to do it manually.

Can we then integrate it into sonic-mgmt container?
i'll also contribute the tests that are using it (some PRs for DASH HA are already raised that make use of this tool)

hdwhdw:
@mgheorghe This is a good idea. Also consider docker-sonic-ptf. Most client side code/utility for testing is there.

aronovic:
> @mgheorghe This is a good idea. Also consider docker-sonic-ptf. Most client side code/utility for testing is there.

Yes, this seems a better place for this utility.

PR #3840, opened 2025-04-09
Changes to make the docker_device_config_mode to unified

vidyac86:
#### What I did
Added migrator changes to change the docker_device_config_mode to unified

#### How to verify it
When the older version had docker_device_config_mode as separated, with this change, the newer version will always have the device_config_mode as unified

UT Logs
--------
[UT_Logs.txt](https://github.com/user-attachments/files/19689837/UT_Logs.txt)

### DIFF
--- a/scripts/db_migrator.py
+++ b/scripts/db_migrator.py
@@ -830,8 +830,6 @@ class DBMigrator():
         """
         log.log_notice('Migrate FRR Config mode configuration')
         metadata = self.configDB.get_entry('DEVICE_METADATA', 'localhost')
-        #device_metadata_data = self.config_src_data["DEVICE_METADATA"]["localhost"]
-        #metadata['docker_router_config_mode'] = "unified"
         if 'docker_routing_config_mode' not in metadata:
             metadata['docker_routing_config_mode'] = "unified"
             self.configDB.set_entry('DEVICE_METADATA', 'localhost', metadata)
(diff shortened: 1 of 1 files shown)


venkatmahalingam:
@eddieruan-alibaba @hasan-brcm @dgsudharsan @vivekrnv We have created this PR for migrating from separated mode to unified and UT logs.

scripts/db_migrator.py
```diff
@@ -822,6 +822,21 @@ def migrate_flex_counter_delay_status_removal(self):
             self.configDB.set_entry('FLEX_COUNTER_TABLE', obj, flex_counter)
 
 
+    def migrate_frr_separate_config_mode (self):
+        """
+        Modify docker_routing_config_mode field.
+        """
+        log.log_notice('Migrate FRR Config mode configuration')
+        metadata = self.configDB.get_entry('DEVICE_METADATA', 'localhost')
+        if 'docker_routing_config_mode' not in metadata:
```
dgsudharsan, line 833: I don't think we need to set docker_routing_config_mode explicitly if not configured since we are already assuming its unified in frr init.
We only need to add configuration when the mode is set as separate. Can you please remove the if condition?

dgsudharsan (changes requested):

StormLiangMS:
hi @vidyac86 could you check the CI failures?

PR #3841, opened 2025-04-09
[SRv6] add support for SRv6 counters

Yakiv-Huryk:
HLD: https://github.com/sonic-net/SONiC/pull/1959

#### What I did
Added support for SRv6 counters CLI

#### How I did it
* extend counterpoll
* extend sonic-clear
* add 'show srv6 stats' command

#### How to verify it
Running the updates tests


scripts/srv6stat
BYGX-wcr, line 1: I would suggest you put the implementation of class SRv6Stat in utilities_common/srv6stat.py and then only keep the scripting logic in scripts/srv6stat. We did that for portstat because we found that some other system components may want to call the API to get counter info. In that scenario, having SRv6Stat in utilities_common will be much easier.

BYGX-wcr:
@qiluo-msft, would you like to take a look?

qiluo-msft:
Could you add the HLD PR link into PR description?

BYGX-wcr:
Hi @qiluo-msft, the link to HLD has been added in the PR description. Are we good to merge this PR?

qiluo-msft:
>     except Exception as e:

Could you use more specific exception type?

---
Refers to: show/main.py:2262 in e90abda. [](commit_id = e90abdaae1909936da3aab9c07f204523c59ff5d, deletion_comment = False)

qiluo-msft:
>         pass

Could you add some logging?

---
Refers to: show/main.py:2263 in e90abda. [](commit_id = e90abdaae1909936da3aab9c07f204523c59ff5d, deletion_comment = False)

utilities_common/srv6stat.py
```diff
@@ -0,0 +1,82 @@
...
+                data[entry] = counter_stats
+        return data
+
+    def fetch_saved_counters(self):
+        if os.path.isfile(self.user_cache_file):
+            try:
+                with open(self.user_cache_file) as f:
+                    return json.load(f)
+            except Exception:
+                pass
```
qiluo-msft, line 50: >pass [](http://example.com/codeflow?start=16&length=4)

Could you add some logging?

utilities_common/srv6stat.py
```diff
@@ -0,0 +1,82 @@
...
+        return {}
+
+    def get_counter_value(self, key, stats, saved_stats, stat_type):
+        old_value = saved_stats.get(key, {}).get(stat_type, '0')
+        db_value = stats[stat_type]
```
qiluo-msft, line 55: >stat_type [](http://example.com/codeflow?start=25&length=9)

Do you need to check whether stat_type exists?

PR #3845, opened 2025-04-11
Chassisd does not wait for the execution to complete for previous admin state change requests

rameshraghupathy:
Fixed issue #22138

#### What I did
Fixed issue #22138
Chassisd does not wait for the execution to complete for previous admin state change requests. Added support to solve this.

#### How I did it
Added a state to indicate the progress of an admin_state transition period (window) of every module in config_db.
When in progress do not allow another admin_state change for that module.
Mark the beginning of state transition in the CLI itself.  Added a timeout also for error cases.
Mark the end of state transition in chassisd.

#### How to verify it
Issue "config chassis module startup DPU0".  This takes around 3 mins. Check the config_db and verify the new variable is set to True.
Issue  another "config chassis module startup DPU0" and also ""config chassis module shutdown DPU0". 
See both are blocked.
Once the DPU operational state is the admin_sate check if a new config is accepted.

### DIFF
--- a/config/chassis_modules.py
+++ b/config/chassis_modules.py
@@ -6,9 +6,10 @@ import re
 import subprocess
 import utilities_common.cli as clicommon
 from utilities_common.chassis import is_smartswitch, get_all_dpus
+from datetime import datetime, timedelta
 
 TIMEOUT_SECS = 10
-
+TRANSITION_TIMEOUT = timedelta(seconds=240)  # 4 minutes
 
 #
 # 'chassis_modules' group ('config chassis_modules ...')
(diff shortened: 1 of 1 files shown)


config/chassis_modules.py
```diff
@@ -36,6 +37,38 @@ def get_config_module_state(db, chassis_module_name):
         return fvs['admin_status']
 
 
+def get_state_transition_in_progress(db, chassis_module_name):
+    fvs = db.cfgdb.get_entry('CHASSIS_MODULE', chassis_module_name)
```
gpunathilell, line 41: In chassisd the changes are deleted from the chassis_state_db, we need to use the same here instead of config_db

rameshraghupathy: @gpunathilell  @vvolam   This is the module admin status and is configured and maintained in configDB. So, this is the right DB.

gpunathilell: In the other PR: https://github.com/sonic-net/sonic-platform-daemons/blob/4c7e1ba132dc6c0e1330b0f33c1ebe26bf2e328e/sonic-chassisd/scripts/chassisd#L713 We are setting the state transition to false in state_db module_table

config/chassis_modules.py
```diff
@@ -149,16 +220,32 @@ def startup_chassis_module(db, chassis_module_name):
...
-        fvs = {'admin_status': 'up'}
+        if get_state_transition_in_progress(db, chassis_module_name) == 'True':
+            if is_transition_timed_out(db, chassis_module_name):
+                set_state_transition_in_progress(db, chassis_module_name, 'False')
+                click.echo(f"Previous transition for module {chassis_module_name} timed out. Proceeding with startup.")
+            else:
+                click.echo(f"Module {chassis_module_name} state transition is already in progress")
+                return
+        else:
+            set_state_transition_in_progress(db, chassis_module_name, 'True')
```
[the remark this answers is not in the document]
rameshraghupathy: @gpunathilell Not necessarily, however any script executing back to back CLI calls may pose issues if I remove this.  So, it is safe to keep it in two places.  If you are particular and if it works for you I can remove it.

gpunathilell:
@rameshraghupathy why was this closed?

PR #3853, opened 2025-04-14
[202411] Stop the hwclock service and timer files before updating CMOS time

saiarcot895:
#### What I did

When rebooting or doing a warm/fast reboot, stop the hwclock service and timer files before running `hwclock -w`. This removes the chance of a race condition where both of them run at the same time, causing one of them to fail.

Fixes sonic-net/sonic-buildimage#22230

#### How to verify it

This is a race condition, and as such, may require a couple hundred reboot attempts to occur. With the fix, it's expected that this race condition will not occur at all.

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -902,6 +902,7 @@ sleep 1
 sync
 
 # sync the current system time to CMOS
+systemctl stop hwclock.service hwclock.timer
 if [ -x /sbin/hwclock ]; then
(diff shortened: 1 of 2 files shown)


yxieca (approved):

yxieca:
@saiarcot895 why do we not need this fix in master branch?

saiarcot895:
@yxieca This change (of having the hwclock service and timer files) is not present in the master branch because Chrony is handling the RTC.

PR #1964, opened 2021-12-09
preserve old order for config reload

arlakshm:
Signed-off-by: Arvindsrinivasan Lakshmi Narasimhan <arlakshm@microsoft.com>

#### What I did
Fixes https://github.com/Azure/sonic-buildimage/issues/9473
#### How I did it
Preserve the old order of the config reload

#### How to verify it
Manual tests
- change `SFLOW` feature state
```
admin@vlab-01:~$ show feature status
Feature         State            AutoRestart     SetOwner
--------------  ---------------  --------------  ----------
bgp             enabled          enabled
database        always_enabled   always_enabled
dhcp_relay      enabled          enabled         local
gbsyncd         enabled          enabled
lldp            enabled          enabled
macsec          disabled         enabled
mgmt-framework  enabled          enabled
mux             always_disabled  enabled
nat             disabled         enabled
pmon            enabled          enabled
radv            enabled          enabled
sflow           disabled         enabled
snmp            enabled          enabled
swss            enabled          enabled
syncd           enabled          enabled
teamd           enabled          enabled
telemetry       enabled          enabled
```
```
admin@vlab-01:~$ sudo config feature state sflow enabled
admin@vlab-01:~$ show feature status
Feature         State            AutoRestart     SetOwner
--------------  ---------------  --------------  ----------
bgp             enabled          enabled
database        always_enabled   always_enabled
dhcp_relay      enabled          enabled         local
gbsyncd         enabled          enabled
lldp            enabled          enabled
macsec          disabled         enabled
mgmt-framework  enabled          enabled
mux             always_disabled  enabled
nat             disabled         enabled
pmon            enabled          enabled
radv            enabled          enabled
sflow           enabled          enabled
snmp            enabled          enabled
swss            enabled          enabled
syncd           enabled          enabled
teamd           enabled          enabled
telemetry       enabled          enabled
```
- save config and do config reload
```
admin@vlab-01:~$ sudo config save
Existing files will be overwritten, continue? [y/N]: y
Running command: /usr/local/bin/sonic-cfggen -d --print-data > /etc/sonic/config_db.json
admin@vlab-01:~$ sudo config reload -y
Running command: rm -rf /tmp/dropstat-*
Disabling container monitoring ...
Stopping SONiC target ...
Running command: /usr/local/bin/sonic-cfggen  -j /etc/sonic/init_cfg.json  -j /etc/sonic/config_db.json  --write-to-db
Running command: /usr/local/bin/db_migrator.py -o migrate
[...]

PR #604, opened 2019-08-14
[neighbor_advertiser] Use full vlan name for vxlan tunnel map programming

pavel-shirshov:
**- What I did**

**- How I did it**
Fixed a bug which caused incorrect programming of vxlan tunnel map.

**- How to verify it**
Run the script and check that syslog doesn't have any errors.

**- Previous command output (if the output of a command-line utility has changed)**

**- New command output (if the output of a command-line utility has changed)**

-->

PR #3857, opened 2025-04-16
[muxcable] Add grpc CLI support for transceiver presence retrieval fr…

mihirpat1:
#### What I did
xcvrd will now move the `status` field from the TRANSCEIVER_STATUS table to the TRANSCEIVER_STATUS_SW table. This is being done via the below PRs

[Update CMIS diagnostics monitoring HLD with TRANSCEIVER_STATUS_SW table by mihirpat1 · Pull Request #1964 · sonic-net/SONiC](https://github.com/sonic-net/SONiC/pull/1964)
[[xcvrd] Re-organize transceiver DOM and STATUS tables by mihirpat1 · Pull Request #604 · sonic-net/sonic-platform-daemons](https://github.com/sonic-net/sonic-platform-daemons/pull/604)

Therefore, the corresponding CLIs that rely on the TRANSCEIVER_STATUS table for retrieving transceiver presence need to be changed.

However, a more robust method to retrieve transceiver presence is by checking if the TRANSCEIVER_INFO table is present. Thus, the relevant code change will be made to address this.

#### How I did it
The CLI handler now looks at the TRANSCEIVER_INFO table to find transceiver presence

### DIFF
--- a/show/muxcable.py
+++ b/show/muxcable.py
@@ -2177,7 +2177,7 @@ def get_grpc_cached_version_mux_direction_per_port(db, port):
     mux_info_full_dict[asic_index] = state_db[asic_index].get_all(
         state_db[asic_index].STATE_DB, 'MUX_CABLE_INFO|{}'.format(port))
     trans_info_full_dict[asic_index] = state_db[asic_index].get_all(
-        state_db[asic_index].STATE_DB, 'TRANSCEIVER_STATUS|{}'.format(port))
+        state_db[asic_index].STATE_DB, 'TRANSCEIVER_INFO|{}'.format(port)) or {}
 
     res_dir = {}
     res_dir = mux_info_full_dict[asic_index]
@@ -2187,9 +2187,8 @@ def get_grpc_cached_version_mux_direction_per_port(db, port):
 
     trans_dir = {}
     trans_dir = trans_info_full_dict[asic_index]
-    
-    status = trans_dir.get("status", "0")
-    presence = "True" if status == "1" else "False"
+
+    presence = "True" if trans_dir else "False"
 
     mux_info_dict["presence"] = presence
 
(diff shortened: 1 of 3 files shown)


prgeor (dismissed):

show/muxcable.py
```diff
@@ -2177,7 +2177,7 @@ def get_grpc_cached_version_mux_direction_per_port(db, port):
     mux_info_full_dict[asic_index] = state_db[asic_index].get_all(
         state_db[asic_index].STATE_DB, 'MUX_CABLE_INFO|{}'.format(port))
     trans_info_full_dict[asic_index] = state_db[asic_index].get_all(
-        state_db[asic_index].STATE_DB, 'TRANSCEIVER_STATUS|{}'.format(port))
+        state_db[asic_index].STATE_DB, 'TRANSCEIVER_INFO|{}'.format(port)) or {}
```
vdahiya12, line 2180: what is the or condition here ?

show/muxcable.py
```diff
@@ -2187,9 +2187,8 @@ def get_grpc_cached_version_mux_direction_per_port(db, port):
 
     trans_dir = {}
     trans_dir = trans_info_full_dict[asic_index]
```
vdahiya12, line 2189: shall  we add a get call here, that would be easier to read the code below
```
trans_dir = trans_info_full_dict.get(asic_index, None)
```

vdahiya12:
Do we have a negative test case, to give presence as False if table not populated?

please check.

show/muxcable.py
mihirpat1: @vdahiya12 The or condition is just to ensure that an empty dictionary is returned rather None to avoid a crash.

show/muxcable.py
mihirpat1: @vdahiya12 Makes sense. Added it now.

vdahiya12 (approved):

PR #3860, opened 2025-04-22
Issue #22407: ConfigReload fails when RADIUS options are enabled

anders-nexthop:
#### What I did

Enabling RADIUS `statistics` adds config entries which are not valid in YANG.  The option adds the True/False boolean values instead of the 'true'/'false' booleans that are defined in the YANG model, which causes failures when trying to reload/replace the config.

Fixes: [#22407](https://github.com/sonic-net/sonic-buildimage/issues/22407)

#### How I did it

Modified relevant config command to use the correct values.

#### How to verify it

Manually verified.

I added a test case for `config radius statistics` command. Note that this test follows the pattern in this file, which is not actually exercising the correct codepath (see https://github.com/sonic-net/sonic-utilities/issues/3754 for details).

### DIFF
--- a/config/aaa.py
+++ b/config/aaa.py
@@ -498,9 +498,9 @@ def statistics(db, option):
         del_table_key(db, 'RADIUS', 'global', 'statistics')
     else:
         if option == 'enable':
-            add_table_kv('RADIUS', 'global', 'statistics', True)
+            add_table_kv(db, 'RADIUS', 'global', 'statistics', 'true')
         elif option == 'disable':
-            add_table_kv('RADIUS', 'global', 'statistics', False)
+            add_table_kv(db, 'RADIUS', 'global', 'statistics', 'false')
 radius.add_command(statistics)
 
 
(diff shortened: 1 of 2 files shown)


bhouse-nexthop:
yeah, likely this would have fixed your issue. As writing a boolean to the database is perfectly legal, the problem is the `config save -y` (which calls sonic-cfggen to write the config) was writing boolean values as True and False in the generated config_db which isn't valid for yang.  So any time you tried to use that config again with an action that performed yang validation, it would fail on any of those values.

It looks like others worked around it by turning those into strings in yang instead of boolean, which is just wrong.

anders-nexthop:
@qiluo-msft can you please review this PR? @prgeor has you tagged in [Issue 22407](https://github.com/sonic-net/sonic-buildimage/issues/22407) as a potential reviewer

qiluo-msft:
@a-barboza Could you help review?

qiluo-msft (approved):
LGTM

bhouse-nexthop:
directly related to this same issue is:
https://github.com/sonic-net/sonic-buildimage/pull/22526
https://github.com/sonic-net/sonic-host-services/pull/252

PR #3861, opened 2025-04-23
Set the timeout to be larger on VS chassis and speedup a bit

BYGX-wcr:
#### What I did

The route_check.py on UpStream LC of virtual chassis takes a longer time to complete while the current timeout limit is 2 mins. Thus, I want to:
1. Set a larger timeout when the device is a virtual chassis.
2. Speedup the route_check process a bit by simplifying the string manipulation.

### DIFF
--- a/scripts/route_check.py
+++ b/scripts/route_check.py
@@ -240,7 +240,7 @@ def checkout_rt_entry(k):
     :return (True, ip) or (False, None)
     """
     if k.startswith(ASIC_KEY_PREFIX):
-        e = k.lower()[len(ASIC_KEY_PREFIX) + 9:].split("\"", 1)[0]
+        e = k.lower()[len(ASIC_KEY_PREFIX) + len('{"dest":"'):].split("\"", 1)[0]
         if not is_local(e):
             return True, e
     return False, None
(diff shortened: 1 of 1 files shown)


scripts/route_check.py
```diff
@@ -240,7 +240,7 @@ def checkout_rt_entry(k):
     :return (True, ip) or (False, None)
     """
     if k.startswith(ASIC_KEY_PREFIX):
-        e = k.lower().split("\"", -1)[3]
+        e = k.lower()[len(ASIC_KEY_PREFIX) + 9:].split("\"", 1)[0]
```
abdosi, line 243: can you explain this one ?

qiluo-msft: same question. 9 is magic number.

BYGX-wcr:
This PR invokes a new API added in https://github.com/sonic-net/sonic-buildimage/pull/22393. So the CI cannot pass until the prior PR is merged.

scripts/route_check.py
qiluo-msft: Could you get 9 from strlen of this IP prefix?

BYGX-wcr: Sure, updated

qiluo-msft (approved):

PR #3862, opened 2025-04-23
Issue #22420: Modify 'config route add' command not to include empty …

anders-nexthop:
#### What I did

The config route commands accept a `nexthop-vrf` as an argument. When that argument is not specified, the commands use the empty string as the default. This violates the yang model, which only accepts vrf names of the form (`default`, `mgmt`, or `Vrf*`). In the case where a `nexthop-vrf` is not defined, we should fall back to using the same vrf given for the prefix we are adding. If that is not specified, then both prefix and nexthop will use the default vrf.

Fixes: [#22420](https://github.com/sonic-net/sonic-buildimage/issues/22420)

#### How I did it

Modify the` add_route()` and `del_route()` codepaths from sonic-utilites/config/main.py to set nexthop-vrf to whatever the vrf is for the associated prefix whenever it is not explicitly specified.

#### How to verify it

All unit tests pass (including all tests in `static_route_test.py`).

I don't think the output of any show commands changes because of this, but it's possible there are combinations of different VRFs which would show up differently. Before this merges I'll set up some routes and nexthops with different VRFs and make sure any changes there may be are shown here.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1248,7 +1248,7 @@ def cli_sroute_to_config(ctx, command_str, strict_nh = True):
     else:
         key = ip_prefix
 
-    return key, config_entry
+    return key, config_entry, vrf_name
 
 def update_sonic_environment():
     """Prepare sonic environment variable using SONiC environment template file.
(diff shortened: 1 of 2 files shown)


prsunny:
@anders-nexthop , do you know if this issue was caused by any recent PRs?

bingwang-ms:
@prsunny Can you please help check again?

prsunny (approved):

PR #3864, opened 2025-04-24
[db_migrator] Add version_202411_02

benle7:
#### What I did
Fixed migration error during upgrade from 202411_02.

#### How I did it
- Added version_202411_02 function in db_migrator to support proper migration during upgrade
(called dynamically via getattr from the migrate function).
- Updated db_migrator_test accordingly.

#### How to verify it
Manual:
- Use canonical setup with 202411 image
- Verify the version in config_db and Redis is version_202411_02
hget VERSIONS|DATABASE VERSION
vim /etc/sonic/config_db.json
- Install master image and reboot
- Check syslog for db_migrator errors or run config reload

Automatic:
- Use canonical setup with 202411 image
- Verify the version in config_db and Redis is version_202411_02
- Run test_deploy-and-upgrade

Additional Option:
- Verify the version in config_db and Redis is version_202411_02
- Run the following command to migrate (to 202505_01):
/usr/local/bin/db_migrator.py -o migrate

### DIFF
--- a/scripts/db_migrator.py
+++ b/scripts/db_migrator.py
@@ -1293,6 +1293,14 @@ class DBMigrator():
         Version 202411_01.
         """
         log.log_info('Handling version_202411_01')
+        self.set_version('version_202411_02')
+        return 'version_202411_02'
+
+    def version_202411_02(self):
+        """
+        Version 202411_02.
+        """
+        log.log_info('Handling version_202411_02')
         self.set_version('version_202505_01')
         return 'version_202505_01'
 
--- a/tests/db_migrator_test.py
+++ b/tests/db_migrator_test.py
@@ -88,9 +88,9 @@ class TestVersionComparison(object):
                                   {'v1': 'version_202311_02', 'v2': 'version_202311_01', 'result': True},
                                   {'v1': 'version_202305_01', 'v2': 'version_202311_01', 'result': False},
                                   {'v1': 'version_202311_01', 'v2': 'version_202305_01', 'result': True},
-                                  {'v1': 'version_202405_01', 'v2': 'version_202411_01', 'result': False},
-                                  {'v1': 'version_202411_01', 'v2': 'version_202405_01', 'result': True},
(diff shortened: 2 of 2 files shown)


keboliu (approved):

stepanblyschak (approved):

benle7:
@qiluo-msft

benle7:
@ryanzhu706 @vaibhavhd

keboliu:
Hi @ryanzhu706 @vaibhavhd would you please help to review?

scripts/db_migrator.py
```diff
@@ -1293,6 +1293,14 @@ def version_202411_01(self):
         Version 202411_01.
         """
         log.log_info('Handling version_202411_01')
+        self.set_version('version_202411_02')
```
vaibhavhd, line 1296: I am confused. Why does 202411 branch have a different code for the DB version 202411_01.

I think the version view for a specific DB versions should be same across the branches. Otherwise it not just gets confusing to follow, but could also mean two different things, on same version. When new branches are cut they will have different config migration for the same version. That's not expected?

https://github.com/sonic-net/sonic-utilities/blob/e6fb8f963c705ff1782d10f39638828b459bb367/scripts/db_migrator.py#L1277

benle7: Hi @vaibhavhd ,

I'm not sure I fully understand what you mean.

I didn't change any existing migration logic for version_202411_01.
I only added version_202411_02() to the migration flow so that if config_db.json starts at version_202411_02,
the migrate() function (which uses getattr(self, version)()) can proceed without error.

Could you help clarify your concern and what you'd suggest as the preferred approach in this case?

Thanks!

vaibhavhd: I am assuming that my question is best answered by @stepanblyschak.
I had it when I check your PR and see difference in version definitions. I think we are missing an enforcement or guideline around this kind of change.

I will sign off on your change as it is not responsible for the confusion that I have.

vaibhavhd (approved):

PR #3872, opened 2025-05-02
Fix warm-reboot script so it can be run via reboot DBus service.

jkmar:
Use whoami if logname failed.

### DIFF
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -1,6 +1,7 @@
 #!/bin/bash -e
 
-REBOOT_USER=$(logname)
+# reboot DBus service, which uses this script, doesn't use a login shell
+REBOOT_USER=$(logname || whoami)
 REBOOT_TIME=$(date)
 REBOOT_CAUSE_FILE="/host/reboot-cause/reboot-cause.txt"
 WARM_DIR=/host/warmboot


jkmar:
Please review @kwangsuk @anand-kumar-subramanian @mbalachandar @amrutasali @sachinholla @sneelam20

vvolam:
@vaibhavhd could you help review this PR

anders-nexthop (commented):
The change makes sense. Could we add a test that would have caught this?

jkmar:
@anders-nexthop I've add a UT

tests/fast_reboot_test.py
```diff
@@ -0,0 +1,25 @@
...
+        test_path = os.path.dirname(os.path.abspath(__file__))
+        fast_reboot = os.path.join(test_path, '..', 'scripts', 'fast-reboot')
+
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            whoami = os.path.join(tmp_dir, 'whoami')
+            with open(whoami, 'w') as f:
+                f.write('echo root')
+                os.chmod(whoami, 0o700)
+
+            shutil.copy('/bin/false', os.path.join(tmp_dir, 'logname'))
```
anders-nexthop, line 17: should we test the `logname` case as well? Or is there an existing test for it?

anders-nexthop (approved):
LGTM. You have a spacing issue that's failing the code analyzer.

scripts/fast-reboot
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash -e
 
-REBOOT_USER=$(logname)
+# reboot DBus service, which uses this script, doesn't use a login shell
+REBOOT_USER=$(logname || whoami)
```
vaibhavhd, line 4: Minor comment - use `REBOOT_USER=$(logname 2>/dev/null || whoami)`?

If logname fails, its error message will be printed before whoami is executed. By redirecting stderr to `null`, you prevent the error message from cluttering the logs or terminal output.

vaibhavhd (approved):
Added a minor suggestion, to prevent error logs that are not really needed.

vvolam:
@Verma-Anukul 202505 might be very late to incorporate this change.

@yejianquan will you be able to include in 202505?

PR #3901, opened 2025-05-29
Issue #22759: Prevent CLI from adding invalid routed interfaces

anders-nexthop:
#### What I did

Currently, config interface ip add will accept EthernetX interfaces which do not have an underlying port or portchannel associated with them. This violates the yang model, so if these entries are added to ConfigDB, 'config replace' and 'config reload' don't work properly. To make matters worse, the implementation of config interface ip remove does not detect these invalid entries, so once they are added they cannot be removed through the CLI. I modified the command to reject invalid cases.

Fixes: [#22759](https://github.com/sonic-net/sonic-buildimage/issues/22759)

#### How I did it

Add a check for this case into the config interface ip add CLI command to avoid invalid cases of this form by checking to ensure that the interface which is passed is associated with a valid port or port channel if it is an Ethernet or PortChannel interface. There is already an existing check for vlan interfaces.

#### How to verify it

Modify existing test cases to handle expected behavior, verify that they all pass.

```
admin@gold113-dut:~$ sudo config interface ip add EthernetNR 1.2.3.4
admin@gold113-dut:~$ sudo config interface ip add EthernetJustMadeUp 1.2.3.4
admin@gold113-dut:~$ sudo config interface ip add Ethernet99999999 1.2.3.4
admin@gold113-dut:~$ sudo config interface ip add Ethernet-totally-not-real 1.2.3.4
admin@gold113-dut:~$ sudo config interface ip remove EthernetNR 1.2.3.4
Cannot find device "EthernetNR"
admin@gold113-dut:~$ sudo config replace -d /etc/sonic/config_db.json
** DRY RUN EXECUTION **
Config Replacer: Config replacement starting.
Config Replacer: Target config length: 50412.
Config Replacer: Getting current config db.
Config Replacer: Generating patch between target config and current config db.
Config Replacer: Applying patch using 'Patch Applier'.
Patch Applier: localhost: Patch application starting.
Patch Applier: localhost: Patch: [{"op": "remove", "path": "/INTERFACE/EthernetJustMadeUp"}, {"op": "remove", "path": "/INTERFACE/Ethernet-totally-not-real"}, {"op": "remove", "path": "/INTERFACE/EthernetNR"}, {"op": "remove", "path": "/INTERFACE/Ethernet99999999"}, {"op": "remove", "path": "/INTERFACE/EthernetJustMadeUp|1.2.3.4~132"}, {"op": "remove", "path": "/INTERFACE/Ethernet99999999|1.2.3.4~132"}, {"op": "remove", "path": "/INTERFACE/Ethernet-totally-not-real|1.2.3.4~132"}, {"op": "remove", "path": "/INTERFACE/EthernetNR|1.2.3.4~132"}]
Patch Applier: localhost getting current config db.
Patch Applier: localhost: simulating the target full config after applying the patch.
Patch Applier: localhost: validating all JsonPatch operations are permitted on the specified fields
Patch Applier: localhost: validating target config does not have empty tables,
Patch Applier: localhost: sorting patch updates.
Failed to replace config
Usage: config replace [OPTIONS] TARGET_FILE_PATH
Try "config replace -h" for help.

Error: Data Loading Failed
Leafref "/sonic-port:sonic-port/sonic-port:PORT/sonic-port:PORT_LIST/sonic-port:name" of value "EthernetNR" points to a non-existing leaf.
```

```
admin@gold113-dut:~$ sudo config interface ip add EthernetNR.10 1.2.3.4
Usage: config interface ip add [OPTIONS] <interface_name> <ip_addr> <default
                               gateway IP address>
Try "config interface ip add -h" for help.

Error: Interface EthernetNR.10 does not exist
```

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -5288,7 +5288,7 @@ def add_interface_ip(ctx, interface_name, ip_addr, gw, secondary):
     # Add a validation to check this interface is not a member in vlan before
     vlan_member_table = config_db.get_table('VLAN_MEMBER')
     if (interface_is_in_vlan(vlan_member_table, base_interface_name)):
-        click.fail("Interface {} is a member of vlan\nAborting!".format(base_interface_name))
+        ctx.fail("Interface {} is a member of vlan\nAborting!".format(base_interface_name))
         return
 
     if base_table_name == "INTERFACE" or base_table_name == "PORTCHANNEL_INTERFACE":
--- a/tests/ip_config_test.py
+++ b/tests/ip_config_test.py
@@ -87,7 +87,6 @@ class TestConfigIP(object):
                 ["Ethernet12", "1.1.1.1/24"],
                 obj=obj)
         print(result.exit_code, result.output)
-        assert result.exit_code == 0
         assert "Interface Ethernet12 is a member of vlan" in result.output
 
         # config int ip add Ethernet12.10 1.1.1.1/24
@@ -104,7 +103,6 @@ class TestConfigIP(object):
                 ["PortChannel1001", "1.1.1.1/24"],
                 obj=obj)
         print(result.exit_code, result.output)
-        assert result.exit_code == 0
         assert "Interface PortChannel1001 is a member of vlan" in result.output
 
         # config int ip add PortChannel1001.10 1.1.1.1/24
@@ -113,7 +111,6 @@ class TestConfigIP(object):
                 ["PortChannel1001.10", "1.1.1.1/24"],
                 obj=obj)
         print(result.exit_code, result.output)
-        assert result.exit_code == 0
         assert "Interface PortChannel1001 is a member of vlan" in result.output
 
 
--- a/tests/vlan_test.py
+++ b/tests/vlan_test.py
@@ -1490,7 +1490,6 @@ class TestVlan(object):
         result = runner.invoke(config.config.commands["interface"].commands["ip"].commands["add"],
                                ["Ethernet4", "10.10.10.1/24"], obj=obj)
         print(result.exit_code, result.output)
-        assert result.exit_code == 0
         assert 'Interface Ethernet4 is a member of vlan\nAborting!\n' in result.output
 
     def test_config_vlan_add_member_of_portchannel(self):


bingwang-ms:
@qiluo-msft Can you please review or find someone to review this change? Thanks

config/main.py
```diff
@@ -5190,39 +5190,60 @@ def add_interface_ip(ctx, interface_name, ip_addr, gw, secondary):
...
-                .format(interface_name))
-
-    # Add a validation to check this interface is in routed mode before
-    # assigning an IP address to it
-
-    sub_intf = False
+    # Add a validation to check this interface is not a member in vlan before
+    vlan_member_table = config_db.get_table('VLAN_MEMBER')
+    if (interface_is_in_vlan(vlan_member_table, base_interface_name)):
+        click.echo("Interface {} is a member of vlan\nAborting!".format(base_interface_name))
```
qiluo-msft: >echo [](http://example.com/codeflow?start=14&length=4)

echo -> fail? #Closed

anders-nexthop:
> > def remove(ctx, interface_name, ip_addr):
> 
> Is there a bug in "config interface ip remove" implementation? should it "detect these invalid entries" ?
> 
> Refers to: config/main.py:5323 in [c4e3812](https://github.com/sonic-net/sonic-utilities/commit/c4e3812644647b08834bb084276d8135a85a1aa3). [](commit_id = [c4e3812](https://github.com/sonic-net/sonic-utilities/commit/c4e3812644647b08834bb084276d8135a85a1aa3), deletion_comment = False)

I thought about whether to address this in `remove()` but decided not to. These entries should never have been put into ConfigDB, and they break YANG validation. Adding code into `remove()` to handle invalid cases seems like the wrong way to address the problem, since it would cover up a real error. Even if all we do is detect these entries and warn about them, that's still a lot of defensive coding.

That was my thinking, but maybe there's a precedent for `config` functions handling invalid configuration data? Is the expectation that functions should try to do something smart in these cases? Even if that's just to try and detect bad config state and warn the user about it?

config/main.py
qiluo-msft: still applicable.

anders-nexthop: done

tests/ip_config_test.py
```diff
@@ -105,6 +140,24 @@ def test_add_del_interface_valid_ipv4(self):
...
+        assert ('PortChannel0001.10', '32.11.10.1/24') in db.cfgdb.get_table('VLAN_SUB_INTERFACE')
+
+        # config int ip add Po0001.10 32.11.10.1/24
+        result = runner.invoke(
+                config.config.commands["interface"].commands["ip"].commands["add"],
+                ["Po0001.11", "32.11.10.1/24"],
+                obj=obj)
+        print(result.exit_code, result.output)
+        assert result.exit_code == 0
+        assert ('Po0001.11', '32.11.10.1/24') in db.cfgdb.get_table('VLAN_SUB_INTERFACE')
```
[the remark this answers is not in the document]
anders-nexthop: Done.

tests/vrrp_test.py
```diff
@@ -36,18 +36,18 @@ def test_add_del_vrrp_instance_without_vip(self):
...
+        assert ('Ethernet60', '9.9.9.1/24') in db.cfgdb.get_table('INTERFACE')
 
-        # config int vrrp remove Ethernet63 7
+        # config int vrrp remove Ethernet60 7
         result = runner.invoke(config.config.commands["interface"].commands["vrrp"].commands["remove"],
-                               ["Ethernet63", "7"], obj=obj)
+                               ["Ethernet60", "7"], obj=obj)
         print(result.exit_code, result.output)
-        assert "Ethernet63 dose not configured the vrrp instance 7!" in result.output
+        assert "Ethernet60 dose not configured the vrrp instance 7!" in result.output
```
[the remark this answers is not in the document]
anders-nexthop: Unfortunately, this typo exists in several places and has been in the wild for some time. Changing it would introduce changes to existing CLI command output, which is a larger scope than this PR needs. It will have to wait for another time.

qiluo-msft (approved):

config/main.py
```diff
@@ -5237,6 +5258,7 @@ def add_interface_ip(ctx, interface_name, ip_addr, gw, secondary):
             click.echo("Interface {} is in {} mode and needs to be in routed mode!".format(
                 interface_name, interface_mode))
             return
+
```
manoharan-nexthop, line 5315: The below block for handling `eth0` should be moved to line 5200 (before the interface name validation).

manoharan-nexthop (changes requested):

config/main.py
anders-nexthop: done.

anders-nexthop:
@manoharan-nexthop thanks for the review, I've made the change you requested. @prsunny @qiluo-msft can I please get another code owner to have a look? thanks!

manoharan-nexthop (approved):

config/main.py
```diff
@@ -5293,7 +5253,65 @@ def add_interface_ip(ctx, interface_name, ip_addr, gw, secondary):
...
-        ctx.fail("'interface_name' is not valid. Valid names [Ethernet/PortChannel/Vlan/Loopback]")
+        ctx.fail(f"{interface_name} is not valid. Valid names [Ethernet/PortChannel/Vlan/Loopback]")
+
+    # Add a validation to check this interface is in routed mode before
+    # assigning an IP address to it. For sub-interfaces, check that the base
+    # interface is in the right mode
+
+    base_interface_name = interface_name
+
+    # Handle short name
```
qiluo-msft, line 5264: >short name [](http://example.com/codeflow?start=13&length=10)

why design this short name?  #Closed

anders-nexthop: This is to conform with existing behaviour. Currently, when adding a subinterface, there is no verification at all that the base interface exists. This is the same underlying issue as the main problem, which is that we are adding an interface/subinterface that does not have a Port associated with it. It's even worse in this case, since we don't even check that the form of the base interface name is correct. This PR addresses that issue as well, since in order to verify that the base interface which a subinterface is attached to is a valid interface which is connected to a port, we must look that interface up and ensure it exists.

Existing tests cover cases like `config int ip add Eth36.10 32.11.10.1/24`, which they expect to work. I don't know if this is intended or is just an oversight, but I erred on the side of ensuring that existing tested behavior was retained. In the original code there is *no* name validation, but the test cases only ever check "short names". I opted to add logic to accept only "Eth/Po" short names, since these are the only ones accepted in the test (well, only "Eth" is used in the test because there were not portchannel tests of this command, but I thought if we support a short name for one interface type we should support if the other). To match existing behavior entirely, we would just check that the given base interface name is a prefix of the full name and accept stuff like "E", "Et", "Ethere", etc. However, since those forms are not tested, I decided to shrink the scope of acceptable subinterface names to just the commonly used ones.

Note that these "short names" are only usable when adding a subinterface; there is an existing check just above this that will fail a regular command that tries to use interface short names:
```
    table_name = get_interface_table_name(interface_name)
    if table_name == "":
        ctx.fail("'interface_name' is not valid. Valid names [Ethernet/PortChannel/Vlan/Loopback]")
        ctx.fail(f"{interface_name} is not valid. Valid names [Ethernet/PortChannel/Vlan/Loopback]")
```
If this is not something we think we want to support, I would gladly get rid of the short name handling for subinterfaces (and remove the corresponding testcases). It doesn't make sense to use short names only in this very particular case, but not accept them anywhere else.

anders-nexthop:
> Please try reduce force-push in your PR commits. We will frequently compare code between commits to understand what is new (especially I have approved it before, and you add new code). Force-pushed code is difficult to review.

Sure, I will retain the commit structure going forward in this and future PRs.

qiluo-msft (approved):

PR #3911, opened 2025-06-10
[sfputil] Use host lane mask as part of rx-output enable/disable

mihirpat1:
#### What I did
We need to use host lane mask instead of media lane mask as part of sfputil debug rx-output EthernetXX enable/disable CLI.
In addition to the above, following needs to be enhanced
1. Need to fail CLI if the tx_disable_channel or rx_disable_channel fails
2. Verify the Rx and Tx output status after enabling/disabling Rx/Tx
3. Modify get_subport function to return subport as 1 if the subport field is not found in PORT table of CONFIG_DB

#### How I did it
Addressed the above issues in `What I did` section

#### How to verify it
1. Verified rx-output disable and enable 
2. Verified tx-output disable and enable 
3. Verified rx-output disable and enable CLI on a device which did not have subport as part of the `PORT` table in `CONFIG_DB`

### DIFF
--- a/sfputil/debug.py
+++ b/sfputil/debug.py
@@ -1,4 +1,5 @@
 import sys
+import time
 import click
 import utilities_common.cli as clicommon
 from utilities_common import platform_sfputil_helper
@@ -19,6 +20,8 @@ ERROR_PORT_CONFIG_LOAD = 4
 ERROR_NOT_IMPLEMENTED = 5
 ERROR_INVALID_PORT = 6
 
+CMIS_MAX_CHANNELS = 8
+TX_RX_OUTPUT_UPDATE_WAIT_TIME = 2  # seconds
 
 @click.group(cls=clicommon.AliasedGroup)
 def debug():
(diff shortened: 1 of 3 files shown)


utilities_common/platform_sfputil_helper.py
```diff
@@ -277,6 +282,8 @@ def get_subport(port_name):
     if subport is None:
         click.echo(f"{port_name}: subport is not present in CONFIG_DB")
         sys.exit(EXIT_FAIL)
+    else:
+        subport = 1
```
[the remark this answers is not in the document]
mihirpat1: This is intentionally done to initialize subport as 1 if the PORT table in CONFIG_DB does not have the `subport` field defined.

sfputil/debug.py
```diff
@@ -82,18 +85,65 @@ def set_output(port_name, enable, direction):
...
         elif direction == "rx":
-            sfp.rx_disable_channel(lane_mask, enable == "disable")
+            lane_count = get_host_lane_count(port_name)
+            disable_func = sfp.rx_disable_channel
+            get_status_func = api.get_rx_output_status
+            status_key = "RxOutputStatus"
+
+        lane_mask = get_subport_lane_mask(int(subport), int(lane_count))
+        if not disable_func(lane_mask, enable == "disable"):
+            click.echo(f"{port_name}: {direction.upper()} disable failed for subport {subport}")
```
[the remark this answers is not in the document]
mihirpat1: This is already part of a try/except block

sfputil/debug.py
```diff
@@ -82,18 +85,65 @@ def set_output(port_name, enable, direction):
...
+                    )
+                    sys.exit(EXIT_FAIL)
+                if enable == "disable":
+                    if lane_status:
+                        click.echo(
+                            f"{port_name}: {direction.upper()} output on lane {lane} is still "
+                            f"enabled on subport {subport}. Restoring state."
+                        )
+                        # Restore the state if disable failed
+                        disable_func(lane_mask, False)
```
mihirpat1: @prgeor Should we have this restoration logic to handle the failure scenario?

prgeor: @mihirpat1 I think user can restore by resetting the optics ...anyways this is for debugging.

mihirpat1: @prgeor Removed the restoration logic now.

prgeor (approved):

PR #3917, opened 2025-06-10
[sonic-package-manager] Save tag that was used to install the application

DavidZagury:
#### What I did
- Added functionality to save and track the Docker image tag used during package installation
- Modified the service creator to include the tag information in container configuration
- This change will improve visibility in `docker ps` output by showing the actual tag used for each container

#### How I did it
- Added a new `tag` field to the `Package` class to store the image tag
- Modified the `PackageSource` class to save the tag from the installed image
- Updated the `ServiceCreator` to include the tag information in the container configuration
- The tag is extracted from the first tag in the image's tags list during installation

#### How to verify it

1. Install a package using sonic-package-manager
2. Run `docker ps` to verify that the container is running with the correct tag
3. Check the container configuration to ensure the tag is properly saved

### DIFF
--- a/sonic_package_manager/database.py
+++ b/sonic_package_manager/database.py
@@ -31,6 +31,7 @@ class PackageEntry:
         built_in: Boolean flag whether the package is built in.
         image_id: Image ID for this package or None if package
                   is not installed.
+        tag: Tag for this package or None if package is not installed.
     """
 
     name: str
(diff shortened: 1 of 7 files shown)


stepanblyschak (approved):

liushilongbuaa (approved):
LGTM

DavidZagury:
@yejianquan yes, we need this in 202505.
This, along with https://github.com/sonic-net/sonic-buildimage/pull/22911 is needed to fix an issue with starting external App Extensions.

PR #3918, opened 2025-06-10
Support reboot cause: Kernel Panic - Out of memory

byu343:
#### What I did
Add new reboot-cause: Kernel Panic - Out of memory

#### How I did it
Existing code already supports "Kernel Panic" as reboot cause. By checking dmesg for kernel panic, overwrite the reboot cause as "Kernel Panic - Out of memory" when possible.

#### How to verify it
- Enable kdump: config kdump enable; config save; reboot
- Trigger OOM: python3 -c "blocks=[]; [blocks.append(bytearray(500*1024*1024)) for _ in iter(int, 1)]"
- Check reboot cause: "show reboot-cause" outputs "Kernel Panic - Out of memory [Time: Tue Jun 10 05:33:16 PM UTC 2025]"  

show reboot-cause
Kernel Panic [Time: Tue Jun 10 03:40:28 PM UTC 2025]

show reboot-cause
Kernel Panic - Out of memory [Time: Tue Jun 10 05:33:16 PM UTC 2025]

### DIFF
--- a/scripts/reboot
+++ b/scripts/reboot
@@ -19,6 +19,10 @@ VMCORE_FILE=/proc/vmcore
 if [ -e $VMCORE_FILE -a -s $VMCORE_FILE ]; then
         echo "We have a /proc/vmcore, then we just kdump'ed"
         echo "Kernel Panic [Time: ${REBOOT_TIME}]" > ${REBOOT_CAUSE_FILE}
+        makedumpfile --dump-dmesg /proc/vmcore /tmp/dmesg
+        if  grep -q "Kernel panic - not syncing: Out of memory" /tmp/dmesg; then
+            echo "Kernel Panic - Out of memory [Time: ${REBOOT_TIME}]" > ${REBOOT_CAUSE_FILE}
+        fi
         sync
         PLATFORM=$(grep -oP 'sonic_platform=\K\S+' /proc/cmdline)
         if [ ! -z "${PLATFORM}" -a -x ${DEVPATH}/${PLATFORM}/${PLAT_REBOOT} ]; then


scripts/reboot
```diff
@@ -19,6 +19,10 @@ VMCORE_FILE=/proc/vmcore
 if [ -e $VMCORE_FILE -a -s $VMCORE_FILE ]; then
         echo "We have a /proc/vmcore, then we just kdump'ed"
         echo "Kernel Panic [Time: ${REBOOT_TIME}]" > ${REBOOT_CAUSE_FILE}
+        makedumpfile --dump-dmesg /proc/vmcore /tmp/dmesg
+        if  grep -q "Kernel panic - not syncing: Out of memory" /tmp/dmesg; then
```
vvolam, line 23: Is it possible to extend to capture other types of kernel panic's as well. Can we add the cause string from "Kernel panic - not syncing: XXXX" to the reboot cause for clear kernel panic type?

vvolam: As this can generalize to other similar panics like "Kernel panic - not syncing: Fatal exception in interrupt", "Kernel panic - not syncing: VFS: Unable to mount root fs", "Kernel panic - not syncing: Watchdog detected hard LOCKUP" ,etc..

liamkearney-msft: this sounds like a good idea - even as a mid-step just having a generic "kernel panic" reboot cause would be nice.

liamkearney-msft:
nice - thanks @byu343 !

liamkearney-msft (approved):

StormLiangMS (approved):
LGTM

PR #3919, opened 2025-06-12
Skip checking offload flags for static routes/sids in route check and add check_sids

BYGX-wcr:
#### What I did

1. Skip checking offload flags for static routes and static SIDs in FRR in check_frr_pending_routes.
2. Add check_sids to check consistency between MY_SID tables of ASIC_DB and APPL_DB

#### How to verify it

Run on a device supporting SRv6

### DIFF
--- a/scripts/route_check.py
+++ b/scripts/route_check.py
@@ -1010,11 +1010,15 @@ def main():
         signal.alarm(0)
         if ret1 < 0 or ret2 < 0:
             ret = -1
+        else:
+            ret = 0
+
+        if res1 is not None or res2 is not None:
             res = dict()
-            if res1 is not None:
-                res.update(res1)
-            if res2 is not None:
-                res.update(res2)
+            res.update(res1 if res1 else {})
+            res.update(res2 if res2 else {})
+        else:
+            res = None
 
         if interval:
             time.sleep(interval)
--- a/scripts/route_check.py
+++ b/scripts/route_check.py
@@ -1003,11 +1003,18 @@ def main():
 
     while True:
         signal.alarm(TIMEOUT_SECONDS)
-        ret, res= check_routes(namespace)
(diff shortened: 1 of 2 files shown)


scripts/route_check.py
```diff
@@ -839,6 +892,72 @@ def check_routes(namespace):
...
+            try:
+                result = future.result()
+                if result:
+                    results[ns] = result
+            except Exception as e:
+                print_message(syslog.LOG_ERR, "Error processing namespace {}: {}".format(ns, e))
+                return -1, results
+
+    if results:
+        print_message(syslog.LOG_WARNING, "SIDs Check Failure results: {",  json.dumps(results, indent=4), "}")
```
r12f, line 949: is the double quote "{" and "}" intentional?

scripts/route_check.py
```diff
@@ -575,7 +628,7 @@ def check_frr_pending_routes(namespace):
 
         for _, entries in frr_routes.items():
             for entry in entries:
-                if entry['protocol'] in ('connected', 'kernel'):
+                if entry['protocol'] in ('connected', 'kernel', 'static'):
```
r12f, line 624: is uSID also counts as "static" here or it only includes the static route

r12f (dismissed):
all logic looking good, hence signed off.

2 minor comments to confirm.

scripts/route_check.py
BYGX-wcr: This is just to follow the original formatting of the failure results. Check https://github.com/sonic-net/sonic-utilities/blob/master/scripts/route_check.py#L833

scripts/route_check.py
BYGX-wcr: static sids also count as static here. For sids associated with BGP EVPN, I am not sure.

r12f (approved):

PR #3922, opened 2025-06-16
fix issue #22476 fix show cmd for bgp

lipxu:
#### What I did
Add an additional check in the cmd of the get_routing_stack function to detect the "frr" or "quagga"

#### How I did it
Modify the CMD, add "grep -E 'quagga|frr'" in CMD

#### How to verify it
quagga
```
admin@bjw-can-7050qx-1:~$ sudo docker ps | grep bgp | grep -E 'quagga|frr' | awk '{print$2}' | cut -d'-' -f3 | cut -d':' -f1 | head -n 1
quagga
admin@bjw-can-7050qx-1:~$ 
```

frr
```
admin@str3-7060-acs-2:/usr/local/lib/python3.11/dist-packages/show$ docker ps
CONTAINER ID   IMAGE                             COMMAND                  CREATED             STATUS             PORTS     NAMES
8a9e151d08bb   docker-sonic-restapi              "/usr/local/bin/supe…"   About an hour ago   Up About an hour             docker-sonic-restapi_but_bgp
21b846f8d085   docker-fpm-frr:latest             "/usr/bin/docker_ini…"   6 hours ago         Up 6 hours                   bgp
...
admin@str3-7060-acs-2:/usr/local/lib/python3.11/dist-packages/show$ 
admin@str3-7060-acs-2:/usr/local/lib/python3.11/dist-packages/show$ sudo docker ps | grep bgp | grep -E 'quagga|frr' | awk '{print$2}' | cut -d'-' -f3 | cut -d':' -f1 | head -n 1
frr
admin@str3-7060-acs-2:/usr/local/lib/python3.11/dist-packages/show$ 
admin@str3-7060-acs-2:/usr/local/lib/python3.11/dist-packages/show$ show ip bgp summary
IPv4 Unicast Summary:
BGP router identifier 10.1.0.32, local AS number 65100 vrf-id 0
BGP table version 6409
RIB entries 12807, using 2868768 bytes of memory
Peers 4, using 2967904 KiB of memory
Peer groups 4, using 256 bytes of memory


Neighbhor      V     AS    MsgRcvd    MsgSent    TblVer    InQ    OutQ  Up/Down      State/PfxRcd  NeighborName
-----------  ---  -----  ---------  ---------  --------  -----  ------  ---------  --------------  --------------
10.0.0.57      4  64600       3609       3548         0      0       0  05:42:55             6400  ARISTA01T1
10.0.0.59      4  64600       3609       3550         0      0       0  05:42:53             6400  ARISTA02T1
10.0.0.61      4  64600       3606       3548         0      0       0  05:42:55             6400  ARISTA03T1
10.0.0.63      4  64600       3607       3548         0      0       0  05:42:55             6400  ARISTA04T1

Total number of neighbors 4
admin@str3-7060-acs-2:/usr/local/lib/python3.11/dist-packages/show$ 
```

### DIFF
--- a/show/main.py
+++ b/show/main.py
@@ -89,7 +89,7 @@ COMMAND_TIMEOUT = 300
 # bash oneliner. To be revisited once routing-stack info is tracked somewhere.
 def get_routing_stack():
     result = None
-    command = "sudo docker ps | grep bgp | awk '{print$2}' | cut -d'-' -f3 | cut -d':' -f1 | head -n 1"
+    command = "sudo docker ps | grep bgp | grep -E 'quagga|frr' | awk '{print$2}' | cut -d'-' -f3 | cut -d':' -f1 | head -n 1"
 
     try:
         stdout = subprocess.check_output(command, shell=True, text=True, timeout=COMMAND_TIMEOUT)


BYGX-wcr:
Is there any known new distributions of SONiC using quagga?

show/main.py
```diff
@@ -89,7 +89,7 @@
 # bash oneliner. To be revisited once routing-stack info is tracked somewhere.
 def get_routing_stack():
     result = None
-    command = "sudo docker ps | grep bgp | awk '{print$2}' | cut -d'-' -f3 | cut -d':' -f1 | head -n 1"
+    command = "sudo docker ps | grep bgp | grep -E 'quagga|frr' | awk '{print$2}' | cut -d'-' -f3 | cut -d':' -f1 | head -n 1"
```
saiarcot895, line 92: Do we even need to keep support for quagga? Current SONiC versions can only be built with FRR.

lipxu: Thanks for your information. @saiarcot895 and @BYGX-wcr , I think for the new function, we no longer need to support quagga.
However, for bug fixes, I believe we should still follow the original logic. Even though this fix likely won't be backported to older version.
The current code still contains checks like 'if routing_stack == "quagga"', so it would be more consistent to includes both FRR and quagga in the earlier condition, thanks.

lipxu:
> Is there any known new distributions of SONiC using quagga?

Thanks for your comments @BYGX-wcr , please see the comments in above reply, thanks.
One more question, could you please help to double confirm this fix would not impact your PR https://github.com/sonic-net/sonic-utilities/pull/3417, thanks

show/main.py
lipxu: Hi, @saiarcot895 , Could you please help to review the PR again when free, if no more comment, I plan to merge it tomorrow, thanks a lot.

qiluo-msft, line 92: >quagga [](http://example.com/codeflow?start=52&length=6)

The extra grep reduce the possibility that one random container name trigger the bug, but not completely remove the possibility.

The random container names come from Kubernetes use case. The deploy container name may be long and random. So if a container name like "xxxbgpxxxfrrxxx" will still trigger the bug.

PR #3932, opened 2025-06-24
Fix reboot cause timezone on new image first reboot

benle7:
Dependent PR: https://github.com/sonic-net/sonic-buildimage/pull/23071

#### What I did
Fix an issue where newly installed SONIC images default the localtime to UTC
even when the current image uses a different timezone.
Besides fixing the specific issue, we also added a framework for image upgrade,
which can be extended in the future for different purposes.

#### How I did it
I added an upgrade_hook_script.py (standalone script) that copies the current /etc/localtime symlink
into the new image's filesystem during installation.
This ensures the timezone is preserved after rebooting into the new image.

#### How to verify it
- Use an image that includes the fix.
- Set a non-UTC timezone on the current image (e.g. sudo timedatectl set-timezone Asia/Jerusalem).
- Run sonic-installer install <image> for a new image that is not already installed and includes the fix.
- Reboot for using the new image.
- Reboot again on the new image.
- Verify timedatectl/date and show reboot-cause reflect the expected timezone.


sonic_installer/main.py
```diff
@@ -282,9 +284,18 @@ def update_sonic_environment(bootloader, binary_image_version):
         try:
             mount_squash_fs(new_image_squashfs_path, new_image_mount)
 
+            # Make sure upper dir and work dir exist to mount the filesystem as rw
+            run_command_or_raise(["mkdir", "-p", new_image_upper_dir])
+            run_command_or_raise(["mkdir", "-p", new_image_work_dir])
+            mount_overlay_fs(new_image_mount, new_image_upper_dir, new_image_work_dir, new_image_mount)
```
judyjoseph, line 290: Are we ok to create a new overlay_fs on same mount point "new_image_mount" ? @saiarcot895  to review as well

sonic_installer/main.py
```diff
@@ -303,6 +314,7 @@ def update_sonic_environment(bootloader, binary_image_version):
                 os.remove(env_file)
                 os.rmdir(env_dir)
         finally:
+            umount(new_image_mount, recursive=True, read_only=False, remove_dir=False, raise_exception=False)
```
judyjoseph, line 317: doing unmount twice ?

sonic_installer/main.py
benle7: Yes, this is valid and intentional: overlayfs supports mounting over an existing mountpoint.
We first mount squashfs, then overlayfs on top to expose a RW view.
The same pattern exists in main.py::migrate_sonic_packages:
mount/unmount twice: once for squashfs, once for overlayfs.

sonic_installer/main.py
benle7: Two unmounts are required:
The first removes the overlay, the second removes the underlying squashfs
(same pattern as in migrate_sonic_packages).

PR #3936, opened 2025-06-25
Integrated Security Cipher for TACPLUS passkey encryption

nmoray:
- What I did
Made use of Security Cipher module to encrypt the TACPLUS passkey.

- How I did it
Implemented the feature by following [HLD](https://github.com/sonic-net/SONiC/pull/1471)

- How to verify it
Unit test logs:
```
1. Encrypt passkey:
sonic# config tacacs passkey nikhil --encrypt
sonic# show run al | jq .TACPLUS
{
  "global": {
    "auth_type": "login",
    "key_encrypt": "True",
    "passkey": "U2FsdGVkX19YTUcG+0r3+d78A5E0zTXEukS3ZNrfyFk=",
    "src_intf": "Loopback0"
  }
}
sonic# cat /etc/pam.d/common-auth-sonic
#THIS IS AN AUTO-GENERATED FILE
#
# /etc/pam.d/common-auth- authentication settings common to all services
# This file is included from other service-specific PAM config files,
# and should contain a list of the authentication modules that define
# the central authentication scheme for use on the system
# (e.g., /etc/shadow, LDAP, Kerberos, etc.). The default is to use the
# traditional Unix authentication mechanisms.
#
# here are the per-package modules (the "Primary" block)

auth	[success=done new_authtok_reqd=done default=ignore auth_err=die]	pam_tacplus.so server=<> secret=nikhil login=login timeout=5   try_first_pass
auth	[success=done new_authtok_reqd=done default=ignore auth_err=die]	pam_tacplus.so server=<> secret=nikhil login=login timeout=5   try_first_pass
auth	[success=1 default=ignore]	pam_unix.so nullok try_first_pass

#
# here's the fallback if no module succeeds
auth    requisite                       pam_deny.so
# prime the stack with a positive return value if there isn't one already;
# this avoids us returning an error just because nothing sets a success code
# since the modules above will each just jump around
auth    required                        pam_permit.so
# and here are more per-package modules (the "Additional" block)
sonic# cat /etc/cipher_pass.json
{
  "TACPLUS": {
    "table_info": [
      "TACPLUS|global"
    ],
    "password": "TEST1"
  }

2. Encrypt passkey and update existing password
sonic# config tacacs passkey nikhil --encrypt --rotate
Password:
sonic# cat /etc/cipher_pass.json
{
  "TACPLUS": {
    "table_info": [
      "TACPLUS|global"
    ],
    "password": "TEST3"
  }
sonic# cat /etc/pam.d/common-auth-sonic
#THIS IS AN AUTO-GENERATED FILE
#
# /etc/pam.d/common-auth- authentication settings common to all services
# This file is included from other service-specific PAM config files,
# and should contain a list of the authentication modules that define
# the central authentication scheme for use on the system
# (e.g., /etc/shadow, LDAP, Kerberos, etc.). The default is to use the
# traditional Unix authentication mechanisms.
#
# here are the per-package modules (the "Primary" block)

auth	[success=done new_authtok_reqd=done default=ignore auth_err=die]	pam_tacplus.so server=<> secret=nikhil login=login timeout=5   try_first_pass
auth	[success=done new_authtok_reqd=done default=ignore auth_err=die]	pam_tacplus.so server=<> secret=nikhil login=login timeout=5   try_first_pass
auth	[success=1 default=ignore]	pam_unix.so nullok try_first_pass

#
# here's the fallback if no module succeeds
auth    requisite                       pam_deny.so
# prime the stack with a positive return value if there isn't one already;
# this avoids us returning an error just because nothing sets a success code
# since the modules above will each just jump around
auth    required                        pam_permit.so
# and here are more per-package modules (the "Additional" block)
sonic# show run al | jq .TACPLUS
{
  "global": {
    "auth_type": "login",
    "key_encrypt": "True",
    "passkey": "U2FsdGVkX18/8nFrAiCe01pOqgPoUUZFaTpmtawdi8I=",
    "src_intf": "Loopback0"
  }
}

3. Revert back to plaintext passkey
sonic# config tacacs passkey ravi
sonic# cat /etc/pam.d/common-auth-sonic
#THIS IS AN AUTO-GENERATED FILE
#
# /etc/pam.d/common-auth- authentication settings common to all services
# This file is included from other service-specific PAM config files,
# and should contain a list of the authentication modules that define
# the central authentication scheme for use on the system
# (e.g., /etc/shadow, LDAP, Kerberos, etc.). The default is to use the
# traditional Unix authentication mechanisms.
#
# here are the per-package modules (the "Primary" block)

auth	[success=done new_authtok_reqd=done default=ignore auth_err=die]	pam_tacplus.so server=<> secret=ravi login=login timeout=5   try_first_pass
auth	[success=done new_authtok_reqd=done default=ignore auth_err=die]	pam_tacplus.so server=<> secret=ravi login=login timeout=5   try_first_pass
auth	[success=1 default=ignore]	pam_unix.so nullok try_first_pass

#
# here's the fallback if no module succeeds
auth    requisite                       pam_deny.so
# prime the stack with a positive return value if there isn't one already;
# this avoids us returning an error just because nothing sets a success code
# since the modules above will each just jump around
auth    required                        pam_permit.so
# and here are more per-package modules (the "Additional" block)
sonic# show run al | jq .TACPLUS
{
  "global": {
    "auth_type": "login",
    "key_encrypt": "False",
    "passkey": "ravi",
    "src_intf": "Loopback0"
  }
}
sonic# cat /etc/cipher_pass.json
{
  "TACPLUS": {
    "table_info": [],
    "password": null
  }
```

- Related PR(s)
- [PR#17201](https://github.com/sonic-net/sonic-buildimage/pull/17201)
- [PR#22711](https://github.com/sonic-net/sonic-buildimage/pull/22711)
- [PR#81](https://github.com/sonic-net/sonic-host-services/pull/81)

Note: This PR is created by closing the old [PR](https://github.com/sonic-net/sonic-utilities/pull/3027) due to several merge conflicts.

### DIFF
--- a/config/aaa.py
+++ b/config/aaa.py
@@ -287,7 +287,7 @@ def passkey(db, ctx, secret, encrypt, rotate):
             # Set new passwd if not set already
             if secure_cipher.is_key_encrypt_enabled("TACPLUS", "global") is False:
                 # Register feature with Security Cipher module for the 1st time
-                secure_cipher.register("TACPLUS", rotate_tacplus_key)
+                secure_cipher.register("TACPLUS", "TACPLUS|global")
                 passwd = getpass.getpass()
                 # Set new password for encryption
                 secure_cipher.set_feature_password("TACPLUS", passwd)
@@ -296,7 +296,7 @@ def passkey(db, ctx, secret, encrypt, rotate):
                 if rotate:
                     passwd = getpass.getpass()
                     # Rotate password for TACPLUS feature and re-encrypt the secret
-                    secure_cipher.rotate_feature_passwd("TACPLUS", "TACPLUS|global", secret, passwd)
+                    secure_cipher.rotate_feature_passwd("TACPLUS", passwd)
                     return
             b64_encoded = secure_cipher.encrypt_passkey("TACPLUS", secret)
             if b64_encoded is not None:
@@ -310,17 +310,19 @@ def passkey(db, ctx, secret, encrypt, rotate):
                 return
         except (EOFError, KeyboardInterrupt):
             # Deregister feature with Security Cipher module
-            secure_cipher.deregister("TACPLUS", rotate_tacplus_key)
+            secure_cipher.deregister("TACPLUS", "TACPLUS|global")
             add_table_kv('TACPLUS', 'global', 'key_encrypt', False)
             click.echo('Input cancelled')
             return
         except Exception as e:
             # Deregister feature with Security Cipher module
-            secure_cipher.deregister("TACPLUS", rotate_tacplus_key)
+            secure_cipher.deregister("TACPLUS", "TACPLUS|global")
             add_table_kv('TACPLUS', 'global', 'key_encrypt', False)
             click.echo('Unexpected error: %s' % e)
             return
     else:
+        if secure_cipher.is_key_encrypt_enabled("TACPLUS", "global"):
+            secure_cipher.deregister("TACPLUS", "TACPLUS|global")
         # Update key_encrypt flag to false
         add_table_kv('TACPLUS', 'global', 'key_encrypt', False)
         add_table_kv('TACPLUS', 'global', 'passkey', secret)
@@ -370,7 +372,7 @@ def add(address, timeout, key, encrypted_key, rotate, auth_type, port, pri, use_
                 # Set new passwd if not set already
                 if secure_cipher.is_key_encrypt_enabled("TACPLUS_SERVER", address) is False:
                     # Register feature with Security Cipher module for the 1st time
-                    secure_cipher.register("TACPLUS", rotate_tacplus_key)
+                    secure_cipher.register("TACPLUS", f"TACPLUS_SERVER|{address}")
                     passwd = getpass.getpass()
                     # Set new password for encryption
                     secure_cipher.set_feature_password("TACPLUS", passwd)
@@ -379,12 +381,7 @@ def add(address, timeout, key, encrypted_key, rotate, auth_type, port, pri, use_
                     if rotate:
                         passwd = getpass.getpass()
                         # Rotate password for TACPLUS feature and re-encrypt the secret
-                        secure_cipher.rotate_feature_passwd(
-                            "TACPLUS",
-                            f"TACPLUS_SERVER|{address}",
-                            encrypted_key,
-                            passwd
-                        )
+                        secure_cipher.rotate_feature_passwd("TACPLUS", passwd)
                         return
                 b64_encoded = secure_cipher.encrypt_passkey("TACPLUS", encrypted_key)
                 if b64_encoded is not None:
@@ -393,22 +390,24 @@ def add(address, timeout, key, encrypted_key, rotate, auth_type, port, pri, use_
                     add_table_kv('TACPLUS_SERVER', address, 'passkey', b64_encoded)
                 else:
                     # Deregister feature with Security Cipher module
-                    secure_cipher.deregister("TACPLUS", rotate_tacplus_key)
+                    secure_cipher.deregister("TACPLUS", f"TACPLUS_SERVER|{address}")
                     click.echo('Passkey encryption failed')
                     return
             except (EOFError, KeyboardInterrupt):
                 # Deregister feature with Security Cipher module
-                secure_cipher.deregister("TACPLUS", rotate_tacplus_key)
+                secure_cipher.deregister("TACPLUS", f"TACPLUS_SERVER|{address}")
                 add_table_kv('TACPLUS_SERVER', address, 'key_encrypt', False)
                 click.echo('Input cancelled')
                 return
             except Exception as e:
                 # Deregister feature with Security Cipher module
-                secure_cipher.deregister("TACPLUS", rotate_tacplus_key)
+                secure_cipher.deregister("TACPLUS", f"TACPLUS_SERVER|{address}")
                 add_table_kv('TACPLUS_SERVER', address, 'key_encrypt', False)
                 click.echo('Unexpected error: %s' % e)
                 return
         else:
+            if secure_cipher.is_key_encrypt_enabled("TACPLUS_SERVER", address):
+                secure_cipher.deregister("TACPLUS", f"TACPLUS_SERVER|{address}")
             if key is not None:
                 # Update key_encrypt flag to false
                 add_table_kv('TACPLUS_SERVER', address, 'key_encrypt', False)
--- a/config/aaa.py
+++ b/config/aaa.py
@@ -6,11 +6,31 @@ from .validated_config_db_connector import ValidatedConfigDBConnector
 from jsonpatch import JsonPatchConflict
 from jsonpointer import JsonPointerException
 import utilities_common.cli as clicommon
+from sonic_py_common.security_cipher import master_key_mgr
+import getpass
 
 ADHOC_VALIDATION = True
 RADIUS_MAXSERVERS = 8
 RADIUS_PASSKEY_MAX_LEN = 65
 VALID_CHARS_MSG = "Valid chars are ASCII printable except SPACE, '#', and ','"
+TACACS_PASSKEY_MAX_LEN = 65
+
+def rotate_tacplus_key(table_info):
+    #Extract table and nested_key names
+    table = table_info.split('|')[0]
+    nested_key = table_info.split('|')[1]
+
+    # Re-encrypt with updated password
+    value = secure_cipher.encrypt_passkey("TACPLUS", secret)
+    add_table_kv(table, nested_key, 'passkey', value)
+
+# Security cipher Callback dir
+# Note: Required for Security Cipher - password rotation feature
+security_cipher_clbk_lookup = {
+        #TACPLUS
+        "rotate_tacplus_key": rotate_tacplus_key
+}
+secure_cipher = master_key_mgr(security_cipher_clbk_lookup)
 
 def is_secret(secret):
     return bool(re.match('^' + '[^ #,]*' + '$', secret))
@@ -122,7 +142,7 @@ authentication.add_command(trace)
 @clicommon.pass_db
 def login(db, auth_protocol):
     """Switch login authentication [ {ldap, radius, tacacs+, local} | default ]"""
-    if len(auth_protocol) is 0:
+    if len(auth_protocol) == 0:
         click.echo('Argument "auth_protocol" is required')
         return
     elif len(auth_protocol) > 2:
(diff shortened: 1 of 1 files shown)


nmoray:
@anders-nexthop please help in adding the reviewers.

config/aaa.py
```diff
@@ -6,11 +6,17 @@
...
+from sonic_py_common.security_cipher import master_key_mgr
+import getpass
 
 ADHOC_VALIDATION = True
 RADIUS_MAXSERVERS = 8
 RADIUS_PASSKEY_MAX_LEN = 65
 VALID_CHARS_MSG = "Valid chars are ASCII printable except SPACE, '#', and ','"
+TACACS_PASSKEY_MAX_LEN = 65
+
+secure_cipher = master_key_mgr(security_cipher_clbk_lookup)
```
anders-nexthop, line 18: where is `security_cipher_cblk_lookup` defined?

config/aaa.py
```diff
@@ -122,7 +128,7 @@ def trace(db, option):
 @clicommon.pass_db
 def login(db, auth_protocol):
     """Switch login authentication [ {ldap, radius, tacacs+, local} | default ]"""
-    if len(auth_protocol) is 0:
+    if len(auth_protocol) == 0:
```
anders-nexthop, line 131: seems unrelated?

config/aaa.py
```diff
@@ -243,16 +249,66 @@ def authtype(db, ctx, type):
...
+                    secure_cipher.rotate_feature_passwd("TACPLUS", passwd)
+                    return
+            b64_encoded = secure_cipher.encrypt_passkey("TACPLUS", secret)
+            if b64_encoded is not None:
+                # Update key_encrypt flag
+                add_table_kv('TACPLUS', 'global', 'key_encrypt', True)
+                add_table_kv('TACPLUS', 'global', 'passkey', b64_encoded)
+            else:
+                # Deregister feature with Security Cipher module
+                secure_cipher.deregister("TACPLUS", rotate_tacplus_key)
```
anders-nexthop, line 291: why would we deregister if the encryption failed? also, where is `rotate_tacplus_key` defined?

config/aaa.py
```diff
@@ -243,16 +249,66 @@ def authtype(db, ctx, type):
...
+            # Set new passwd if not set already
+            if secure_cipher.is_key_encrypt_enabled("TACPLUS", "global") is False:
+                # Register feature with Security Cipher module for the 1st time
+                secure_cipher.register("TACPLUS", "TACPLUS|global")
+                passwd = getpass.getpass()
+                # Set new password for encryption
+                secure_cipher.set_feature_password("TACPLUS", passwd)
+            else:
+                # Check if password rotation is enabled
+                if rotate:
```
anders-nexthop, line 279: nit: could be else-if with else statement above

config/aaa.py
```diff
@@ -243,16 +249,66 @@ def authtype(db, ctx, type):
...
+                    passwd = getpass.getpass()
+                    # Rotate password for TACPLUS feature and re-encrypt the secret
+                    secure_cipher.rotate_feature_passwd("TACPLUS", passwd)
```
anders-nexthop, line 282: don't we need to encrypt the secret here as well? maybe this shouldn't return?

config/aaa.py
```diff
@@ -243,16 +249,66 @@ def authtype(db, ctx, type):
...
+            else:
+                # Deregister feature with Security Cipher module
+                secure_cipher.deregister("TACPLUS", rotate_tacplus_key)
+                click.echo('Passkey encryption failed')
+                return
+        except (EOFError, KeyboardInterrupt):
+            # Deregister feature with Security Cipher module
+            secure_cipher.deregister("TACPLUS", "TACPLUS|global")
+            add_table_kv('TACPLUS', 'global', 'key_encrypt', False)
+            click.echo('Input cancelled')
```
anders-nexthop, line 298: where would we get the EOFError from? inputting nothing for the password entry maybe? also do we actually want to deregister the table in all cases? If we do that then we will get the following behavior:

-> run with --encrypt for the first time, user inputs password to register table
-> run with --encrypt for second time, user cancels command while it's running (table is deregistered)
-> run with --encrypt for the third time, user has to input password to register table AGAIN

do we really need change the feature password in this case?

config/aaa.py
```diff
@@ -243,16 +249,66 @@ def authtype(db, ctx, type):
...
+@click.option('-e', '--encrypt', help='Enable secret encryption', is_flag=True)
+@click.option('-r', '--rotate', help='Rotate encryption secret', is_flag=True)
 @click.pass_context
 @clicommon.pass_db
-def passkey(db, ctx, secret):
+def passkey(db, ctx, secret, encrypt, rotate):
     """Specify TACACS+ server global passkey <STRING>"""
     if ctx.obj == 'default':
         del_table_key(db, 'TACPLUS', 'global', 'passkey')
     elif secret:
```
anders-nexthop, line 260: in secret is not defined? then what? is the idea that an empty secret disables the passkey entirely? what happens if we pass an empty secret and --encrypt?

config/aaa.py
```diff
@@ -243,16 +249,66 @@ def authtype(db, ctx, type):
...
+        except Exception as e:
+            # Deregister feature with Security Cipher module
+            secure_cipher.deregister("TACPLUS", "TACPLUS|global")
+            add_table_kv('TACPLUS', 'global', 'key_encrypt', False)
+            click.echo('Unexpected error: %s' % e)
+            return
     else:
-        click.echo('Argument "secret" is required')
+        if secure_cipher.is_key_encrypt_enabled("TACPLUS", "global"):
+            secure_cipher.deregister("TACPLUS", "TACPLUS|global")
```
anders-nexthop, line 308: so if we run without --encrypt, we will remove the associated key encryption password and set `key_encrypt` to false, since the user is no longer wanting encryption. this seems correct.

What about the --rotate option though? if a user passes --rotate without --encrypt, what then? we should at least throw an error in that case, rather than removing the encryption.

config/aaa.py
```diff
@@ -261,13 +317,15 @@ def passkey(db, ctx, secret):
...
+@click.option('-k', '--key', help='Shared secret, stored in plaintext')
+@click.option('-K', '--encrypted_key', help='Shared secret, stored in encrypted format')
+@click.option('-r', '--rotate', help='Rotate encryption secret', is_flag=True)
 @click.option('-a', '--auth_type', help='Authentication type, default pap', type=click.Choice(["chap", "pap", "mschap", "login"]))
 @click.option('-o', '--port', help='TCP port range is 1 to 65535, default 49', type=click.IntRange(1, 65535), default=49)
 @click.option('-p', '--pri', help="Priority, default 1", type=click.IntRange(1, 64), default=1)
 @click.option('-m', '--use-mgmt-vrf', help="Management vrf, default is no vrf", is_flag=True)
 @clicommon.pass_db
-def add(db, address, timeout, key, auth_type, port, pri, use_mgmt_vrf):
+def add(address, timeout, key, encrypted_key, rotate, auth_type, port, pri, use_mgmt_vrf, encrypt):
```
anders-nexthop, line 328: I don't see --encrypt as an arg?

config/aaa.py
```diff
@@ -288,8 +346,56 @@ def add(db, address, timeout, key, auth_type, port, pri, use_mgmt_vrf):
...
         if timeout is not None:
             data['timeout'] = str(timeout)
-        if key is not None:
-            data['passkey'] = key
+
+        if key and encrypted_key:
+            raise click.UsageError("You must provide either --key or --encrypted_key")
+
+        if encrypted_key is not None:
+            try:
```
anders-nexthop, line 354: this is almost the same as the block above, lets refactor them to re-use the logic

config/aaa.py
```diff
@@ -288,8 +346,56 @@ def add(db, address, timeout, key, auth_type, port, pri, use_mgmt_vrf):
             data['auth_type'] = auth_type
         if timeout is not None:
             data['timeout'] = str(timeout)
-        if key is not None:
-            data['passkey'] = key
+
+        if key and encrypted_key:
```
anders-nexthop, line 350: would it make more sense to keep just --key and have an --encrypt argument, instead of --encrypted_key? --encrypted_key sounds like you are passing in a key that is already encrypted, which is not true; you are passing in a plaintext key that you WANT to be encrypted.

anders-nexthop (changes requested):
Needs some work still, but seems much cleaner after your changes to the underlying key_encrypt system.

config/aaa.py
```diff
@@ -243,16 +249,66 @@ def authtype(db, ctx, type):
...
 @click.argument('secret', metavar='<secret_string>', required=False)
+@click.option('-e', '--encrypt', help='Enable secret encryption', is_flag=True)
+@click.option('-r', '--rotate', help='Rotate encryption secret', is_flag=True)
 @click.pass_context
 @clicommon.pass_db
-def passkey(db, ctx, secret):
+def passkey(db, ctx, secret, encrypt, rotate):
     """Specify TACACS+ server global passkey <STRING>"""
     if ctx.obj == 'default':
         del_table_key(db, 'TACPLUS', 'global', 'passkey')
```
yenlu-keith, line 259: should we also remove key_encrypt?

config/aaa.py
```diff
@@ -243,16 +249,66 @@ def authtype(db, ctx, type):
...
+                secure_cipher.deregister("TACPLUS", rotate_tacplus_key)
+                click.echo('Passkey encryption failed')
+                return
+        except (EOFError, KeyboardInterrupt):
+            # Deregister feature with Security Cipher module
+            secure_cipher.deregister("TACPLUS", "TACPLUS|global")
+            add_table_kv('TACPLUS', 'global', 'key_encrypt', False)
+            click.echo('Input cancelled')
+            return
+        except Exception as e:
```
qiluo-msft, line 300: `except Exception` is too broad. Narrow to the specific exceptions `encrypt_passkey()` can raise: `subprocess.CalledProcessError`, `ValueError`, `OSError`. Same for the identical block in `add()` at line 385.

config/aaa.py
```diff
@@ -243,16 +249,66 @@ def authtype(db, ctx, type):
...
+
+    if encrypt:
+        try:
+            # Set new passwd if not set already
+            if secure_cipher.is_key_encrypt_enabled("TACPLUS", "global") is False:
+                # Register feature with Security Cipher module for the 1st time
+                secure_cipher.register("TACPLUS", "TACPLUS|global")
+                passwd = getpass.getpass()
+                # Set new password for encryption
+                secure_cipher.set_feature_password("TACPLUS", passwd)
```
qiluo-msft, line 276: `set_feature_password` silently skips if a password is already set (see its implementation). So if `config tacacs passkey --encrypt` ran first, the password prompt here in `add -K` is silently ignored (and vice versa). Please print a message to the user in that case.

qiluo-msft (commented):
Please add unit tests — coverage checker will block this PR without them. Need to cover at least: first-time encrypt, rotate, revert to plaintext, and the `--encrypt` without `secret` error path.

config/aaa.py
qiluo-msft, line 18: Crash at import time: `security_cipher_clbk_lookup` is undefined. `master_key_mgr.__init__` also takes no arguments (singleton). Should be: `secure_cipher = master_key_mgr()`

config/aaa.py
qiluo-msft, line 291: Crash: `rotate_tacplus_key` is undefined. Should be the string `"TACPLUS|global"`.

config/aaa.py
qiluo-msft, line 328: Crash: `db` is missing from the signature even though `@clicommon.pass_db` is still present. Also `encrypt` is in the signature but has no `@click.option` declaration. All `add_table_kv` calls inside this function are also missing the `db` first argument.

config/aaa.py
```diff
@@ -243,16 +249,66 @@ def authtype(db, ctx, type):
...
     elif secret:
-        add_table_kv(db, 'TACPLUS', 'global', 'passkey', secret)
+        if len(secret) > TACACS_PASSKEY_MAX_LEN:
+            click.echo('Maximum of %d chars can be configured' % TACACS_PASSKEY_MAX_LEN)
+            return
+        elif not is_secret(secret):
+            click.echo(VALID_CHARS_MSG)
+            return
+
+    if encrypt:
```
qiluo-msft, line 268: Crash: `secret` is `required=False`, so a user can run `config tacacs passkey --encrypt` without providing a secret. In that case `elif secret:` is skipped and `encrypt_passkey("TACPLUS", None)` is called here. Validate early: if `encrypt` and not `secret`, print an error and return.

config/aaa.py
```diff
@@ -243,16 +249,66 @@ def authtype(db, ctx, type):
...
+                passwd = getpass.getpass()
+                # Set new password for encryption
+                secure_cipher.set_feature_password("TACPLUS", passwd)
+            else:
+                # Check if password rotation is enabled
+                if rotate:
+                    passwd = getpass.getpass()
+                    # Rotate password for TACPLUS feature and re-encrypt the secret
+                    secure_cipher.rotate_feature_passwd("TACPLUS", passwd)
+                    return
```
qiluo-msft, line 283: Logic error: after rotating the password, the function returns immediately without encrypting and storing `secret`. If the user runs `config tacacs passkey <new_secret> --encrypt --rotate`, the new secret is silently discarded — only the encryption password is rotated. Either remove the `return` and fall through to `encrypt_passkey`, or clarify in the help text that `--rotate` does not update the passkey.

config/aaa.py
```diff
@@ -243,16 +249,66 @@ def authtype(db, ctx, type):
...
+            return
+        except Exception as e:
+            # Deregister feature with Security Cipher module
+            secure_cipher.deregister("TACPLUS", "TACPLUS|global")
+            add_table_kv('TACPLUS', 'global', 'key_encrypt', False)
+            click.echo('Unexpected error: %s' % e)
+            return
     else:
-        click.echo('Argument "secret" is required')
+        if secure_cipher.is_key_encrypt_enabled("TACPLUS", "global"):
```
qiluo-msft, line 307: Logic error: if the user passes `--rotate` without `--encrypt`, execution falls into this `else` branch, which silently deregisters the cipher and disables encryption. A typo (`--rotate` instead of `--encrypt --rotate`) would unknowingly remove the encryption setup. Add a guard at the top of the function: `if rotate and not encrypt: click.echo('--rotate requires --encrypt'); return`

PR #3938, opened 2025-06-27
[sfputil debug] Fix issue: do not check output status when CMIS version is lower than 5.0

Junchao-Mellanox:
#### What I did

OutputStatusTx and OutputStatusRx are only supported since CMIS 5.0. `sfputil debug tx-output` and `sfputil debug rx-output` should check them only if CMIS rev is higher or equal to 5.0.

#### How I did it

1. Use `get_tx_disable` and `get_rx_disable` to check if the corresponding bits have been set by `sfputil debug` command
2. Check the CMIS rev, ignore checking of OutputStatusTx and OutputStatusRx if CMIS rev is below 5.0.

#### How to verify it

1. Manual test
2. unit test

### DIFF
--- a/sfputil/debug.py
+++ b/sfputil/debug.py
@@ -93,6 +93,20 @@ def set_output(port_name, enable, direction):
 
     subport = get_subport(port_name)
 
+    if hasattr(api, 'get_cmis_rev'):
+        cmis_rev = api.get_cmis_rev()
+        if cmis_rev is None:
+            click.echo(f"{port_name}: CMIS revision not available for subport {subport}")
+            sys.exit(EXIT_FAIL)
+
+        # OutputStatusRx and OutputStatusTx are supported from CMIS 5.0
+        if float(cmis_rev) < 5.0:
+            click.echo(
+                f"{port_name}: This functionality is not supported"
+                f" with CMIS version {cmis_rev}, requires CMIS 5.0 and above"
+            )
+            sys.exit(EXIT_FAIL)
+
     try:
         if direction == "tx":
             lane_count = get_media_lane_count(port_name)
--- a/tests/sfputil_test.py
+++ b/tests/sfputil_test.py
@@ -1822,23 +1822,29 @@ EEPROM hexdump for port Ethernet4
         assert result.exit_code == EXIT_FAIL
 
     @pytest.mark.parametrize(
-        "direction, lane_count, enable, disable_func_result, output_dict, expected_echo, expected_exit",
+        "direction, lane_count, enable, disable_func_result, cmis_version, output_dict, expected_echo, expected_exit",
         [
             # TX disable success
             (
-                "tx", 2, "disable", True, {"TxOutputStatus1": False, "TxOutputStatus2": False},
+                "tx", 2, "disable", True, "5.3", {"TxOutputStatus1": False, "TxOutputStatus2": False},
                 "TX output disabled", None
             ),
             # RX enable success
-            ("rx", 1, "enable", True, {"RxOutputStatus1": True}, "RX output enabled", None),
+            ("rx", 1, "enable", True, "5.0", {"RxOutputStatus1": True}, "RX output enabled", None),
             # TX disable fails to disable
-            ("tx", 1, "disable", True, {"TxOutputStatus1": True}, "TX output on lane 1 is still enabled", SystemExit),
+            (
+                "tx", 1, "disable", True, "5.0", {"TxOutputStatus1": True},
+                "TX output on lane 1 is still enabled", SystemExit
+            ),
             # RX enable fails to enable
-            ("rx", 1, "enable", True, {"RxOutputStatus1": False}, "RX output on lane 1 is still disabled", SystemExit),
-            # TX disable_func returns False
-            ("tx", 1, "disable", False, {}, "TX disable failed", SystemExit),
-            # RX output_dict is None
-            ("rx", 1, "disable", True, None, "RX output status not available", SystemExit),
+            (
+                "rx", 1, "enable", True, "5.0", {"RxOutputStatus1": False},
+                "RX output on lane 1 is still disabled", SystemExit
+            ),
+            # CMIS version is None
+            ("tx", 1, "disable", False, None, {}, "CMIS revision not available", SystemExit),
+            # CMIS version is below 5.0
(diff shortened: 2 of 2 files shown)


prgeor:
@Junchao-Mellanox draft?

keboliu (approved):

Junchao-Mellanox:
Hi @prgeor , @mihirpat1 , could you please help review it? thanks!

sfputil/debug.py
```diff
@@ -112,6 +114,39 @@ def set_output(port_name, enable, direction):
...
+        if hasattr(api, 'get_cmis_rev'):
+            cmis_rev = api.get_cmis_rev()
+            if cmis_rev is None:
+                click.echo(f"{port_name}: CMIS revision not available for subport {subport}")
+                sys.exit(EXIT_FAIL)
+
+            # OutputStatusRx and OutputStatusTx are supported from CMIS 5.0
+            if float(cmis_rev) < 5.0:
+                click.echo(
+                    f"{port_name}: {direction.upper()} output "
```
prgeor: @Junchao-Mellanox user is not clear why failed from the log message

prgeor (commented):
@Junchao-Mellanox am I reading the wrong spec?
txdisable

prgeor (changes requested):
seems we don't need this change.

sfputil/debug.py
Junchao-Mellanox: Hi @prgeor , this is not a failure but an expected code path when the CMIS version is below 5.0. The log message confirms that the operation completes following a version-specific check. Note that this check differs from CMIS 5.2, as it does not include rxOutputStatus or txOutputStatus parameters.

prgeor: @Junchao-Mellanox looks like CMIS < 5.0 does not have support for `OutputStatusRx<i>` and `OutputStatusTx<i>`
In that case, I would suggest to check for CMIS version at the begining of `set_output()`  and return failure.

sfputil/debug.py
```diff
@@ -112,6 +114,39 @@ def set_output(port_name, enable, direction):
 
         time.sleep(TX_RX_OUTPUT_UPDATE_WAIT_TIME)
 
+        disable_status_list = get_disable_status_func()
+        if not disable_status_list:
+            click.echo(f"{port_name}: {direction.upper()} disable status not available for subport {subport}")
+            sys.exit(EXIT_FAIL)
+
+        for lane in range(1, CMIS_MAX_CHANNELS + 1):
```
prgeor: @Junchao-Mellanox There is another loop duplicate of doing the same for loop(in same set_ouptut() below) can you make it into a function check?

sfputil/debug.py
Junchao-Mellanox: @prgeor, let me clarify here.
- When CMIS < 5.0:
    - `tx_disable` and `rx_disable` are supported
    - `OutputStatusRx<i>` and `OutputStatusTx<i>` **are not supported**
    - `get_tx_disable` and `get_rx_disable` are supported
    - we use `get_tx_disable` and `get_rx_disable` to verify the status
- When CMIS >= 5.0:
    - `tx_disable` and `rx_disable` are supported
    - `OutputStatusRx<i>` and `OutputStatusTx<i>` **are supported**
    - get_tx_disable` and `get_rx_disable` are supported
    - we use `get_tx_disable`, `get_rx_disable`, `OutputStatusRx<i>` and `OutputStatusTx<i>` to verify the status

Based on the above, we don't need to fail the command when CMIS < 5.0

sfputil/debug.py
Junchao-Mellanox: the 2 for loops are different, the first one is checking a list, the second one is checking a dict. Also, the conditions in the for loops are different. Although we still can make them a function, but it would not be so readable. The function will look like this with a few callback (`status_getter, status_validator, status_disable_checker`) arguments:

```
def verify_lane_status(status_container, status_getter, status_validator, status_disable_checker,
                       port_name, enable, direction, subport, lane_mask):
    for lane in range(1, CMIS_MAX_CHANNELS + 1):
        if lane_mask & (1 << (lane - 1)):
            status = status_getter(status_container, lane)
            if not status_validator(status):
                click.echo(f"{port_name}: {direction.upper()} output status not available for lane {lane}")
                sys.exit(EXIT_FAIL)
            if enable == "disable":
                if status_disable_checker(status):
                    click.echo(f"{port_name}: {direction.upper()} output on lane {lane} is still "
                               f"enabled on subport {subport}. Restoring state.")
                    sys.exit(EXIT_FAIL)
            else:
                if not status_disable_checker(status):
                    click.echo(f"{port_name}: {direction.upper()} output on lane {lane} is still "
                               f"disabled on subport {subport}. Restoring state.")
                    sys.exit(EXIT_FAIL)

# for the first loop, we call it like this:
verify_lane_status(
            disable_status_list,
            lambda status_list, lane: status_list[lane - 1],
            lambda status: status and status != "N/A",
            lambda status: not status,
            port_name,
            enable,
            direction,
            subport,
            lane_mask
        )

# for the second loop, we call it like this:
verify_lane_status(
            output_dict,
            lambda status_dict, lane: status_dict.get(f'{status_key}{lane}'),
            lambda status: status is not None,
            lambda status: status,
            port_name,
            enable,
            direction,
            subport,
            lane_mask
        )
```
Should I go with this change?

sfputil/debug.py
prgeor: @Junchao-Mellanox Why do we not check for `OutputStatusRx` and `OutputStatusTx` for CMIS version < 5.0? I know its not available in < 5.0. Checking only Tx disable bits does not mean the module has muted the output

Junchao-Mellanox: Hi @prgeor , I am not sure I understand. `OutputStatusRx` and `OutputStatusTx` are not supported when CMIS < 5.0, how can we check it?

Junchao-Mellanox:
For the 202505 cherry-pick conflict, need take https://github.com/sonic-net/sonic-utilities/pull/3911 first, then there will be no conflict. I have added label 202505 to PR  https://github.com/sonic-net/sonic-utilities/pull/3911. @prgeor @mihirpat1 for awareness.

yejianquan:
Hi @Junchao-Mellanox , I've synced up with @prgeor , this PR is not suggested to be included 202505 branch,
hence remove the request label for now, please sync up with Prince, describe business requirement if this PR is still needed of 202505 branch.

PR #3933, opened 2025-06-24
fix issue #22476 remove quagga in show bgp cmd

lipxu:
#### What I did
with the PR https://github.com/sonic-net/sonic-utilities/pull/3922 , there is still a little chance to hit the issue.
Remove quagga support in show bgp cmd

#### How I did it
remove the quagga related code

PR #3947, opened 2025-07-02
fix issue #22476 remove quagga in show bgp cmd #3933

lipxu:


clear/main.py
```diff
@@ -79,20 +79,22 @@ def get_command(self, ctx, cmd_name):
...
+    cmd4 = ["head", "-n", "1"]
 
     try:
         _, result = getstatusoutput_noshell_pipe(cmd0, cmd1, cmd2, cmd3, cmd4)
 
     except OSError as e:
-        raise OSError("Cannot detect routing-stack")
-
-    return (result)
+        # If detection fails, return "frr" as default
```
qiluo-msft: >default [](http://example.com/codeflow?start=46&length=7)

This default value will hide bugs in above code. Suggest not to provide default value.

lipxu: Thanks for your reminder, @qiluo-msft , Actually, I tried to remove the check and return the frr directly at first, but caused failures in UT cases due to "add_command", So I updated the query cmd to match the exact bgp docker name, and set the default value to frr to avoid any potential issues. 
Please let me know if any other concerns, and please let me know if you want to keep the raising exception to catch the issue, thanks a lot.

lipxu: revert the commit and raise the exception as previous to catch the issue in advance, thanks

qiluo-msft: I do not understand why "but caused failures in UT cases due to "add_command"".

You raise another new OSError object, and lose original `e`. If you really need this `except` block, could you just `raise` without object?

lipxu: "I do not understand why "but caused failures in UT cases due to "add_command"""
---It's for the first solution, I tried to remove the quagga related code, but failed with UT case, seems related to the dynamic python module import, quagga or frr.

lipxu: You raise another new OSError object, and lose original e. If you really need this except block, could you just raise without object?
--- In this PR, the only change should be the shell CMD, I did not modify the exception handling code, as I think it was working well previously. thanks.

PR #3948, opened 2025-07-02
[config show]BGP Suppress fib pending config and display for multi-asic

vganesan-nokia:
#### What I did
CLI commands part of the fix for issue (https://github.com/sonic-net/sonic-buildimage/issues/19022)

#### How I did it
- To send the suppress-fib-pending configuration to all asics in multi-asic system
- Show command to diplay current suppress-fib-pending configuration for all available asics
- Unit tests to test the multi-asic support of suppress-fib-pending configuration

#### How to verify it
- In a multi-asic device configure suppress-fib-pending using command "sudo config suppress-fib-pending enabled"
- Check the config db of all asics to verify whether the "suppress-fib-pending": "enabled" fv pair is avialable in the config db DEVICE_METADATA
- Run the same test for "disabled" 

    admin@sonic:~$ show suppress-fib-pending 
    Enabled

    admin@sonic:~$ show suppress-fib-pending 
    asic0: Enabled
    asic1: Enabled
    admin@sonic:~$

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -2567,8 +2567,8 @@ def synchronous_mode(sync_mode):
 #
 @config.command('suppress-fib-pending')
 @click.argument('state', metavar='<enabled|disabled>', required=True, type=click.Choice(['enabled', 'disabled']))
-@click.option('-n', '--namespace', help='Namespace name',
-             required=True if multi_asic.is_multi_asic() else False, type=click.Choice(multi_asic.get_namespace_list()))
+@click.option('-n', '--namespace', help='Namespace name', required=True if multi_asic.is_multi_asic() else False,
+              type=click.Choice(multi_asic.get_namespace_list()))
 @clicommon.pass_db
 def suppress_pending_fib(db, namespace, state):
     ''' Enable or disable pending FIB suppression. Once enabled,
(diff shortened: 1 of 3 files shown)


config/main.py
```diff
@@ -2567,12 +2567,18 @@ def synchronous_mode(sync_mode):
 #
 @config.command('suppress-fib-pending')
 @click.argument('state', metavar='<enabled|disabled>', required=True, type=click.Choice(['enabled', 'disabled']))
+@click.option('-n', '--namespace', help='Namespace name', required=True if multi_asic.is_multi_asic() else False,
```
deepak-singhal0408: @vganesan-nokia I think we should not give namespace option to user during config.. Instead, for multi-asic, we should just enable on all Asics? 
Cant think of a usecase, where user would want to enable this feature on one asic and not on other.. 

Otherwise, this feature would get very complicated to test on multi-asic systems.

vganesan-nokia: O.k. Thanks @deepak-singhal0408 for the feedback. We can have the "namespace" option but as an optional one instead of required. With this In multi-asic system, if the name space is not given, the configuration will be applicable for all the asics in the system. Or, do we want this "--namespace" option to be completely removed?

deepak-singhal0408: I would prefer the --namespace option to be completely removed from config (otherwise someone might use that and we will have inconsistent config across asics).. We can just make changes to configure on all asics when this command is executed..
In show command also, I would suggest to remove --namespace option.. just display for all asics on multi-asic systems..

PR #3958, opened 2025-07-08
Add BlockingMode for Reboot script

lityu:
#### What I did
Add blocking-mode for the reboot script.
#### How I did it
https://github.com/sonic-net/SONiC/pull/2016
#### How to verify it
Local test and introduce test case for reboot script
reboot 
reboot [-b]

### DIFF
--- a/scripts/reboot
+++ b/scripts/reboot
@@ -2,7 +2,7 @@
 
 declare -r EXIT_SUCCESS=0
 declare -r EXIT_ERROR=1
-declare -r EXIT_TIMEOUT=1
+declare -r EXIT_TIMEOUT=2
 declare -r WATCHDOG_UTIL="/usr/local/bin/watchdogutil"
 declare -r PRE_REBOOT_HOOK="pre_reboot_hook"
 
@@ -151,6 +151,7 @@ function show_help_and_exit()
     echo "        -d : DPU module name on a smart switch, option is invalid when on DPU"
     echo "        -p : Pre-shutdown steps on DPU, invalid on NPU"
     echo "        -b : Enable the command to run in blocking mode"
+    echo "        -v : Print out detail logs. For blocking mode will print dots to identify the script is running"
 
     exit ${EXIT_SUCCESS}
 }


scripts/reboot
```diff
@@ -145,6 +149,7 @@ function show_help_and_exit()
     echo "        -h, -? : getting this help"
     echo "        -d : DPU module name on a smart switch, option is invalid when on DPU"
     echo "        -p : Pre-shutdown steps on DPU, invalid on NPU"
+    echo "        -b : Enable the command to run in blocking mode"
```
lizhijianrd, line 153: Should we add help for `-v`?

scripts/reboot
```diff
@@ -341,4 +375,37 @@ if [ -x ${DEVPATH}/${PLATFORM}/${PLAT_REBOOT} ]; then
...
+                    let BLOCKING_TIMER++
+                    echo -n "."
+                fi
+            fi
+
+            CURRENT_TIME=$(date +%s)
+            ELAPSED=$((CURRENT_TIME - REBOOT_START_TIME))
+            if (( ELAPSED >= BLOCKING_MODE_TIMEOUT_IN_SECOND )); then
+                echo "The reboot used $(ELAPSED)seconds while the timeout is configed as $(BLOCKING_MODE_TIMEOUT_IN_SECOND)seconds"
+                break
```
lizhijianrd: We should define a ERR code for timeout scenario, e.g.,
```bash
declare -r EXIT_TIMEOUT=2
```
Then replace `break` with `exit EXIT_TIMEOUT`

scripts/reboot
```diff
@@ -341,4 +375,37 @@ if [ -x ${DEVPATH}/${PLATFORM}/${PLAT_REBOOT} ]; then
...
+                else
+                    let BLOCKING_TIMER++
+                    echo -n "."
+                fi
+            fi
+
+            CURRENT_TIME=$(date +%s)
+            ELAPSED=$((CURRENT_TIME - REBOOT_START_TIME))
+            if (( ELAPSED >= BLOCKING_MODE_TIMEOUT_IN_SECOND )); then
+                echo "The reboot used $(ELAPSED)seconds while the timeout is configed as $(BLOCKING_MODE_TIMEOUT_IN_SECOND)seconds"
```
lizhijianrd: Maybe this is better?
```bash
echo "[REBOOT FAILED] The reboot used $(ELAPSED)seconds while the timeout is configed as $(BLOCKING_MODE_TIMEOUT_IN_SECOND)seconds"
```

PR #3972, opened 2025-07-14
[db_migrator] Fix parse_xml fails when minigraph has SonicQosProfile

wangxin:
#### What I did
This change is to fix https://github.com/sonic-net/sonic-buildimage/issues/22120 "Bug: [db_migrator] parse_xml fails when minigraph has SonicQosProfile".

If there is `SonicQosProfile` in minigraph.xml, parse_xml fails with below error:
```
2025 Jul 14 06:49:23.590302 vlab-01 ERR db_migrator: Caught exception while trying to parse minigraph: join() argument must be str, bytes, or os.PathLike object, not 'NoneType'
```
It's because select_mmu_profiles needs platform.
https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-config-engine/minigraph.py#L1925

#### How I did it
This change added code in db_migrator.py to get platform info. While calling `parse_xml`, pass in the `platform` argument.

#### How to verify it
Add SonicQosProfile to minigraph:
```
    <a:DeviceProperty>
        <a:Name>SonicQosProfile</a:Name>
        <a:Reference i:nil="true"/>
        <a:Value>Balanced</a:Value>
    </a:DeviceProperty>
```

Run `db_migrator.py -o migrate`. Check syslog, migration passed without exception.

### DIFF
--- a/scripts/db_migrator.py
+++ b/scripts/db_migrator.py
@@ -64,6 +64,8 @@ class DBMigrator():
         self.TABLE_KEY       = 'DATABASE'
         self.TABLE_FIELD     = 'VERSION'
 
+        self.platform = device_info.get_platform_info().get('platform')
+
         # Generate config_src_data from minigraph and golden config
         self.generate_config_src(namespace)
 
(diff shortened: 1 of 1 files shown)


anamehra:
Hi @cyw233 , @abdosi , this PR is required for db-migratory error in 202405-msft branch.

PR #3982, opened 2025-07-17
remove quagga for bgp show cli

lipxu:
#### What I did
The PR https://github.com/sonic-net/sonic-utilities/pull/3960 check the docker container's name.
Check the docker's name whether is exactly same with "bgp"
However, on multiple asic devices, there are several bgp containers, and named bgp0, bgp1 ....

#### How I did it
In master, return frr directly since there would no quagga in new release version and Remove quagga related UT code


BYGX-wcr (dismissed):
Makes the code cleaner

show/main.py
```diff
@@ -88,17 +88,7 @@
...
-
-    try:
-        stdout = subprocess.check_output(command, shell=True, text=True, timeout=COMMAND_TIMEOUT)
-        result = stdout.rstrip('\n')
-    except Exception as err:
-        click.echo('Failed to get routing stack: {}'.format(err), err=True)
-
-    return result
-
+    return "frr"
```
BYGX-wcr, line 91: Can we move the get_routing-stack function to utilities_common?

show/main.py
```diff
@@ -88,17 +88,7 @@
 # location (configdb?), so that we prevent the continous execution of this
 # bash oneliner. To be revisited once routing-stack info is tracked somewhere.
 def get_routing_stack():
```
BYGX-wcr, line 90: If we directly return frr, the show ip bgp command will become callable on any SONiC device even if the BGP container is not running on that device. This is different than before. For example, on the DPUs of smartswitches, are show ip bgp commands valid to execute?

lipxu: Good point, thanks for reminder, let me check it first. thanks.

PR #3988, opened 2025-07-19
Exclude Smart Switch from modular chassis operations/checks

tirupatihemanth:
#### Why I did it
When running the `test_fwutil_update_current` test in  https://github.com/sonic-net/sonic-mgmt/blob/master/tests/platform_tests/fwutil/test_fwutil.py we encounter the following issue when we run the command
```
admin@sonic:$ sudo fwutil update chassis component ONIE fw -y

Error: Failed to parse "platform_components.json": invalid platform schema: "module" key hasn't been found. Aborting...

Aborted!
```
This is because platform_components.json for smartswitch doesn't have module key in it and this should be ignored for smartswitch.

#### What I did
Avoid any modular chassis operations on smart-switch.

#### How I did it
Added a check to verify the device is not a smart switch before treating it as a modular chassis.

### DIFF
--- a/fwutil/lib.py
+++ b/fwutil/lib.py
@@ -196,7 +196,7 @@ class PlatformDataProvider(object):
 
     def get_chassis(self):
         return self.__chassis
-    
+
     def is_smart_switch(self):
         return self.__chassis.is_smartswitch()
 
--- a/fwutil/lib.py
+++ b/fwutil/lib.py
@@ -196,6 +196,9 @@ class PlatformDataProvider(object):
 
     def get_chassis(self):
         return self.__chassis
+    
+    def is_smart_switch(self):
+        return self.__chassis.is_smartswitch()
 
     def is_modular_chassis(self):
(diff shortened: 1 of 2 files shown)


fwutil/lib.py
```diff
@@ -535,18 +538,28 @@ def __init__(self, root_path=None):
...
         self.__root_path = root_path
-
-        self.__pcp = PlatformComponentsParser(self.is_modular_chassis())
+        is_modular_chassis = self.should_treat_as_modular_chassis()
+        self.__pcp = PlatformComponentsParser(is_modular_chassis)
         self.__pcp.parse_platform_components(root_path)
 
         self.__validate_platform_schema(self.__pcp)
 
+    def should_treat_as_modular_chassis(self):
```
vvolam: @tirupatihemanth I think you could ignore this copilot review comment and move to older check as it is used only at one place. Your initial changes are better readable.

vvolam (approved):

PR #3989, opened 2025-07-21
Secureboot: Image signing verification enhancements

bhouse-nexthop:
#### What I did

The current signature verification of sonic images assumes the DB Keys are all Root CAs.  The secureboot standard says nothing about this, the DBKeys are explicitly trusted by signing them with the KEK, and that signing method does not follow the standard X.509 PKI architecture.  Therefore the DB Key is not guaranteed to be a CA Root (aka not self-signed).  It is possible the DB Key was created as an intermediate, but since it is explicitly trusted that is ok.

Fixes https://github.com/sonic-net/sonic-buildimage/issues/23406

#### How I did it

This adds this explicit trust of the DB Key as a secondary signing verification if the original verification fails.  It disables looking inside the pkcs7 container for any keys at all and assumes the key specified is the exact key for the signature.

#### How to verify it

Build secureboot image signed with a DB Key that is not self-signed, then run through the sonic-installer install with that image and see the verification succeeds.

### DIFF
--- a/scripts/verify_image_sign_common.sh
+++ b/scripts/verify_image_sign_common.sh
@@ -15,35 +15,27 @@ verify_image_sign_common() {
         no_check_time="-noattr"
     fi
     
-    # This command verifies the signing as a complete certificate chain root of trust and requires the DB Key to
-    # be a self-signed root, but the image signed with an intermediate embedded into the certificate.
     EFI_CERTS_DIR=/tmp/efi_certs
     RESULT="CMS Verification Failure"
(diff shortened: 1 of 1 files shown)


bhouse-nexthop:
@ycoheNvidia since you contributed the signing validation for sonic images, can you review this PR modification?

ycoheNvidia:
Adding @davidpil2002 @DavidZagury @Yarden-Z as well

tjchadaga:
@qiluo-msft - could you please help with the review?

PR #3990, opened 2025-07-21
`vnet_route_check.py` should not report VNET routes in APP DB but not in STATE DB and ASIC DB as mismatches

mramezani95:
#### What I did
Changed `vnet_route_check.py` so that when a VNET route is missing in STATE DB, then it is treated as an inactive route. This means that when the script is called without `-a` and `--all` options, if a route in APP DB is missing in both STATE DB and ASIC DB, it will not be reported as a mismatch.

#### How I did it
1. Changed the logic in `vnet_route_check.py`.
2. Added a new test and modified an existing test in `vnet_route_check_test.py` to verify the desired behavior.

#### How to verify it
Run the tests in `vnet_route_check_test.py`.

N/A

N/A

### DIFF
--- a/scripts/vnet_route_check.py
+++ b/scripts/vnet_route_check.py
@@ -50,7 +50,7 @@ RC_OK = 0
 RC_ERR = -1
 default_vrf_oid = ""
 
-report_level = syslog.LOG_ERR
+report_level = syslog.LOG_WARNING
 write_to_syslog = True
 
 
@@ -374,7 +374,7 @@ def filter_active_vnet_routes(vnet_routes: dict):
             exists, fvs = vnet_route_tunnel_table.get(key)
             if not exists:
                 print_message(syslog.LOG_WARNING, f"VNET_ROUTE_TUNNEL_TABLE|{key} does not exist in STATE DB.")
-                active_routes.append(prefix)  # Treating "prefix" as an active route
+                # Treating "prefix" as an inactive route
                 continue
             fvs_dict = dict(fvs)
             if fvs_dict.get("state") == "active":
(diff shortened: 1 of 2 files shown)


scripts/vnet_route_check.py
```diff
@@ -374,7 +374,7 @@ def filter_active_vnet_routes(vnet_routes: dict):
             exists, fvs = vnet_route_tunnel_table.get(key)
             if not exists:
                 print_message(syslog.LOG_WARNING, f"VNET_ROUTE_TUNNEL_TABLE|{key} does not exist in STATE DB.")
-                active_routes.append(prefix)  # Treating "prefix" as an active route
```
prsunny, line 377: I think the issue is that routes in STATE_DB exists but state as "empty". Can you get an example case?

scripts/vnet_route_check.py
```diff
@@ -50,7 +50,7 @@
 RC_ERR = -1
 default_vrf_oid = ""
 
-report_level = syslog.LOG_ERR
+report_level = syslog.LOG_WARNING
```
prsunny, line 53: i wouldn't recommend change the report_level as this may have backward compatibility issues

scripts/vnet_route_check.py
mramezani95: AFAIK, if a route is in STATE DB, then its `state` field is set to either `active` or `inactive`. The problem is that some routes may not be in STATE DB, so they will have an empty status in the output of `show vnet routes all`.

prsunny:
@qiluo-msft , please merge. @yejianquan for viz

PR #3998, opened 2025-07-24
[SPM] Rename the variable tag to docker-image-reference

DavidZagury:
#### What I did
Rename the variable tag to docker_image_reference since it can also hold image digest.

#### How I did it
Change the variable name to match the name agreed on  https://github.com/sonic-net/sonic-buildimage/pull/22911

#### How to verify it
Install App Extension

### DIFF
--- a/sonic_package_manager/database.py
+++ b/sonic_package_manager/database.py
@@ -31,7 +31,8 @@ class PackageEntry:
         built_in: Boolean flag whether the package is built in.
         image_id: Image ID for this package or None if package
                   is not installed.
-        tag: Tag for this package or None if package is not installed.
+        docker_image_reference: Docker image reference for this package or None if package
+                                is not installed.
     """
 
     name: str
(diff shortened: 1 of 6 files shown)


qiluo-msft (approved):

yejianquan:
Create https://github.com/sonic-net/sonic-utilities/pull/4024 to revert this change to fix the issue, @DavidZagury @qiluo-msft

DavidZagury:
@yejianquan thank you for the quick revert once issue has been found.
Can you look at this new PR - https://github.com/sonic-net/sonic-utilities/pull/4019

I believe the mistake was with using docker_image_references for the Image class, which is a class coming from docker API, and should still use tags.

I have also added a check that the class has the variable I am going to use - just to be sure :)

yejianquan:
> @yejianquan thank you for the quick revert once issue has been found. Can you look at this new PR - #4019
> 
> I believe the mistake was with using docker_image_references for the Image class, which is a class coming from docker API, and should still use tags.
> 
> I have also added a check that the class has the variable I am going to use - just to be sure :)

redirect to @qiluo-msft for reviewing the new PR

PR #3999, opened 2025-07-24
restart USB to Ethernet management interface if not RUNNING

eladnachman:
PROBLEM:
Sometimes after "config reload" command the management interface (eth0) connected over ETH-USB device is
  UP/RUNNING on a remote peer side, but
  stays UP/not-running on the local side.
Well-known negotiation problem of ETH-over-USB devices (ASIX).

SOLUTION:
If the management interface is a USB device and is not RUNNING (/sys/class/net/${if}/operstate is not "up") at the end of the "config reload" command
bring it down and up with delay to trigger renegotiation.

#### What I did
restart USB to Ethernet management interface if not at RUNNING state

#### How I did it
If the management interface is a USB device and is not at RUNNING state at the end of the config reload command, bring it down and up with delay

#### How to verify it
Manual test

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1019,6 +1019,157 @@ def wait_service_restart_finish(service, last_timestamp, timeout=30):
     log.log_warning("Service: {} does not restart in {} seconds, stop waiting".format(service, timeout))
 
 
+def _iface_vrf(iface: str) -> str | None:
+    try:
+        out = subprocess.check_output(["ip", "-o", "link", "show", "dev", iface], text=True)
+        m = re.search(r"master\s+(\S+)", out)
+        return m.group(1) if m else None
+    except subprocess.CalledProcessError:
+        return None
+
+
+def _capture_default_routes(iface: str):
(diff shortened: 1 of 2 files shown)


config/main.py
```diff
@@ -1018,6 +1018,45 @@ def wait_service_restart_finish(service, last_timestamp, timeout=30):
...
+    try:
+        with open(f"/sys/class/net/{iface}/operstate") as f:
+            operstate = f.read().strip()
+        if operstate == "up":
+            return  # Already RUNNING
+    except Exception:
+        pass  # Continue to attempt reset
+
+    click.echo("Reset USB-based mgmt interface for re-negotiation")
+    subprocess.run(["ip", "link", "set", iface, "down"], check=True)
```
saiarcot895, line 1164: Bringing the interface down would cause the default route on the mgmt interface to be cleared.

eladnachman: fixed  - now the default interface is saved before bringing down the interface and restored after bringing the interface up again

PR #4003, opened 2025-07-29
arm64 kdump support - add support in kdump config

eladnachman:
Add support for configuring the crash kernel parameters in the u-boot environment. This requires a set of python function to process, parse and update the u-boot environment as well as calling those instead of grub2 for arm64 platforms.

#### What I did
Add support for configuring the crash kernel parameters in the u-boot environment. 
#### How I did it
Adding a set of python function to process, parse and update the u-boot environment as well as calling those instead of grub2 for arm64 platforms.
#### How to verify it
Manual testing (configuring kdump under arm64 and crashing the kernel)


eladnachman:
Also required to be merge into 202505

PR #4004, opened 2025-07-29
[dhcp_relay]: sonic dhcp relay agent for IPv4

cshivashgit:
Changes for SONiC DHCPv4 Relay feature

Currently SONiC uses the 'isc-dhcp-relay' package to allow DHCP relay functionality on IPv4 networks.
With this PR we are adding sonic dhcp relay agent for IPv4 as described in this HLD(https://github.com/sonic-net/SONiC/pull/1938).

#### What I did
* Added dependency checks in SONiC config CLI to prevent deletion of interfaces, VRFs, and VLANs that are in use by DHCPv4 relay configurations.
* Added/updated unit tests.

#### How I did it
* Implemented a new function check_dhcpv4_relay_dependencies to check if an interface or VRF is referenced in the DHCPV4_RELAY table.
* Called this function before allowing deletion of portchannels, loopbacks, VRFs, and VLANs.
* Modified the VLAN deletion logic to block removal if the VLAN is present in the DHCPV4_RELAY table.
* Extended the test suite to verify that deletion is blocked when dependencies exist and allowed after cleanup.

#### How to verify it
* Run the updated unit tests: they now include cases where deletion of interfaces, VRFs, or VLANs in use by DHCPv4 relay is attempted and should fail.
* Manually test by configuring DHCPv4 relay on an interface/VRF/VLAN, then attempt to delete it using the CLI; the command should fail with an appropriate error message.
* After removing the DHCPv4 relay configuration, deletion should succeed.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1622,8 +1622,8 @@ def config_file_yang_validation(filename):
 def check_dhcpv4_relay_dependencies(db, object_name, object_type):
     """Checks if to be deleted interface/VRF is used in DHCPV4_RELAY table."""
     # Check if has_sonic_dhcpv4_relay flag is enabled
-    feature_table = db.get_table("FEATURE")
-    dhcp_relay_feature = feature_table.get("dhcp_relay", {})
+    feature_table = db.get_table("DEVICE_METADATA")
+    dhcp_relay_feature = feature_table.get("localhost", {})
     if dhcp_relay_feature.get("has_sonic_dhcpv4_relay") != "True":
         return
 
--- a/scripts/db_migrator.py
+++ b/scripts/db_migrator.py
@@ -1266,8 +1266,8 @@ class DBMigrator():
         return 'version_202305_01'
 
     def check_has_sonic_dhcpv4_relay_flag(self):
-        device_metadata_table = self.configDB.get_table("FEATURE")
-        dhcp_relay_feature = device_metadata_table.get("dhcp_relay", {})
+        device_metadata_table = self.configDB.get_table("DEVICE_METADATA")
+        dhcp_relay_feature = device_metadata_table.get("localhost", {})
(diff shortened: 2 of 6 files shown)


config/main.py
```diff
@@ -1619,6 +1619,22 @@ def config_file_yang_validation(filename):
     return True
 
 
+def check_dhcpv4_relay_dependencies(db, object_name, object_type):
+    """Checks if to be deleted interface/VRF is used in DHCPV4_RELAY table."""
+    for relay_vlan, data in db.get_table('DHCPV4_RELAY').items():
```
Blueve, line 1630: How about check feature table first and skip this check if the new dhcp relay is not enabled? 

The isc-dhcp does not has the hard dependencies as it always requires a container restart.
Add this dependency might break existing assumption when using not move to SONiC dhcp.

scripts/db_migrator.py
```diff
@@ -1231,12 +1265,23 @@ def version_4_0_3(self):
         self.set_version('version_202305_01')
         return 'version_202305_01'
 
+    def check_has_sonic_dhcpv4_relay_flag(self):
+        device_metadata_table = self.configDB.get_table("DEVICE_METADATA")
```
Blueve, line 1269: I think you need read from FEATURE table based on newest implementation?
https://github.com/sonic-net/sonic-buildimage/pull/22486/files#diff-13f96ff98435683f7aef9df0fe2753511594dff98ab4d6f66ab8e4234c89f76aR52

shchitra: Addressed.

config/main.py
shchitra: Addressed

Blueve (dismissed):

Blueve (approved):

PR #4007, opened 2025-07-31
Adapt 'show muxcable tunnel-route' for prefix route based mux neighbors

manamand2020:
#### What I did
Mux neighbors now use prefix-based routes based on changes from the following PR:
https://github.com/sonic-net/sonic-swss/pull/3722
Presence of route prefix for neighbors does not mean tunnel-route use anymore. Show command now uses nexthop type to distinguish between tunnel and normal neighbor route.

#### How to verify it
Verified output on hardware.


zjswhhh (commented):
lgtm mostly!

Can you also add UT to cover the case that when ROUTE exists, but next hop is non tunnel.

lolyu (commented):
LGTM, so no change to the command line output format right?

PR #4029, opened 2025-08-11
Fixed cli command for ECN config on voq switch to set the WRED_PROFILE for all Voqs

saksarav-nokia:
#### What I did
 Set ECN ON/OFF for all the Voq's in the asic in the Voq system to fix https://github.com/sonic-net/sonic-buildimage/issues/22856
#### How I did it
 The existing code was setting ECN ON/OFF for only local front panel ports Voq's. However in the Voq system, the ECN should be ON/OFF for all the Voqs in the namespace including the Voq's of the remote asics.
#### How to verify it
Verified the ECN functionality in the Voq chassis.
None
None

### DIFF
--- a/scripts/ecnconfig
+++ b/scripts/ecnconfig
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/python3
 
 """
 ecnconfig is the utility to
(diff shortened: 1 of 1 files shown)


saksarav-nokia:
@vmittal-msft , pls review this

vmittal-msft (approved):

PR #567, opened 2019-07-02
sfpshow: prevent 'show int trans eeprom --dom' from crashing

ds952811:
When some specific 25G DAC attached to the system, the 'dom_info_dict'
could be a None object, as a result, the 'show int transceiver eeprom --dom'
will crash.

And thus, in this commit, we'll add an additional checker for the None object.

Signed-off-by: Dante (Kuo-Jung) Su <dante.su@broadcom.com>
Change-Id: Ic2b5eb38ddf98b98cce7eddf161595013a19c720

**- What I did**
Add a None object checker to prevent CLI from crashing

**- How I did it**
Add a None object checker

**- How to verify it**
Attach a 25G DAC to the system

**- Previous command output (if the output of a command-line utility has changed)**

**- New command output (if the output of a command-line utility has changed)**

-->

PR #4031, opened 2025-08-13
Module graceful shutdown support 

rameshraghupathy:
Provide support for SmartSwitch DPU module graceful shutdown.

Why I did it
Please refer to the HLD and related PRs:
HLD: # 1991 https://github.com/sonic-net/SONiC/pull/1991
sonic-platform-common: #567 https://github.com/sonic-net/sonic-platform-common/pull/567
sonic-host-services: https://github.com/sonic-net/sonic-platform-common/issues/255

How to verify it
Issue the "config chassis modules shutdown DPUx" command
Verify the DPU module is gracefully shut by checking the logs in /var/log/syslog on both NPU and DPU

### DIFF
--- a/config/chassis_modules.py
+++ b/config/chassis_modules.py
@@ -65,19 +65,28 @@ def get_state_transition_in_progress(db, chassis_module_name):
     value = fvs.get('state_transition_in_progress', 'False') if fvs else 'False'
     return value
 
-
 def set_state_transition_in_progress(db, chassis_module_name, value):
     ensure_statedb_connected(db)
     state_db = db.statedb
-    entry = state_db.get_entry('CHASSIS_MODULE_TABLE', chassis_module_name) or {}
+
+    # Write to the correct table the daemon watches
+    table = 'CHASSIS_MODULE_INFO_TABLE'
+
+    entry = state_db.get_entry(table, chassis_module_name) or {}
     entry['state_transition_in_progress'] = value
+
     if value == 'True':
         entry['transition_start_time'] = datetime.utcnow().isoformat()
+        # Record the type so the daemon knows this is a shutdown
+        entry['transition_type'] = 'shutdown'
     else:
         entry.pop('transition_start_time', None)
-        state_db.delete_field('CHASSIS_MODULE_TABLE', chassis_module_name, 'transition_start_time')
-    state_db.set_entry('CHASSIS_MODULE_TABLE', chassis_module_name, entry)
+        entry.pop('transition_type', None)
+        # Explicit deletes using the correct table
+        state_db.delete_field(table, chassis_module_name, 'transition_start_time')
+        state_db.delete_field(table, chassis_module_name, 'transition_type')
 
+    state_db.set_entry(table, chassis_module_name, entry)
 
 def is_transition_timed_out(db, chassis_module_name):
     ensure_statedb_connected(db)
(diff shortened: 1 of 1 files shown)


gpunathilell:
@rameshraghupathy The table being updated from chassisd is CHASSIS_MODULE_TABLE (and not CHASSIS_MODULE_INFO_TABLE) is that implementation also being changed?
https://github.com/gpunathilell/sonic-platform-daemons/blob/1feb0024fec65bd59be6dfb2a30f105ac6428e1e/sonic-chassisd/scripts/chassisd#L1346
https://github.com/gpunathilell/sonic-platform-daemons/blob/3e3caa0421212157a1529b5025044c8a0853397c/sonic-chassisd/scripts/chassisd#L50

gpunathilell:
@rameshraghupathy are we going to set the transition timeout from sonic-utilities or from module_base.py?

config/chassis_modules.py
```diff
@@ -41,6 +53,70 @@ def modules():
...
+
+
+# shared helper
+def _block_if_conflicting_transition(chassis_module_name: str, conflict_type: str, target_oper_status: str) -> bool:
+    """
+    Return True if a conflicting transition is in progress and the module has
+    not yet reached the target oper status; otherwise False.
+
+    Uses centralized ModuleBase transition API via _transition_entry() for
+    consistency, and reads oper_status from CHASSIS_MODULE_TABLE.
```
vvolam: Could you add summary the function what it is achieving exactly?

config/chassis_modules.py
```diff
@@ -41,6 +53,70 @@ def modules():
...
+
+
+def _transition_entry(module_name: str) -> dict:
+    """Read the transition entry for a module via ModuleBase centralized API."""
+    mb = ModuleBase()
+    conn = _state_db_conn()
+    return mb.get_module_state_transition(conn, module_name) or {}
+
+
+def _transition_in_progress(module_name: str) -> bool:
```
vvolam, line 81: Please a comment here as well on what function is doing?

vvolam (commented):
Other than cosmetic, these changes LGTM

config/chassis_modules.py
rameshraghupathy: @vvolam done

config/chassis_modules.py
rameshraghupathy: @vvolam Done

config/chassis_modules.py
```diff
@@ -8,9 +8,30 @@
 from utilities_common.chassis import is_smartswitch, get_all_dpus
 from datetime import datetime, timedelta
 
+# New imports to use centralized APIs
+try:
+    # Prefer swsscommon SonicV2Connector
+    from swsscommon.swsscommon import SonicV2Connector
+except ImportError:
+    # Fallback (if ever needed)
+    from swsssdk import SonicV2Connector
```
gpunathilell: why is this fallback import needed?

rameshraghupathy: @gpunathilell SONiC has two Python bindings for RedisDB: the newer swsscommon and the older swsssdk. Different images/containers (or older platform builds) may ship only one of them.

gpunathilell: older platform builds will use older branches, and not master. Do we need to support swssdk usage? This would mask any import issues and using swssdk connectors (which has not been in development for over two years) may cause issues

rameshraghupathy: @gpunathilell I'm ok removing  the fallback option.  @vvolam Hope you are ok as well?

PR #4034, opened 2025-08-19
Adding cleanup for BGP_PEER_CONFIGURED_TABLE when config reload is done

sudarshankumar4893:
#### What I did
When config reload is done, BGP_PEER_CONFIGURED_TABLE is not cleared of pervious entries of BGP peers. Added logic to clean that up during config reload
#### How I did it
Connected to state db and cleared up all keys starting with BGP_PEER_CONFIGURED_TABLE|*
#### How to verify it
delete a peer from config_db.json and perform config reload. Peer should be deleted from state db table BGP_PEER_CONFIGURED_TABLE

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1432,6 +1432,13 @@ def delete_transceiver_tables():
         state_db.delete_all_by_pattern(state_db.STATE_DB, table + state_db_del_pattern)
 
 
+def delete_bgp_peer_table():
+    # Delete all entries in BGP_PEER_CONFIGURED_TABLE in state db.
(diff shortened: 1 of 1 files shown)


StormLiangMS:
hi @sudarshankumar4893  
Could you help clarify the use case for deleting the BGP_PEER_CONFIGURED_TABLE after a config reload?

My understanding is that entries in the State DB should be added when they are set in the CONFIG_DB and removed when they are deleted from the CONFIG_DB. Based on this model, I am trying to understand why additional cleanup logic is required directly in the State DB.

It would be helpful to get more context on the specific scenario this cleanup is intended to address.

PR #4037, opened 2025-08-20
Add sonic-error-report tool for structured error reporting

hdwhdw:
Add sonic-error-report tool for structured error reporting

- Add error_reporter.py library module to utilities_common/
- Add sonic-error-report CLI script to scripts/
- Add comprehensive unit tests to tests/
- Update setup.py to include new script in installation

This provides crash-resilient JSON error reporting for SONiC operations, preparing for future integration with reboot and upgrade scripts.

#### What I did

Added `sonic-error-report` tool to provide structured, crash-resilient error reporting for SONiC operations.

#### How I did it

- Created `utilities_common/error_reporter.py` with `SonicErrorReportManager` class for the core functionality
- Added `scripts/sonic-error-report` CLI tool that uses the library
- Implemented staged reporting that persists even if operations crash (writes JSON with timeout defaults, then updates on success/failure)
- Added path sanitization to prevent directory traversal attacks

#### How to verify it

Run the unit tests:
```bash
pytest tests/sonic_error_report_test.py
```

Example CLI usage:
```bash
# Initialize a report (creates staged JSON with timeout defaults)
sonic-error-report --scenario update_firmware -e uuid-123 init firmware-update

# Mark as failed with specific error
sonic-error-report --scenario update_firmware -e uuid-123 fail 126 --reason "Disk space error"

# Mark as successful
sonic-error-report --scenario update_firmware -e uuid-123 success --duration 60
```

N/A - New command

Creates JSON reports in format: `<scenario>.<eventGuid>.json`
Example: `update_firmware.uuid-123.json`

### DIFF
--- a/utilities_common/error_reporter.py
+++ b/utilities_common/error_reporter.py
@@ -70,7 +70,7 @@ class SonicErrorReportManager:
 
     def _load_report_template(self):
         """Load the default report template from JSON file."""
-        template_path = os.path.join(os.path.dirname(__file__), 
+        template_path = os.path.join(os.path.dirname(__file__),
                                      'error_report_template.json')
         try:
             with open(template_path, 'r') as f:
@@ -173,10 +173,10 @@ class SonicErrorReportManager:
             guid = str(uuid.uuid4())
         else:
             guid = str(guid)  # Convert to string in case UUID object is passed
-            
+
         # Load template from JSON file
         report = self._load_report_template()
-        
+
         # Fill in dynamic values
         report["sonic_upgrade_summary"]["script_name"] = "{}".format(operation_type)
         report["sonic_upgrade_summary"]["guid"] = guid


scripts/sonic-error-report
vaibhavhd, line 1: Why is this file not committed as `scripts/sonic-error-report.py`?

utilities_common/error_reporter.py
```diff
@@ -0,0 +1,282 @@
...
+
+        Returns:
+            str: The GUID used for this report (provided or generated)
+        """
+        # Generate GUID if not provided, otherwise use the provided one
+        if guid is None:
+            guid = str(uuid.uuid4())
+        else:
+            guid = str(guid)  # Convert to string in case UUID object is passed
+        report = {
```
vaibhavhd: Should this dictionary move to another file, to define the VARS, constants?

Alternatively, you could add the JSON schema here, and update the schema with the default values what you have here.
That way we achieve two things at once? -- 1) define a schema for standardization 2) use the same schema to have default values and avoid any other definition.

scripts/sonic-error-report
hdwhdw: I think the general rule of thumb here is you drop the `.py` if you expect people to run your script directly, and not drop it if it is either a helper script to be imported, or something that is supposed to be run using `python script.py`. Here we are the former.

hdwhdw: https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html

vaibhavhd: Thanks. So CLIs are to be name w/o `.py`. Please keep in mind the future the future integration w/ other scripts and APIs - bash, Python, GNxI. I don't see a concern, but mentioning this so that you double check.

utilities_common/error_reporter.py
hdwhdw: I have mixed feeling for this. For this step (getting the default script) we really don't want any failure/error so hardcoding inline is actually the more sensible choice. But you do have a point for standardization. Hence I would create a template/schema.json, use a script (commited) to generate the default json, and also commit the default json.

This still introduce a point of failure were our script can't load the default.json, while avoiding any failure in generating such script for the schema.

utilities_common/error_report_template.json
vaibhavhd, line 1: Can the template be derived from the schema?

hdwhdw: Yes. I committed the code for generating the template from schema. But this is mostly just for software development (generate the schema offline). But I am not sure about generating this on the fly. We don't want something in the exception handler like this have any chance of failing.

vaibhavhd: Ok. I suggest to remove `generate_template_from_schema.py` then. If you do not suggest to use the code then it is best to not commit it. Either use it or don't commit it -  I am saying that as unused code is often more expensive (maintenance, etc.) and the its adhoc benefits are overshadowed.

PR #4042, opened 2025-08-22
Add python wheels for GCU

ganglyu:
#### What I did
Microsoft ADO: 33497730
Create new python wheels for GCU

#### How I did it
Added a new setup.py for GCU to support packaging and installation.
Will integrate a Makefile into sonic-buildimage for streamlined build support.
Introduced a gcu.py script to enable direct invocation of GCU via gNMI.
Utilized nsenter for executing systemctl commands, allowing GCU to run seamlessly within the container environment.
SONiC CLI still uses config/main.py to invoke GCU, so this change is backward compatible.

#### How to verify it
Run build pipeline.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -183,5 +183,5 @@ stages:
       displayName: 'Build Python 3 wheel for GCU'
 
     - publish: '$(System.DefaultWorkingDirectory)/gcu/dist/'
-      artifact: wheels
+      artifact: gcu_wheels
       displayName: "Publish Python wheels for GCU"


azure-pipelines.yml
```diff
@@ -172,6 +172,12 @@ stages:
         python3 -m build -n
       displayName: 'Build Python 3 wheel'
 
+    - script: |
+        set -e
+        pushd gcu
+        python3 -m build -n
+      displayName: 'Build Python 3 wheel for GCU'
```
wen587, line 183: Should we add a publish step for gcu wheel?

wen587 (approved):

PR #4045, opened 2025-08-29
Fix sonic-kdump-config for running commands with pipe

byu343:
#### What I did
The exisingg code does not work because the commands with pipe are only able to run in shell

### DIFF
--- a/scripts/sonic-kdump-config
+++ b/scripts/sonic-kdump-config
@@ -371,7 +371,7 @@ def get_kdump_ssh_path():
 #
 #  @return The integer value X from USE_KDUMP=X in /etc/default/kdump-tools
 def read_use_kdump():
-    (rc, lines, err_str) = run_command("grep 'USE_KDUMP=.*' %s | cut -d = -f 2" % kdump_cfg, use_shell=False);
+    (rc, lines, err_str) = run_command("grep 'USE_KDUMP=.*' %s | cut -d = -f 2" % kdump_cfg, use_shell=True);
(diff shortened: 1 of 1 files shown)


scripts/sonic-kdump-config
```diff
@@ -371,7 +371,7 @@ def get_kdump_ssh_path():
 #
 #  @return The integer value X from USE_KDUMP=X in /etc/default/kdump-tools
 def read_use_kdump():
-    (rc, lines, err_str) = run_command("grep 'USE_KDUMP=.*' %s | cut -d = -f 2" % kdump_cfg, use_shell=False);
```
qiluo-msft, line 374: >use_shell [](http://example.com/codeflow?start=93&length=9)

There is security related alert on this line by semgrep. ref: [https://semgrep.dev/docs/cheat-sheets/python-command-injection#1b-shelltrue](https://semgrep.dev/docs/cheat-sheets/python-command-injection#1b-shelltrue).

However in this case you may argue that the command is static constructed and no user input is allowed. It may be not future proof if the command become dynamic constructed.

To make static code analysis easier, let's prevent using ` shell=True`. In this case, it is easy to completely python implementing both grep and cut. No need to run command. Could you explore?

qiluo-msft: Alternative, you may want to use some options:
1. getstatusoutput_noshell_pipe
2. SysInfoProvider.run_command

saiarcot895 (approved):

PR #4049, opened 2025-09-03
Fix issue that dynamic/static threshold 0 can not be configured using mmuconfig

stephenxs:
#### What I did

Fix issue that dynamic/static threshold 0 can not be configured using mmuconfig

### DIFF
--- a/tests/mmuconfig_input/mmuconfig_test_vectors.py
+++ b/tests/mmuconfig_input/mmuconfig_test_vectors.py
@@ -269,27 +269,30 @@ testData = {
                                    'cmp_args': [',ingress_lossless_profile_hbm,static_th,12121213'],
                                    'rc_msg': ''
                                    },
-             'mmu_cfg_static_th_zero': {'cmd': ['config'],
-                                   'args': ['-p', 'ingress_lossless_profile_hbm', '-s', '0'],
-                                   'rc': 0,
-                                   'db_table': 'BUFFER_PROFILE',
-                                   'cmp_args': [',ingress_lossless_profile_hbm,static_th,0'],
-                                   'rc_msg': ''
-                                   },
-             'mmu_cfg_alpha' :    {'cmd' : ['config'],
-                                   'args' : ['-p', 'alpha_profile', '-a', '2'],
-                                   'rc' : 0,
-                                   'db_table' : 'BUFFER_PROFILE',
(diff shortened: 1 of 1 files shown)


stephenxs:
Seems that pretest static analysis is checking the old code.

```
stephens@6459730f24d1:/sonic/src/sonic-utilities$ git add tests/mmuconfig_input/mmuconfig_test_vectors.py 
stephens@6459730f24d1:/sonic/src/sonic-utilities$ git commit -s
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

tests/mmuconfig_input/mmuconfig_test_vectors.py:288:29: E203 whitespace before ':'
tests/mmuconfig_input/mmuconfig_test_vectors.py:288:41: E203 whitespace before ':'
tests/mmuconfig_input/mmuconfig_test_vectors.py:289:42: E203 whitespace before ':'
tests/mmuconfig_input/mmuconfig_test_vectors.py:290:40: E203 whitespace before ':'
tests/mmuconfig_input/mmuconfig_test_vectors.py:291:46: E203 whitespace before ':'
```

scripts/mmuconfig
```diff
@@ -196,9 +196,9 @@ def main(show_config, profile, alpha, staticth, trim, namespace, verbose):
             mmu_cfg.list()
         # Buffershow cannot modify profiles
         elif config and profile:
-            if alpha:
+            if alpha is not None:
                 mmu_cfg.set(profile, DYNAMIC_THRESHOLD, alpha)
-            elif staticth:
+            elif staticth is not None:
```
qiluo-msft, line 201: >staticth [](http://example.com/codeflow?start=17&length=8)

If `alpha is not None`, why you skip `staticth` checking branch? I think they should be independent with each other.

qiluo-msft (commented):
As commented

scripts/mmuconfig
stephenxs: @qiluo-msft static th is for static threshold. Dynamic threshold is for dynamic threshold 
Imo a buffer profile cannot be both dynamic and static at the same time.

qiluo-msft: Same question to trim.

stephenxs: @qiluo-msft trim is the legacy logic. I didn't change them.

qiluo-msft (approved):

PR #4064, opened 2025-09-22
Fixing state_db not having delete_field attribute causing a crash when DPUs in bad state

rameshraghupathy:
#### What I did
Fixing state_db not having delete_field attribute causing a crash when DPUs in bad state. Removed the line of code in set_state_transition_in_progress, which was deleting the field "transition_start_time".  This information is redundant.  Leaving the  'transition_start_time' in the CHASSIS_MODULE_TABLE is no harm.  This part of the code is  refactored and that PR is almost ready to be merged.  This is just an interim fix mainly for 202505/06 branches.

#### How I did it
Removed the line of code in set_state_transition_in_progress, which was deleting the field "transition_start_time'.  This information is redundant. 

#### How to verify it
Make the DPU midplane unreachable for one of the DPUs.  Toggle the DPU ON/OFF state a couple of times using "config chassis  modules startup/shutdown DPUx"

Which release branch to backport (provide reason below if selected)
[x] 202505
[x] 202506

### DIFF
--- a/config/chassis_modules.py
+++ b/config/chassis_modules.py
@@ -75,7 +75,6 @@ def set_state_transition_in_progress(db, chassis_module_name, value):
         entry['transition_start_time'] = datetime.utcnow().isoformat()
     else:
         entry.pop('transition_start_time', None)
-        state_db.delete_field('CHASSIS_MODULE_TABLE', chassis_module_name, 'transition_start_time')
     state_db.set_entry('CHASSIS_MODULE_TABLE', chassis_module_name, entry)
 
 


config/chassis_modules.py
```diff
@@ -75,7 +75,6 @@ def set_state_transition_in_progress(db, chassis_module_name, value):
         entry['transition_start_time'] = datetime.utcnow().isoformat()
     else:
         entry.pop('transition_start_time', None)
```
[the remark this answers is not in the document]
vvolam: @rameshraghupathy could you address this comment and also update PR description accordingly to create clarity?

vvolam:
@prsunny I am waiting on Ramesh to address the comments.

config/chassis_modules.py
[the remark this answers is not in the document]
rameshraghupathy: @vvolam Done

config/chassis_modules.py
```diff
@@ -89,9 +96,18 @@ def is_transition_timed_out(db, chassis_module_name):
     if not start_time_str:
         return False
     try:
+        # Try ISO format first
         start_time = datetime.fromisoformat(start_time_str)
     except ValueError:
-        return False
+        try:
+            start_time = datetime.strptime(start_time_str, '%Y-%m-%d %H:%M:%S.%f')
+        except ValueError:
```
vvolam: Is this exception really required here?

rameshraghupathy: @vvolum Given this diversity of formats, removing the exception block would be risky because the code needs to handle these different timestamp formats robustly. Simplified it to work.  Moreover, this logic is refactored and this will be short lived.

config/chassis_modules.py
```diff
@@ -89,9 +96,18 @@ def is_transition_timed_out(db, chassis_module_name):
     if not start_time_str:
         return False
     try:
+        # Handle Zulu time format more robustly
+        if start_time_str.endswith('Z'):
```
vvolam: Let's use UTC everywhere.

rameshraghupathy: @vvolam We're using UTC timezone in ISO format everywhere - It's the best of both worlds for consistency and interoperability.  Why both?
UTC ensures all timestamps are in the same timezone (no confusion about local vs UTC time)
ISO format provides a standardized string representation that: is suitable for storing as a string, Is human-readable, Includes timezone information, can be parsed back to datetime objects reliably, Is widely supported across systems ( APIs, JSON payloads)

config/chassis_modules.py
```diff
@@ -92,7 +99,14 @@ def is_transition_timed_out(db, chassis_module_name):
         start_time = datetime.fromisoformat(start_time_str)
     except ValueError:
         return False
-    return datetime.utcnow() - start_time > TRANSITION_TIMEOUT
+
+    # Use UTC everywhere for consistent comparison
+    current_time = datetime.now(timezone.utc)
+    if start_time.tzinfo is None:
```
gpunathilell, line 105: @rameshraghupathy do we need to handle a case when tzinfo is none? as it is always updated by chassisd and module_base. why is this if condition needed?

rameshraghupathy: @gpunathilell It is a T1 switch and there is nothing wrong in doing the additional check to be resilient.

config/chassis_modules.py
```diff
@@ -27,6 +27,12 @@ def set_entry(self, table, key, entry):
         for field, value in entry.items():
             self.db.set("STATE_DB", redis_key, field, value)
 
+    def delete_field(self, table, key, field):
+        """Delete a specific field from table|key."""
+        redis_key = f"{table}|{key}"
+        client = self.db.get_redis_client("STATE_DB")
+        return client.hdel(redis_key, field)
```
gpunathilell, line 34: Can we use the same redis sonic dbconnector abstraction to delete https://github.com/rameshraghupathy/sonic-utilities/blob/fc016432a4d79dbb9eebd31f7f008759dc737122/config/chassis_modules.py#L85
Since we are using the db connector abstractions for the other two functions
https://github.com/rameshraghupathy/sonic-utilities/blob/fc016432a4d79dbb9eebd31f7f008759dc737122/config/chassis_modules.py#L28

rameshraghupathy: @gpunathilell Thank you for the suggestion about using the database connector abstraction consistently. While I agree this would improve consistency, this logic is currently undergoing broader refactoring in PRs #4031 and related PRs across sonic-platform-common, sonic-host-services, and sonic-platform-daemons. To avoid conflicts and duplicated effort, I believe the database access pattern consistency should be addressed as part of that broader refactoring work rather than making incremental changes here.

gpunathilell: Since refactoring will be done, please do not resolve the review comments, I will approve as this is blocking bug fix.  Thank you

PR #4067, opened 2025-09-26
[DO NOT MERGE] Trixie extra

saiarcot895:


scripts/coredump-compress
```diff
@@ -9,7 +9,7 @@ done
 
 CONTAINER_ID=""
 if [ $# > 0 ]; then
-    CONTAINER_ID=$(xargs -0 -L1 -a /proc/${1}/cgroup | grep -oP "pids:/docker/\K\w+")
+    CONTAINER_ID=$(xargs -0 -L1 -a /proc/${1}/cgroup | grep -oP "system.slice/docker-\K\w+")
```
qiluo-msft: Does it break backward compatible? #Closed

saiarcot895: Yes, but the Trixie upgrade in sonic-buildimage will be going in very soon.

saiarcot895: Updated to support both.

qiluo-msft (approved):

PR #4068, opened 2025-09-26
add cli to show liquid cooling leakage status

yuazhe:
#### What I did

Following this HLD for liquid cooling https://github.com/sonic-net/SONiC/pull/2032, I created this pr to add relevant CLI for checking liquid cooling status

### DIFF
--- a/scripts/leakageshow
+++ b/scripts/leakageshow
@@ -1,0 +1,62 @@
+#!/usr/bin/env python3
+
+"""
+    Script to show leakage status.
+"""
+
+import os
+import sys
+from tabulate import tabulate
+from natsort import natsorted
+
+# mock the redis for unit test purposes #
+try:
+    if os.environ["UTILITIES_UNIT_TESTING"] == "1":
+        modules_path = os.path.join(os.path.dirname(__file__), "..")
+        test_path = os.path.join(modules_path, "tests")
+        sys.path.insert(0, modules_path)
+        sys.path.insert(0, test_path)
+        import mock_tables.dbconnector
+except KeyError:
+    pass
+
+from swsscommon.swsscommon import SonicV2Connector
+
+header = ['Name', 'Leak']
+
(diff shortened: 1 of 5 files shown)


keboliu:
@judyjoseph would you please help to review this PR?

yuazhe:
> @yuazhe please add sample o/p fof the new CLI added in PR description

thanks, I have updated it accordingly.

scripts/leakageshow
```diff
@@ -0,0 +1,62 @@
...
+        test_path = os.path.join(modules_path, "tests")
+        sys.path.insert(0, modules_path)
+        sys.path.insert(0, test_path)
+        import mock_tables.dbconnector
+except KeyError:
+    pass
+
+from swsscommon.swsscommon import SonicV2Connector
+
+header = ['Name', 'Leak']
```
nate-nexthop: Can we change the header for the leak sensor status to `Leaking` from `Leak`?
```
show platform leakage status
Name       Leaking  
-----------------
leak_sensors1    NO   
leak_sensors2    NO   
...
leak_sensorsX    Yes   
```

I think this reads better. You would ask the question, "Is this sensor leaking?". Not, "Is this sensor leak?".

nate-nexthop (changes requested):
Overall LGTM.

scripts/leakageshow
liat-grozovik: Agree. @yuazhe please handle and sync team internally . then i can move on and merge it

judyjoseph (approved):
LGTM, please resolve comments as well

scripts/leakageshow
yuazhe: @nate-nexthop  thanks for the correction, I have updated the PR

nate-nexthop: Thanks! Please update the PR description with the new text

yuazhe: done @liat-grozovik we should be good to move forward here.

nate-nexthop: Thank you!

PR #4082, opened 2025-10-07
Checker to check if the lag-ids are in sync with chassisdb.

arlakshm:

### DIFF
--- a/scripts/chassis_lag_id_checker
+++ b/scripts/chassis_lag_id_checker
@@ -1,0 +1,166 @@
+#!/usr/bin/env python3
+
+"""
+chassis_lag_id_checker
+
+This script checks for synchronization of LAG (Link Aggregation Group) IDs
+between the chassis_db and asic_db on VOQ chassis Linecard.
+This script is intended to be run by Monit.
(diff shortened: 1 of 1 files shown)


scripts/chassis_lag_id_checker
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env python3
```
deepak-singhal0408: why you have 2 files?

scripts/chassis_db_consistency_checker.py
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env python3
```
deepak-singhal0408, line 1: is there a plan to extend this to other DBs consistency check as well in future? if so, do you want to instead call it as chassis_db_consistency_checker?

arlakshm: renamed the file

scripts/chassis_lag_id_checker
arlakshm: removed the extra file

deepak-singhal0408 (dismissed):
LGTM. thanks,

PR #4089, opened 2025-10-11
Fetch capability of mirror before configuring it

stephenxs:
#### What I did

Fetch capability of ingress/egress mirror before configuring it and avoid configuring ingress/egress mirror on a platform that does not support it.

#### How I did it

Check the capability in `PORT_INGRESS_MIRROR_CAPABLE` and `PORT_EGRESS_MIRROR_CAPABLE` in `STATE_DB` table `SWITCH_CAPABILITY`.
The capability of ingress/egress mirror is inserted to STATE_DB by orchagent during initialization.

#### How to verify it

Manual test and unit test

### DIFF
--- a/tests/config_mirror_session_test.py
+++ b/tests/config_mirror_session_test.py
@@ -473,13 +473,3 @@ def test_mirror_session_capability_function():
         assert config.is_port_mirror_capability_supported("tx") is False
         assert config.is_port_mirror_capability_supported("both") is False
         assert config.is_port_mirror_capability_supported(None) is False
-
-    # Test 4: Test exception handling (backward compatibility)
-    with mock.patch('config.main.SonicV2Connector') as mock_connector:
-        mock_connector.side_effect = Exception("Connection failed")
-
-        # Should return True for backward compatibility when exception occurs
-        assert config.is_port_mirror_capability_supported("rx") is True
-        assert config.is_port_mirror_capability_supported("tx") is True
-        assert config.is_port_mirror_capability_supported("both") is True
-        assert config.is_port_mirror_capability_supported(None) is True


config/main.py
```diff
@@ -1118,6 +1118,35 @@ def interface_has_mirror_config(ctx, mirror_table, dst_port, src_port, direction
...
+    try:
+        state_db = SonicV2Connector(host='127.0.0.1')
+        state_db.connect(state_db.STATE_DB, False)
+        entry_name = "SWITCH_CAPABILITY|switch"
+
+        # If no direction is specified, check both ingress and egress capabilities
+        if not direction:
+            ingress_supported = state_db.get(state_db.STATE_DB, entry_name, "PORT_INGRESS_MIRROR_CAPABLE")
+            egress_supported = state_db.get(state_db.STATE_DB, entry_name, "PORT_EGRESS_MIRROR_CAPABLE")
+            return ingress_supported == "true" and egress_supported == "true"
```
bingwang-ms: You may also need to handle the case if the result for `PORT_INGRESS_MIRROR_CAPABLE` or `PORT_EGRESS_MIRROR_CAPABLE` is missing.
It can be handled in swss or utility. Otherwise it can break other platform that doesn't have this capability check yet

stephenxs: Hi @bingwang-ms 
These fields are always inserted by switch orch during initialization.
However, the question is what should we assume in case a vendor does not support such capability to be polled. We have two options
1. We assume such a vendor does not support `PORT_INGRESS_MIRROR_CAPABLE` or `PORT_EGRESS_MIRROR_CAPABLE`
2. We assume such a vendor does support the above attributes but does not support to fetch whether they are supported.

Currently, we take 1. see code [here](https://github.com/sonic-net/sonic-swss/pull/3934/files#diff-24f98a56c9b86f83c343860373973def43ea3b64e498630f81efe809ee39e819R1876).
We can take 2 but it's a bit strange to me .. In that case we just need to change `false` to `true`.

bingwang-ms: My suggestion is to align with current behavior. Then we should go with option 1 and this is aligned with HLD. But it seems the code in sonic-swss is not aligned with this. I also left a comment in sonic-swss PR.

stephenxs: Ok. It should be option 2 that aligns with the current behavior, right

bingwang-ms: Resolving this as the issue has been done in swss code.
So sonic-swss PR has to be merged first

bingwang-ms (approved):

bingwang-ms:
@qiluo-msft Can you help review this change? Thanks

config/main.py
```diff
@@ -1118,6 +1118,35 @@ def interface_has_mirror_config(ctx, mirror_table, dst_port, src_port, direction
 
     return False
 
+
+def is_port_mirror_capability_supported(direction, namespace=None):
+    """ Check if port mirror capability is supported for the given direction """
+    try:
+        state_db = SonicV2Connector(host='127.0.0.1')
```
qiluo-msft: Could you use unix socket which has better performance? #Closed

config/main.py
```diff
@@ -1118,6 +1118,35 @@ def interface_has_mirror_config(ctx, mirror_table, dst_port, src_port, direction
...
+            if ingress_supported != "true":
+                return False
+
+        if direction in ['tx', 'both']:
+            egress_supported = state_db.get(state_db.STATE_DB, entry_name, "PORT_EGRESS_MIRROR_CAPABLE")
+            if egress_supported != "true":
+                return False
+
+        return True
+    except Exception:
```
qiluo-msft: Could you use more specific exception type? #Closed

config/main.py
stephenxs: Fixed

config/main.py
stephenxs: Removed as we do not expect any exception here.

stephenxs:
It's weird that pretest passed locally but failed in the github
```
stephens@675ff3c81f2e:/sonic/src/sonic-utilities/config$ git commit -s
[INFO] Initializing environment for https://github.com/PyCQA/flake8.
[INFO] Installing environment for https://github.com/PyCQA/flake8.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
flake8...................................................................Passed
[detached HEAD 8bf2b0d7] Fix community review comments
 1 file changed, 19 insertions(+), 23 deletions(-)
stephens@675ff3c81f2e:/sonic/src/sonic-utilities/config$ git log
```

liat-grozovik:
@qiluo-msft can you please check if you are ok with comments handling and approve this PR?

yxieca:
@stephenxs  this change causes test failure when advancing sonic-utilities submodule. We need to revert to move forward.

yxieca:
Issue: https://github.com/sonic-net/sonic-mgmt/issues/21690

Test failure: https://elastictest.org/scheduler/testplan/693d87a8be086527515c111d

PR #4093, opened 2025-10-15
generate_dump: add interface FEC stats

fraserg-arista:
Add FEC stats to the tech support tarball produced by "show tech". The stats can be found in files named "interface.counters.fec-stats_$idx".


StormLiangMS:
hi @prgeor could you hep to take a look?

prgeor (approved):
@fraserg-arista will be good to include fec histogram output as well

`show interfaces counters fec-histogram EthernetXX`

StormLiangMS:
hi @fraserg-arista could you check the comment?

PR #4100, opened 2025-10-23
cli support igmp snooping

jasper-hou-micas:

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -10691,7 +10691,7 @@ def mld_snooping_q_interval(ctx, vid, q_time):
         qmr_time = 10
     qmr_time = int(qmr_time)
     if q_time <= qmr_time:
-        ctx.fail(f"query-interval {q_time} must be >= query-max-response-time {qmr_time}.")
+        ctx.fail(f"query-interval {q_time} must be > query-max-response-time {qmr_time}.")
     q_time_name = format(q_time)
     db.mod_entry('MLD_L2MC', vlan_name, {'query-interval': q_time_name})
 
--- a/config/main.py
+++ b/config/main.py
@@ -10084,6 +10084,9 @@ def igmp_snooping_enable(ctx, vid):
     vlan = db.get_entry('VLAN', vlan_name)
     if len(vlan) == 0:
         ctx.fail("{} doesn't exist".format(vlan_name))
+    l2mc_cfg = db.get_entry('MLD_L2MC', vlan_name)
+    if l2mc_cfg  and l2mc_cfg['enabled'] == 'true':
+        ctx.fail(f"{vlan_name} does not enable IGMP Snooping.please first disable MLD Snooping")
     l2mc_enable = db.get_entry('L2MC', vlan_name)
     if len(l2mc_enable) < 1:
         igmp_snooping_fvs = {
@@ -10112,6 +10115,21 @@ def igmp_snooping_disable(ctx, vid):
     vlan = db.get_entry('VLAN', vlan_name)
     if len(vlan) == 0:
         ctx.fail("{} doesn't exist".format(vlan_name))
+    l2mc_mrtrs = db.get_table('L2MC_MROUTER')
+    l2mc_mrtr_keys = l2mc_mrtrs.keys()
+    for key in l2mc_mrtr_keys:
+        if vlan_name == key[0] :
+            db.set_entry('L2MC_MROUTER', key, None)
+    l2mc_static_grps = db.get_table('L2MC_STATIC_MEMBER')
+    l2mc_static_grp_keys = l2mc_static_grps.keys()
+    for key in l2mc_static_grp_keys:
+        if vlan_name == key[0] :
+            db.set_entry('L2MC_STATIC_MEMBER', key, None)
+    static_grp_table = db.get_table('L2MC_STATIC_GROUP')
+    static_grp_keys = static_grp_table.keys()
+    for key in static_grp_keys:
+        if vlan_name == key[0] :
+            db.set_entry('L2MC_STATIC_GROUP', key, None)
     db.set_entry('L2MC', vlan_name, None)
 
 @igmp_snooping.command('last-member-query-interval')
@@ -10145,6 +10163,16 @@ def igmp_snooping_qmr_interval(ctx, vid, qmr_time):
     vlan = db.get_entry('VLAN', vlan_name)
     if len(vlan) == 0:
         ctx.fail("{} doesn't exist".format(vlan_name))
+    l2mc_cfg = db.get_entry('L2MC', vlan_name)
+    if not l2mc_cfg:
+        ctx.fail(f"{vlan_name} does not enable MLD Snooping.")
+    q_time = l2mc_cfg.get('query-interval')
+    if q_time is None:
+        q_time = 125
+    q_time = int(q_time)
+
+    if q_time <= qmr_time:
+        ctx.fail(f"query-max-response-time {qmr_time} must be < query-interval {q_time}.")
     qmr_time_name = format(qmr_time)
     db.mod_entry('L2MC', vlan_name, {'query-max-response-time': qmr_time_name})
 
@@ -10162,6 +10190,16 @@ def igmp_snooping_q_interval(ctx, vid, q_time):
     vlan = db.get_entry('VLAN', vlan_name)
     if len(vlan) == 0:
         ctx.fail("{} doesn't exist".format(vlan_name))
+    l2mc_cfg = db.get_entry('L2MC', vlan_name)
+    if not l2mc_cfg:
+        ctx.fail(f"{vlan_name} does not enable MLD Snooping.")
+    qmr_time = l2mc_cfg.get('query-max-response-time')
+    if qmr_time is None:
+        qmr_time = 10
+    qmr_time = int(qmr_time)
+
+    if q_time <= qmr_time:
+        ctx.fail(f"query-interval {q_time} must be > query-max-response-time {qmr_time}.")
     q_time_name = format(q_time)
     db.mod_entry('L2MC', vlan_name, {'query-interval': q_time_name})
 
@@ -10186,7 +10224,7 @@ def igmp_snooping_optimised_multicast_flood(ctx, vid, optimised_multicast_flood)
     if len(l2mc_enable) < 1:
         ctx.fail("{} doesn't enable IGMP Snooping".format(vlan_name))
     
-    db.mod_entry('L2MC_SUPPRESS', vlan_name, {'optimised-multicast-flood': optimised_multicast_flood})
+    db.mod_entry('L2MC_SUPPRESS', vlan_name, {'ipv4-optimised-multicast-flood': optimised_multicast_flood})
 
 @igmp_snooping.command('link_local_groups_suppression')
 @click.argument('link_local_groups_suppression', metavar='<enable|disable>', required=True)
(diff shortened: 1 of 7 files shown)


config/main.py
```diff
@@ -9861,5 +9861,942 @@ def del_vnet_route(ctx, vnet_name, prefix):
...
+    if len(l2mc_enable) < 1:
+        ctx.fail("{} doesn't enable IGMP Snooping".format(vlan_name))
+    
+    db.mod_entry('L2MC_SUPPRESS', vlan_name, {'ipv4-optimised-multicast-flood': optimised_multicast_flood})
+
+@igmp_snooping.command('link_local_groups_suppression')
+@click.argument('link_local_groups_suppression', metavar='<enable|disable>', required=True)
+@click.argument('vid', metavar='<vlan>', required=True, type=int)
+@click.pass_context
+def igmp_snooping_optimised_multicast_flood(ctx, vid, link_local_groups_suppression):
```
aidan-gallagher: **Critical:** This function is also named `igmp_snooping_optimised_multicast_flood` -- same as the one at line 10036. Python silently overwrites the first definition, making the `optimised_multicast_flood` command unreachable (this one handles `link_local_groups_suppression` instead).

Same issue exists for MLD at lines 10492/10517, and in `show/main.py` at lines 2887/3016 (`igmp_snooping_vlan`) and 3343/3469 (`mld_snooping_vlan`).

config/main.py
```diff
@@ -9861,5 +9861,942 @@ def del_vnet_route(ctx, vnet_name, prefix):
...
+    except ValueError as err:
+        ctx.fail("IP address is not valid: {}".format(err))
+
+    if not ip_address.is_multicast:
+        ctx.fail("IP address {} is not multicast".format(str(ip_address)))
+
+    if ip_address < ipaddress.IPv4Address("224.0.0.0") or ip_address > ipaddress.IPv4Address("239.255.255.255"):
+        ctx.fail("IP address {} is outside the valid multicast range".format(str(ip_address)))
+
+    if ip_address < ipaddress.IPv4Address("224.0.0.0") or ip_address <= ipaddress.IPv4Address("224.0.0.255"):
```
aidan-gallagher: **Critical:** The `or` here should be `and`. As written, the condition is: `ip < 224.0.0.0 OR ip <= 224.0.0.255`. The first condition is true for everything below the multicast range, and the second is true for the reserved range *and* everything below it. The result happens to be correct by accident for addresses > 224.0.0.255, but the logic does not express the intended range check (`224.0.0.0 <= ip <= 224.0.0.255`).

show/main.py
```diff
@@ -2875,5 +2876,924 @@ def banner(db):
...
+                    app_mrouter_dict [str(mrouter_vlan_id)] = [mrouter_interface] 
+     
+            #click.echo ("APP_DB: {}".format(app_mrouter_dict)) 
+     
+        key = "Vlan"+str(vlan_id) 
+        data = config_db.get_entry('MLD_L2MC', key) 
+        app_l2mc_supress_key = "L2MC_SUPPRESS_TABLE:Vlan" + str(vlan_id)
+
+     
+        if not data:
```
aidan-gallagher, line 3582: **Critical:** `return` here exits the entire `mld_snooping_all` function on the first VLAN that is disabled or empty, skipping all subsequent VLANs. The IGMP equivalent at lines 3071-3075 correctly uses `continue`. This is a copy-paste bug -- should be `continue`.

config/vlan.py
```diff
@@ -147,6 +147,20 @@ def del_vlan(db, vid, multiple, no_restart_dhcp_relay):
 
     config_db = ValidatedConfigDBConnector(db.cfgdb)
 
+    keys = config_db.get_keys("L2MC")
```
aidan-gallagher: **High:** This L2MC guard rail is ineffective when `--multiple` is used. `vid` is the raw CLI argument (e.g., `"100-200"`), so `f"Vlan{vid}"` produces `"Vlan100-200"` which never matches any CONFIG_DB key. The check is bypassed entirely in multiple mode.

Same issue in `del_vlan_member` at line 387.

config/main.py
```diff
@@ -9861,5 +9861,942 @@ def del_vnet_route(ctx, vnet_name, prefix):
...
+        ctx.fail("Invalid VlanId!!")
+    vlan_name = 'Vlan{}'.format(vid)
+    vlan = db.get_entry('VLAN', vlan_name)
+    if len(vlan) == 0:
+        ctx.fail("{} doesn't exist".format(vlan_name))
+    ver_name = format(version)
+    db.mod_entry('L2MC', vlan_name, {'version': ver_name})
+
+@igmp_snooping.group('querier')
+@clicommon.pass_db
```
aidan-gallagher: **High:** This subgroup uses `@clicommon.pass_db` which sets `ctx.obj` to a `Db` wrapper instance. However, the parent `igmp_snooping` group (line 9867) sets `ctx.obj = {'db': config_db}` (a dict with a `ConfigDBConnector`). Child commands that do `ctx.obj['db']` will get a `TypeError: 'Db' object is not subscriptable` when called through this subgroup path.

Also applies to subgroups at lines 10133 and 10169.

aidan-gallagher (commented):
This PR adds IGMP/MLD snooping CLI commands (config, show, debug). The IGMP and MLD implementations are nearly identical copy-paste with minimal abstraction, which has already produced several bugs where MLD code says "IGMP" and vice versa.

| Critical | High | Medium | Low | Total |
|----------|------|--------|-----|-------|
| 3 | 7 | 10 | 5 | **25** |

**No test coverage (High):** ~2,000 lines of new CLI code with zero tests. The vendor's test results only document MLD snooping hardware tests -- no IGMP CLI tests.

**Unrelated commits bundled (Medium):** A kdump fix and `fast-reboot` change are included in this PR.

Additional findings not called out inline: `static-group` and `mrouter add` commands don't verify snooping is enabled on the VLAN before adding config (config/main.py:10179-10223, 10269-10292); `mrouter add` doesn't verify the port is a VLAN member; MLD `all` show uses variable named `igmp_snooping_fvs` (show/main.py:3488); command names use underscores (`optimised_multicast_flood`, `link_local_groups_suppression`) where SONiC convention is hyphens.

show/main.py
```diff
@@ -2875,5 +2876,924 @@ def banner(db):
...
+    config_db = ConfigDBConnector()
+    config_db.connect()
+    vlan_dict = {}
+
+    l2mc_keys = config_db.get_keys('L2MC')
+    if not l2mc_keys:
+        return
+
+    for key in l2mc_keys:
+        vlan_id = str(key.strip("Vlan"))
```
aidan-gallagher: **High:** `.strip("Vlan")` strips individual characters from the set `{'V','l','a','n'}`, not the substring `"Vlan"`. This works by coincidence for numeric VLAN IDs but is semantically wrong. There are 20+ occurrences across the show commands. Consider using `key[4:]` or `key.removeprefix("Vlan")` instead.

debug/main.py
```diff
@@ -279,6 +279,20 @@ def rib():
...
+@cli.group()
+def IGMP_Snooping():
+    """debug IGMP-Snooping group"""
+    pass
+
+@IGMP_Snooping.command('loglevel')
+@click.argument('log', metavar='<0-6>', required=True, type=click.IntRange(0, 6))
+def loglevel(log):
+    """Set loglevel for L2MCD (0-6)"""
+    command = ['docker', 'exec', '-it','l2mcd', 'l2mcd_ctl', '-d', str(log)]
```
aidan-gallagher: **High:** The `-it` flag requires an interactive terminal (TTY). This command will fail when run in non-interactive contexts like scripts, automation, or CI. Consider using `-i` only or omitting the flag.

config/main.py
```diff
@@ -9861,5 +9861,942 @@ def del_vnet_route(ctx, vnet_name, prefix):
...
+    vlan_valid = clicommon.is_vlanid_in_range(int(vid))
+    if vlan_valid == False:
+        ctx.fail("Invalid VlanId!!")
+    vlan_name = 'Vlan{}'.format(vid)
+    vlan = db.get_entry('VLAN', vlan_name)
+    if len(vlan) == 0:
+        ctx.fail("{} doesn't exist".format(vlan_name))
+    l2mc_cfg = db.get_entry('L2MC', vlan_name)
+    if not l2mc_cfg:
+        ctx.fail(f"{vlan_name} does not enable MLD Snooping.")
```
aidan-gallagher: **Medium:** This IGMP snooping command's error message says "MLD Snooping" instead of "IGMP Snooping". Same issue at line 10020.

The reverse is also present: MLD commands at lines 10592, 10599, and 10614 have docstrings saying "IGMP" instead of "MLD". These copy-paste errors suggest the MLD implementation was cloned from the IGMP code.

show/main.py
```diff
@@ -2875,5 +2876,924 @@ def banner(db):
...
+                'version' : '2', 
+                'fast-leave': 'false', 
+                'query-interval':'125', 
+                'query-max-response-time': '10', 
+                'last-member-query-interval': '1000', 
+                'ipv4-optimised-multicast-flood' : 'false',
+                'ipv4-link-local-groups-suppression' : 'false',
+                }  
+     
+        app_db = SonicV2Connector(host='127.0.0.1')
```
aidan-gallagher: **Medium:** A new `SonicV2Connector` is created on every iteration of the VLAN loop. This establishes a new Redis connection per VLAN, which is unnecessary. The connection should be created once before the loop. Same issue at line 3500 in the MLD path.

show/main.py
```diff
@@ -2875,5 +2876,924 @@ def banner(db):
...
+    """ IGMP snooping groups vlan <vlan-id>"""
+    app_db = SonicV2Connector(host='127.0.0.1')
+    app_db.connect(app_db.APPL_DB) 
+
+    member_type_dict = {}
+    member_table_dict = {}
+    mrouter_type_dict = {}
+    mrouter_table_dict = {}
+
+    member_table_keys = app_db.keys(app_db.APPL_DB, 'L2MC_MEMBER_TABLE:*')
```
aidan-gallagher: **Medium:** `app_db.keys()` can return `None` when no keys match the pattern. Iterating over `None` at line 3268 will raise `TypeError`. Add a guard like `member_table_keys = ... or []`. Same issue at lines 3177-3180.

config/main.py
jasper-hou-micas: Will further revise and optimize accordingly

config/main.py
jasper-hou-micas: Will further revise and optimize accordingly

show/main.py
jasper-hou-micas: Will further revise and optimize accordingly

config/vlan.py
jasper-hou-micas: Will further revise and optimize accordingly

config/main.py
jasper-hou-micas: Will further revise and optimize accordingly

show/main.py
jasper-hou-micas: Will further revise and optimize accordingly

debug/main.py
jasper-hou-micas: Will further revise and optimize accordingly

config/main.py
jasper-hou-micas: Will further revise and optimize accordingly

show/main.py
jasper-hou-micas: Will further revise and optimize accordingly

show/main.py
jasper-hou-micas: Will further revise and optimize accordingly

jasper-hou-micas:
> **No test coverage (High):** ~2,000 lines of new CLI code with zero tests. The vendor's test results only document MLD snooping hardware tests -- no IGMP CLI tests.

@aidan-gallagher 
The IGMP test cases were provided in the first email, and the MLD test cases are included in the latest attachment, with the duplicate items removed.
The CLI code has been covered by system and functional testing.

PR #4118, opened 2025-11-04
Remove direct dependency on libyang

### LINKED ISSUE #22385: Enhancement: Upgrade to libyang3
### Is it platform specific

generic

### Importance or Severity

High

SONiC currently uses 3 versions of libyang: 1.0.73, 1.0.184, and 2.1.148. All are end of life and potentially have security issues. The current release is 3.12.2 and that is what we are targeting here.

There are also numerous libyang related bugs that have been "worked around" in SONiC due to the shortcomings of libyang1. Hopefully upgrading to a modern version will resolve such issues.

### Current Behavior

3 libyang versions

### Proposed Behavior

single libyang version

### PR Tracking Table
| PR | description | dependencies |
|---------|--------|--------|
| #21679 | Remove libyang 1.0.184. build libyang 3.12.2, libyang-python 3.1.0.  Migrate FRR to libyang 3.  Update sonic-yang-models test cases specifically to make it easier to see libyang3-specific changes once those land. **THIS CAN BE MERGED WITHOUT ANY POSSIBLE IMPACT** | None |
| #21719 | Port sonic-yang-models test cases **THIS CAN BE MERGED WITHOUT ANY POSSIBLE IMPACT** | #21679 |
| #21716 | Port sonic-yang-mgmt. | #21679, #21719 |
| #21718 | Update all subprojects that use libyang1 to depend on libyang3 | #21679, #21719, #21716 |
| https://github.com/sonic-net/sonic-utilities/pull/3766 | sonic-utilities | #21679, #21719, #21716, #21718 |
| #21752, https://github.com/sonic-net/sonic-swss-common/pull/973 | swss-common | #21679, #21719, #21716, #21718 |
| #21789, https://github.com/sonic-net/sonic-mgmt-common/pull/157 | sonic-mgmt-common (GOlang) | #21679, #21718 |
| https://github.com/sonic-net/sonic-platform-daemons/pull/589 , https://github.com/sonic-net/sonic-swss/pull/3518 , https://github.com/sonic-net/sonic-dbsyncd/pull/74 , https://github.com/sonic-net/sonic-restapi/pull/157 , https://github.com/sonic-net/sonic-snmpagent/pull/348 , https://github.com/sonic-net/sonic-platform-common/pull/544 | azure pipelines: update various external repos to pull in the libyang3 debs that are otherwise not in need of modification | all of the above |
| #21790 | cleanup / remove libyang1 | all of the above |

bhouse-nexthop:
#### What I did

sonic-utilities should never call directly into libyang.  All functionality it needs should be provided in sonic-yang-mgmt.  This resolves issues when libyang changes API/ABI such as happened between libyang1 and libyang2, and again between libyang2 and libyang3.

Depends on https://github.com/sonic-net/sonic-buildimage/pull/24414 and https://github.com/sonic-net/sonic-buildimage/pull/26563
Fixes https://github.com/sonic-net/sonic-buildimage/pull/22385

#### How I did it

Removed calls to libyang and instead uses new functionality introduced in sonic-yang-mgmt.

Also installs both libyang1 and libyang3 for tests to maintain compatibility with both during the migration.

#### How to verify it

After https://github.com/sonic-net/sonic-buildimage/pull/24414 merges observe test cases still pass.

N/A

N/A

### DIFF
--- a/config/config_mgmt.py
+++ b/config/config_mgmt.py
@@ -8,7 +8,6 @@ import re
 import shutil
 import syslog
 import tempfile
-import yang as ly
 from json import load
 from sys import flags
 from time import sleep as tsleep
@@ -35,8 +34,7 @@ class ConfigMgmt():
     to verify config for the commands which are capable of change in config DB.
     '''
 
-    def __init__(self, source="configDB", debug=False, allowTablesWithoutYang=True,
-                 sonicYangOptions=0, configdb=None):
+    def __init__(self, source="configDB", debug=False, allowTablesWithoutYang=True, configdb=None):
         '''
         Initialise the class, --read the config, --load in data tree.
 
@@ -55,7 +53,6 @@ class ConfigMgmt():
             self.configdbJsonOut = None
             self.source = source
             self.allowTablesWithoutYang = allowTablesWithoutYang
-            self.sonicYangOptions = sonicYangOptions
             self.configdb = configdb
 
             # logging vars
@@ -71,7 +68,7 @@ class ConfigMgmt():
         return
 
     def __init_sonic_yang(self):
-        self.sy = sonic_yang.SonicYang(YANG_DIR, debug=self.DEBUG, sonic_yang_options=self.sonicYangOptions)
+        self.sy = sonic_yang.SonicYang(YANG_DIR, debug=self.DEBUG)
         # load yang models
         self.sy.loadYangModel()
         # load jIn from config DB or from config DB json file.
@@ -291,8 +288,7 @@ class ConfigMgmt():
 
         # Instantiate new context since parse_module_mem() loads the module into context.
         sy = sonic_yang.SonicYang(YANG_DIR)
-        module = sy.ctx.parse_module_mem(yang_module_str, ly.LYS_IN_YANG)
-        return module.name()
+        return sy.load_module_str_name(yang_module_str)
 
 
 # End of Class ConfigMgmt
--- a/sonic_package_manager/manager.py
+++ b/sonic_package_manager/manager.py
@@ -6,7 +6,6 @@ import os
 import pkgutil
 import subprocess
 import tempfile
-import yang as ly
 from inspect import signature
 from typing import Any, Iterable, List, Callable, Dict, Optional
 
@@ -1344,7 +1343,7 @@ class PackageManager:
         docker_api = DockerApi(docker.from_env(), ProgressManager())
         registry_resolver = RegistryResolver()
         metadata_resolver = MetadataResolver(docker_api, registry_resolver)
-        cfg_mgmt = config_mgmt.ConfigMgmt(source=INIT_CFG_JSON, sonicYangOptions=ly.LY_CTX_DISABLE_SEARCHDIR_CWD)
+        cfg_mgmt = config_mgmt.ConfigMgmt(source=INIT_CFG_JSON)
         cli_generator = CliGenerator(log)
         feature_registry = FeatureRegistry(SonicDB)
         service_creator = ServiceCreator(feature_registry,
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -18,4 +18,4 @@ jobs:
       - uses: actions/checkout@v3
       - run: semgrep ci
         env:
-           SEMGREP_RULES: "p/default r/python.lang.security.audit.dangerous-system-call-audit.dangerous-system-call-audit"
+           SEMGREP_RULES: "p/default r/python.lang.security.audit.dangerous-system-call-audit.dangerous-system-call-audit .semgrep/"
--- a/.semgrep/no-direct-libyang.yml
+++ b/.semgrep/no-direct-libyang.yml
@@ -1,0 +1,21 @@
+rules:
+  - id: no-direct-libyang-import
+    patterns:
+      - pattern-either:
+          - pattern: import yang
+          - pattern: import yang as $X
+          - pattern: from yang import $X
+          - pattern: from yang import $X as $Y
+          - pattern: import libyang
+          - pattern: import libyang as $X
+          - pattern: from libyang import $X
+          - pattern: from libyang import $X as $Y
+    message: >-
+      Do not import libyang directly. sonic-utilities must go through
+      sonic_yang / sonic_yang_ext so that libyang API/ABI changes (libyang1
+      vs libyang2 vs libyang3) are isolated to sonic-yang-mgmt.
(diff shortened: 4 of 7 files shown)


config/config_mgmt.py
```diff
@@ -71,7 +68,7 @@ def __init__(self, source="configDB", debug=False, allowTablesWithoutYang=True,
         return
 
     def __init_sonic_yang(self):
-        self.sy = sonic_yang.SonicYang(YANG_DIR, debug=self.DEBUG, sonic_yang_options=self.sonicYangOptions)
```
saiarcot895, line 74: Why can't options be used? Because the underlying enums/constants could change across libyang versions?

bhouse-nexthop: correct, the point is to remove anything that is libyang-specific.  The options here were meaningless anyhow.

bhouse-nexthop:
@saiarcot895 @qiluo-msft this is the next libyang3 blocker

sonic_package_manager/manager.py
```diff
@@ -1344,7 +1343,7 @@ def get_manager() -> 'PackageManager':
         docker_api = DockerApi(docker.from_env(), ProgressManager())
         registry_resolver = RegistryResolver()
         metadata_resolver = MetadataResolver(docker_api, registry_resolver)
-        cfg_mgmt = config_mgmt.ConfigMgmt(source=INIT_CFG_JSON, sonicYangOptions=ly.LY_CTX_DISABLE_SEARCHDIR_CWD)
```
saiarcot895, line 1347: Wouldn't this option mean that the current working directory can't be used to resolve any paths in the config being loaded, that it would have to (I guess) all be from the root directory?

bhouse-nexthop: The existing code disabled searching from CWD.

In the libyang3 port of sonic-yang-mgmt we never search CWD for yang models as the python bindings disable searching by default: https://github.com/CESNET/libyang-python/blob/a69eb4a6865069de1a8b73d76b2d7427893214cf/libyang/context.py#L201

And we don't override it when we initialize the context:
https://github.com/bhouse-nexthop/sonic-buildimage/blob/4390075149ae6eb60c2430e74b416f83c33d2e94/src/sonic-yang-mgmt/sonic_yang.py#L56

So the end result is identical.

saiarcot895 (dismissed):

rookie-who:
@bhouse-nexthop Two things:

1. The `find_ref_paths()` change is behavioral — old code walked leaf xpaths in Python and called `find_data_dependencies()` per leaf, new code passes the parent xpath and relies on sonic-yang-mgmt to recurse. Is there a test covering this equivalence? If not, worth adding one to catch any divergence.

2. What prevents a future PR from re-introducing `import yang as ly`? A lint rule or CI check (e.g. grep for `import yang` / `from yang`) would make this a hard gate rather than a convention.

rookie-who:
Re: preventing future `import yang` — a semgrep rule would slot right into the existing CI:

```yaml
# .semgrep/no-direct-libyang.yml
rules:
  - id: no-direct-libyang-import
    patterns:
      - pattern-either:
          - pattern: import yang
          - pattern: import yang as $X
          - pattern: from yang import $X
    message: "Do not import libyang directly. Use sonic_yang / sonic_yang_ext instead."
    languages: [python]
    severity: ERROR
    paths:
      exclude:
        - tests/
```

This would fail CI on any PR that reintroduces the dependency. Could be added in this PR or as a follow-up.

rookie-who:
Correction on my earlier comment — the existing `find_ref_paths` tests in `gu_common_test.py` do cover non-leaf inputs (`path="/PORT"`, `path=""`), which exercises the changed code path. These should catch any divergence between the old Python leaf-iteration and the new sonic-yang-mgmt recursive call.

The `load_module_str_name()` path in `get_module_name()` is still untested though — worth a quick unit test to cover that.

bhouse-nexthop:
@rookie-who

> @bhouse-nexthop Two things:
> 
> 1. The `find_ref_paths()` change is behavioral — old code walked leaf xpaths in Python and called `find_data_dependencies()` per leaf, new code passes the parent xpath and relies on sonic-yang-mgmt to recurse. Is there a test covering this equivalence? If not, worth adding one to catch any divergence.

Yes, the tests most definitely cover this.  Both in the pre-existing sonic-utilities tests as well as tests specifically added to sonic-yang-mgmt for this explicit case.

One of the sonic-yang-mgmt tests that covers this exact scenario is:
https://github.com/sonic-net/sonic-buildimage/blob/406302a9a060a7a40fa05ecbbc2b8c2e4e805943/src/sonic-yang-mgmt/tests/libyang-python-tests/test_SonicYang.json#L68-L76

> 2. What prevents a future PR from re-introducing `import yang as ly`? A lint rule or CI check (e.g. grep for `import yang` / `from yang`) would make this a hard gate rather than a convention.

That's a good point, I can look into how to specifically prevent that.

bradh352:
@rookie-who
> Correction on my earlier comment — the existing `find_ref_paths` tests in `gu_common_test.py` do cover non-leaf inputs (`path="/PORT"`, `path=""`), which exercises the changed code path. These should catch any divergence between the old Python leaf-iteration and the new sonic-yang-mgmt recursive call.
> 
> The `load_module_str_name()` path in `get_module_name()` is still untested though — worth a quick unit test to cover that.

That's covered here: https://github.com/bhouse-nexthop/sonic-utilities/blob/86f1c8ab03eef19531ceca43a6943dbe29b785bd/tests/config_mgmt_test.py#L30

bradh352:
@rookie-who
> Re: preventing future `import yang` — a semgrep rule would slot right into the existing CI:
> 
> ```yaml
> # .semgrep/no-direct-libyang.yml
> rules:
>   - id: no-direct-libyang-import
>     patterns:
>       - pattern-either:
>           - pattern: import yang
>           - pattern: import yang as $X
>           - pattern: from yang import $X
>     message: "Do not import libyang directly. Use sonic_yang / sonic_yang_ext instead."
>     languages: [python]
>     severity: ERROR
>     paths:
>       exclude:
>         - tests/
> ```
> 
> This would fail CI on any PR that reintroduces the dependency. Could be added in this PR or as a follow-up.

Thanks, mostly used this, but also had to add libyang as a pattern since libyang3 uses that import name.

rookie-who (approved):
Semgrep rule looks good — covers both `yang` and `libyang` imports. Thanks for picking that up.

LGTM.

vaibhavhd (approved):

saiarcot895 (approved):

PR #4123, opened 2025-11-10
Issue 24036: Add help for 'show ip route' subcommands

### LINKED ISSUE #24036: Enhancement: Show options for `show ip route`
### Is it platform specific

generic

### Importance or Severity

Medium

the command `show ip route` currently does not show any information on possible command options in its help text:
`
user@sonic-host:~$ show ip route -h
Usage: show ip route [OPTIONS] [IPADDRESS] [vrf <vrf_name>] [...]

  Show IP (IPv4) routing table

Options:
  -d, --display TEXT    all|frontend
  -n, --namespace TEXT  Namespace name or all
  --verbose             Enable verbose output
  -?, -h, --help        Show this message and exit.
`

For example, show ip route summary is an option but no info is in the help text

### Current Behavior

current behaviour shows no info for subcommands

### Proposed Behavior

display subcommands in help text

anders-nexthop:
#### What I did

Add a new `VtyshCommand` class that extends click CLI commands with vtysh help text and shell completions. The class calls into vtysh to get subcommand information and generates help text following the existing SONiC CLI format.

Closes: [#24036](https://github.com/sonic-net/sonic-buildimage/issues/24036)

#### How I did it

Call into vtysh to extract subcommands and descriptions, and use that information to populate help text and completions. A `VtyshCommand` wrapper class handles everything so other commands can easily be extended.

**Update from original submission:** Replaced the old `click._bashcomplete` monkey-patch (removed in Click 8) with Click 8's native shell completion API — a `VtyshParamType` class that implements `shell_complete()` using `CompletionItem`. This is cleaner and forward-compatible.

An important consideration is whether it's acceptable to call into vtysh to retrieve this information. The underlying issue is that the `show ip route` command tree is implemented in FRR, not SONiC. The current CLI is a thin shim. This approach is a general solution that works for any show command passing through to vtysh, and automatically supports new vtysh commands.

Note: This PR does not handle description strings for user-defined arguments (IPADDRESS, vrf NAME, etc). That can be added in a follow-up.

#### How to verify it

Manually verify that subcommand help works, existing help works, tab completions work, and existing commands still function.

Unit tests:

```
tests/vtysh_help_test.py::TestVtyshHelpCommands::test_vtysh_help_basic_functionality PASSED
tests/vtysh_help_test.py::TestVtyshHelpCommands::test_vtysh_help_with_subcommands PASSED
tests/vtysh_help_test.py::TestVtyshHelpCommands::test_vtysh_help_without_subcommands PASSED
tests/vtysh_help_test.py::TestVtyshHelpCommands::test_nested_command_help PASSED
tests/vtysh_help_test.py::TestVtyshHelpCommands::test_basic_invalid_command_error_handling PASSED
tests/vtysh_help_test.py::TestVtyshHelpCommands::test_nested_invalid_command_error_handling PASSED
tests/vtysh_help_test.py::TestVtyshHelpCommands::test_vtysh_caching_functionality PASSED
tests/vtysh_help_test.py::TestVtyshCompletionCommands::test_basic_completion_functionality PASSED
tests/vtysh_help_test.py::TestVtyshCompletionCommands::test_nested_completion_functionality PASSED
tests/vtysh_help_test.py::TestVtyshCompletionCommands::test_completion_with_no_results PASSED
tests/vtysh_help_test.py::TestVtyshCompletionCommands::test_completion_with_vtysh_error PASSED
tests/vtysh_help_test.py::TestVtyshCompletionCommands::test_completion_with_whitespace_handling PASSED
tests/vtysh_help_test.py::TestVtyshCompletionCommands::test_nested_command_help_with_completion PASSED
tests/vtysh_help_test.py::TestVtyshCompletionCommands::test_nested_command_help_with_inline_completion PASSED
tests/vtysh_help_test.py::TestVtyshCompletionCommands::test_completion_false_uses_space_before_question_mark PASSED
tests/vtysh_help_test.py::TestVtyshCompletionCommands::test_completion_true_uses_no_space_before_question_mark PASSED
tests/vtysh_help_test.py::TestVtyshCompletionCommands::test_shell_complete_end_to_end PASSED
tests/vtysh_help_test.py::TestVtyshCompletionCommands::test_shell_complete_end_to_end_no_filter PASSED
```

#### New command output

```
admin@vlab-01:~$ show ip route -h
Usage: show ip route [OPTIONS] COMMAND [ARGS]...

Show IP (IPv4) routing table

Options:
  -d, --display TEXT    all|frontend
  -n, --namespace TEXT  Namespace name or all
  --verbose             Enable verbose output
  -?, -h, --help        Show this message and exit.

Commands:
  A.B.C.D         Network in the IP routing table to display
  A.B.C.D/M       IP prefix <network>/<length>, e.g., 35.0.0.0/8
  babel           Babel routing protocol (Babel)
  bgp             Border Gateway Protocol (BGP)
  connected       Connected routes (directly attached subnet or host)
  ...
  vrf             Specify the VRF
```

### DIFF
--- a/show/main.py
+++ b/show/main.py
@@ -40,6 +40,7 @@ except KeyError:
 
 from . import acl
 from . import bgp_common
+from .vtysh_helper import vtysh_command
 from . import chassis_modules
 from . import dropcounters
 from . import fabric
@@ -1411,8 +1412,9 @@ def loopback_action():
 # 'route' subcommand ("show ip route")
 #
 
-@ip.command()
-@click.argument('args', metavar='[IPADDRESS] [vrf <vrf_name>] [...]', nargs=-1, required=False)
+
+@ip.command(cls=vtysh_command("show ip route"))
+@click.argument('args', nargs=-1, required=False)
 @click.option('--display', '-d', 'display', default=None, show_default=False, type=str, help='all|frontend')
 @click.option('--namespace', '-n', 'namespace', default=None, type=str, show_default=False, help='Namespace name or all')
 @click.option('--verbose', is_flag=True, help="Enable verbose output")
@@ -1508,16 +1510,15 @@ def interfaces(namespace, display):
 # 'route' subcommand ("show ipv6 route")
 #
 
-@ipv6.command()
-@click.argument('args', metavar='[IPADDRESS] [vrf <vrf_name>] [...]', nargs=-1, required=False)
+@ipv6.command(cls=vtysh_command("show ipv6 route"))
+@click.argument('args', nargs=-1, required=False)
 @click.option('--display', '-d', 'display', default=None, show_default=False, type=str, help='all|frontend')
 @click.option('--namespace', '-n', 'namespace', default=None, type=str, show_default=False, help='Namespace name or all')
 @click.option('--verbose', is_flag=True, help="Enable verbose output")
 def route(args, namespace, display, verbose):
     """Show IPv6 routing table"""
-    # Call common handler to handle the show ipv6 route cmd
+    # Call common handler to handle the show ip route cmd
     bgp_common.show_routes(args, namespace, display, verbose, "ipv6")
-
 
 # 'protocol' command
 @ipv6.command()
--- a/show/main.py
+++ b/show/main.py
@@ -1517,7 +1517,7 @@ def interfaces(namespace, display):
 @click.option('--verbose', is_flag=True, help="Enable verbose output")
 def route(args, namespace, display, verbose):
     """Show IPv6 routing table"""
-    # Call common handler to handle the show ip route cmd
+    # Call common handler to handle the show ipv6 route cmd
     bgp_common.show_routes(args, namespace, display, verbose, "ipv6")
 
 # 'protocol' command
--- a/show/vtysh_helper.py
+++ b/show/vtysh_helper.py
@@ -1,0 +1,262 @@
+import click
+import functools
+import subprocess
+
+from click.shell_completion import CompletionItem
+
+
+class VtyshParamType(click.types.StringParamType):
+    """Custom type for vtysh command arguments that provides shell completion."""
+
+    def shell_complete(self, ctx, param, incomplete):
+        cmd = ctx.command
+        if not isinstance(cmd, VtyshCommand):
+            return []
+
+        args = ctx.params.get(param.name or "args", ())
+        if args:
+            cmd_prefix = f"{cmd.vtysh_command_prefix} {' '.join(args)}"
+        else:
+            cmd_prefix = cmd.vtysh_command_prefix
+
+        completions = cmd.get_vtysh_completions(cmd_prefix, completion=False)
+
+        if incomplete:
+            completions = [c for c in completions if c.startswith(incomplete)]
+
+        return [CompletionItem(c) for c in completions]
+
+
+class VtyshCommand(click.Command):
+    """
+    Custom Click command class that integrates vtysh help functionality.
+    This provides enhanced help by showing both Click help and vtysh subcommands.
+    """
+
+    # List of vtysh commands that support completion
+    vtysh_completion_commands = []
+
+    def __init__(self, name, vtysh_command_prefix, **kwargs):
+        """
+        Initialize the VtyshCommand.
+
+        Args:
+            name: Command name
+            vtysh_command_prefix: The vtysh command prefix (e.g., "show ip route")
+            **kwargs: Other Click command arguments
+        """
+        self.vtysh_command_prefix = vtysh_command_prefix
+        super().__init__(name, **kwargs)
+
+        # Patch variadic arguments to use VtyshParamType for shell completion
+        for param in self.params:
+            if isinstance(param, click.Argument) and param.nargs == -1:
+                param.type = VtyshParamType()
+
+    def parse_args(self, ctx, args):
+        """Track vtysh command args for later use"""
+        help_options = ['-h', '--help', '-?', '?']
+        self.raw_args = []
+        for arg in args:
+            if arg in help_options:
+                break
+            self.raw_args.append(arg)
+        # SONiC CLI accepts '?' as a hidden help option, handle it explicitly here
+        if '?' in args:
+            click.echo(ctx.get_help())
+            ctx.exit()
+        return super().parse_args(ctx, args)
+
+    def get_help(self, ctx):
+        """Override Click's get_help to provide enhanced vtysh help."""
+        formatter = click.HelpFormatter()
+
+        # Try the full command first
+        is_valid = False
+        last_valid_command = self.vtysh_command_prefix
+        if len(self.raw_args) == 0:
+            is_valid = True
+        else:
+            arg_prefix = ' '.join(self.raw_args[:-1])
+            full_command_prefix = f"{self.vtysh_command_prefix}"
+            if arg_prefix != "":
+                full_command_prefix += f" {arg_prefix}"
+            full_command = f"{self.vtysh_command_prefix} {' '.join(self.raw_args)}"
+            # Handle partial commands (ie, "show ip route sum")
+            completions = self.get_vtysh_completions(full_command)
+            if len(completions) == 1:
+                last_valid_command = f"{full_command_prefix} {completions[0]}"
+                is_valid = True
+
+        if not is_valid:
+            # If the full command failed, work backwards to find last valid command prefix
+            last_valid_command = self.vtysh_command_prefix
+            for arg in self.raw_args:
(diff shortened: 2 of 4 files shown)


show/vtysh_helper.py
```diff
@@ -0,0 +1,262 @@
...
+
+    @functools.lru_cache()
+    def get_vtysh_help(self, cmd_prefix, completion=False):
+        """
+        Get help for a vtysh command.
+        """
+        try:
+            help_command = f"{cmd_prefix}"
+            help_command += "?" if completion else " ?"
+            result = subprocess.run(
```
deepak-singhal0408, line 268: This calls `subprocess.run(["vtysh", "-c", ...])` directly, but the existing command execution path in `run_bgp_command()` ([bgp_util.py:224](https://github.com/sonic-net/sonic-utilities/blob/master/utilities_common/bgp_util.py#L224)) uses `rvtysh` with `-n <asic_id>` for namespace targeting.

On multi-ASIC platforms, plain `vtysh` only queries the default namespace's FRR instance. On a chassis supervisor with no local FRR, it fails entirely — and since the static metavar was removed, help degrades to bare `[ARGS]...` (worse than today).

This should use `run_bgp_command()` or `rvtysh` with namespace awareness to stay consistent with the command execution path. A static fallback metavar should also be kept for cases where vtysh is unavailable.

show/main.py
```diff
@@ -1508,17 +1510,16 @@ def interfaces(namespace, display):
...
-@click.argument('args', metavar='[IPADDRESS] [vrf <vrf_name>] [...]', nargs=-1, required=False)
+@ipv6.command(cls=vtysh_command("show ipv6 route"))
+@click.argument('args', nargs=-1, required=False)
 @click.option('--display', '-d', 'display', default=None, show_default=False, type=str, help='all|frontend')
 @click.option('--namespace', '-n', 'namespace', default=None, type=str, show_default=False, help='Namespace name or all')
 @click.option('--verbose', is_flag=True, help="Enable verbose output")
 def route(args, namespace, display, verbose):
     """Show IPv6 routing table"""
-    # Call common handler to handle the show ipv6 route cmd
+    # Call common handler to handle the show ip route cmd
```
deepak-singhal0408: Nit: This comment says "show ip route" but this is the ipv6 route handler — should be "show ipv6 route".

tests/vtysh_help_test.py
```diff
@@ -0,0 +1,531 @@
+import os
```
deepak-singhal0408, line 1: The tests cover single-namespace scenarios well. Worth adding test cases for multi-ASIC behavior — e.g., verifying graceful degradation when `vtysh` is unavailable (chassis supervisor scenario) and namespace-specific help queries.

deepak-singhal0408 (commented):
Good feature — dynamic vtysh help and tab completion for `show ip route` / `show ipv6 route` is a solid UX improvement. The `VtyshCommand` + `VtyshParamType` design using Click 8's `shell_complete()` API is clean.

Main issue is the namespace gap — see [inline comment on vtysh_helper.py](https://github.com/sonic-net/sonic-utilities/pull/4123#discussion_r2904266614).

Also, since this introduces a new user-facing behavior, `doc/Command-Reference.md` should be updated:
- Under [`show ip route`](https://github.com/sonic-net/sonic-utilities/blob/master/doc/Command-Reference.md#L6802) and [`show ipv6 route`](https://github.com/sonic-net/sonic-utilities/blob/master/doc/Command-Reference.md#L6932): mention that `--help` now shows dynamic FRR subcommands and TAB completion is supported
- Under [`Help For Show Commands`](https://github.com/sonic-net/sonic-utilities/blob/master/doc/Command-Reference.md#L436): note that certain commands provide enhanced FRR-sourced help

show/vtysh_helper.py
anders-nexthop: Thanks for the catch on the namespace gap.

**On the namespace:** switched the subprocess call from `vtysh` to `rvtysh -n <asic_id>` (asic id read from the Click `--namespace` option; falls back to default namespace for the tab-completion path, which is fine because FRR's command tree is identical per ASIC). I kept our own `subprocess.run` rather than reusing `bgp_util.run_bgp_command` — that helper does `click.echo(output)` on any non-zero return, which would dump vtysh's `?` output into the user's terminal as if it were an error (vtysh routinely exits non-zero on `?` queries that nonetheless produced valid help). Happy to switch if there's an established way around that I missed.

**On the static metavar:** I'd push back on restoring `[IPADDRESS] [vrf <vrf_name>] [...]` specifically — it was always an approximation that didn't mention `bgp`, `connected`, `static`, `babel`, or the other half of FRR's tree. With the dynamic help working, we now have the real list. When vtysh is unreachable (chassis supervisor without local FRR), `show ip route` itself doesn't work, so the `-h` regression from the hardcoded metavar to Click's default `[ARGS]...` only bites in a scenario where the command is dead either way; tab completion is also dead in that scenario regardless of metavar.

show/main.py
anders-nexthop: Fixed.

tests/vtysh_help_test.py
anders-nexthop: Added four new test cases:

- `test_help_with_vtysh_unavailable` — chassis-supervisor case (rvtysh returns nothing → help renders without crashing, no Commands section).
- `test_namespace_passed_through_to_rvtysh` — on multi-ASIC, `--namespace asic0` produces `rvtysh -n 0`.
- `test_namespace_ignored_on_single_asic` — on single-ASIC, `--namespace` is silently ignored for the help path (SONiC's host `vtysh` wrapper doesn't consume `-n` on single-ASIC, so passing it would yield empty help).
- `test_no_namespace_omits_n_flag` — locks in that the default path stays bare `rvtysh`.

anders-nexthop:
Thanks for the review! Addressed all four items in the latest push:

- **Namespace gap**: switched to `rvtysh -n <asic_id>` — namespace threaded from the Click `--namespace` option. See inline reply on `vtysh_helper.py`.
- **Typo in `show/main.py`**: fixed.
- **Tests**: added four new cases — chassis-supervisor, multi-ASIC namespace pass-through, single-ASIC namespace-ignored, and default-no-namespace.
- **Docs**: updated `Command-Reference.md` under `Help For Show Commands`, `show ip route`, and `show ipv6 route` to note the dynamic FRR-sourced help and TAB completion.

Single push-back on restoring the static `[IPADDRESS] [vrf <vrf_name>] [...]` metavar — reasoning in the inline reply. Let me know what you think.

*— anders-bot (AI-assisted, on behalf of @anders-nexthop)*

show/vtysh_helper.py
deepak-singhal0408: ack. thanks,

deepak-singhal0408 (dismissed):

anders-nexthop:
@deepak-singhal0408 I pushed one more commit to fix some test cases that had `vtysh` hardcoded instead of `rvtysh`. CI/CD is passing now but I need another approval.

deepak-singhal0408 (approved):

anders-nexthop:
Thanks @deepak-singhal0408 , I have merged this. This change was for `show ip route` only, but the approach could be used with any vtysh command. Is there desire to extend this feature to other commands? If so I can take that effort up next.

deepak-singhal0408:
Thanks for the offer @anders-nexthop ! `show ip bgp` seems like another high-value target — the current SONiC CLI `show ip bgp summary` doesn't expose received route counts, and operators frequently need to discover subcommands like `community`, `regexp`, `neighbors <IP> received-routes` that aren't surfaced today.

That said, I'd suggest letting this `show ip route` change soak for a bit first — see how it gets consumed, whether any issues surface, and let feature asks come in organically before expanding scope. Over-exposing vtysh commands could have side effects (users relying on FRR-specific options that change across versions, or discovering commands that interact poorly with SONiC's state model).

cc: @yxieca and @lguohan as well for viz

anders-nexthop:
> Thanks for the offer @anders-nexthop ! `show ip bgp` seems like another high-value target — the current SONiC CLI `show ip bgp summary` doesn't expose received route counts, and operators frequently need to discover subcommands like `community`, `regexp`, `neighbors <IP> received-routes` that aren't surfaced today.
> 
> That said, I'd suggest letting this `show ip route` change soak for a bit first — see how it gets consumed, whether any issues surface, and let feature asks come in organically before expanding scope. Over-exposing vtysh commands could have side effects (users relying on FRR-specific options that change across versions, or discovering commands that interact poorly with SONiC's state model).
> 
> cc: @yxieca and @lguohan as well for viz

Sure, makes sense. Thanks for the review!

PR #4127, opened 2025-11-19
Fixing MSTP global command enable issue

divyachandralekha:
What I did
Added support to display MSTP output for "show spanning-tree" command 
Fixed the issue on MSTP global enable command issue 
Added testcases to cover mstp display command changes


vganesan-nokia:
@divyachandralekha, currently all the commands starting with "config spanning-tree mst instance" expects an entry for given instance already exists in STP_MST_INST table in config db. If an entry for the given instance does not already exist in config db, there is failure. 
We would like to suggest to have a command to newly create an entry in "STP_MST_INST" table. We can use the command "config spanning-tree mst instance priority <instance id> <priority>" to create an entry in STP_MST_INST table if one does not exist already before setting the priority.

show/stp.py
```diff
@@ -401,3 +408,321 @@ def show_stp_vlan_statistics(ctx, vlanid):
...
+    if not mst_entry:
+        return
+    
+    mst_global = get_mst_global_info()
+    
+    # Print instance header
+    vlan_str = mst_entry.get('vlan@', '')
+    vlan_list = format_vlan_list(vlan_str)
+    click.echo("")
+    click.echo("#######  MST{:<8} Vlans mapped : {}".format(mst_id, vlan_list))
```
vganesan-nokia, line 579: If the instance is 0 we can add the word (CIST) as propsed in the HLD

show/stp.py
```diff
@@ -167,19 +167,26 @@ def spanning_tree(ctx):
...
-        for key in keys:
-            result = re.search('.STP_VLAN_TABLE:Vlan(.*)', key)
-            vlanid = result.group(1)
-            vlan_list.append(int(vlanid))
-        vlan_list.sort()
-        for vlanid in vlan_list:
-            ctx.invoke(show_stp_vlan, vlanid=vlanid)
+        # For MSTP mode, use new format
+        if g_stp_mode == 'MSTP':
+            ctx.invoke(show_mstp_summary)
```
vganesan-nokia, line 176: Suggestion for alternate approach: 
If we can do this similar to how it is done for PVST we can show spanning tree for all MST instances and also for a specific instance (for "show spanning-tree mst instance <instance id> command)  using the same function show_mst_instnace_detail().

show/stp.py
```diff
@@ -401,3 +408,321 @@ def show_stp_vlan_statistics(ctx, vlanid):
...
+    
+    # Display each MST instance
+    for mst_id in mst_list:
+        show_mst_instance_detail(mst_id)
+    
+    # Display MST region information at the end
+    show_mst_region_info()
+
+
+def show_mst_instance_detail(mst_id):
```
vganesan-nokia, line 566: When I compare the implemented show output to the show information in HLD, I see that the information shown here is for the summary. i..e, the interface information is shown in single row for each interface. Should we rename this function to "show_mst_instance_summary()" to make the name approriately mean what is done in the function?

show/stp.py
```diff
@@ -401,3 +408,321 @@ def show_stp_vlan_statistics(ctx, vlanid):
...
+        
+        # Sort by numeric part
+        eth_list.sort(key=lambda x: int(re.search(r'\d+', x).group()))
+        po_list.sort(key=lambda x: int(re.search(r'\d+', x).group()))
+        
+        for ifname in eth_list + po_list:
+            show_mst_port_info(mst_id, ifname)
+
+
+def show_mst_port_info(mst_id, ifname):
```
vganesan-nokia, line 656: We may need another function like "show_mst_port_info_detail()" which can be used for showing MST instance detail where the each port info is displayed in deatil (as per "show spanning-tree mst detail" info in the HLD)

show/stp.py
```diff
@@ -401,3 +408,321 @@ def show_stp_vlan_statistics(ctx, vlanid):
...
+    
+    # Bridge timers
+    click.echo("Bridge Timers                   : MaxAge {}s Hello {}s FwdDly {}s MaxHops {}".format(
+        mst_global['max_age'], mst_global['hello_time'], 
+        mst_global['forward_delay'], mst_global['max_hops']))
+    
+    # CIST Root timers (typically same as bridge timers)
+    click.echo("CIST Root Timers                : MaxAge {}s Hello {}s FwdDly {}s MaxHops {}".format(
+        mst_global['max_age'], mst_global['hello_time'], 
+        mst_global['forward_delay'], mst_global['max_hops']))
```
vganesan-nokia, line 717: Recommendation: It is better to implement the "show spanning-tree statistics [mst <instance id>]" as part of this PR. We already have some statistics (such as bpd rx, bpdu tx and num fwd transitions) in the STP_MST_PORT_TABLE entries. We can also implement "show spanning-tree mst interface <interface>" as part of the PR to meet all the show commands specified in HLD

tests/test_config_mstp.py
```diff
@@ -954,7 +954,7 @@ def test_mstp_interface_edge_port_invalid_state(self, mock_db):
 
             assert result.exit_code != 0
             assert "Invalid value" in result.output
-            assert "is not one of 'enable', 'disable'" in result.output
+            assert "choose from enable, disable" in result.output
```
vganesan-nokia: A generic commnent not for this line. 
We need unit test for the show commands impelemented in show/stp.py.

show/stp.py
divyachandralekha: it seems the MSTP show commands are missed during implementation.
Currently I have enhanced the existing show command to verify the MSTP. Rest all show MSTP related changes will add in next PR.

show/stp.py
divyachandralekha: I will take care in next PR.

show/stp.py
divyachandralekha: I will take care in next PR.

show/stp.py
divyachandralekha: I will take care in next PR.

show/stp.py
divyachandralekha: I will take care in next PR.

vganesan-nokia (approved):
Thanks for respoinding to the review comments. Looking forward to the next PR.

divyachandralekha:
@qiluo-msft  , Please merge the changes

satyprak-nokia:
looking to integrate this PR. When can we expect this to be merged?

PR #4129, opened 2025-11-21
Changes for DB migration to release 202511

vmittal-msft:
#### What I did
Updated DB migration for release 202511

#### How I did it
Updated the db_migration.py script to include release 202511

### DIFF
--- a/scripts/db_migrator.py
+++ b/scripts/db_migrator.py
@@ -58,7 +58,7 @@ class DBMigrator():
                      none-zero values.
               build: sequentially increase within a minor version domain.
         """
-        self.CURRENT_VERSION = 'version_202511_01'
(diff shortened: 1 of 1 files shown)


scripts/db_migrator.py
```diff
@@ -58,7 +58,7 @@ def __init__(self, namespace, socket=None):
                      none-zero values.
               build: sequentially increase within a minor version domain.
         """
-        self.CURRENT_VERSION = 'version_202505_01'
+        self.CURRENT_VERSION = 'version_202605_01'
```
vaibhavhd: Shouldn't this be `version_202511_01`

yxieca: 202605 is the correct one :-)

PR #4131, opened 2025-11-23
[Mellanox] Fix generate_dump sysfs copy to copy only files with permission

noaOrMlnx:
#### What I did
Fix generate dump script to only copy valid files: 
- If the module is not present, copy only presence files. 
error: ERR kernel: sxd_kernel: [error] show_module_frequency: __sx_core_get_module_frequency failed (-14) for module64
- If frequency_support value is 0, don't copy frequency sysfs
error: ERR kernel: [  858.655221] sxd_kernel: [error] __sx_core_get_module_frequency: Frequency read is not supported.
#### How I did it
check the above files content before copying all
#### How to verify it
run 'show techsupport' on Nvidia switch and validate output

### DIFF
--- a/scripts/generate_dump
+++ b/scripts/generate_dump
@@ -932,7 +932,7 @@ save_sdk_sysfs() {
         return 1
     }
 
-    copy_file() {
+    copy_file_content() {
         local f="$1"
         local target="$2"
         if $NOOP; then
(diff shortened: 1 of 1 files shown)


noaOrMlnx:
@vmittal-msft can you please help cherry-pick this PR to 202511?

PR #4132, opened 2025-11-24
[mellanox] [db_migrator] add a migration for tunnel ecn mode

Yakiv-Huryk:
#### What I did
In the 202511, the tunnel ECN mode for Mellanox platforms has changed from standard to copy_from_outer.
This commit adds a migration that aligns APPL_DB to this change.

#### How I did it
Added a new migration for the Mellanox platform.

#### How to verify it
New unit test, manual tests.

### DIFF
--- a/scripts/db_migrator.py
+++ b/scripts/db_migrator.py
@@ -935,6 +935,36 @@ class DBMigrator():
             except Exception as e:
                 log.log_error(f"Failed to migrate DHCP servers for {vlan_key}: {str(e)}")
 
+    def migrate_ipinip_tunnel_ecn_mode_mellanox(self):
+        """
+        Migrate IPINIP_TUNNEL ecn mode to copy_from_outer.
+
+        Prior to 202511, on Mellanox platform, the ecn mode of IPINIP tunnel was set to standard.
+        """
+
+        table_name = 'TUNNEL_DECAP_TABLE'
+        keys_to_migrate = [
+            'IPINIP_SUBNET',
+            'IPINIP_TUNNEL',
+            'IPINIP_SUBNET_V6',
+            'IPINIP_V6_TUNNEL'
+        ]
+
+        for k, v in self.appDB.get_table(table_name).items():
+            if k in keys_to_migrate:
+                log.log_info(f"Migrating APPL_DB {k}, ecn mode: {v.get('ecn_mode')} -> copy_from_outer")
+                v['ecn_mode'] = 'copy_from_outer'
+                self.appDB.set_entry(table_name, k, v)
+
+        state_db_keys = self.stateDB.keys(self.stateDB.STATE_DB, f"{table_name}:*")
+        if not state_db_keys:
+            return
+        for key in keys_to_migrate:
+            state_db_key = f"{table_name}:{key}"
+            if state_db_key in state_db_keys:
+                ecn_mode = self.stateDB.get(self.stateDB.STATE_DB, state_db_key, 'ecn_mode')
+                log.log_info(f"Migrating STATE_DB {state_db_key}, ecn mode: {ecn_mode} -> copy_from_outer")
+                self.stateDB.set(self.stateDB.STATE_DB, state_db_key, 'ecn_mode', 'copy_from_outer')
 
     def version_unknown(self):
         """
@@ -1396,6 +1426,8 @@ class DBMigrator():
         Version 202511_01
         """
         log.log_info('Handling version_202511_01')
+        if self.asic_type == "mellanox":
+            self.migrate_ipinip_tunnel_ecn_mode_mellanox()
         self.set_version('version_202605_01')
         return 'version_202605_01'
 
--- a/scripts/db_migrator.py
+++ b/scripts/db_migrator.py
@@ -956,11 +956,11 @@ class DBMigrator():
                 v['ecn_mode'] = 'copy_from_outer'
                 self.appDB.set_entry(table_name, k, v)
 
-        state_db_keys = self.stateDB.keys(self.stateDB.STATE_DB, f"{table_name}:*")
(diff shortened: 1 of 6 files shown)


scripts/db_migrator.py
```diff
@@ -935,6 +935,27 @@ def migrate_dhcp_servers_to_dhcpv4_relay(self):
...
+            'IPINIP_SUBNET_V6',
+            'IPINIP_V6_TUNNEL'
+        ]
+
+        table = self.appDB.get_table(table_name)
+        for k, v in table.items():
+            if k in keys_to_migrate:
+                log.log_info(f"Migrating {k}, ecn mode: {v.get('ecn_mode')} -> copy_from_outer")
+                v['ecn_mode'] = 'copy_from_outer'
+                self.appDB.set_entry(table_name, k, v)
```
developfast, line 957: setting the entry in the app db is enough? Will asic catch this?

Yakiv-Huryk: Yes, setting it into APPL_DB is enough, it will be configured in the _OrchDaemon::warmRestoreAndSyncUp()_

Also, ASIC db is flushed, so it's not relevant: https://github.com/sonic-net/sonic-utilities/blob/67232b63ac6486a7f320fd23b6e85245ad938b61/scripts/fast-reboot#L257

Regarding STATE db, I'm not sure I understand why it's relevant for this fix.

dgsudharsan (approved):

scripts/db_migrator.py
kperumalbfn, line 957: Does this trigger set attribute SAI API?

vmittal-msft:
@Yakiv-Huryk  please help raise pr for 202511

scripts/db_migrator.py
developfast: @Yakiv-Huryk To clarify, the ask here is that values across STATE, ASIC and APPL DB should be synced before and after warm-boot i.e. here after warm-boot `ecn_mode` is changing to `standard` after warm-boot, therefore after a warm-boot you are required to ensure that the `ecn_mode` value is `standard` across all 3 of these DBs. We have a db consistency check that we run after warm-boot that will fail if these DBs are not synced.

Yakiv-Huryk: This doesn't trigger the set attribute API, it changes the parameter of a tunnel that will be created during warm boot restore:
`|c|SAI_OBJECT_TYPE_TUNNEL:oid:0x2a000000000b61|
SAI_TUNNEL_ATTR_TYPE=SAI_TUNNEL_TYPE_IPINIP|SAI_TUNNEL_ATTR_OVERLAY_INTERFACE=oid:0x6000000000b60|
SAI_TUNNEL_ATTR_UNDERLAY_INTERFACE=oid:0x600000000000a|
SAI_TUNNEL_ATTR_DECAP_ECN_MODE=SAI_TUNNEL_DECAP_ECN_MODE_COPY_FROM_OUTER|
SAI_TUNNEL_ATTR_DECAP_TTL_MODE=SAI_TUNNEL_TTL_MODE_PIPE_MODEL|
SAI_TUNNEL_ATTR_DECAP_DSCP_MODE=SAI_TUNNEL_DSCP_MODE_UNIFORM_MODEL`

The set attribute flow will happen as a part of ipinip.json processing (in _apply_ipinip_subset()_). However, by that time, the tunnel will already be created with a new ecn mode.

Yakiv-Huryk: @developfast, I've added a migration for the STATE DB. Regarding the ASIC DB, since it's flushed before migration (and will be filled after), there is nothing to migrate. Please correct me if I'm wrong.

Yakiv-Huryk: @kperumalbfn, kind reminder to check this

kperumalbfn: Thanks @Yakiv-Huryk, makes sense with the above sairedis log.

kperumalbfn: What will happen if bulk routine is used for the tunnel obect?

Yakiv-Huryk: Since we update the APPL DB config, I don't see how bulk would cause any difference. Just to remind, the ASIC DB is not migrated, we flush it before the warmboot: https://github.com/sonic-net/sonic-utilities/blob/67232b63ac6486a7f320fd23b6e85245ad938b61/scripts/fast-reboot#L257

kperumalbfn (dismissed):

developfast (dismissed):
lgtm, thanks for the output and changes @Yakiv-Huryk

scripts/db_migrator.py
```diff
@@ -935,6 +935,36 @@ def migrate_dhcp_servers_to_dhcpv4_relay(self):
...
+            'IPINIP_V6_TUNNEL'
+        ]
+
+        for k, v in self.appDB.get_table(table_name).items():
+            if k in keys_to_migrate:
+                log.log_info(f"Migrating APPL_DB {k}, ecn mode: {v.get('ecn_mode')} -> copy_from_outer")
+                v['ecn_mode'] = 'copy_from_outer'
+                self.appDB.set_entry(table_name, k, v)
+
+        state_db_keys = self.stateDB.keys(self.stateDB.STATE_DB, f"{table_name}:*")
```
qiluo-msft: >: [](http://example.com/codeflow?start=79&length=1)

STATE_DB is using `|` instead of `:` as separator. Could you double check and add test coverage for this line? #Closed

Yakiv-Huryk: Thanks, fixed: https://github.com/sonic-net/sonic-utilities/pull/4132/changes/82f2382ba81ba446b330506491cb66048ee5e728

developfast (approved):
lgtm

qiluo-msft (approved):

PR #4137, opened 2025-11-27
Add nexthop group usage check in route_check

xwjiang-ms:
#### What I did
Previously, due to a bug in Broadcom SAI, the system incorrectly created next-hop groups with a size of 16.
This resulted in excessive next-hop group consumption, which eventually limited the number of available ECMP routes and caused traffic impact.
To prevent similar issues from going unnoticed, I added a next-hop group usage check to route_check.
If the next-hop group usage exceeds 80%, the script will report an error.
#### How I did it
Get CRM stats from counters DB, and get nexthop usage, then compare the usage with threshold, report error if usage exceeded threshold.
Fixed some format error.
#### How to verify it
Verified on lab device.

### DIFF
--- a/scripts/route_check.py
+++ b/scripts/route_check.py
@@ -213,8 +213,8 @@ def diff_sorted_lists(t1, t2):
     t1_x = t2_x = 0
     t1_miss = []
     t2_miss = []
-    t1_len = len(t1)
-    t2_len = len(t2)
+    t1_len = len(t1);
+    t2_len = len(t2);
     while t1_x < t1_len and t2_x < t2_len:
         d = cmps(t1[t1_x], t2[t2_x])
         if (d == 0):


StormLiangMS:
@xwjiang-ms how this trigger an alert? By syslog err?

StormLiangMS:
could you also check the UT failures?

StormLiangMS:
@xwjiang-ms could you check the PR failures?

scripts/route_check.py
```diff
@@ -209,8 +213,8 @@ def diff_sorted_lists(t1, t2):
     t1_x = t2_x = 0
     t1_miss = []
     t2_miss = []
-    t1_len = len(t1);
-    t2_len = len(t2);
+    t1_len = len(t1)
```
StormLiangMS: @xwjiang-ms why touch this?

xwjiang-ms: Understood, this PR shouldn't contain the change

StormLiangMS (approved):
LGTM

prsunny:
How is this different from CRM monitoring reporting "threshold exceeded" error thats also spewed periodically?

StormLiangMS:
> How is this different from CRM monitoring reporting "threshold exceeded" error thats also spewed periodically?

Good point, then the question why we don't have the alert in prod when hit this issue? @xwjiang-ms  could you take a look at the CRM alert for the threthold alert? If any gap between SONiC and net assist?

PR #4142, opened 2025-11-28
feat: validate crm high and low threshold input

JaviEst:
#### What I did

Resolved issue https://github.com/sonic-net/sonic-utilities/issues/2931 by validating the high and low threshold input values.

#### How I did it

Updated the high and low threshold config functions to validate both thresholds are between 0 and 100 for percentage type as well as validate the low threshold if smaller than the high threshold.

#### How to verify it

Run new implemented tests to verify the validation logic is enforced.

### DIFF
--- a/crm/main.py
+++ b/crm/main.py
@@ -428,13 +428,13 @@ def low(ctx, value):
     if threshold_type == 'percentage':
         if value < 0 or value > 100:
             raise click.ClickException(
-                'Error: Low threshold value must be between 0 and 100 for '
+                'Low threshold value must be between 0 and 100 for '
                 'percentage type.')
 
         high_value = crm_info.get(high_attr, None)
         if high_value is not None and value >= int(high_value):
             raise click.ClickException(
-                'Error: Low threshold value must be less than high threshold '
+                'Low threshold value must be less than high threshold '
                 f'value: {high_value} for percentage type.')
 
     attr = ''
@@ -464,13 +464,13 @@ def high(ctx, value):
     if threshold_type == 'percentage':
         if value < 0 or value > 100:
             raise click.ClickException(
-                'Error: High threshold value must be between 0 and 100 for '
+                'High threshold value must be between 0 and 100 for '
                 'percentage type.')
 
         low_value = crm_info.get(low_attr, None)
         if low_value is not None and value <= int(low_value):
             raise click.ClickException(
-                'Error: High threshold value must be greater than low threshold '
+                'High threshold value must be greater than low threshold '
                 f'value: {low_value} for percentage type.')
 
     attr = ''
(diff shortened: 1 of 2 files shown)


JaviEst:
Hoping I can get some reviews please @qiluo-msft @mssonicbld @yxieca

crm/main.py
```diff
@@ -418,27 +412,71 @@ def type(ctx, value):
...
 
-    attr += ctx.obj["crm"].res_type + '_' + 'low_threshold'
+    high_attr = ''
+    if ctx.obj["crm"].addr_family != None:
+        high_attr += ctx.obj["crm"].addr_family + '_'
+    high_attr += ctx.obj["crm"].res_type + '_' + 'high_threshold'
+
+    crm_info = ctx.obj["crm"].get_crm_config()
+    threshold_type = crm_info.get(type_attr, 'percentage')
+    if threshold_type == 'percentage':
```
anders-nexthop, line 428: this `type` logic wasn't there in the original version. The yang model looks for [crm_threshold_type](https://github.com/sonic-net/sonic-buildimage/blob/0facff54617dd434577b8e5c452efeb33452dbf8/src/sonic-yang-models/yang-templates/sonic-types.yang.j2#L117C1-L122C6) looks like this:
```
    typedef crm_threshold_type {
        type string {
            length 1..64;
            pattern "percentage|used|free|PERCENTAGE|USED|FREE";
        }
    }
```
It looks like your change restricts the type check to only be done when the value is `percentage`, which seems like the right thing to do. But what *should* we be doing in the `used`/`free` case? are those not also percents? maybe we should be doing this check regardless of type?

anders-nexthop, line 428: you need to accept `PERCENTAGE` as well, as per the [yang-model](https://github.com/sonic-net/sonic-buildimage/blob/0facff54617dd434577b8e5c452efeb33452dbf8/src/sonic-yang-models/yang-templates/sonic-types.yang.j2#L120)

crm/main.py
```diff
@@ -418,27 +412,71 @@ def type(ctx, value):
...
+        if value < 0 or value > 100:
+            raise click.ClickException(
+                'Low threshold value must be between 0 and 100 for '
+                'percentage type.')
+
+        high_value = crm_info.get(high_attr, None)
```
anders-nexthop, line 434: if you're checking the type for the low value, shouldn't you check it for the high value as well? or is it assumed that if the low value is a percentage, the high value must be as well? if so, is that assumption captured in the yang-model somewhere?

tests/crm_test.py
```diff
@@ -1922,6 +1922,108 @@ def test_crm_multi_asic_show_resources_srv6_nexthop(self):
...
+            obj=db,
+        )
+        print(sys.stderr, result.output)
+        assert result.exit_code == 1
+        assert (
+            "Error: High threshold value must be greater than low threshold value: "
+            f"{low_value} for percentage type." in result.output
+        )
+
+    def test_crm_config_thresholds_valid_values(self):
```
anders-nexthop, line 2004: either here or in another test case, you should check the  corner cases, where thresholds are 0/100, and when they are equal (so for low/high threshold, you could check 0/100, 0/0, 100/100)

PR #4154, opened 2025-12-11
improve unit test coverage for fast-reboot-dump script

maipbui:
for hackathon week
#### What I did
enhances unit test coverage for scripts/fast-reboot-dump.py
- generate_default_route_entries
- generate_media_config
- generate_fdb_entries
- generate_neighbor_entries

This partially fixes https://github.com/sonic-net/sonic-utilities/issues/1175 by increasing coverage of the fast-reboot-dump, coverage improving from 48.5% to 70%
#### How to verify it
coverage after this unit test
```
scripts/fast-reboot-dump.py                              270     71     82     12    70%
```

### DIFF
--- a/tests/fast_reboot_dump_test.py
+++ b/tests/fast_reboot_dump_test.py
@@ -48,9 +48,12 @@ class TestFastRebootDump(object):
 
         conn.connect(conn.APPL_DB)
 
+        # Included: (Vlan2, mac) is present in all_available_macs
         key_included = 'NEIGH_TABLE:Vlan2:192.168.0.2'
         conn.set(conn.APPL_DB, key_included, 'neigh', '52:54:00:5D:FC:B7')
 
+        # Excluded: exists in DB, but (Vlan3, mac) is NOT present in all_available_macs
+        # so generate_neighbor_entries() skip it
         key_excluded = 'NEIGH_TABLE:Vlan3:192.168.0.3'
         conn.set(conn.APPL_DB, key_excluded, 'neigh', 'aa:bb:cc:dd:ee:ff')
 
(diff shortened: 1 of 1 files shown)


tests/fast_reboot_dump_test.py
```diff
@@ -36,7 +39,124 @@ def test_generate_fdb_entries_vlan_portcahnnel_member(self):
...
+    def test_generate_neighbor_entries(self, mock_sonicv2, _mock_syslog):
+        conn = dbconnector.SonicV2Connector()
+        mock_sonicv2.return_value = conn
+
+        conn.connect(conn.APPL_DB)
+
+        key_included = 'NEIGH_TABLE:Vlan2:192.168.0.2'
+        conn.set(conn.APPL_DB, key_included, 'neigh', '52:54:00:5D:FC:B7')
+
+        key_excluded = 'NEIGH_TABLE:Vlan3:192.168.0.3'
```
tahmed-dev, line 57: what role does `key_excluded` play in this test? I do not see we are testing for it not being in the generated neighbor list. Also, a little hint on why it is excluded would help here.

maipbui: It's for negative test case; I've updated the test to confirm that entry is skipped and doesn’t show up in the output.

tests/fast_reboot_dump_test.py
```diff
@@ -36,7 +39,128 @@ def test_generate_fdb_entries_vlan_portcahnnel_member(self):
...
+        assert 'speed' not in attrs
+
+        file_entry = next(item for item in file_data if key in item)
+        assert file_entry[key] == attrs
+        assert entry['OP'] == 'SET'
+        assert file_entry['OP'] == 'SET'
+
+    @mock.patch.object(fast_reboot_dump, "get_vlan_ifaces", return_value=[])
+    @mock.patch.object(fast_reboot_dump, "SonicV2Connector")
+    def test_generate_fdb_entries(self, mock_sonicv2, _mock_get_vlan_ifaces):
```
tahmed-dev, line 146: it would be good if we could parameterize this test and test for conditions like:
1. empty fdp table
2. combination of IPv4 and IPv6 fdb entries

maipbui: discussed offline, non-blocking comment, will address the comment later

qiluo-msft (approved):

PR #4155, opened 2025-12-12
[chassis] Fixed a bug in "show fabric isolation" output

saksarav-nokia:
#### What I did
https://github.com/sonic-net/sonic-buildimage/issues/24829
Fixed a bug in the "show fabric isolation" command output. 
When the fabric port sees CRC/FEC-Uncorrectable error, the Fabric monitor feature Isolates the port and sets the ISOALTED=1 and also AUTO_ISOLATED=1 in the STATE_DB for that port in the FABRIC_PORT_TABLE. The field ISOLATED and CONFIG_ISOLATED are always present for all the fabric ports in FABRIC_PORT_TABLE. However the field AUTO_ISOLATED is not present always and added only when the port is auto ISOLATED.
Due to the bug in the FabricIsolation cli script, the  Auto Isolated is shown 1 for all the ports printed in the show command after the port which is actually isolated. In the output shown below, the port 165 is isolated, however Auto-isolated is shown 1 for all the ports after 165 which is wrong.
image

#### How I did it
Initialized the variable correctly inside the loop.
#### How to verify it
Induced the CRC error for one of the port and after that port is isolated, verified the "show fabric isolation" output shows the correct output.

image

### DIFF
--- a/scripts/fabricstat
+++ b/scripts/fabricstat
@@ -381,11 +381,11 @@ class FabricIsolation(FabricStat):
                 port_dict.update({port_number: port_data})
         # Create ordered table of fabric ports.
         header = ["Local Link", "Auto Isolated", "Manual Isolated", "Isolated"]
-        auto_isolated = 0
-        manual_isolated = 0
-        isolated = 0
         body = []
         for port_number in sorted(port_dict.keys()):
+            auto_isolated = 0
+            manual_isolated = 0
+            isolated = 0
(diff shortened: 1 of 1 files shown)


saksarav-nokia:
@kenneth-arista @ysmanman @arlakshm @vmittal-msft @judyjoseph for viz

saksarav-nokia:
@jfeng-arista @arlakshm please review this

PR #4163, opened 2025-12-18
Update bash completions for sonic-utilities commands

saiarcot895:
#### What I did

Update the bash completion files for all sonic-utilities commands to make them compatible with the current Click version.

Fixes sonic-net/sonic-buildimage#24594.

#### How I did it

Use [Click's documentation](https://click.palletsprojects.com/en/stable/shell-completion/) to generate the bash completion script for each command that is packaged from sonic-utilities and uses Click.

#### How to verify it

Tested in KVM in Trixie image.

```
admin@vlab-01:~$ sonic-package-manager
install     list        manifests   migrate     repository  reset       show        uninstall   update
admin@vlab-01:~$ sonic-package-manager
install     list        manifests   migrate     repository  reset       show        uninstall   update
admin@vlab-01:~$ sonic-package-manager
install     list        manifests   migrate     repository  reset       show        uninstall   update
admin@vlab-01:~$ spm
install     list        manifests   migrate     repository  reset       show        uninstall   update
admin@vlab-01:~$ spm ^C
admin@vlab-01:~$ show
Display all 105 possibilities? (y or n)
aaa                       buffer_pool               environment               icmp                      macsec                    passw-hardening           runningconfiguration      suppress-fib-pending      vlan
acl                       chassis                   event-counters            interfaces                management_interface      pbh                       serial_console            switch                    vnet
arp                       clock                     fabric                    ip                        mgmt-vrf                  pfc                       services                  switch-hash               vrf
asic-sdk-health-event     copp                      feature                   ipv6                      mirror_session            pfcwd                     sflow                     switch-trimming           vrrp
auto-techsupport          dhcp4relay-counters       fg-nhg                    kdump                     mmu                       platform                  snmpagentaddress          syslog                    vrrp6
auto-techsupport-feature  dhcp6relay_counters       fg-nhg-member             kubernetes                muxcable                  policer                   snmptrap                  system-health             vxlan
banner                    dhcp_relay                fg-nhg-prefix             ldap                      nat                       priority-group            spanning-tree             system-memory             warm_restart
bfd                       dhcp_server               fgnhg                     ldap-server               ndp                       processes                 srv6                      tacacs                    watermark
bgp                       dhcprelay_helper          flowcnt-route             line                      ntp                       queue                     ssh                       techsupport               ztp
bmp                       dns                       flowcnt-trap              lldp                      nvgre-tunnel              radius                    startupconfiguration      uptime
boot                      dropcounters              headroom-pool             logging                   nvgre-tunnel-map          reboot-cause              storm-control             users
buffer                    ecn                       history                   mac                       p4-table                  route-map                 subinterfaces             version
admin@vlab-01:~$ config
aaa                       cbf                       dropcounters              interface_naming_mode     loopback                  nvgre-tunnel-map          reload                    spanning-tree             unique-ip
acl                       chassis                   ecn                       ipv6                      macsec                    override-config-table     replace                   ssh                       vlan
apply-patch               checkpoint                fabric                    kdump                     mclag                     passw-hardening           rollback                  subinterface              vnet
asic-sdk-health-event     clock                     feature                   kubernetes                member                    pbh                       route                     suppress-fib-pending      vrf
auto-techsupport          console                   fg-nhg                    ldap                      mirror_session            pfcwd                     save                      switch-hash               vxlan
auto-techsupport-feature  delete-checkpoint         fg-nhg-member             ldap-server               mmu                       platform                  serial_console            switch-trimming           warm_restart
banner                    dhcp_relay                fg-nhg-prefix             list-checkpoints          muxcable                  portchannel               sflow                     switchport                watermark
bgp                       dhcp_server               flowcnt-route             load                      nat                       qos                       snmp                      synchronous_mode          yang_config_validation
bmp                       dhcpv4_relay              hostname                  load_mgmt_config          ntp                       radius                    snmpagentaddress          syslog                    ztp
buffer                    dns                       interface                 load_minigraph            nvgre-tunnel              rate                      snmptrap                  tacac
```

Note that these commands don't have a completion script generated, likely because an exception is being raised when just importing that module:

```
Cannot generate completion for counterpoll.main:cli!
Cannot generate completion for debug.main:cli!
Cannot generate completion for fwutil.main:cli!
Cannot generate completion for psuutil.main:cli!
Cannot generate completion for sfputil.main:cli!
Cannot generate completion for undebug.main:cli!
```

### DIFF
--- a/tests/watchdogutil_test.py
+++ b/tests/watchdogutil_test.py
@@ -1,0 +1,26 @@
+from click.testing import CliRunner
+import watchdogutil.main as watchdogutil
+from unittest.mock import patch, MagicMock
+
+
+class TestWatchdog(object):
+    @patch('os.geteuid', MagicMock(return_value=1000))
+    def test_non_root_fails(self):
+        runner = CliRunner()
+        result = runner.invoke(watchdogutil.watchdogutil, ["version"])
+        assert result.exit_code != 0
+        assert "Root privileges are required" in result.output
+
+    @patch('os.geteuid', MagicMock(return_value=0))
+    def test_import_fails(self):
+        runner = CliRunner()
+        result = runner.invoke(watchdogutil.watchdogutil, ["version"])
+        assert result.exit_code != 0
+        assert result.exception
+
+    def test_version(self):
+        runner = CliRunner()
+        result = runner.invoke(watchdogutil.watchdogutil.commands["version"],
+                               [])
+        assert result.exit_code == 0
+        assert watchdogutil.VERSION in result.output


anders-nexthop (approved):
Looks okay to me but I agree with @qiluo-msft that having a way to generate these would be desirable. It would be pretty easy to miss adding the completion script for new commands.

show/main.py
```diff
@@ -269,14 +269,17 @@ def is_gearbox_configured():
...
-    if any(re.match(GEARBOX_TABLE_PHY_PATTERN, key) for key in keys):
-        return True
-    else:
+        # If any _GEARBOX_TABLE:phy:* records present in APPL_DB, then the gearbox is configured
+        if any(re.match(GEARBOX_TABLE_PHY_PATTERN, key) for key in keys):
+            return True
+        else:
+            return False
+    except RuntimeError:
         return False
```
qiluo-msft, line 283: >False [](http://example.com/codeflow?start=15&length=5)

returning false if connecting redis failed seems dangerous.

saiarcot895: This function is to determine if the gearbox is configured (which will then impact which CLIs are available). If we can't determine that (because, say, we can't connect to redis), then `False` seems like a safe default?

saiarcot895: Looks like it's _only_ used to determine if the CLI should be made visible or not, so then it becomes a question of whether or not the gearbox CLI shows up if it cannot be determined if it is enabled or not.

sonic-utilities-data/debian/install
mlok-nokia, line 1: This change seems cause build image failed

PR #4167, opened 2025-12-19
[202511] [mellanox] [db_migrator] add a migration for tunnel ecn mode (#4132)

Yakiv-Huryk:
Cherry-pick https://github.com/sonic-net/sonic-utilities/pull/4132

What I did
In the 202511, the tunnel ECN mode for Mellanox platforms has changed from standard to copy_from_outer. This commit adds a migration that aligns APPL_DB to this change.

How I did it
Added a new migration for the Mellanox platform.

How to verify it
New unit test, manual tests.

### DIFF
--- a/scripts/db_migrator.py
+++ b/scripts/db_migrator.py
@@ -902,6 +902,37 @@ class DBMigrator():
             if keys:
                 self.configDB.delete(self.configDB.CONFIG_DB, authorization_key)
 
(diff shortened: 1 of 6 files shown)


developfast:
> 4 BGP unit tests are failing. It's not related to the changes, and they are also failing on the other 202511 PR.

understood, i think most of the team is ooo so this will be looked at probably in the new year. have a great holiday!

PR #2145, opened 2022-04-28
[GCU] Validate peer_group_range ip_range are correct

ghooo:
#### What I did

Fixes #2119

#### Previous command output (if the output of a command-line utility has changed)
Sorting output:
```
Patch Applier: Applying 4 changes in order:
Patch Applier:   * [{"op": "remove", "path": "/BGP_PEER_RANGE/BGPSLBPassiveV6"}]
Patch Applier:   * [{"op": "add", "path": "/BGP_PEER_RANGE/BGPVac", "value": {"ip_range": ["192.168.0.0/21"], "name": "BGPVac", "src_address": "10.1.0.32"}}]
Patch Applier:   * [{"op": "remove", "path": "/BGP_PEER_RANGE/BGPSLBPassive"}]
Patch Applier:   * [{"op": "add", "path": "/BGP_PEER_RANGE/BGPSLBPassive", "value": {"ip_range": ["10.255.0.0/25"], "name": "BGPSLBPassive", "src_address": "10.1.0.32"}}]

```
#### New command output (if the output of a command-line utility has changed)

Sorting output:
```
Patch Applier: Applying 4 changes in order:
Patch Applier:   * [{"op": "remove", "path": "/BGP_PEER_RANGE/BGPSLBPassiveV6"}]
Patch Applier:   * [{"op": "remove", "path": "/BGP_PEER_RANGE"}]
Patch Applier:   * [{"op": "add", "path": "/BGP_PEER_RANGE", "value": {"BGPSLBPassive": {"ip_range": ["10.255.0.0/25"], "name": "BGPSLBPassive", "src_address": "10.1.0.32"}}}]
Patch Applier:   * [{"op": "add", "path": "/BGP_PEER_RANGE/BGPVac", "value": {"ip_range": ["192.168.0.0/21"], "name": "BGPVac", "src_address": "10.1.0.32"}}]
```

PR #4179, opened 2026-01-06
Skip IP range duplicate check in validate_bgp_peer_group for different vnets

alawing:
#### What I did
Skip IP range duplicate check in `validate_bgp_peer_group` for different vnets
Related issue: #2145

#### How I did it
Check duplicates only when both `vnet_name` and `ip_range` match

#### How to verify it
testbed tests and unit tests

### DIFF
--- a/tests/generic_config_updater/gu_common_test.py
+++ b/tests/generic_config_updater/gu_common_test.py
@@ -254,6 +254,9 @@ class TestConfigWrapper(unittest.TestCase):
                 },
                 "VnetB|WLPARTNER_PASSIVE_V4": {
                     "ip_range": ["1.1.1.1/31", "10.10.10.10/16", "100.100.100.100/24"]
+                },
+                "WLPARTNER_PASSIVE_V4": {
+                    "ip_range": ["1.1.1.1/31", "10.10.10.10/16", "100.100.100.100/24"]
                 }
             }
         }


generic_config_updater/gu_common.py
```diff
@@ -307,10 +307,16 @@ def validate_bgp_peer_group(self, config_db):
             # TODO: convert string to IpAddress object for better handling of IPs
             # TODO: validate range intersection
             ip_range = peer_group["ip_range"]
+
+            # Use "default" if vrf_name is missing or falsy (e.g., empty string)
+            vrf_name = peer_group.get("vrf_name") or "default"
```
anish-n: I don't think peer_group has a "vrf_name", the vrf_name needs to be deciphered from the peer_group_name and has a format as follows: ```<vrf-name>|<peer-group-name>```, in addition the default vrf will be one where there is ```<peer-group-name>``` but not ```<vrf-name>|<peer-group-name>```

alawing: resolved and passed tests

generic_config_updater/gu_common.py
```diff
@@ -307,10 +307,17 @@ def validate_bgp_peer_group(self, config_db):
             # TODO: convert string to IpAddress object for better handling of IPs
             # TODO: validate range intersection
             ip_range = peer_group["ip_range"]
+
+            # Use "default" vnet name if not specified
+            name_split = peer_group_name.split('|')
+            vnet_name = name_split[0] if len(name_split) > 1 else "default"
```
anish-n: reword vnet to vrf, vrf is a more general name applicable to vnet and vrf

tests/generic_config_updater/gu_common_test.py
```diff
@@ -243,6 +243,28 @@ def test_validate_bgp_peer_group__valid_non_intersecting_ip_ranges__returns_true
         self.assertTrue(actual)
         self.assertIsNone(error)
 
+    def test_validate_bgp_peer_group__valid_intersecting_ip_ranges_in_different_vnets__returns_true(self):
```
anish-n: rename vnet to vrf

tests/generic_config_updater/gu_common_test.py
```diff
@@ -243,6 +243,28 @@ def test_validate_bgp_peer_group__valid_non_intersecting_ip_ranges__returns_true
...
+        config_wrapper = gu_common.ConfigWrapper()
+        config = {
+            "BGP_PEER_RANGE":
+            {
+                "VnetA|WLPARTNER_PASSIVE_V4": {
+                    "ip_range": ["1.1.1.1/31", "10.10.10.10/16", "100.100.100.100/24"]
+                },
+                "VnetB|WLPARTNER_PASSIVE_V4": {
+                    "ip_range": ["1.1.1.1/31", "10.10.10.10/16", "100.100.100.100/24"]
+                }
```
anish-n, line 260: could you also add default vrf with same ip space here

tests/generic_config_updater/gu_common_test.py
```diff
@@ -274,15 +296,26 @@ def test_validate_bgp_peer_group__same_ip_prefix__return_false(self):
             duplicated_ip="fc00:1::32/16")
 
     def check_validate_bgp_peer_group(self, ip_range, other_ip_range=[], duplicated_ip=None):
+        # Check both same vnet and different vnet scenarios
+        self._check_validate_bgp_peer_group(ip_range, other_ip_range, duplicated_ip, same_vnet=False)
+        self._check_validate_bgp_peer_group(ip_range, other_ip_range, duplicated_ip, same_vnet=True)
+
+    def _check_validate_bgp_peer_group(self, ip_range, other_ip_range=[], duplicated_ip=None, same_vnet=False):
```
anish-n: can we change same_vnet to is_default_vrf

tests/generic_config_updater/gu_common_test.py
alawing: This case is in following function:
test_validate_bgp_peer_group__same_ip_prefix__return_false

alawing: Addressed this

PR #4203, opened 2026-01-15
[acl-loader]: Wait for rules to be removed before installing new ones

MavChtz:
When running `acl-loader update full` multiple times consecutively, with an ACL rule set that exceeds the TCAM's capacity, the installed ACL rules differ between iterations.

#### What I did
full_update() now waits for the existing rules to be removed.
incremental_update() has also been modified as it handles dataplane rules in the same way.

This results in a deterministic rule set between runs.

#### How I did it
By polling the State DB before installing the new ones. 

#### How to verify it
1. Have a /etc/sonic/config.db with a large amount of rules.
2. Run `acl-loader update full` with a JSON file containing a large amount of rules, multiple times.
3. If the rule set provided is large enough to saturate the TCAM's capacity, you'll see differences between runs in `show acl rule`.

### DIFF
--- a/acl_loader/main.py
+++ b/acl_loader/main.py
@@ -5,6 +5,7 @@ import ipaddress
 import json
 import syslog
 import operator
+import time
 
 import openconfig_acl
 import tabulate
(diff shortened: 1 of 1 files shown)


MavChtz:
Hey @saiarcot895, do you mind having a look into this?

PR #4205, opened 2026-01-16
Fix route_check.py to not hog a lot of memory

venkit-nexthop:
#### What I did
The route_check.py script launches a sonic "show ip route json" to detect any missing routes between FRR and app-db. This causes the show python program to hog a lot of memory (in one test I have noticed it went to 15G), causing the system to OOM. The issue is due to the fact that the show script is reading all the json content (using vtysh) in one shot and is attempting to interpret the json (using json.loads()).

#### How I did it
This diff modifies the route_check.py to not
invoke "show" and rather invoke the vtysh cmd directly. It then attempt to interpret one route at
a time in a paginated manner. This prevents a sudden transient memory buildup. The zebra process already does the right thing and backs off when the output socket buffers are full. There is probably scope to improve that further
(Refer to
https://sonicfoundation.dev/2025-sonic-hackathon-most-impactful-award-spotlight-optimizing-output-buffer-memory-for-show-commands/)

#### How to verify it
Earlier with a very high route-scale (or path scale when ECMP was high), route_check.py used to hog a lot of memory even causing OOMs. With this fix, the same route-scale is operating just fine.

### DIFF
--- a/scripts/route_check.py
+++ b/scripts/route_check.py
@@ -37,6 +37,7 @@ To verify:
 import argparse
 from enum import Enum
 import ipaddress
+import ijson
 import json
 import os
 import re
@@ -394,7 +395,7 @@ def is_suppress_fib_pending_enabled(namespace):
 def fetch_routes(ipv6=False, namespace=multi_asic.DEFAULT_NAMESPACE):
     """
     Fetch routes using the given command.
-    Uses chunk-based reading with incremental parsing of individual prefix entries.
+    Uses ijson for streaming JSON parsing to handle large route tables efficiently.
     Parses each prefix entry as soon as it becomes available instead of waiting for the entire JSON.
     """
     missing_routes = []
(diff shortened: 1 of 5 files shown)


venkit-nexthop:
@deepak-singhal0408 
@prsunny 
@eddieruan-alibaba

As discussed in the PM routing WG meeting on Jan 15 2026. Here is the PR for handling memory hog coming from route-check.py

scripts/route_check.py
```diff
@@ -61,12 +61,15 @@
 
 SUBSCRIBE_WAIT_SECS = 1
 
-# Max of 2 minutes on normal devices and 5 minutes on virtual chassis
-TIMEOUT_SECONDS = 120 if not device_info.is_virtual_chassis() else 360
+# Max of 10 mins
+TIMEOUT_SECONDS = 600
```
eswaran-nexthop: comments from the chassis meeting from Deepak Singhal - 

1. Why did this timeout have to change? Was there an increase in time due to this change?
2. monit script invokes route_check.py every 60s. Do we need to change that?

deepak-singhal0408: also could we get some benchmarking data from a scaled setup? say with 50k v4+50kv6 routes, how much time does this script takes before and after your change.

venkit-nexthop: The original timeout had an assumption about low scale. I just bumped it up so that it would be future proof for much larger scale. The actual time for executing is the same (if not better). In fact the transient memory utilization is a lot better. With a certain scale, the original route_check.py was going up to 15G of transient memory utilization and causing OOM.

I will try to capture some benchmark data.

venkit-nexthop: Upon further thought, maybe I will set the timeout back to the original value. We can revisit the timeouts later.

prsunny: afaik, monit is invoking every 5 minutes. is it different for route_check?

deepak-singhal0408: thanks @prsunny for correction.. yes, looks like its invoked every 5 minutes..

https://github.com/sonic-net/sonic-buildimage/blob/master/files/image_config/monit/conf.d/sonic-host#L30

venkit-nexthop: Yes, that is right. The monitrc file has

`set daemon 60`

This represents each cycle for monit. 

The route check is controlled by the following monit config

```  
  # route_check.py Verify routes between APPL-DB & ASIC-DB are in sync.
  # For any discrepancy, details are logged and a non-zero code is returned
  # which would trigger a monit alert.
  # Hence for any discrepancy, there will be log messages for "ERR" level
  # from both route_check.py & monit.
  #
  check program routeCheck with path "/usr/local/bin/route_check.py"
         every 5 cycles
         if status != 0 for 3 cycle then alert repeat every 1 cycles
```

So it checks every 5 cycles (i.e every 5 minutes).

The monitrc does not override any limits. Hence the default limit for `programTimeout` is 300 seconds (5 mins). So if the route_check.py does not exit in 5 mins, it will get killed. 

Irrespective of what the TIMEOUT value is set within the script itself, the monit will kill it if it runs > 300secs.

venkit-nexthop: @prabhataravind 

```
$ cat route-check
$ show ip route summary | grep Total
Totals               50011                50010
$ show ipv6 route summary | grep Total
Totals               50026                50026


NEW
====
$ /usr/bin/time -v route_check.py
:
:
Command exited with non-zero status 255
        Command being timed: "route_check.py"
        User time (seconds): 13.60
        System time (seconds): 3.51
        Percent of CPU this job got: 77%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:22.01
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 120900
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 1
        Minor (reclaiming a frame) page faults: 47548
        Voluntary context switches: 272236
        Involuntary context switches: 273
        Swaps: 0
        File system inputs: 132
        File system outputs: 0
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 255
        
OLD
===
$ /usr/bin/time -v /tmp/route_check.py
:
:
Command exited with non-zero status 255
        Command being timed: "/tmp/route_check.py"
        User time (seconds): 15.79
        System time (seconds): 4.55
        Percent of CPU this job got: 100%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:20.33
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 1307688
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 88
        Minor (reclaiming a frame) page faults: 709034
        Voluntary context switches: 286976
        Involuntary context switches: 483
        Swaps: 0
        File system inputs: 798
        File system outputs: 0
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 255
```

venkit-nexthop: NEW
```
        User time (seconds): 13.60
        System time (seconds): 3.51
        Percent of CPU this job got: 77%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:22.01
```
OLD
```
        User time (seconds): 15.79
        System time (seconds): 4.55
        Percent of CPU this job got: 100%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:20.33
```

NEW : `Maximum resident set size (kbytes): 120900`
OLD  : `Maximum resident set size (kbytes): 1307688`

scripts/route_check.py
```diff
@@ -385,38 +391,225 @@ def is_suppress_fib_pending_enabled(namespace):
...
+    def extract_prefix_entry(buffer_str, start_idx=0):
+        """
+        Extract a single prefix entry from the buffer.
+        Returns: (prefix_json_str, end_position) or (None, start_idx) if no complete entry found.
+
+        The JSON structure is: {"prefix1": [route_entries], "prefix2": [route_entries], ...}
+        We want to extract individual "prefix": [route_entries] pairs.
+        """
+        # Skip whitespace and opening brace of outer object
+        i = start_idx
```
deepak-singhal0408: @venkit-nexthop , not sure if you considered https://pypi.org/project/ijson/  instead of writing a custom parser?

venkit-nexthop: Thanks @deepak-singhal0408 I was not aware of ijson. The logic to parse the route prefixes from the json stream is simple and the upside is that it does not need any new python package. I have unit tests to cover the parsing logic as well.

But if you strongly feel that we should be using ijson, I can try to rework my PR to use that.

deepak-singhal0408: > Thanks @deepak-singhal0408 I was not aware of ijson. The logic to parse the route prefixes from the json stream is simple and the upside is that it does not need any new python package. I have unit tests to cover the parsing logic as well.
> 
> But if you strongly feel that we should be using ijson, I can try to rework my PR to use that.

I feel we should make use of available library.. it seems widely used.. 
image
Also the route_check code will simplify a lot(will reduce maintenance burden as well).. Also will be future proof..

venkit-nexthop: sounds good. I will rework the PR. Thanks for the suggestion.

venkit-nexthop: I have posted an update to the PR replacing the earlier parser code with ijson.

Here is the benchmark for the same route scale

```
        Command being timed: "/tmp/route_check.py"
        User time (seconds): 6.14
        System time (seconds): 2.70
        Percent of CPU this job got: 64%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:13.80
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 128268
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 0
        Minor (reclaiming a frame) page faults: 40620
        Voluntary context switches: 269117
        Involuntary context switches: 182
        Swaps: 0
        File system inputs: 108
        File system outputs: 0
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
```

deepak-singhal0408 (approved):
LGTM.. Thanks,

venkit-nexthop:
@prsunny could you help me merge this. 
@deepak-singhal0408 has approved the change.

venkit-nexthop:
@deepak-singhal0408 this has not merged yet after your approval. Does it need some other maintainer to approve as well?

saiarcot895:
sonic-utilities PRs just run unit tests that are located in this repository. However, when we do submodule updates in sonic-buildimage (and an image is built with this change), that includes sonic-mgmt tests running in a KVM environment, which is why the breakage doesn't appear here, but does appear there.

venkit-nexthop:
> sonic-utilities PRs just run unit tests that are located in this repository. However, when we do submodule updates in sonic-buildimage (and an image is built with this change), that includes sonic-mgmt tests running in a KVM environment, which is why the breakage doesn't appear here, but does appear there.

Thanks @saiarcot895 I will post a PR to fix this now.

saiarcot895:
Hold off on that, there's something stranger going on.

venkit-nexthop:
> Hold off on that, there's something stranger going on.

Sure

deepak-singhal0408:
This change is already available in 202511. removed the cherry-pick flags.

PR #4213, opened 2026-01-20
[bash_completion.d] Update completion for newer click

### LINKED ISSUE #4211: Bash completion for "show" and "config" seems to have stopped working with the update to Trixie
Locally we've been tracking the 202511 branch of sonic-buildimage and have pulled in the Trixie update. The completion for "show" and "config" seems to have stopped working.

cpackham-atlnz:
As of click 8.0.0 the way shell completion works for bash hash changed. Update the completion files for the new method.

Fixes #4211


cpackham-atlnz:
Looks like this is obsoleted by other changes

PR #4217, opened 2026-01-22
Fix route_check.py redis client high memory usage

### LINKED ISSUE #4216: route_check.py redis client memory usage causing dbmemory errors
Recently we observed 'dbmemory' errors in sonic-mgmt tests. These errors are triggered when a redis client queue consumes more than 10MB of memory. The failures were attributed to `route_check.py`, which is executed by monit every 5 minutes.

Root cause:
When `route_check.py` runs after a config reload, the `check_frr_pending_routes()` routine can take a long time to execute (30+ seconds). This delay occurs either because the `show ip route json` command is slow when there are a lots of routes, or because it times out and retries multiple times (FRR_CHECK_RETRIES=3, FRR_WAIT_TIME=15 seconds).

During this delay, the ASIC_DB subscriber remains open and route updates continue to accumulate in the subscriber queue, causing redis client
memory to grow beyond the 10MB threshold.

Solution:
Close the subscriber immediately after reading ASIC_DB updates and before calling `check_frr_pending_routes()`. This prevents queue accumulation during the slow FRR check.

roeebar-arista:
#### What I did
Close ASIC_DB subscriber in `route_check.py` after collecting updates to prevent queue accumulation. Previously, the subscriber remained open during the `check_frr_pending_routes`, causing route updates to accumulate in the redis subscriber queue and increase memory usage. That high memory usage cause `dbmemory` error in sonic-mgmt tests.

#### How I did it
Delete the subscriber object reference after it is no longer needed. This should kill the subscriber redis connection.

#### How to verify it
No more 'dbmemory' error after a config reload (see issue #4216). Also, you can monitor client memory usage using:
```while true; do date; redis-cli CLIENT LIST | grep -v "omem=0"; sleep 1; done ``` and verify no significant memory spikes when running `route_check.py`.

fixes #4216

### DIFF
--- a/scripts/route_check.py
+++ b/scripts/route_check.py
@@ -877,6 +877,11 @@ def check_routes_for_namespace(namespace):
         # Look for subscribe updates for a second
         adds, deletes = get_subscribe_updates(selector, subs)
 
+    # Release the subscriber. If we keep the subscriber open then route updates will accumulate in the subscriber queue
+    # causing high client memory usage in redis.
+    del subs
+    del selector
+
     # Drop all those for which SET received
     rt_appl_miss, _ = diff_sorted_lists(rt_appl_miss, adds)
 


venkit-nexthop (approved):

saravanan-nexthop (approved):
It will be good to know how many routes were on the dut and how many times did route_check ran to confirm the high memory usage is gone and no new regressions.

deepak-singhal0408 (approved):
LGTM.. thanks,

PR #4237, opened 2026-01-31
Fix JsonMove._get_value to Support Both String and Integer List Indices

xincunli-sonic:
### What I did:

Issue: https://github.com/sonic-net/sonic-utilities/issues/4221

- Updated [JsonMove._get_value](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) to handle both string and integer indices when traversing lists in config data.
- Adjusted related unit tests to reflect the new behavior.

### How I did it:

- Modified the traversal logic to convert string tokens to integers when accessing lists, allowing both "1" and 1 as valid indices.
- Removed the test expecting a TypeError for integer indices and added assertions for both string and integer index access.

### How to verify it:

Patched change in lab device, confirmed. 

```shell
admin@STR-SN5640-RDMA-1:~$ cat /usr/local/lib/python3.11/dist-packages/generic_config_updater/patch_sorter.py | grep -C 2 "int(token)"
        for token in tokens:
            if isinstance(config, list):
                token = int(token)
            config = config[token]

admin@STR-SN5640-RDMA-1:~$ cat t_tc_to_queue_map_modify.json 
[
  {
    "op": "replace",
    "path": "/TC_TO_QUEUE_MAP/AZURE/8",
    "value": "8"
  },
  {
    "op": "add",
    "path": "/TC_TO_QUEUE_MAP/AZURE/7",
    "value": "7"
  }
]

admin@STR-SN5640-RDMA-1:~$ sudo config apply-patch -v t_tc_to_queue_map_modify.json
Patch Applier: localhost: Patch application starting.
Patch Applier: localhost: Patch: [{"op": "replace", "path": "/TC_TO_QUEUE_MAP/AZURE/8", "value": "8"}, {"op": "add", "path": "/TC_TO_QUEUE_MAP/AZURE/7", "value": "7"}]
Patch Applier: localhost getting current config db.
Patch Applier: localhost: simulating the target full config after applying the patch.
Patch Applier: localhost: validating all JsonPatch operations are permitted on the specified fields
Patch Applier: localhost: validating target config does not have empty tables,
Patch Applier: localhost: sorting patch updates.
Patch Sorter - Strict: Validating patch is not making changes to tables without YANG models.
Patch Sorter - Strict: Validating target config according to YANG models.
Patch Sorter - Strict: Sorting patch updates.
Patch Applier: The localhost patch was converted into 1 change:
Patch Applier: localhost: applying 1 change in order:
Patch Applier:   * [{"op": "replace", "path": "/TC_TO_QUEUE_MAP/AZURE/7", "value": "7"}, {"op": "replace", "path": "/TC_TO_QUEUE_MAP/AZURE/8", "value": "8"}]
Patch Applier: localhost: verifying patch updates are reflected on ConfigDB.
Patch Applier: localhost patch application completed.
Patch applied successfully.
```
Also run the updated unit tests and all tests should pass, confirming the fix.

### DIFF
--- a/tests/generic_config_updater/test_patch_sorter_get_value.py
+++ b/tests/generic_config_updater/test_patch_sorter_get_value.py
@@ -48,3 +48,8 @@ class TestJsonMoveGetValue(unittest.TestCase):
         tokens = ["table1", "key1", "field", "extra"]
         with self.assertRaises(TypeError):
             JsonMove._get_value(self.config, tokens)
+
+    def test_get_value_dict_numeric_keys(self):
+        self.config["7"] = {"8": "30"}
+        tokens = ["7", "8"]
+        self.assertEqual(JsonMove._get_value(self.config, tokens), "30")
--- a/tests/generic_config_updater/test_patch_sorter_get_value.py
+++ b/tests/generic_config_updater/test_patch_sorter_get_value.py
@@ -1,5 +1,6 @@
 import unittest
-from generic_config_updater.patch_sorter import JsonMove, GenericConfigUpdaterError
+from generic_config_updater.patch_sorter import JsonMove
+
 
 class TestJsonMoveGetValue(unittest.TestCase):
     def setUp(self):
@@ -49,5 +50,3 @@ class TestJsonMoveGetValue(unittest.TestCase):
         with self.assertRaises(TypeError):
             JsonMove._get_value(self.config, tokens)
 
-if __name__ == "__main__":
-    unittest.main()
--- a/tests/generic_config_updater/test_patch_sorter_get_value.py
+++ b/tests/generic_config_updater/test_patch_sorter_get_value.py
@@ -28,10 +28,9 @@ class TestJsonMoveGetValue(unittest.TestCase):
         self.assertEqual(JsonMove._get_value(self.config, tokens), "value2")
 
     def test_get_value_list(self):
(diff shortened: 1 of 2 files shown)


bhouse-nexthop (approved):
Thanks for tackling this, yes, this looks right to me.

I think the only questions that come to mind are if tuples can be represented in the config (I don't think so).

tests/generic_config_updater/test_patch_sorter_get_value.py
```diff
@@ -0,0 +1,50 @@
+import unittest
+from generic_config_updater.patch_sorter import JsonMove
+
+
+class TestJsonMoveGetValue(unittest.TestCase):
+    def setUp(self):
+        self.config = {
```
developfast, line 7: @xincunli-sonic can you also add a case where both the `key` and `field` are also a number in string format ex. `"7" : {"8": "30"}`

developfast (approved):
lgtm

wangxin (approved):

vmittal-msft:
@xincunli-sonic therer is cherry-pick conflict for 202511. please raise manual pr

vmittal-msft:
@xincunli-sonic please create manual pr if must have for 20511 else please remove 202511 tags

PR #4249, opened 2026-02-05
show ipv6 bgp summary fixes for uncompressed addresses

prashanth-nexthop:
#### What I did
There is an inconsistency in the outputs of "show ipv6 bgp summary" and "vtysh -c "show bgp ipv6 summary" in the neighbor name column

The way the CLI "show ipv6 bgp summary" works is,

It in turn runs the vtysh command "show bgp ipv6 summary json" to get the data
This data is keyed by the neighbor's IP address
In order to display neighbor name, it gets BGP_NEIGHBOR information and uses that IP address as the key to look up the data it got from vtysh
But if the sonic config uses IP address string that is either not contracted or incorrectly contracted, then the lookup fails causing the output to miss the neighbor name field. Because the FRR output seems to be representing IPv6 addresses correctly.
For ex, if the neighbor IP in BGP_NEIGHBOR config is is 2000:0000:0000:0000:0000:0000:0000:0001, FRR contracts this to 2000::1 and when 2000:0000:0000:0000:0000:0000:0000:0001 is used as key, the lookup fails. Similarly 2000::0 would fail because FRR contracts it to just 2000::

The problem is not unique to IPv6 addresses ending in 0, it can happen when contraction is technically incorrect.

#### How I did it
I am normalizing the IPv6 address used in python library to its canonical form

#### How to verify it
Added unit tests to verify the changes.

```
WIth BGP_NEIGHBOR config using
"default|fc00:0000:0000:0000:0000:0000:0000:0002":{
"name": "Neighbor1",
"peer_group_name": "PEER_V6"
}
"default|2001::0": {
"default|2001::0": {
"name": "Neighbor2",
"peer_group_name": "PEER_V6"
}
Output of show ipv6 bgp summary without the fix
Neighbhor V AS MsgRcvd MsgSent TblVer InQ OutQ Up/Down State/PfxRcd NeighborName

2001:: 4 65200 808 807 4007 0 0 00:01:21 1594
fc00::2 4 65200 808 807 4007 0 0 00:01:21 1594

** The same output with the fix**
Neighbhor V AS MsgRcvd MsgSent TblVer InQ OutQ Up/Down State/PfxRcd NeighborName

2001:: 4 65200 816 814 4007 0 0 00:08:48 1594 Neighbor2
fc00::2 4 65200 817 814 4007 0 0 00:08:48 1594 Neighbor1
```

### DIFF
--- a/tests/bgp_commands_test.py
+++ b/tests/bgp_commands_test.py
@@ -938,6 +938,96 @@ class TestBgpCommandsSingleAsic(object):
         dbconnector.load_database_config()
 
 
+class TestBgpUtilIpNormalization(object):
+    """
+    Test cases for IPv6 address normalization in bgp_util.
+    These tests verify that non-canonical IPv6 addresses are properly normalized
+    to canonical form for consistent neighbor lookups.
+    """
+
+    def test_normalize_ip_address(self):
+        """Test IP address normalization for IPv6, IPv4, and non-IP strings."""
+        test_cases = [
+            ('2000::0', '2000::'),
+            ('fc00::0001', 'fc00::1'),
+            ('2001:0db8:0000:0000:0000:0000:0000:0001', '2001:db8::1'),
+            ('fc00::42', 'fc00::42'),
+            ('::1', '::1'),
+            ('10.0.0.1', '10.0.0.1'),
+            ('Ethernet0', 'Ethernet0'),
+            ('', ''),
+        ]
+        for input_val, expected in test_cases:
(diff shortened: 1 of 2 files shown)


prashanth-nexthop:
@rlhui @ZhaohuiS @StormLiangMS

prashanth-nexthop:
@rlhui @ZhaohuiS @StormLiangMS can I get a merge approval for this simple CLI fix ?

PR #4266, opened 2026-02-10
[console] add config console escape

Xichen96:
#### What I did
Add escape char config to console feature.
In CONFIG_DB, an optional new entry under CONSOLE_SWITCH, "console_mgmt", is added, "default_escape_char" as the key and we expect a lower case letter as value. Similarly, an optional new entry under CONSOLE_PORT, and then each line, "escape_char" as the key and we also expect a lower case letter as value.
Commands are:
1. config console default_escape (upper and lower case letters, or clear), e.g. config console default_escape k, config console default_escape B, config console default_escape clear.
2. config console escape [LINE_NUMBER] (upper and lower case letters, or clear), e.g. config console escape 6 C, config console escape 5 clear
3. show line --show-escape, e.g. show line --show-escape, show line -s -b
Connect command work as usual, except that it will add --escape <escape_char> in the underlying picocom command if it is configured.

Additionally,
1. some style change to old code to pass pre-commit check
2. unit tests for new functions
3. enhancement to old unit tests and its bug fix

#### How I did it
Implement them

#### How to verify it
Run the commands in dut.

### DIFF
--- a/config/console.py
+++ b/config/console.py
@@ -1,4 +1,5 @@
 import click
+import string
 import utilities_common.cli as clicommon
 from .validated_config_db_connector import ValidatedConfigDBConnector
 from jsonpatch import JsonPatchConflict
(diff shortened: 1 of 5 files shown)


Blueve:
Do you have YANG model update for this change?

config/console.py
```diff
@@ -129,11 +174,11 @@ def upate_console_remote_device_name(db, linenum, devicename):
         if dataKey in data and devicename == data[dataKey]:
             # do nothing if the device name is same with existing configurtion
             return
-        elif not devicename:
+        elif not devicename and dataKey in data:
```
Blueve: This changes the reset path semantics for `config console remote_device <line>`. If the entry exists but does not currently have `remote_device`, this condition is false and the code falls through to the final `else`, which writes `remote_device = None`. `ValidatedConfigDBConnector.mod_entry()` stringifies values, so this ends up storing the literal string `"None"` in CONFIG_DB instead of preserving the old no-op behavior.

config/console.py
```diff
@@ -208,6 +255,38 @@ def update_console_flow_control(db, mode, linenum):
...
+    """Update escape character for a console line"""
+    config_db = ValidatedConfigDBConnector(db.cfgdb)
+    ctx = click.get_current_context()
+
+    table = "CONSOLE_PORT"
+    dataKey = 'escape_char'
+
+    data = config_db.get_entry(table, linenum)
+    if data:
+        if escape == "clear" and dataKey in data:
```
Blueve: There is still a bad `clear` path here when the line exists but does not already have `escape_char`: the condition is false, so the `else` branch stores `escape.lower()`, i.e. the literal string `"clear"`, into CONFIG_DB. That will later make `consutil connect` pass `picocom -e clear ...` instead of treating `clear` as a no-op reset.

Blueve (approved):

PR #4289, opened 2026-02-19
Add multi-asic namespace support for IPv6 link-local commands

william8545:
#### What I did

Added multi-ASIC namespace support for IPv6 link-local `config` and `show` commands. Previously these commands only worked against the default (single) ASIC namespace.

#### How I did it

- Added `-n/--namespace` option to the `config ipv6` command group so all subcommands (`enable/disable link-local`, per-interface `use-link-local-only`) operate on the correct namespace-specific config DB.
- Added `-n/--namespace` option to `show ipv6 link-local-mode` so it can display interfaces from a single namespace or iterate over all namespaces.
- Removed redundant `ConfigDBConnector` creation in subcommands — they now reuse the DB connection established by the parent group.
- Added new multi-ASIC unit tests and updated existing single-ASIC tests accordingly.

#### How to verify it

- On a multi-ASIC device:
  - `show ipv6 link-local-mode` — shows interfaces from all ASICs
  - `show ipv6 link-local-mode -n asic0` — shows only asic0 interfaces
  - `sudo config ipv6 -n asic0 enable link-local` — enables link-local on asic0 only
  - `sudo config ipv6 -n asic0 disable link-local` — disables link-local on asic0 only

### DIFF
--- a/show/main.py
+++ b/show/main.py
@@ -1570,7 +1570,7 @@ def link_local_mode(namespace, verbose):
     ns_list = masic.get_ns_list_based_on_options()
 
     for ns in ns_list:
-        config_db = ConfigDBConnector(use_unix_socket_path=True, namespace=ns)
+        config_db = ConfigDBConnector(namespace=ns)
         config_db.connect()
         interface = ""
 


oleksandrivantsiv (approved):

cyw233:
Hi @oleksandrivantsiv , I'm seeing multiple PR tests failure after your recent multi-asic tests improvement, for example, this no-op PR https://github.com/sonic-net/sonic-utilities/pull/4344 has the following failed tests. Could you please check? Thanks

image

PR #4310, opened 2026-02-27
Create GCU wheel

xincunli-sonic:
### What I did

Created a standalone GCU (Generic Config Updater) wheel that can be deployed and updated independently of the host sonic-utilities package via the GCU container.

### How I did it

- Moved apply-patch logic from config/main.py into generic_config_updater/main.py
- Added GCU_STANDALONE_BIN = "/opt/sonic/gcu/current/bin/gcu-standalone" in main.py and a run_gcu_standalone() helper that forwards all flags to the subprocess
- The apply-patch Click command checks os.path.exists(GCU_STANDALONE_BIN) first; if present it delegates to run_gcu_standalone(), otherwise falls through to the built-in implementation
- Registered gcu-standalone = generic_config_updater.main:main as a console_scripts entry point in setup.py
Added main_test.py with 12 test classes covering all public functions in main.py

### How to verify it

Manual verification:

- Without standalone binary — config apply-patch patch.json applies the patch via the built-in implementation
- With standalone binary — install the GCU wheel at /opt/sonic/gcu/current/; config apply-patch patch.json redirects to gcu-standalone apply-patch patch.json

### DIFF
--- a/generic_config_updater/setup.py
+++ b/generic_config_updater/setup.py
@@ -26,6 +26,7 @@ setup(
     url='https://github.com/Azure/sonic-utilities/generic_config_updater',
     maintainer='Xincun Li',
     maintainer_email='xincun.li@microsoft.com',
+    package_dir={'generic_config_updater': '.'},
     packages=[
         'generic_config_updater',
     ],


config/main.py
```diff
@@ -1957,97 +1819,59 @@ def apply_patch(
...
-                # Wait for all tasks to complete
-                concurrent.futures.wait(futures)
+    trace_file = None
+    try:
+        if path_trace:
+            # Open the trace file and call GenericUpdater directly so the
+            # caller-supplied file handle is forwarded as trace_io.
+            trace_file = open(path_trace, 'w')
+            with open(patch_file_path, 'r') as fh:
+                patch_ops = json.loads(fh.read())
```
saiarcot895: For this section, maybe something like the following will be a bit cleaner? You won't need the `finally` block at least:

```python
try:
    if path_trace:
        with open(path_trace, 'w') as trace_file:
            with open(patch_file_path, 'r') as fh:
                patch_ops = json.loads(fh.read())
            config_format = ConfigFormat[format.upper()]
```

Or too many levels of indentation?

generic_config_updater/field_operation_validators.py
```diff
@@ -6,7 +6,10 @@
 from sonic_py_common import device_info
 from .gu_common import GenericConfigUpdaterError
 from swsscommon import swsscommon
-from utilities_common.constants import DEFAULT_SUPPORTED_FECS_LIST
+
+# Default FEC modes when STATE_DB does not advertise supported_fecs for a port.
+# Kept local to avoid pulling utilities_common into the GCU wheel.
+DEFAULT_SUPPORTED_FECS_LIST = ['rs', 'fc', 'none', 'auto']
```
saiarcot895, line 14: You might want to make a command in `utilities_common.constants` saying to update this list as well; otherwise, the two will get out-of-sync.

config/main.py
```diff
@@ -1920,6 +1755,33 @@ def print_dry_run_message(dry_run):
...
+    if parallel:
+        cmd.append("--parallel")
+    if ignore_non_yang_tables:
+        cmd.append("--ignore-non-yang-tables")
+    for path in ignore_path:
+        cmd += ["--ignore-path", path]
+    if verbose:
+        cmd.append("--verbose")
+    if path_trace:
+        cmd += ["--path-trace", path_trace]
```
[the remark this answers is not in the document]
xincunli-sonic: run_gcu_standalone is redirecting command to GCU_STANDALONE_BIN which consumes path_trace as trace_io

PR #4314, opened 2026-02-27
Fix issue: pmon services's restart count is not cleared during config reload

stephenxs:
#### What I did

Currently, when "config reload" is executed, services' restart count are cleared to avoid reaching restart limit. This is done by listing all services using command systemctl list-dependencies --plain .target. However, this doesn't include pmon service, neither all other services that don't have WantedBy=sonic.target, which means pmon's start count is not cleared.

#### How I did it

Sometimes pmon fails to restart due to reaching start limit (3 times in 1200 seconds). The pmon service can be started by featured, syncd during config reload. Before multi-ASIC, pmon depends on syncd. The dependency is removed after multi-ASIC, which means pmon can restart immediately triggered by sonic.target which is once more restarting. As a result the pmon service is more likely to reach the restart limit.

#### How to verify it

Clear restart count also for services that have reverse dependency on sonic.target.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1005,14 +1005,21 @@ def _stop_services():
     clicommon.run_command(['sudo', 'systemctl', 'stop', 'sonic.target', '--job-mode', 'replace-irreversibly'])
 
 
-def _get_sonic_services():
+def _get_sonic_services(reverse=False):
     cmd = ['systemctl', 'list-dependencies', '--plain', 'sonic.target']
+    if reverse:
+        cmd.append('--reverse')
+
     out, _ = clicommon.run_command(cmd, return_cmd=True)
     out = out.strip().split('\n')[1:]
     return (unit.strip() for unit in out)
 
+
 def _reset_failed_services():
-    for service in _get_sonic_services():
+    services = set(_get_sonic_services())
+    services.update(set(_get_sonic_services(True)))
+
+    for service in services:
         clicommon.run_command(['systemctl', 'reset-failed', str(service)])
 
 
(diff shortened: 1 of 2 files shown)


keboliu (approved):

StormLiangMS (dismissed):
LGTM

config/main.py
```diff
@@ -1005,14 +1005,21 @@ def _stop_services():
     clicommon.run_command(['sudo', 'systemctl', 'stop', 'sonic.target', '--job-mode', 'replace-irreversibly'])
 
 
-def _get_sonic_services():
+def _get_sonic_services(reverse=False):
     cmd = ['systemctl', 'list-dependencies', '--plain', 'sonic.target']
+    if reverse:
+        cmd.append('--reverse')
```
saiarcot895, line 1011: I would suggest also adding `--type=service`. when adding `--reverse`. Otherwise, this will also include `multi-user.target` and `graphical.target`, and while there probably wouldn't be any negative effects from resetting the failed state of the units, it might be better to restrict it to SONiC services.

stephenxs: Fixed

liat-grozovik (approved):

stephenxs:
> The change is not in 202511 yet. @stephenxs, please manually create the cherry pick PR for branch 202511.You can ping the release branch owner(github account: ) to approve your cherry pick PR.If this change is already in 202511, please comment "already in 202511". Thanks!
> 
> ---Powered by SONiC BuildBot

202511 PR was merged a month ago https://github.com/sonic-net/sonic-utilities/pull/4336

PR #4315, opened 2026-02-27
fix scapy delayed import when we have large routes

tirupatihemanth:
#### What I did
Fix scapy import delay in teamd_increase_retry_count.py and scripts/pcmping script when there are a lot of routes installed
#### How I did it
Turn off scapy to autoload routes
#### why I did it
- This is because scapy 2.6.1 in Debian 13 tries to create internal python objects for all the routes in route table. This process can take long time when we have a lot of routes installed. 
- Fix in teamd_increase_retry_count.py script reduced the warm-reboot script delay

### DIFF
--- a/scripts/pcmping
+++ b/scripts/pcmping
@@ -13,6 +13,10 @@ import sys
 import time
 
 from swsscommon.swsscommon import ConfigDBConnector
+from scapy.config import conf
+# Avoid long startup delays when route tables are large.
+conf.route_autoload = False
+conf.route6_autoload = False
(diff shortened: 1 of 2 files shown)


dgsudharsan (approved):

saiarcot895 (approved):

tirupatihemanth:
already in 202511

tirupatihemanth:
Here's 202511 PR already merged: https://github.com/sonic-net/sonic-utilities/pull/4347

PR #4319, opened 2026-02-28
Fix legacy mirror_session add failing capability check

### LINKED ISSUE #4318: Legacy 'config mirror_session add' missing direction parameter after PR #4089
## Description

PR #4089 added mirror capability checking via `is_port_mirror_capability_supported()` in `validate_mirror_session_config()`, and added the `direction` parameter to `config mirror_session erspan add` and `config mirror_session span add`.

However, the legacy `config mirror_session add` command (line ~2976 in config/main.py) was not updated. It still has the old signature:

```python
def add(session_name, src_ip, dst_ip, dscp, ttl, gre_type, queue, policer):
    add_erspan(session_name, src_ip, dst_ip, dscp, ttl, gre_type, queue, policer)
```

This means:
1. `add_erspan()` is called with `direction=None`
2. `validate_mirror_session_config(direction=None)` checks BOTH ingress and egress capabilities
3. On platforms where orchagent does not populate `PORT_INGRESS_MIRROR_CAPABLE` / `PORT_EGRESS_MIRROR_CAPABLE` in STATE_DB, `state_db.get()` returns `None`, and `None != "true"` causes the command to fail with: `Error: Port mirror direction 'both' is not supported by the ASIC`

This breaks basic ERSPAN mirror session creation on platforms that previously worked fine, since the legacy command never used directional mirroring.

## Steps to Reproduce

On a platform where orchagent does not populate mirror capability keys in STATE_DB:

```
config mirror_session add test_session 1.1.1.1 2.2.2.2 8 4 35006
Error: Port mirror direction 'both' is not supported by the ASIC
```

## Expected Behavior

The legacy `config mirror_session add` command should either:
1. Accept a `direction` parameter (like `erspan add` does), or
2. Skip the capability check when `direction` is not specified and `src_port` is not specified (legacy ERSPAN path)

## Affected Version

All images built after PR #4089 was merged (Nov 17, 2025).

## Related

- sonic-net/sonic-utilities#4089
- sonic-net/sonic-mgmt#22661

yxieca:
PR #4089 added port mirror capability checking via `is_port_mirror_capability_supported()` in `validate_mirror_session_config()`. However, the legacy `config mirror_session add` command calls `add_erspan()` with `src_port=None` and `direction=None`. When `direction` is `None`, the capability check queries both `PORT_INGRESS_MIRROR_CAPABLE` and `PORT_EGRESS_MIRROR_CAPABLE` in STATE_DB. On platforms where orchagent does not populate these keys, `get()` returns `None`, and the check fails with:

```
Error: Port mirror direction 'both' is not supported by the ASIC
```

This breaks basic ERSPAN mirror session creation on platforms that previously worked fine.

Fixes #4318

#### How did you do it?
Skip the port mirror capability check in `validate_mirror_session_config()` when `src_port` is not specified. Legacy ERSPAN sessions without `src_port` don't use port-based mirroring and should not be subject to the capability check. The check is still performed when `src_port` is provided (SPAN and port-based ERSPAN paths).

Added unit tests in `tests/config_mirror_session_test.py`:
- `test_legacy_mirror_session_add_skips_capability_check`: Verifies legacy `config mirror_session add` succeeds without capability mocking
- `test_erspan_add_without_src_port_skips_capability_check`: Verifies `erspan add` without src_port also skips the check
- `test_erspan_add_with_src_port_checks_capability`: Verifies capability check still runs when src_port is specified
- Added test for STATE_DB returning None (orchagent not populating keys)

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1218,10 +1218,9 @@ def validate_mirror_session_config(config_db, session_name, dst_port, src_port, 
             ctx.fail("Error: Direction {} is invalid".format(direction))
 
     # Check port mirror capability before allowing configuration.
-    # Only check when src_port is specified (port-based mirroring).
-    # Legacy ERSPAN sessions without src_port don't use port-based mirroring
-    # and should not be subject to the capability check.
-    if src_port:
+    # Only check when direction is specified (port-based mirroring).
+    # Legacy ERSPAN sessions have direction=None and don't need this check.
+    if direction:
         for ns in namespace_set:
             if not is_port_mirror_capability_supported(direction, namespace=ns):
                 ctx.fail("Error: Port mirror direction '{}' is not supported by the ASIC".format(
@@ -3155,12 +3154,10 @@ def mirror_session():
 @click.argument('ttl', metavar='<ttl>', type=TTL_RANGE, required=True)
 @click.argument('gre_type', metavar='[gre_type]', callback=validate_gre_type, required=False)
 @click.argument('queue', metavar='[queue]', type=QUEUE_RANGE, required=False)
-@click.argument('src_port', metavar='[src_port]', required=False)
-@click.argument('direction', metavar='[direction]', required=False)
 @click.option('--policer')
-def add(session_name, src_ip, dst_ip, dscp, ttl, gre_type, queue, policer, src_port, direction):
+def add(session_name, src_ip, dst_ip, dscp, ttl, gre_type, queue, policer):
     """ Add ERSPAN mirror session.(Legacy support) """
-    add_erspan(session_name, src_ip, dst_ip, dscp, ttl, gre_type, queue, policer, src_port, direction)
+    add_erspan(session_name, src_ip, dst_ip, dscp, ttl, gre_type, queue, policer)
 
 @mirror_session.group(cls=clicommon.AbbreviationGroup, name='erspan')
 @click.pass_context
--- a/tests/config_mirror_session_test.py
+++ b/tests/config_mirror_session_test.py
@@ -77,32 +77,25 @@ def test_mirror_session_add():
                 config.config.commands["mirror_session"].commands["add"],
                 ["test_session", "100.1.1.1", "2.2.2.2", "8", "63", "10", "100"])
 
-        mocked.assert_called_with("test_session", "100.1.1.1", "2.2.2.2", 8, 63, 10, 100, None, None, None)
+        mocked.assert_called_with("test_session", "100.1.1.1", "2.2.2.2", 8, 63, 10, 100, None)
 
         result = runner.invoke(
                 config.config.commands["mirror_session"].commands["add"],
                 ["test_session", "100.1.1.1", "2.2.2.2", "8", "63", "0X1234", "100"])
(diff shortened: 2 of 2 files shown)


tests/config_mirror_session_test.py
```diff
@@ -473,3 +495,70 @@ def test_mirror_session_capability_function():
...
+
+def test_legacy_mirror_session_add_with_direction():
+    """Test that legacy 'config mirror_session add' accepts the new
+    src_port and direction parameters."""
+    config.ADHOC_VALIDATION = True
+    runner = CliRunner()
+
+    with mock.patch('config.main.add_erspan') as mock_add_erspan:
+        result = runner.invoke(
+                config.config.commands["mirror_session"].commands["add"],
```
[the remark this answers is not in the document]
yxieca: Good catch! Fixed in c36d06bc — the tests now call `validate_mirror_session_config` directly and assert that `is_port_mirror_capability_supported` is NOT called when `src_port` is None, rather than mocking `add_erspan` which bypassed the validation entirely.

rustiqly (approved):
LGTM. The fix is clean and well-scoped:

1. **Root cause is correct**: `validate_mirror_session_config()` runs the capability check even for legacy ERSPAN sessions that don't use port-based mirroring (`src_port=None`). Gating on `if src_port:` is the right approach — ERSPAN sessions without `src_port` never needed this check.

2. **Legacy CLI extended properly**: Adding `src_port` and `direction` args to the legacy `config mirror_session add` command maintains backward compatibility (both are optional) while enabling the full feature set through the legacy path.

3. **Good test coverage**: Tests verify both the skip path (no `src_port`) and the check path (with `src_port`), plus the `None` return from STATE_DB case.

4. **Consistent with #22663**: The test-side fix checks capabilities before calling the CLI, while this PR fixes the CLI itself. Both are needed — belt and suspenders.

config/main.py
```diff
@@ -3153,10 +3155,12 @@ def mirror_session():
 @click.argument('ttl', metavar='<ttl>', type=TTL_RANGE, required=True)
 @click.argument('gre_type', metavar='[gre_type]', callback=validate_gre_type, required=False)
 @click.argument('queue', metavar='[queue]', type=QUEUE_RANGE, required=False)
+@click.argument('src_port', metavar='[src_port]', required=False)
+@click.argument('direction', metavar='[direction]', required=False)
```
bingwang-ms: This would change the CLI, which doesn't look like a good idea since the CLI has been there for years.
To tell if a mirror session is port mirror or erspan (everflow), we can check if `direction` is None.
For erspan, it's always `None`

bingwang-ms: I raised another PR to fix the same issue https://github.com/sonic-net/sonic-utilities/pull/4323

yxieca: Good point! Fixed in c6a6961a — removed src_port and direction from the legacy CLI entirely, keeping it unchanged. Now using `if direction:` as the guard in `validate_mirror_session_config()` instead of `if src_port:`, since ERSPAN sessions always have `direction=None`. Tests updated to match.

PR #4323, opened 2026-03-02
fix: skip PORT_INGRESS/EGRESS_MIRROR_CAPABLE check for ERSPAN mirror sessions

bingwang-ms:
## What I did

Fixed is_port_mirror_capability_supported() so that ERSPAN sessions (direction=None) are not blocked by the PORT_INGRESS_MIRROR_CAPABLE / PORT_EGRESS_MIRROR_CAPABLE capability check.

## Root cause

PR #4089 added a capability check that reads PORT_INGRESS_MIRROR_CAPABLE and PORT_EGRESS_MIRROR_CAPABLE from STATE_DB SWITCH_CAPABILITY|switch. For ERSPAN sessions, direction=None was treated as 'check both', but:
1. These capability flags only apply to SPAN (port mirror) sessions, not ERSPAN (which uses source/destination IPs, not ports)
2. Platforms that don't populate these STATE_DB keys return None, which != 'true', so the function incorrectly returns False (unsupported)

PR #4159 partially addressed the multi-ASIC namespace issue but did not fix the fundamental problem for ERSPAN sessions with no src/dst port specified.

## How I fixed it

- **For ERSPAN (direction=None)**: Return True immediately. PORT_INGRESS/EGRESS_MIRROR_CAPABLE does not apply to ERSPAN sessions.
- **For SPAN (direction != None)**: Treat absent STATE_DB key (None value) as 'supported' for backward compatibility with platforms that don't populate SWITCH_CAPABILITY table entries.

## How to verify it

Unit tests updated in 	ests/config_mirror_session_test.py:
- Added Test 4 to verify behavior when STATE_DB keys are absent (all return True)
- Updated Test 2 and Test 3 assertions for direction=None to expect True

Fixes: https://github.com/sonic-net/sonic-mgmt/issues/21690

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1144,26 +1144,27 @@ def interface_has_mirror_config(ctx, mirror_table, dst_port, src_port, direction
 
 
 def is_port_mirror_capability_supported(direction, namespace=None):
-    """ Check if port mirror capability is supported for the given direction """
-    # ERSPAN sessions have direction=None; PORT_INGRESS/EGRESS_MIRROR_CAPABLE only applies
-    # to SPAN (port mirror) sessions. Skip the capability check for ERSPAN.
-    if not direction:
-        return True
+    """ Check if port mirror capability is supported for the given direction.
 
+    PORT_INGRESS_MIRROR_CAPABLE / PORT_EGRESS_MIRROR_CAPABLE only apply to SPAN
+    (port mirror) sessions. Callers should not invoke this for ERSPAN sessions.
+    Absent STATE_DB keys (None) are treated as supported for backward compatibility
+    with platforms that do not populate the SWITCH_CAPABILITY table.
+    """
     state_db = SonicV2Connector(use_unix_socket_path=True, namespace=namespace)
     state_db.connect(state_db.STATE_DB, False)
     entry_name = "SWITCH_CAPABILITY|switch"
 
-    if direction in ['rx', 'both']:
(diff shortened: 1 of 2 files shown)


config/main.py
```diff
@@ -1217,13 +1219,15 @@ def validate_mirror_session_config(config_db, session_name, dst_port, src_port,
...
-
-    for ns in namespace_set:
-        if not is_port_mirror_capability_supported(direction, namespace=ns):
-            ctx.fail("Error: Port mirror direction '{}' is not supported by the ASIC".format(
-                direction if direction else 'both'))
+    # Check port mirror capability before allowing configuration.
+    # ERSPAN sessions (dst_port=None) use src/dst IPs, not ports; the
+    # PORT_INGRESS/EGRESS_MIRROR_CAPABLE flags only apply to SPAN sessions.
+    is_erspan = dst_port is None
+    if not is_erspan:
```
stephenxs, line 1226: so basically, the fix is to avoid checking mirror capabilities when dst port is None, right?

bingwang-ms: Correct. It's because we can't assume all vendors will set `PORT_INGRESS/EGRESS_MIRROR_CAPABLE` to `true`. So it probably leads to the issue we saw on vs.
Since we already have capability check for erspan, I think we can remove the additional checker for erspan.

qiluo-msft (approved):

PR #4345, opened 2026-03-10
Fix unit tests

william8545:
#### What I did

Fixed unit test failures caused by mock table changes from multiple PRs.

#### How I did it

Updated test expected outputs to match the current mock config_db data (additional VLANs).
Updated test assertions to handle the actual error messages from the current mock data state.
Switched test target interfaces to ones that are not VLAN/portchannel members in the current mock data.

#### How to verify it

Run the full unit test suite and confirm all tests pass.


oleksandrivantsiv:
We are trying to understand this now

PR #4348, opened 2026-03-11
config: allow golden config to override mac, platform, asic_id

securely1g:
## Description
Currently `generate_sysinfo()` unconditionally overwrites `mac`, `platform`, and `asic_id` in the golden config with values from the running config or hardware detection. This prevents users from intentionally overriding these fields via `config load_minigraph -o` or `config override-config-table`.

### Problem
When a user provides an explicit `mac` in their golden config (e.g., for MAC migration, virtual environments, or chassis scenarios), it gets silently overwritten before the override is applied:

1. `load_minigraph` calls `sonic-cfggen -H` which sets MAC from hardware
2. `override_config_by()` → `override_config_table()` → `generate_sysinfo()`
3. `generate_sysinfo()` reads MAC from running config (set by step 1) and **unconditionally overwrites** the golden config's MAC

### Fix
Change `generate_sysinfo()` to only backfill `mac`, `platform`, and `asic_id` when they are **not explicitly present** in the golden config. If the golden config provides these values, they are preserved.

This maintains backward compatibility — golden configs without these fields (the common case) still get auto-populated from hardware/running config.

## Motivation and Context
Users cannot override `DEVICE_METADATA.localhost.mac` via golden config override, even when explicitly specified.

## How Has This Been Tested?
Code review and manual verification of the logic change.

## Type of change
- [x] Bug fix

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -2635,10 +2635,14 @@ def generate_sysinfo(cur_config, config_input, ns=None):
         asic_name = multi_asic.get_asic_id_from_name(ns)
         asic_id = multi_asic.get_asic_device_id(asic_name)
 
-    device_metadata['localhost']['mac'] = mac.rstrip('\n')
-    device_metadata['localhost']['platform'] = platform.rstrip('\n')
+    # Only backfill sysinfo fields if not explicitly provided in golden config
+    if 'mac' not in device_metadata['localhost']:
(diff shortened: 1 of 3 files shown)


securely1g:
@BYGX-wcr Could you please review this PR? Thanks!

BYGX-wcr (approved):
Good

PR #4354, opened 2026-03-13
Add CLI commands for bgp aggregate feature

nanali-msft:
#### What I did
Add CLI commands for bgp aggregate feature
The HLD design is at https://github.com/sonic-net/SONiC/blob/master/doc/BGP/BGP-route-aggregation-with-bbr-awareness.md#cli-design
image

ADO:37528013
#### How I did it
Files Modified
	1. utilities_common/bgp.py — Added CFG_BGP_AGGREGATE_ADDRESS = "BGP_AGGREGATE_ADDRESS" constant
	2. config/bgp_cli.py — Added:
	• validate_ip_prefix() callback for IP prefix validation
	• AGGREGATE_ADDRESS click group (config bgp aggregate-address)
	• add command with options: --bbr-required, --summary-only, --as-set, --aggregate-address-prefix-list, --contributing-address-prefix-list
	• remove command taking an address argument
	• Both commands check for duplicate/missing entries before modifying Config DB
	3. config/main.py — Registered bgp_cli.AGGREGATE_ADDRESS under the bgp group
	4. show/bgp_cli.py — Added show_aggregate_address(db, af) helper that:
	  • Reads BGP_AGGREGATE_ADDRESS from Config DB
	  • Reads state from State DB when available
	  • Filters entries by address family (ipv4/ipv6)
	  • Displays flags: A=As Set, B=BBR Required, S=Summary Only
	5. show/bgp_frr_v4.py — Added aggregate-address subcommand under show ip bgp
	6. show/bgp_frr_v6.py — Added aggregate-address subcommand under show ipv6 bgp

Files Created
	1. tests/bgp_aggregate_address_test.py — Unit tests for add/remove/show commands (10 test cases)
	2. tests/bgp_input/mock_config/aggregate_address.json — Mock Config DB with v4+v6 entries
	3. tests/bgp_input/mock_config/aggregate_address_empty.json — Empty mock Config DB
	4. tests/bgp_input/assert_show_output.py — Expected show output strings for v4, v6, and empty cases

#### How to verify it

1. Add UT
2. Verified on DUT devices:
```
admin@sonic:~$ show ip bgp aggregate-address
Flags: A - As Set, B - BBR Required, S - Summary Only

Prefix          State    Option Flags    Aggregate Address Prefix List    Contributing Address Prefix List
--------------  -------  --------------  -------------------------------  ----------------------------------
192.168.0.0/24  Active   B
admin@sonic:~$ sudo config bgp aggregate-address add 192.168.0.0/24 --summary-only --as-set
Usage: config bgp aggregate-address add [OPTIONS] ADDRESS
Try 'config bgp aggregate-address add -h' for help.

Error: Aggregate address '192.168.0.0/24' already exists

admin@sonic:~$ sudo config bgp aggregate-address remove 192.168.0.0/24
admin@sonic:~$ sudo config bgp aggregate-address add 192.168.0.0/24 --summary-only --as-set
admin@sonic:~$ show ip bgp aggregate-address
Flags: A - As Set, B - BBR Required, S - Summary Only

Prefix          State    Option Flags    Aggregate Address Prefix List    Contributing Address Prefix List
--------------  -------  --------------  -------------------------------  ----------------------------------
192.168.0.0/24  Active   A,S

admin@sonic:~$ sudo config bgp aggregate-address add fc00:1::/64 --bbr-required --aggregate-address-prefix-list AGG_ROUTE_V4 --contributing-address-prefix-list CONTRIBUTING_ROUTE_V4
admin@sonic:~$ show ip bgp aggregate-address
Flags: A - As Set, B - BBR Required, S - Summary Only

Prefix          State    Option Flags    Aggregate Address Prefix List    Contributing Address Prefix List
--------------  -------  --------------  -------------------------------  ----------------------------------
192.168.0.0/24  Active   A,S

admin@sonic:~$ show ipv6 bgp aggregate-address
Flags: A - As Set, B - BBR Required, S - Summary Only

Prefix       State    Option Flags    Aggregate Address Prefix List    Contributing Address Prefix List
-----------  -------  --------------  -------------------------------  ----------------------------------
fc00:1::/64  Active   B               AGG_ROUTE_V4                     CONTRIBUTING_ROUTE_V4
admin@sonic:~$ sudo config bgp aggregate-address remove fc00:1::/64
admin@sonic:~$ show ipv6 bgp aggregate-address
Flags: A - As Set, B - BBR Required, S - Summary Only

Prefix    State    Option Flags    Aggregate Address Prefix List    Contributing Address Prefix List
--------  -------  --------------  -------------------------------  ----------------------------------
admin@sonic:~$ sudo config bgp aggregate-address add fc00:1::/64 --bbr-required --aggregate-address-prefix-list AGG_ROUTE_V4 --contributing-address-prefix-list CONTRIBUTING_ROUTE_V4
admin@sonic:~$ sudo config bgp aggregate-address add fc00:2::/64 --bbr-required --aggregate-address-prefix-list AGG_ROUTE_V4 --contributing-address-prefix-list CONTRIBUTING_ROUTE_V4

admin@sonic:~$sudo config bgp aggregate-address add fc00:9::/64   --bbr-required   --aggregate-address-prefix-list AGG_ROUTE_V4   --contributing-address-prefix-list *$CONTRIBUTING_ROUTE_V4_#$#$#
Error: '*#00' is invalid for contributing-address-prefix-list: only alphanumeric characters, underscores and hyphens are allowed

admin@sonic:~$ show ipv6 bgp aggregate-address
Flags: A - As Set, B - BBR Required, S - Summary Only

Prefix       State    Option Flags    Aggregate Address Prefix List    Contributing Address Prefix List
-----------  -------  --------------  -------------------------------  ----------------------------------
fc00:1::/64  Active   B               AGG_ROUTE_V4                     CONTRIBUTING_ROUTE_V4
fc00:2::/64  Active   B               AGG_ROUTE_V4                     CONTRIBUTING_ROUTE_V4
admin@sonic:~$ show ipv6 bgp -h
Usage: show ipv6 bgp [OPTIONS] COMMAND [ARGS]...

  Show IPv6 BGP (Border Gateway Protocol) information

Options:
  -?, -h, --help  Show this message and exit.

Commands:
  aggregate-address  Show IPv6 BGP aggregate addresses
  neighbors          Show IPv6 BGP neighbors
  network            Show BGP ipv6 network
  summary            Show summarized information of IPv6 BGP state
  vrf                Show IPv6 BGP information for a given VRF
admin@sonic:~$ sudo config bgp -h
Usage: config bgp [OPTIONS] COMMAND [ARGS]...

  BGP-related configuration tasks

Options:
  -?, -h, --help  Show this message and exit.

Commands:
  aggregate-address  Configure BGP aggregate addresses
  device-global      Configure BGP device global state
  remove             Remove BGP neighbor configuration from the device
  shutdown           Shut down BGP session(s)
  startup            Start up BGP session(s)
```

### DIFF
--- a/config/bgp_cli.py
+++ b/config/bgp_cli.py
@@ -1,3 +1,4 @@
+import re
 import click
 import ipaddress
 import utilities_common.cli as clicommon
@@ -199,12 +200,30 @@ def DEVICE_GLOBAL_WCMP_DISABLED(ctx, db):
 #
 
 
+PREFIX_LIST_PATTERN = re.compile(r'^[0-9a-zA-Z_-]*$')
+PREFIX_LIST_MAX_LEN = 128
+
+
 def validate_ip_prefix(ctx, param, value):
     """ Validate that the argument is a valid IP prefix """
     try:
         return str(ipaddress.ip_network(value, strict=False))
     except ValueError:
         raise click.BadParameter("'{}' is not a valid IP prefix".format(value))
+
+
+def validate_prefix_list_name(name, field_name):
+    """ Validate prefix list name against YANG schema constraints.
+    Pattern: [0-9a-zA-Z_-]*, max length: 128
+    """
+    if len(name) > PREFIX_LIST_MAX_LEN:
+        raise click.ClickException(
+            "'{}' is invalid for {}: length exceeds {}".format(
+                name, field_name, PREFIX_LIST_MAX_LEN))
+    if not PREFIX_LIST_PATTERN.match(name):
+        raise click.ClickException(
+            "'{}' is invalid for {}: only alphanumeric characters, "
+            "underscores and hyphens are allowed".format(name, field_name))
 
 
 @click.group(
@@ -239,6 +258,12 @@ def AGGREGATE_ADDRESS_ADD(ctx, db, address, bbr_required, summary_only, as_set,
 
     table = CFG_BGP_AGGREGATE_ADDRESS
     key = address
+
+    # Validate prefix list names against YANG schema
+    validate_prefix_list_name(aggregate_address_prefix_list,
+                              "aggregate-address-prefix-list")
+    validate_prefix_list_name(contributing_address_prefix_list,
+                              "contributing-address-prefix-list")
 
     # Check if entry already exists
     cfg = db.cfgdb.get_config()
--- a/tests/bgp_aggregate_address_test.py
+++ b/tests/bgp_aggregate_address_test.py
@@ -362,4 +362,124 @@ class TestBgpAggregateAddress:
 
         assert result.exit_code != SUCCESS
 
+    # ---------- PREFIX LIST NAME VALIDATION ---------- #
 
+    def test_config_aggregate_address_add_invalid_prefix_list_pattern(self):
+        """Verify that prefix list names with invalid characters are rejected"""
+        db = Db()
+        runner = CliRunner()
+
+        result = runner.invoke(
+            config.config.commands["bgp"].commands["aggregate-address"].
+            commands["add"],
+            ["192.168.0.0/24",
+             "--aggregate-address-prefix-list", "bad!name@#"],
+            obj=db
+        )
+
+        logger.debug("\n" + result.output)
+        logger.debug(result.exit_code)
+
+        assert result.exit_code != SUCCESS
+        assert "invalid" in result.output.lower()
+
+    def test_config_aggregate_address_add_invalid_contributing_prefix_list_pattern(self):
+        """Verify that contributing prefix list names with invalid characters are rejected"""
+        db = Db()
+        runner = CliRunner()
+
+        result = runner.invoke(
+            config.config.commands["bgp"].commands["aggregate-address"].
+            commands["add"],
+            ["192.168.0.0/24",
+             "--contributing-address-prefix-list", "has spaces"],
+            obj=db
+        )
+
+        logger.debug("\n" + result.output)
+        logger.debug(result.exit_code)
+
+        assert result.exit_code != SUCCESS
+        assert "invalid" in result.output.lower()
+
+    def test_config_aggregate_address_add_prefix_list_too_long(self):
+        """Verify that prefix list names exceeding 128 characters are rejected"""
+        db = Db()
+        runner = CliRunner()
+
+        long_name = "a" * 129
+        result = runner.invoke(
+            config.config.commands["bgp"].commands["aggregate-address"].
+            commands["add"],
+            ["192.168.0.0/24",
+             "--aggregate-address-prefix-list", long_name],
+            obj=db
+        )
+
+        logger.debug("\n" + result.output)
+        logger.debug(result.exit_code)
+
+        assert result.exit_code != SUCCESS
+        assert "length" in result.output.lower()
+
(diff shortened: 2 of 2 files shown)


config/bgp_cli.py
```diff
@@ -190,3 +192,96 @@ def DEVICE_GLOBAL_WCMP_DISABLED(ctx, db):
...
+
+#
+# BGP aggregate-address ------------------------------------------------------------------------------------------------
+#
+
+
+def validate_ip_prefix(ctx, param, value):
+    """ Validate that the argument is a valid IP prefix """
+    try:
+        ipaddress.ip_network(value, strict=False)
```
Blueve: This validates with `ipaddress.ip_network(..., strict=False)` but returns the original string unchanged. That means a user can add `10.1.1.5/24`, which is accepted, but the DB key becomes `10.1.1.5/24` instead of the canonical network form. Later removing `10.1.1.0/24` will fail even though it is the same network. Please either normalize here (for example `return str(ipaddress.ip_network(value, strict=False))`) or reject host bits by using `strict=True`.

config/bgp_cli.py
```diff
@@ -190,3 +192,96 @@ def DEVICE_GLOBAL_WCMP_DISABLED(ctx, db):
...
+    data = {
+        "bbr-required": "true" if bbr_required else "false",
+        "summary-only": "true" if summary_only else "false",
+        "as-set": "true" if as_set else "false",
+        "aggregate-address-prefix-list": aggregate_address_prefix_list,
+        "contributing-address-prefix-list": contributing_address_prefix_list,
+    }
+
+    try:
+        db.cfgdb.set_entry(table, key, data)
```
Blueve, line 282: These aggregate-address add/remove handlers bypass the existing `update_entry_validated()` helper in this module and write directly with `db.cfgdb.set_entry()`. That skips the validated update path this file already uses for BGP config and can let invalid values reach CONFIG_DB (for example bad prefix-list names/lengths) until some later component rejects them. Please route these writes through the validated helper so the new commands get the same schema enforcement as the rest of this module.

config/bgp_cli.py
nali1126: good suggestion, will fix soon

config/bgp_cli.py
nali1126: Thanks for the review. update_entry_validated() doesn't actually perform YANG schema validation — it's an update-in-place helper for modifying fields on an existing entry (used by TSA/WCMP on the singleton BGP_DEVICE_GLOBAL key). Both code paths ultimately call the same db.set_entry().

The aggregate-address commands need different semantics:

- add must fail if the entry already exists (create-or-error), whereas update_entry_validated(create_if_not_exists=True) would silently merge into an existing entry
- remove must delete the entire entry, which the helper doesn't support

Routing these through the helper would require either reworking it to handle create/delete semantics (breaking its current contract) or adding special-case logic that duplicates what the handlers already do. The current approach matches the pattern used by other table-entry CRUD commands in config/main.py.

nali1126: Added validate_prefix_list_name align the YANG Model schema definition of aggregate-address-prefix-list and contributing-address-prefix-list

PR #4355, opened 2026-03-13
clear: make --namespace optional for arp and ndp commands

oleksandrivantsiv:
- Set required=False for -n/--namespace option on arp and ndp
- Add test_clear_arp_multi_asic_without_namespace and test_clear_ndp_multi_asic_without_namespace to verify default namespace when -n is omitted

#### What I did
Made the `-n`/`--namespace` option **optional** for `sonic-clear arp` and `sonic-clear ndp` on multi-ASIC. When `-n` is omitted, the command runs in the default namespace (same behavior as before).

#### How I did it
- In `clear/main.py`, set `required=False` for the `--namespace` option on the `arp` and `ndp` commands (replacing `required=multi_asic.is_multi_asic()`).
- In `tests/clear_test.py`, added:
  - `test_clear_arp_multi_asic_without_namespace` – multi-ASIC, no `-n`, expects ARP flush in default namespace.
  - `test_clear_ndp_multi_asic_without_namespace` – multi-ASIC, no `-n`, expects NDP flush in default namespace.

#### How to verify it
- **Single-ASIC:** Run `sonic-clear arp` and `sonic-clear ndp` (with or without `-n`); behavior unchanged.
- **Multi-ASIC:** Run `sonic-clear arp` and `sonic-clear ndp` without `-n` – they run in the default namespace without error; run with `-n asic0` (or another namespace) to target a specific namespace.
- Run tests: `pytest tests/clear_test.py::TestClearArp tests/clear_test.py::TestClearNdp -v`

N/A — no change to command output; only `-n` is now optional on multi-ASIC.

N/A — no change to command output.


saiarcot895:
Thank you!

PR #4361, opened 2026-03-16
Inform user that reload is required after suppress-fib-pending change

mike-dubrovsky:
## Summary

suppress-fib-pending config changes require a reload to take effect.

Changes:
- config/main.py: inform user that reload is required after
  changing suppress-fib-pending

This change needs to go together with
https://github.com/sonic-net/sonic-swss/pull/4333
https://github.com/sonic-net/sonic-buildimage/pull/26151
https://github.com/sonic-net/sonic-mgmt/pull/22916
https://github.com/sonic-net/SONiC/pull/2335

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -2866,9 +2866,22 @@ def suppress_pending_fib(db, state):
     if multi_asic.get_num_asics() > 1:
         namespace_list = multi_asic.get_namespaces_from_linux()
 
+    changed = False
     for ns in namespace_list:
         config_db = db.cfgdb_clients[ns]
+        entry = config_db.get_entry('DEVICE_METADATA', 'localhost')
+        if entry.get('suppress-fib-pending') == state:
+            continue
         config_db.mod_entry('DEVICE_METADATA', 'localhost', {"suppress-fib-pending": state})
+        changed = True
+
+    if changed:
+        click.echo(f"""suppress-fib-pending set to '{state}' in CONFIG_DB. This does NOT take effect until a config reload.
+Persist and apply with:
+    config save -y
+    config reload -y""")
+    else:
+        click.echo(f"suppress-fib-pending is already '{state}'; no change made.")
 
 #
 # 'yang_config_validation' command ('config yang_config_validation ...')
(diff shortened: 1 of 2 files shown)


config/main.py
```diff
@@ -2870,6 +2870,10 @@ def suppress_pending_fib(db, state):
         config_db = db.cfgdb_clients[ns]
         config_db.mod_entry('DEVICE_METADATA', 'localhost', {"suppress-fib-pending": state})
 
+    click.echo("""Wrote %s suppress-fib-pending into CONFIG_DB, reload required to apply the configuration:
```
deepak-singhal0408: Wording: "reload required to apply the configuration" can be misread as a **device reboot**. Suggest saying **config reload** explicitly, which also makes the "doesn't take effect until reload" contract clear at the point of use. Proposed:

```
suppress-fib-pending set to '<state>' in CONFIG_DB. This does NOT take effect until a config reload.
Persist and apply with:
    config save -y
    config reload -y
```

deepak-singhal0408: This notice prints on every invocation, including when `state` already matches the current CONFIG_DB value — which can misleadingly prompt the operator to reload when nothing changed. Consider reading the current `suppress-fib-pending` value first and, when it already equals the requested `state`, skip the write and instead print e.g. `suppress-fib-pending is already <state>, no change` (reserving the reload notice for an actual change).

mike-dubrovsky: done

mike-dubrovsky: done

config/main.py
```diff
@@ -2870,9 +2866,20 @@ def suppress_pending_fib(db, state):
     if multi_asic.get_num_asics() > 1:
         namespace_list = multi_asic.get_namespaces_from_linux()
 
+    changed = False
     for ns in namespace_list:
         config_db = db.cfgdb_clients[ns]
+        entry = config_db.get_entry('DEVICE_METADATA', 'localhost')
+        if entry.get('suppress-fib-pending') == state:
```
deepak-singhal0408, line 2873: Suggest an explicit message here, 
e.g. 
suppress-fib-pending is already '{state}'; no change made.

mike-dubrovsky: done

deepak-singhal0408:
@mike-dubrovsky , please fix the Ci failure. 
the new testcase contradicts the no-change early-return.

deepak-singhal0408 (approved):

PR #4365, opened 2026-03-17
Fix migrate_sonic_packages() crash on symlink resolv.conf

william8545:
#### What I did

Fixed `sonic-installer install` failing during `migrate_sonic_packages()` when `/etc/resolv.conf` in the new image is a symlink to `/run/resolvconf/resolv.conf`.

The failure occurs because the `cp` command at `main.py:386` follows the symlink through the overlay mount. Since the symlink target is an absolute path, it resolves to the **host's** `/run/resolvconf/resolv.conf` — the same file as the source. `cp` detects same source and destination inode and exits with:

```
cp: '/etc/resolv.conf' and '/tmp/image-<version>-fs/etc/resolv.conf' are the same file
```

This was introduced by the `build_debian.sh` change that replaced `touch` with `ln -sf /run/resolvconf/resolv.conf` for `/etc/resolv.conf` in the image filesystem.

#### How I did it

Check whether `/etc/resolv.conf` in the chroot is a symlink or a regular file, and handle each case appropriately:

- **Symlink** (images with `resolvconf` package installed): Read the symlink target via `readlink` (e.g. `/run/resolvconf/resolv.conf`), then create the target file inside the chroot with the host's DNS content. The symlink then resolves correctly inside the chroot. This avoids touching the symlink itself, so the overlay upper dir's `etc/resolv.conf` is never modified and the new image boots with the symlink intact. No cleanup is needed — the target lives under `/run`, which is a tmpfs recreated at every boot.

- **Regular file** (images without `resolvconf`, or where the build process explicitly creates a regular file via `touch`): Overwrite directly with the host's DNS content. No backup/restore is needed — the original file is empty (cleared during build), and after reboot the `resolv-config` service reconfigures DNS from CONFIG_DB.

The previous backup-overwrite-restore pattern has been removed since it is unnecessary in both cases.

#### How to verify it

1. Start with a switch running an image where `/etc/resolv.conf` is a symlink:
   ```bash
   # Confirm symlink exists
   ls -la /etc/resolv.conf
   # Expected: /etc/resolv.conf -> /run/resolvconf/resolv.conf

   # Ensure only one image is installed (clean state)
   sudo sonic-installer list
   # If the target image is already present, remove it:
   sudo sonic-installer remove <target-image> -y
   ```

2. Run `sonic-installer install` with an image that also has the symlink:
   ```bash
   sudo sonic-installer install <image-path> -y
   ```

3. Verify:
   - Installation completes without `cp: ... are the same file` error
   - `sonic-installer list` shows the new image as default
   - After reboot, `/etc/resolv.conf` is still a symlink to `/run/resolvconf/resolv.conf`
   - DNS resolution works

### DIFF
--- a/sonic_installer/main.py
+++ b/sonic_installer/main.py
@@ -325,7 +325,6 @@ def migrate_sonic_packages(bootloader, binary_image_version):
     DOCKERD_SOCK = "docker.sock"
     VAR_RUN_PATH = "/var/run/"
     RESOLV_CONF_FILE = os.path.join("etc", "resolv.conf")
-    RESOLV_CONF_BACKUP_FILE = os.path.join("/", TMP_DIR, "resolv.conf.backup")
 
     packages_file = "packages.json"
     packages_path = os.path.join(PACKAGE_MANAGER_DIR, packages_file)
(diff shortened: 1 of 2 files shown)


vmittal-msft:
@saiarcot895 please help approve

PR #3001, opened 2023-09-28
Link event damping config CLI.

Ashish1805:
- Add the config cli for link event damping.

#### What I did
Added new config cli for link event damping and 

#### How I did it
Added damping group in interface and added sub-commands algo and aied-param under damping group.

#### How to verify it
Unit test

#### Previous command output (if the output of a command-line utility has changed)
N/A
#### New command output (if the output of a command-line utility has changed)
N/A

PR #4367, opened 2026-03-17
[Link Event Damping] Add config/show/clear CLI commands

DendroLabs:
## Summary

Supersedes #3001 by @Ashish1805. Adds complete CLI support for the Link Event Damping feature ([HLD: sonic-net/SONiC#1071](https://github.com/sonic-net/SONiC/pull/1071)).

**All review feedback from #3001 has been addressed:**
- Uses `click.IntRange` for parameter validation (@Junchao-Mellanox)
- Rebased onto current master (@prsunny)
- Added `show` command that was flagged as missing (@Junchao-Mellanox)

**Commands added:**

`config interface dampening <interface> enable [options]`
- Enables AIED dampening with industry-standard defaults (half-life=5s, reuse=1000, suppress=2000, max-suppress=20s, penalty=1000)
- `--monitor` flag enables monitor-only mode (RFC 7196 "Calculate But Do Not Damp")
- Validates: interface exists, reuse < suppress, half-life <= max-suppress-time

`config interface dampening <interface> disable`
- Disables dampening on the interface

`show interfaces dampening [interface_name]`
- Displays config parameters and operational state (current penalty, suppressed status, time remaining)
- Shows "(Mon)" for monitor-only mode interfaces
- Shows per-interface dampening counters when querying a specific port

`sonic-clear interfaces dampening [interface_name]`
- Resets penalty to 0 and immediately unsuppresses
- Standard across Cisco/Juniper; recommended by RFC 2439

**New features not in any other NOS vendor:**
- **Monitor-only mode** (`--monitor`): Calculates penalties and emits syslog but does NOT suppress events. Allows safe parameter tuning in production. Recommended by RFC 7196; no vendor has implemented this.
- **Syslog integration**: Syncd emits WARNING on suppress, NOTICE on unsuppress (addresses #1 industry complaint about silent suppression)

## Test plan
- [ ] `config interface dampening enable` writes correct fields to CONFIG_DB PORT table
- [ ] `config interface dampening disable` resets all fields
- [ ] `config interface dampening enable --monitor` sets algorithm to "aied-monitor"
- [ ] Parameter validation rejects invalid combinations (reuse >= suppress, etc.)
- [ ] `show interfaces dampening` displays configured interfaces
- [ ] `show interfaces dampening <port>` shows counters
- [ ] `sonic-clear interfaces dampening` writes clear flag to STATE_DB
- [ ] Existing sonic-utilities tests pass (no regressions)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Ashish Singh <ashish.singh@google.com>

### DIFF
--- a/clear/main.py
+++ b/clear/main.py
@@ -787,9 +787,15 @@ def dampening(interface_name):
     port_table = config_db.get_table("PORT")
 
     if interface_name:
+        if clicommon.get_interface_naming_mode() == "alias":
+            alias = interface_name
+            interface_name = clicommon.InterfaceAliasConverter().alias_to_name(interface_name)
+            if interface_name == alias:
+                click.echo("Error: invalid interface alias {}".format(alias))
+                sys.exit(1)
         if interface_name not in port_table:
             click.echo("Error: Interface {} does not exist".format(interface_name))
-            raise SystemExit(1)
+            sys.exit(1)
         ports_to_clear = [interface_name]
     else:
         ports_to_clear = []
--- a/clear/main.py
+++ b/clear/main.py
@@ -11,6 +11,7 @@ from flow_counter_util.route import exit_if_route_flow_counter_not_support
 from utilities_common import util_base
 from show.plugins.pbh import read_pbh_counters
 from config.plugins.pbh import serialize_pbh_counters
+from swsscommon.swsscommon import SonicV2Connector, ConfigDBConnector
 from . import plugins
 from . import stp
 # This is from the aliases example:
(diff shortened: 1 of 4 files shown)


DendroLabs:
Hi @prsunny @Junchao-Mellanox — friendly ping on this PR. The companion sairedis PRs are progressing (sonic-net/sonic-sairedis#1798 merged, sonic-net/sonic-sairedis#1935 open for review), and this CLI PR addresses all the feedback from the original #3001. CI was green when submitted. Would appreciate a review when you get a chance. Thanks!

prsunny (approved):

prsunny:
@DendroLabs , could you check copilot comments and address them.

clear/main.py
```diff
@@ -763,5 +764,61 @@ def asic_sdk_health_event(db, namespace):
...
+
+@interfaces.command()
+@click.argument('interface_name', metavar='<interface_name>', required=False)
+def dampening(interface_name):
+    """Clear link event dampening state (reset penalty to 0).
+
+    Without an interface name, clears dampening on all interfaces.
+    The interface is immediately unsuppressed if currently damped.
+    """
+    config_db = ConfigDBConnector()
```
qiluo-msft, line 785: host CONFIG_DB only — on multi-ASIC, PORT lives in per-namespace CONFIG_DB and that's where orchagent reads. Same issue with `ctx.obj['config_db']` in the `enable`/`disable` commands. See how `config interface ip add` resolves the namespace via `multi_asic.get_port_namespace(port)` / `db.cfgdb_clients[ns]`.

prsunny:
@DendroLabs , could you address comment?

DendroLabs:
Thanks @prsunny. Before I push the namespace fixes, I want to surface something I ran into while wiring the CLI against the now merged syncd implementation (#1906), because I think it is a design question worth settling together rather than me papering over it in the CLI.

**What I am trying to do:** make `show`/`clear interfaces dampening` work end to end as the operator facing front door for this feature, the same way operators already use `show`/`clear` for every other port level function.

**Where the CLI and the merged syncd schema do not line up today:**

1. **STATE_DB key format.** #1906 writes the `LINK_EVENT_DAMPING_STATS` table keyed by the raw SAI port VID (`oid:0x...`). Every other operator facing STATE_DB port table (`PORT_TABLE`, transceiver, and so on) is keyed by port name (`Ethernet0`). Keying by VID means the CLI has to go through `COUNTERS_PORT_NAME_MAP` just to find a row, and anyone inspecting STATE_DB by hand cannot map entries back to ports. I would like to propose the stats table be keyed by port name for consistency and operator usability.

2. **Manual clear.** RFC 2439 section 4.8.6 calls out operator ability to manually clear damping, and Cisco and Juniper both expose it (`clear interface dampening`). The merged syncd does not yet have a handler for a clear request, so `sonic-clear interfaces dampening` has nothing to talk to. I would like to add that path so the clear command is real.

3. **Monitor only mode.** RFC 7196 recommends a dry run capability so operators can validate thresholds in production before any real suppression. My earlier work implemented this as an `aied-monitor` algorithm value (penalties and syslog computed, events not suppressed). The merged swss algorithm map only accepts `disabled` and `aied`, so there is currently no home for it. I think it is a meaningful safety feature and would like to find the right place for it.

**Why it matters:** without these, the CLI either reports `N/A` against a real switch (show), silently does nothing (clear), or rejects a documented mode (monitor). I would rather get the schema right than ship a front end that quietly does not match the back end.

That said, I am very happy to defer to you and @sivat6 on how to land this. If you would prefer I align the CLI to the existing VID keyed schema for now and treat the clear and monitor pieces as later follow ups, I am glad to do that and keep moving on the namespace fixes in parallel. If you would rather settle the schema first, I am happy to go whichever way fits best with how you see this feature evolving. Just let me know the direction you would like and I will follow it.

prsunny:
@DendroLabs , my ask is to address copilot comments and from @qiluo-msft .  not sure which one specifically you are referring to.

prsunny:
@DendroLabs , seems this PR is still pending on addressing comment. We would like to get this for 202611 branch. Can you fix the PR this week?

PR #4370, opened 2026-03-19
Fix docker daemon crash during package migration due to -H fd://

tirupatihemanth:
#### Why I did
[PR #25398](https://github.com/sonic-net/sonic-buildimage/pull/25398) changed the host's Docker config to use -H fd://. During image installation, sonic-installer copies these host options to run a temporary dockerd inside a chroot. Since systemd isn't running in the chroot to create and handover a socket dockerd crashes, causing sonic-package-manager migrate to fail.

#### What I did
Sanitized the copied DOCKER_OPTS so the temporary dockerd can start safely without systemd socket activation.
#### How I did it
- In sonic_installer/main.py (get_docker_opts), replaced fd:// with unix://.
- Updated mock arguments in test_sonic_installer.py to verify the sanitization logic.

#### How to verify it

Run sudo sonic-installer install <image.bin> and verify the installation completes successfully without failing during the package migration step.

### DIFF
--- a/tests/test_sonic_installer.py
+++ b/tests/test_sonic_installer.py
@@ -86,7 +86,12 @@ def test_install(run_command, run_command_or_raise, get_bootloader, swap, fs):
         call(["chroot", mounted_image_folder, "mount", "proc", "/proc", "-t", "proc"]),
         call(["chroot", mounted_image_folder, "mount", "sysfs", "/sys", "-t", "sysfs"]),
         call(["cp", f"{mounted_image_folder}/etc/default/docker", f"{mounted_image_folder}/tmp/docker_config_backup"]),
-        call(["sh", "-c", f"echo 'DOCKER_OPTS=\"$DOCKER_OPTS {' '.join(expected_dockerd_opts)}\"' >> {mounted_image_folder}/etc/default/docker"]), # dockerd started with added options as host dockerd
+        # dockerd started with unix:// instead of fd://
+        call([
+            "sh", "-c",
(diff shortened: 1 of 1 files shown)


dgsudharsan (approved):

tests/test_sonic_installer.py
```diff
@@ -35,6 +35,9 @@ def test_install(run_command, run_command_or_raise, get_bootloader, swap, fs):
     fs.create_file("/var/run/docker.pid", contents="15")
     fs.create_file("/proc/15/cmdline", contents="\x00".join(["dockerd"] + dockerd_opts))
 
+    # Expect fd:// to be replaced by unix://
+    expected_dockerd_opts = [opt if opt != "fd://" else "unix://" for opt in dockerd_opts]
```
saiarcot895, line 39: This is making the logic messier. Today, it's `fd://`, but what if there's something else tomorrow?

What about `unshare` as suggested in https://github.com/sonic-net/sonic-buildimage/issues/26152#issuecomment-4078825143?

StormLiangMS:
@tirupatihemanth could you check the comment from Saikrishna?

tests/test_sonic_installer.py
tirupatihemanth: HI @saiarcot895 this would cause regression when 
```
new_package.installed = False
new_package.default_reference = "not None"
```
We fetch the default reference from registry. This network access would break. 

https://github.com/sonic-net/sonic-utilities/blob/02b9e4856789337f9f69efeeb8ba7d9677d22a7c/sonic_package_manager/manager.py#L822

saiarcot895: I see, then this approach is the best way we have currently.

PR #4376, opened 2026-03-20
Fix counterpoll CLI when running in a namespace

peterbailey-arista:
Fixes issue https://github.com/sonic-net/sonic-utilities/issues/4375

Currently when using `sudo ip netns exec <namespace> counterpoll <args>` it will ignore the namespace and use the default namespace. This patch fixes that behavior to use namespace that the command is running in.

This is how the command worked previously but a regression was introduced breaking this behaviour.

#### What I did
I changed the default value of the -n arg to be the namespace we are running the command in.

#### How to verify it

I verified that with my change the namespace being run in will be the default namespace chosen.

```
$ sonic-db-cli -n asic0 CONFIG_DB hgetall "FLEX_COUNTER_TABLE|PG_DROP"
{'FLEX_COUNTER_STATUS': 'enable'}
$ sudo ip netns exec asic0 counterpoll pg-drop disable
$ sonic-db-cli -n asic0 CONFIG_DB hgetall "FLEX_COUNTER_TABLE|PG_DROP"
{'FLEX_COUNTER_STATUS': 'disable'}
```

 I also verified that the command exits gracefully if the user attempts to confuse it with 2 namespaces like so:
```
sudo ip netns exec asic0 counterpoll pg-drop -n asic1 disable
Usage: counterpoll pg-drop [OPTIONS] COMMAND [ARGS]...
Try 'counterpoll pg-drop --help' for help.

Error: Invalid value for '-n' / '--namespace': 'asic1' is not 'asic0'.
```

```
$ sonic-db-cli -n asic0 CONFIG_DB hgetall "FLEX_COUNTER_TABLE|PG_DROP"
{'FLEX_COUNTER_STATUS': 'enable'}
$ sudo ip netns exec asic0 counterpoll pg-drop disable
$ sonic-db-cli -n asic0 CONFIG_DB hgetall "FLEX_COUNTER_TABLE|PG_DROP"
{'FLEX_COUNTER_STATUS': 'enable'}
```

Seen above

### DIFF
--- a/counterpoll/main.py
+++ b/counterpoll/main.py
@@ -31,7 +31,7 @@ def is_dpu(db):
 
 
 def connect_to_db(namespace):
-    if namespace is None:
+    if not namespace:
         namespace = DEFAULT_NAMESPACE
     else:
         if not SonicDBConfig.isGlobalInit():
@@ -50,7 +50,8 @@ def cli():
 @cli.group()
 @click.option('-n', '--namespace', help='Namespace name',
               required=False,
-              type=click.Choice(multi_asic.get_namespace_list()))
+              type=click.Choice(multi_asic.get_namespace_list()),
+              default=multi_asic.get_current_namespace())
 @click.pass_context
 def queue(ctx, namespace):
     """ Queue counter commands """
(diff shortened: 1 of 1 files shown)


peterbailey-arista:
@saksarav-nokia There is sonic mgmt infrastructure that expects to be able to control the appropriate asics by prepending the network namespace.

https://github.com/sonic-net/sonic-mgmt/blob/98d24345964a12ec7c9174bc0687afc802630e74/tests/common/devices/sonic_asic.py#L38

The utilities CLIs are expected to work in this fashion

PR #4412, opened 2026-04-01
Allow batch flapping of interfaces on multi-asic systems via CLI

### LINKED ISSUE #4781: Enhancement: Add support for batch flapping of interfaces on multi-asic platforms
Currently, bulk interface startup and shutdown is not supported on multi-asic

As an enhancement, we could add support for shutdown and startup a batch of input interfaces on multi-asic systems. 

This will also fix lldp batch flap test case failure in sonic-mgmt : https://github.com/sonic-net/sonic-mgmt/issues/27579

arista-setu:
Context : currently, shutdown and startup of a batch of interfaces is not allowed multi asic. We have checks to return error saying it is not allowed. 

#### What I did

Enabled batch interface flapping (startup/shutdown) on multi-ASIC platforms by removing the blanket restriction and adding proper namespace validation instead.

Fixes: #4781 

### Backport request 

Requesting backport of same to 202511 and 202605 branch

#### How I did it

1. Created a new helper function `validate_interface_names_in_namespace()` that validates:
     - Interface names are valid and exist in config_db
     - On multi-ASIC systems, all interfaces belong to the correct namespace

  2. Refactored `startup()` and `shutdown()` commands to:
     - Replace the blanket multi-ASIC batch rejection with targeted namespace validation
     - Use a generic loop with `get_port_table_name()` helper to handle all interface types
  (PORT, PORTCHANNEL, VLAN_SUB_INTERFACE, LOOPBACK_INTERFACE) instead of duplicate
  table-by-table checks
     - Reduce code duplication across both functions

#### How to verify it

Verified command functions as expected on multi-asic

Previously the command errors out as follows : 
```
$ config interface startup Ethernet0,Ethernet4
Error: Interface range not supported in multi-asic platforms !!
```

Multi-asic behavior

Only success case : provided input interfaces ( single interface, comma-seperated list or range-based list ) that exist on provided namespace ( -n arg ) will successfully complete

```
admin@qzd554:~$ sudo config interface -n asic0 shutdown Ethernet0-16,Ethernet40
admin@qzd554:~$ show interfaces status Ethernet0-16,Ethernet40
  Interface                            Lanes    Speed    MTU    FEC    Alias    Vlan    Oper    Admin    Medium     Type           Vendor            Model    Asym PFC    Flaps
-----------  -------------------------------  -------  -----  -----  -------  ------  ------  -------  --------  -------  ---------------  ---------------  ----------  -------
  Ethernet0          72,73,74,75,76,77,78,79     400G   9100     rs     etp1  routed    down     down    Copper  QSFP-DD  Arista Networks  CAB-D-D-400G-1M
    off        4
  Ethernet8          80,81,82,83,84,85,86,87     400G   9100     rs     etp2  routed    down     down    Copper  QSFP-DD  Arista Networks  CAB-D-D-400G-1M
    off        4
 Ethernet16          88,89,90,91,92,93,94,95     400G   9100     rs     etp3  routed    down     down    Copper  QSFP-DD  Arista Networks  CAB-D-D-400G-1M
    off        4
 Ethernet40  112,113,114,115,116,117,118,119     400G   9100     rs     etp6  routed    down     down    Copper  QSFP-DD  Arista Networks  CAB-D-D-400G-1M
    off        4

admin@qzd554:~$ sudo config interface -n asic0 startup Ethernet0-16,Ethernet40
admin@qzd554:~$ show interfaces status Ethernet0-16,Ethernet40
  Interface                            Lanes    Speed    MTU    FEC    Alias    Vlan    Oper    Admin    Medium     Type           Vendor            Model    Asym PFC    Flaps
-----------  -------------------------------  -------  -----  -----  -------  ------  ------  -------  --------  -------  ---------------  ---------------  ----------  -------
  Ethernet0          72,73,74,75,76,77,78,79     400G   9100     rs     etp1  routed      up       up    Copper  QSFP-DD  Arista Networks  CAB-D-D-400G-1M
    off        3
  Ethernet8          80,81,82,83,84,85,86,87     400G   9100     rs     etp2  routed      up       up    Copper  QSFP-DD  Arista Networks  CAB-D-D-400G-1M
    off        3
 Ethernet16          88,89,90,91,92,93,94,95     400G   9100     rs     etp3  routed      up       up    Copper  QSFP-DD  Arista Networks  CAB-D-D-400G-1M
    off        3
 Ethernet40  112,113,114,115,116,117,118,119     400G   9100     rs     etp6  routed      up       up    Copper  QSFP-DD  Arista Networks  CAB-D-D-400G-1M
    off        3
```

Failure cases:

When resolved valid interfaces from input interfaces does not exists on provided namespace

```
admin@qzd554:~$ sudo config interface -n asic0 shutdown Ethernet0,Ethernet40,Ethernet128-152
Usage: config interface shutdown [OPTIONS] <interface_name>
Try 'config interface shutdown -h' for help.

Error: Interface(s) not in provided namespace 'asic0'. Interface(s) Ethernet144, Ethernet152 exist(s) on namespace 'asic1'

admin@qzd554:~$ sudo config interface -n asic0 shutdown Ethernet128-152
Usage: config interface shutdown [OPTIONS] <interface_name>
Try 'config interface shutdown -h' for help.

Error: Interface(s) not in provided namespace 'asic0'. Interface(s) Ethernet144, Ethernet152 exist(s) on namespace 'asic1'

sudo config interface -n asic0 shutdown Ethernet0,Ethernet8,Ethernet200
Usage: config interface shutdown [OPTIONS] <interface_name>
Try 'config interface shutdown -h' for help.

Error: Interface(s) not in provided namespace 'asic0'. Interface(s) Ethernet200 exist(s) on namespace 'asic1'
```

When resolved interfaces from input interfaces does not have any valid interfaces

```
admin@qzd554:~$ sudo config interface -n asic0 shutdown Ethernet1-7
Usage: config interface shutdown [OPTIONS] <interface_name>
Try 'config interface shutdown -h' for help.

Error: Provided interface(s) : 'Ethernet1, Ethernet2, Ethernet3, Ethernet4, Ethernet5, Ethernet6, Ethernet7' are invalid in namespace 'asic0'

admin@qzd554:~$ sudo config interface -n asic0 shutdown Ethernet137-143
Usage: config interface shutdown [OPTIONS] <interface_name>
Try 'config interface shutdown -h' for help.

Error: Provided interface(s) : 'Ethernet137, Ethernet138, Ethernet139, Ethernet140, Ethernet141, Ethernet142, Ethernet143' are invalid in namespace 'asic0'
```

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -36,7 +36,7 @@ from swsscommon import swsscommon
 from swsscommon.swsscommon import SonicV2Connector, ConfigDBConnector, ConfigDBPipeConnector, \
                                 isInterfaceNameValid, IFACE_NAME_MAX_LEN
 from utilities_common.db import Db
-from utilities_common.intf_filter import parse_interface_in_filter
+from utilities_common.intf_filter import expand_single_intf_filter
 from utilities_common import bgp_util
 import utilities_common.cli as clicommon
 from utilities_common.helper import get_port_pbh_binding, get_port_acl_binding, update_config
@@ -588,6 +588,109 @@ def get_port_namespace(port):
                 return namespace
 
     return None
+
+
+def _load_other_namespace_ports(namespace):
+    """Build a port-to-namespace map for all namespaces other than provided namespace."""
+    other_ns_ports = {}
+    if not multi_asic.is_multi_asic():
+        return other_ns_ports
+    ns_list = multi_asic.get_all_namespaces()
+    for ns in ns_list['front_ns'] + ns_list['back_ns']:
+        if ns == namespace:
+            continue
+        ns_db = ConfigDBConnector(use_unix_socket_path=True, namespace=ns)
+        ns_db.connect()
+        for port in ns_db.get_table('PORT'):
+            other_ns_ports[port] = ns
+        for pc in ns_db.get_table('PORTCHANNEL'):
+            other_ns_ports[pc] = ns
+    return other_ns_ports
+
+
+def get_interface_names_in_namespace(ctx, interface_name):
+    """Return existing interfaces selected by filter in the provided namespace."""
+    config_db = ctx.obj.get('config_db')
+    namespace = ctx.obj.get('namespace', DEFAULT_NAMESPACE)
+    cross_ns_intfs = []
+    invalid_intfs = []
+    valid_intfs = []
+    seen_intfs = set()
+    other_ns_ports = None
+
+    # Expand all possible comma seperated input interface(s).
+    for intf_filter in interface_name.split(','):
+        intf_filter = intf_filter.strip()
+        try:
+            intf_fs, is_range = expand_single_intf_filter(intf_filter)
+        except ValueError as e:
+            log.log_warning("Invalid interface filter '{}' in namespace '{}': {}".format(
+                intf_filter, namespace, str(e)))
+            ctx.fail(str(e))
+
+        missing_intfs = []
+        valid_in_filter = []
+        cross_ns_in_filter = []
+        for intf in intf_fs:
+            if intf in seen_intfs:
+                continue
+            seen_intfs.add(intf)
+            # Check if interface is in target namespace.
+            if interface_name_is_valid(config_db, intf):
+                valid_in_filter.append(intf)
+            else:
+                if other_ns_ports is None:
+                    # Lazy init on first miss to skip uneccessary config_db connections
+                    # when all input interfaces are valid.
+                    other_ns_ports = _load_other_namespace_ports(namespace)
+                # Case when interface is present in other namespace.
+                if intf in other_ns_ports:
+                    cross_ns_in_filter.append((intf, other_ns_ports[intf]))
+                # Sparse gap in range which is not present in any namespace.
+                # This interfaces are tolerated if range has other valid hits.
+                elif is_range:
+                    missing_intfs.append(intf)
+                else:
+                    invalid_intfs.append(intf)
+
+        cross_ns_intfs.extend(cross_ns_in_filter)
+
+        # Range that matched nothing — promote missing to invalid for proper error reporting.
+        if is_range and not valid_in_filter and not cross_ns_in_filter:
+            invalid_intfs.extend(missing_intfs)
+            missing_intfs = []
+
+        valid_intfs.extend(valid_in_filter)
+
+        if missing_intfs:
+            log.log_info("Interface range '{}' in namespace '{}' skipped missing interface(s): {}".format(
+                intf_filter, namespace, ', '.join(missing_intfs)))
+
+        if valid_in_filter:
+            log.log_info("Interface filter '{}' in namespace '{}' resolved to existing interface(s): {}".format(
+                intf_filter, namespace, ', '.join(valid_in_filter)))
+
+    # Fail with clear error if valid interfaces existed on different namespace.
+    if cross_ns_intfs:
+        by_ns = {}
+        for intf, ns in cross_ns_intfs:
+            by_ns.setdefault(ns, []).append(intf)
+        parts = []
+        for ns, intfs in by_ns.items():
+            parts.append("Interface(s) {} exist(s) on namespace '{}'".format(', '.join(intfs), ns))
+        msg = "Interface(s) not in provided namespace '{}'. {}".format(namespace, '. '.join(parts))
+        log.log_warning(msg)
+        ctx.fail(msg)
+
+    # Fail when input includes interface(s) that don't exist on any namespace.
+    if invalid_intfs:
+        log.log_warning("Invalid interface(s) in namespace '{}': {}".format(
+            namespace, ', '.join(invalid_intfs)))
+        ctx.fail("Provided interface(s) : '{}' are invalid in namespace '{}'".format(
+            ', '.join(invalid_intfs), namespace))
+
+    return valid_intfs
+
 
 def del_interface_bind_to_vrf(config_db, vrf_name):
     """del interface bind to vrf
@@ -5337,6 +5440,8 @@ def fast_linkup(ctx, interface_name, mode, verbose):
     if verbose:
         command += ['-vv']
     clicommon.run_command(command, display_cmd=verbose)
+
+
 #
 # 'startup' subcommand
 #
@@ -5354,37 +5459,13 @@ def startup(ctx, interface_name):
         if interface_name is None:
             ctx.fail("'interface_name' is None!")
 
-    try:
-        intf_fs = parse_interface_in_filter(interface_name)
-    except ValueError as e:
-        ctx.fail(str(e))
-
-    if len(intf_fs) > 1 and multi_asic.is_multi_asic():
-        ctx.fail("Interface range not supported in multi-asic platforms !!")
-
-    if len(intf_fs) == 1 and interface_name_is_valid(config_db, interface_name) is False:
-        ctx.fail("Interface name is invalid. Please enter a valid interface name!!")
-
+    intf_fs = get_interface_names_in_namespace(ctx, interface_name)
     log.log_info("'interface startup {}' executing...".format(interface_name))
-    port_dict = config_db.get_table('PORT')
-    for port_name in port_dict:
-        if port_name in intf_fs:
-            config_db.mod_entry("PORT", port_name, {"admin_status": "up"})
-
-    portchannel_list = config_db.get_table("PORTCHANNEL")
-    for po_name in portchannel_list:
-        if po_name in intf_fs:
-            config_db.mod_entry("PORTCHANNEL", po_name, {"admin_status": "up"})
-
-    subport_list = config_db.get_table("VLAN_SUB_INTERFACE")
-    for sp_name in subport_list:
-        if sp_name in intf_fs:
-            config_db.mod_entry("VLAN_SUB_INTERFACE", sp_name, {"admin_status": "up"})
-
-    lo_list = config_db.get_table("LOOPBACK_INTERFACE")
-    for lo in lo_list:
-        if lo in intf_fs:
-            config_db.mod_entry("LOOPBACK_INTERFACE", lo, {"admin_status": "up"})
+    for intf in intf_fs:
+        table_name = get_port_table_name(intf)
+        if not table_name:
+            continue
+        config_db.mod_entry(table_name, intf, {"admin_status": "up"})
 
 #
 # 'shutdown' subcommand
@@ -5395,7 +5476,6 @@ def startup(ctx, interface_name):
 @click.pass_context
 def shutdown(ctx, interface_name):
     """Shut down interface"""
-    log.log_info("'interface shutdown {}' executing...".format(interface_name))
     # Get the config_db connector
     config_db = ctx.obj['config_db']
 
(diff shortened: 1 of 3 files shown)


arlakshm:
@arista-setu, can you please add some UT?

config/main.py
```diff
@@ -586,6 +586,22 @@ def get_port_namespace(port):
 
     return None
 
+
+def validate_interface_names_in_namespace(ctx, intf_fs):
```
anamehra: could you please add logs for all options like range Ethernet0-Ethernet10 which is supported on single asics?

config/main.py
anamehra: parse_interface_in_filter() expands Ethernet16-20 into Ethernet16, Ethernet17, Ethernet18, Ethernet19, Ethernet20. The old startup/shutdown code iterated the actual PORT table and only updated existing ports, so this worked on common sparse SONiC port layouts like Ethernet0, Ethernet4, Ethernet8, Ethernet12, Ethernet16, etc. The new validate_interface_names_in_namespace() rejects the whole command as soon as any generated intermediate name is absent.

config/main.py
```diff
@@ -5302,31 +5300,13 @@ def shutdown(ctx, interface_name):
     except ValueError as e:
         ctx.fail(str(e))
 
-    if len(intf_fs) > 1 and multi_asic.is_multi_asic():
-        ctx.fail("Interface range not supported in multi-asic platforms !!")
+    validate_interface_names_in_namespace(ctx, intf_fs)
```
anamehra: same as above. please validate all supported options on single asic.

kenneth-arista:
@abdosi

config/main.py
arista-setu: Thanks for the review. I'll address it and update the PR shortly. Thanks

arista-setu: I've added relevant logs with latest revision

config/main.py
arista-setu: Thanks for the review,

I've address this gap in latest revision. I also refactored utilities_common/intf_filter.py to reuse existing code to parse the range based input for its use-case in get_interface_names_in_namespace function. Now the function will parse and return the valid interfaces in the range based input as well non-range based input ( which will then be operated for shutdown/startup operations )

config/main.py
arista-setu: Addressed with the latest review, Thanks.

config/main.py
```diff
@@ -589,6 +589,54 @@ def get_port_namespace(port):
...
+
+        valid_intfs_in_filter = []
+        invalid_intfs_in_filter = []
+        for intf in intf_fs:
+            if interface_name_is_valid(config_db, intf):
+                valid_intfs_in_filter.append(intf)
+            else:
+                invalid_intfs_in_filter.append(intf)
+
+        if is_range:
```
anamehra: Could you confirm the expected behavior for multi-ASIC range inputs when some generated members are valid interfaces but belong to a different namespace?

With the current range path, if at least one member exists in the selected namespace, the other generated names are skipped/logged as missing. That preserves sparse single-ASIC behavior, but for multi-ASIC it may silently skip a valid interface from another namespace instead of failing namespace validation.

tests/config_test.py
```diff
@@ -4499,6 +4554,124 @@ def teardown_method(self):
...
+
+        result = runner.invoke(config.config.commands['interface'].commands['startup'],
+                               ['Ethernet0,Ethernet4'], obj=obj)
+        assert result.exit_code == 0
+
+        port_table = config_db.get_table('PORT')
+        assert port_table['Ethernet0']['admin_status'] == 'up'
+        assert port_table['Ethernet4']['admin_status'] == 'up'
+
+    def test_startup_shutdown_sparse_port_range(self, get_cmd_module, setup_multi_broadcom_masic):
```
anamehra, line 4653: Could you please add coverage for the documented mixed input form, for example `Ethernet8,Ethernet16-20,Ethernet32`, and also a multi-ASIC range or mixed input that includes an interface from another namespace?

That would confirm both the documented single-ASIC command form and the namespace-validation behavior after range expansion.

rlhui:
@arista-setu, there are more comments to address, thanks

tests/config_test.py
arista-setu: Added coverage in unit test for cover mixed input ( comma-separated + range ) for both single and multi-asic

config/main.py
arista-setu: Thanks for pointing out it does silently skipped valid interfaces from other namespace. I've addressed this gap now.

The expected behavior on multi-asic should be based on `all-or-nothing` for successful completion. When filters interface from entire input is process, if interface is valid but belongs to different namespace we should report it.

In the latest revision, I've updated logging and error reporting mechanism.
- Earlier, a generic message like `Provided interface(s) : <interfaces> are either invalid or not in provided namespace <ns>` was printed out in error case.
- In updated version, I've updated to give out specific reason for the error. I've update description of this PR with sample output on mutli-asic.

anamehra (dismissed):
LGTM

peterbailey-arista (approved):

peterbailey-arista (approved):
LGTM, approved internally

rlhui:
@arista-setu , would you please help address pr build failures

arista-setu:
The build failures are unrelated to changes in the PR. The build fails because failure in tests/pfcstat_test.py which is also seen on other sonic-utilities PRs.

I will kick of a build again.

arista-setu:
@anamehra @abdosi Can you please help review the PR? Thanks.

anamehra:
> @anamehra @abdosi Can you please help review the PR? Thanks.

Hi @arista-setu , I will review this again. I approved it past but looks like there was a force review which removed all previous commit history. We should prevent force commits to maintain the commit history and reviews in PR.

anamehra:
Thanks for addressing the previous review comments. I re-checked the latest revision, and the earlier gaps around sparse ranges, mixed input, and multi-ASIC cross-namespace validation look handled now.

I noticed one new behavior from the recent parser refactor that we should fix before merge. In `utilities_common/intf_filter.py`, `Ethernet-Rec*` and `Ethernet-IB*` currently return an empty list instead of preserving the input or failing validation.

From CLI POV, this can make the command succeed without doing anything:

```bash
sudo config interface shutdown Ethernet-IB0
```

or silently skip one requested member:

```bash
sudo config interface shutdown Ethernet0,Ethernet-IB0
```

In the second case, `Ethernet0` may be processed, but `Ethernet-IB0` is dropped because it expands to `[]`. The original parser raised a `ValueError` for these inputs, so this looks like a regression from the PR change.

Could you please update this path so these inputs either remain as explicit interface names and go through normal validation, or fail clearly instead of becoming a successful no-op?

peterbailey-arista:
> From CLI POV, this can make the command succeed without doing anything:

@anamehra 

I think this was the CLI's original functionality before Setu's change. @arista-setu can confirm for us but I believe he does it this way in order to preserve the behaviour of startup / shutdown on these internal always-up ports (which is just skip, no-op)

It is safer to preserve the original behaviour where he can, while still adding the batch-flapping for multi asic. Assuming I am correct about this, we could propose and change how the CLI should handle these special ports in a follow up PR and design discussion instead of here.
Does that work for you?

arista-setu:
I've preserved older behavior of startup / shutdown on these internal ports. Based on my testing previous behavior was skip them at parsing stage which is consistent with newer refactor in `intf_filter.py`. Since these are skipped, these will not get flapped.

Therefore, this refactor safely preserves older behavior with new functionality on multi-asic. It should not introduce product side regressions. Since this is larger diff already, handling of special ports in followup PR if needed will be much better.

anamehra:
> I've preserved older behavior of startup / shutdown on these internal ports. Based on my testing previous behavior was skip them at parsing stage which is consistent with newer refactor in `intf_filter.py`. Since these are skipped, these will not get flapped.
> 
> Therefore, this refactor safely preserves older behavior with new functionality on multi-asic. It should not introduce product side regressions. Since this is larger diff already, handling of special ports in followup PR if needed will be much better.

Hi @arista-setu , @peterbailey-arista , my understanding was that the original code was not handling special ports other than -BP and throwing errors. But new code may ignore those. I do not have a chassis to validate those. For -BP ports it looks fine, so I will approve for that. Please validate old and new behavior for other special ports and add logs.

anamehra (dismissed):
LGTM fo rchassis-packet. Approving based on chassis-packet behavior.

vmittal-msft:
@saksarav-nokia please help review

config/main.py
```diff
@@ -589,6 +589,109 @@ def get_port_namespace(port):
...
+        ctx.fail(msg)
+
+    # Fail when input includes interface(s) that don't exist on any namespace.
+    if invalid_intfs:
+        log.log_warning("Invalid interface(s) in namespace '{}': {}".format(
+            namespace, ', '.join(invalid_intfs)))
+        ctx.fail("Provided interface(s) : '{}' are invalid in namespace '{}'".format(
+            ', '.join(invalid_intfs), namespace))
+
+    return valid_intfs
```
saksarav-nokia, line 692: If f expand_single_intf_filter() ,  returns intf_fs == [] and then nothing added to valid_intfs, invalid_intfs, cross_ns_intfs .
So no error is printed and command retunrs silently without doing anything?

config/main.py
saksarav-nokia, line 5458: Alias conversion runs on the whole filter string. Is this ok?

config/main.py
```diff
@@ -589,6 +589,109 @@ def get_port_namespace(port):
...
+        for intf in intf_fs:
+            if intf in seen_intfs:
+                continue
+            seen_intfs.add(intf)
+            # Check if interface is in target namespace.
+            if interface_name_is_valid(config_db, intf):
+                valid_in_filter.append(intf)
+            else:
+                if other_ns_ports is None:
+                    # Lazy init on first miss to skip uneccessary config_db connections
```
saksarav-nokia: Fix the typo in "uneccessary "

config/main.py
arista-setu: Here when expand_single_intf_filter returns empty list, the intf is essentially skipped and not processed further. 

It will return empty list for special internal ports ( Rec0 and IB0/1) only. Yes it is silently skipping it without error message or failing the command to keep it consistent with CLI's original functionality/behaviour before this refactor.

As discussed earlier, this refactor safely preserves older behavior with new functionality on multi-asic. It should not introduce product side regressions. Since this is larger diff already, handling of special ports in followup PR if needed will be much better.

config/main.py
arista-setu: Alias conversion only ever worked for single input interface only which is working after this refactor. 
```
admin@qzd368:~$ sudo env SONIC_CLI_IFACE_MODE=alias config interface -n asic0 shutdown etp1
admin@qzd368:~$ sudo env SONIC_CLI_IFACE_MODE=alias config interface -n asic0 startup etp1
admin@qzd368:~$ sudo env SONIC_CLI_IFACE_MODE=alias show interfaces status etp1
  Interface                    Lanes    Speed    MTU    FEC    Alias    Vlan    Oper    Admin    Medium     Type           Vendor            Model    Asym PFC    Flaps
-----------  -----------------------  -------  -----  -----  -------  ------  ------  -------  --------  -------  ---------------  ---------------  ----------  -------
  Ethernet0  72,73,74,75,76,77,78,79     400G   9100     rs     etp1  routed    up       up     Optic  QSFP-DD  Arista Networks  AOC-D-D-400G-1M         off        6
```

Support for alias-range or comma separated alias is not present and it fails with error message. This behavior is consistent after this refactor. I think we can address this issue in future PR.

anamehra (approved):

peterbailey-arista (approved):
LGTM, tested and approved internally as well

saksarav-nokia:
@mlok-nokia for viz

saksarav-nokia (approved):

arista-setu:
@rlhui can you help merging this PR? Thanks!

PR #4439, opened 2026-04-08
Parallelize sonic-utilities tests -- too large a portion of overall sonic build time

bhouse-nexthop:
#### What I did

sonic-utilities takes a long time to build, and it builds twice (bookworm + trixie).  It is often the last thing left building before building the final image. 

This PR Parallelizes the entire sonic-utilities test suite using pytest-xdist. All 251 test files (4618 tests) now run fully in parallel with zero failures across all execution modes: `-n auto`, `-n 1`, and pure serial (`-p no:xdist`).

setup/teardown logic was very similar, the most common have been moved to happen automatically so tests only don't need to worry about it in a vast majority of cases.  There were a lot of cases where tests either weren't cleaning up fully, or actually depending on what a prior test file set up erroneously which breaks parallelism.  Centralizing this solves all those issues.  This has caused a net reduction of 258 lines for this entire PR. 

**Performance: ~13 min → ~96 sec (92% reduction)**

**NOTE**: This also corrects latent issues in tests that ended up actually being order-dependent tests.

#### How I did it

**Infrastructure:**
- Added pytest-xdist with `--dist loadfile` scheduling to `pytest.ini` and `setup.py`
- Created per-xdist-worker sandbox copies of `mock_tables/` and cache directories to eliminate cross-worker file races
- Added `_reset_between_files()` hook in `tests/conftest.py` that restores single-asic defaults (env vars, mock DB state, multi_asic module functions, SonicDBConfig, CliRunner.invoke) on every test file transition
- Added `LazyChoice(click.Choice)` in `utilities_common/multi_asic.py` to defer namespace list evaluation from import time to validation time

**Categories of test fixes required:**

1. **Stale class references after module reload** — `@mock.patch.object` captures the class at decoration time; after `importlib.reload()` creates a new class, the mock becomes invisible. Converted to string-based `@mock.patch()` which resolves at call time. (bgp_commands_test.py, ip_show_routes_voq_chassis_test.py)

2. **Cross-worker `/tmp/` file races** — Tests backed up mock DB files to shared `/tmp/` paths. Parallel workers overwrote each other's backups. Fixed by backing up within the per-worker sandbox directory. (pgdropstat_test.py, portstat_test.py, mmuconfig_test.py)

3. **Module-level mock pollution** — Tests assigned mocks directly to module attributes without restoring. Refactored to use `with mock.patch.object(...)` context managers or save/restore patterns. (sflow_test.py, route_check_test.py, sfp_test.py, watchdogutil_test.py, vlan_test.py, vrrp_test.py)

4. **Missing/broken setup_class and teardown_class** — Some test classes lacked explicit setup, relying on execution order. Others had broken teardowns (e.g. `dbconnector.load_database_config` without `()` — a silent no-op). Added self-contained setup/teardown to 16 test files.

5. **Frozen Click.Choice at import time** — `config vlan` namespace option froze `get_namespace_list()` at import. Added `LazyChoice` to evaluate at validation time. (utilities_common/multi_asic.py, config/vlan.py, multi_asic_vlan_test.py)

6. **YANG model directory races** — cli_autogen tests modified `/usr/local/yang-models` in place. Changed to per-test temp directories. (cli_autogen_common.py, cli_autogen_test.py, cli_autogen_yang_parser_test.py)

**Cleanup:**
- Removed dead legacy YANG helper functions using `os.system('sudo ...')`
- Removed duplicate `@mock.patch` decorator in pcieutil_test.py
- Removed 14 redundant inline `get_all_namespaces` lambdas already handled by mock_multi_asic.py reload

**Production code changes (4 files):**
- `utilities_common/multi_asic.py` — LazyChoice class + apply to namespace options
- `utilities_common/cli.py` — UserCache reads `SONIC_CLI_CACHE_DIR` env var
- `config/vlan.py` — Use `multi_asic_click_option_namespace` instead of frozen Choice
- `scripts/mmuconfig` — Read `MMUCONFIG_FILE` env var instead of hardcoding `/tmp/mmuconfig`

#### How to verify it

```bash
# Full parallel (default)
python3 -m pytest tests/ -n auto --dist loadfile

# Single xdist worker (exercises reset logic sequentially)
python3 -m pytest tests/ -n 1 --dist loadfile

# Pure serial, no xdist (backwards compatibility)
python3 -m pytest tests/ -p no:xdist

# All three modes: 4618 passed, 0 failed
```

N/A — no command output changes. All production code changes are limited to test isolation plumbing (env var reads, lazy evaluation).

N/A


bhouse-nexthop:
@saiarcot895 do you have perms to enable the different worker pool of sonicso1ES-amd64 ?  The pool it was previously using only had 1 cpu core so couldn't prove the tests actually run in parallel. 

I don't know how often you build sonic-buildimage from scratch with no cache, but we often see sonic-utilities being one of the last 2 things to finish building for a few minutes (thermalctld is another).  Since it builds twice, its a build performance hog accounting for over 10% of the overall time.

bhouse-nexthop:
@saiarcot895 can this be merged before conflicts are introduced?

saiarcot895 (dismissed):

saiarcot895 (approved):

PR #4442, opened 2026-04-09
Clarify in the help string how negative values can be specified when configuring transceiver tx-power

maheeppartap-nexthop:
#### What I did
The config interface transceiver tx_power command is built using click, and tx_power is a argument instead of an option. This means this command does not accept a value, and so a negative number being passed in after this variable gets parsed as a flag which leads to the user not being able to set negative numbers. There is a comment explaining how to pass in neg numbers, but is not part of the docstring, and so is omitted from the help message. I moved the text to be part of the docstring

#### How I did it

Changed the comment to be part of the docstring. 

#### How to verify it
```
sudo config interface transceiver tx_power --help
Usage: config interface transceiver tx_power [OPTIONS] <interface_name> <tx-
                                             power>

  Set transciever (only for 400G-ZR) Tx laser power.

  For negative values, you must insert ``--`` before the value so that Click
  treats it as a positional argument instead of an option. For example:

    config interface transceiver tx_power Ethernet0 -- -11

Options:
  -?, -h, --help  Show this message and exit.
```
```
sudo config interface transceiver tx_power --help
Usage: config interface transceiver tx_power [OPTIONS] <interface_name> <tx-
                                             power>

  Set transciever (only for 400G-ZR) Tx laser power

Options:
  -?, -h, --help  Show this message and exit.
```

```
sudo config interface transceiver tx_power --help
Usage: config interface transceiver tx_power [OPTIONS] <interface_name> <tx-
                                             power>

  Set transciever (only for 400G-ZR) Tx laser power.

  For negative values, you must insert ``--`` before the value so that Click
  treats it as a positional argument instead of an option. For example:

    config interface transceiver tx_power Ethernet0 -- -11

Options:
  -?, -h, --help  Show this message and exit.
```

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -6179,17 +6179,19 @@ def frequency(ctx, interface_name, frequency):
     clicommon.run_command(command)
 
 
-#
-# 'tx_power' subcommand ('config interface transceiver tx_power ...')
-# For negative float use:-
-# config interface transceiver tx_power Ethernet0 -- -27.4"
-#
 @transceiver.command('tx_power')
 @click.pass_context
 @click.argument('interface_name', metavar='<interface_name>', required=True)
 @click.argument('tx-power', metavar='<tx-power>', required=True, type=float)
 def tx_power(ctx, interface_name, tx_power):
-    """Set transceiver (only for 400G-ZR) Tx laser power"""
+    """Set transciever (only for 400G-ZR) Tx laser power.
+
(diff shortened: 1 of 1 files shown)


prgeor (approved):
@maheeppartap-nexthop BTW this is already captured in documentation here:-

Image


Doc:https://github.com/sonic-net/sonic-utilities/blob/master/doc/Command-Reference.md

PR #4467, opened 2026-04-17
Added Sonic CLI commands to configure VxLAN ports

aakashtiwari2630-max:
#### What I did
Added three new Sonic CLI commands to configure VxLAN ports:
1. configure vxlan destination-port
2. configure vxlan source-port
3. configure vxlan source-port-mask

After configuring these commands, SWITCH_TABLE in APP_DB will be updated dynamically so that the VxLAN port values are active for installed VxLAN tunnels. 

Also, after configuring these commands, the port values will be stored in DEVICE_METADATA, so that in case of config reload or reboot, the same values are replayed using vxlanmgrd. Refer to:
https://github.com/sonic-net/sonic-swss/pull/4492 

Please note: After configuring these commands, config save is necessary for the port settings saved under DEVICE_METADATA to persist.

#### How to verify it
Added 3 new tests.
tests/vxlan_test.py::TestVxlan::test_config_vxlan_destination_port PASSED                                                                                                                                                                                               [ 91%] 
tests/vxlan_test.py::TestVxlan::test_config_vxlan_source_port PASSED                                                                                                                                                                                                    [ 95%] 
tests/vxlan_test.py::TestVxlan::test_config_vxlan_source_port_mask PASSED                                                                                                                                                                                               [100%] 

root@sonic:/home/admin# config vxlan destination-port 51234                                                                                                                     
VXLAN destination port set to 51234
root@sonic:/home/admin# config vxlan source-port 1234                                                                                  
VXLAN source port set to 1234                                      
root@sonic:/home/admin# config vxlan source-port-mask 7                                                                                                                                                        
VXLAN source port mask set to 7

### DIFF
--- a/config/vxlan.py
+++ b/config/vxlan.py
@@ -5,9 +5,19 @@ from utilities_common.constants import DEFAULT_NAMESPACE
 
 from jsonpatch import JsonPatchConflict
 from .validated_config_db_connector import ValidatedConfigDBConnector
-from swsscommon.swsscommon import isInterfaceNameValid, IFACE_NAME_MAX_LEN
+from swsscommon.swsscommon import isInterfaceNameValid, IFACE_NAME_MAX_LEN, ConfigDBConnector, SonicV2Connector, ProducerStateTable
 
 ADHOC_VALIDATION = True
+
+
+def _write_switch_table_to_appl_db(field, value):
+    """Write a field to APP_DB SWITCH_TABLE via ProducerStateTable so orchagent is notified."""
+    app_db = SonicV2Connector()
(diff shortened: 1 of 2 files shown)


prsunny:
I dont think DEVICE_METADATA is the right place to save this vxlan configs

aakashtiwari2630-max:
> I dont think DEVICE_METADATA is the right place to save this vxlan configs

I followed the example of other similar SWITCH_TABLE properties like ecmp_hash_offset and ecmp_hash_seed for this. But let me know if you can think of any other option.

PR #4472, opened 2026-04-18
Fix Pensando DPU reboot handling in smartswitch operations

SahilChaudhari:
Modified reboot scripts to properly handle Pensando ASIC type during smartswitch reboot operations to prevent unnecessary service disruptions.

Changes:
scripts/reboot:
   - Exclude Pensando from ASIC shutdown sequence (similar to Mellanox)
   - Skip stopping swss service for Pensando during ASAN cleanup

### DIFF
--- a/scripts/reboot
+++ b/scripts/reboot
@@ -98,7 +98,7 @@ function stop_sonic_services()
         systemctl stop mux
     fi
 
-    if [[ x"$ASIC_TYPE" != x"mellanox" ]]; then
+    if [[ x"$ASIC_TYPE" != x"mellanox" && x"$ASIC_TYPE" != x"pensando" ]]; then
         ASIC_CONF=${DEVPATH}/$PLATFORM/asic.conf
         if [ -f "$ASIC_CONF" ]; then
             source $ASIC_CONF
@@ -122,7 +122,9 @@ function stop_sonic_services()
 function stop_services_asan()
 {
     debug "Stopping swss for ASAN"
-    systemctl stop swss
+    if [[ x"$ASIC_TYPE" != x"pensando" ]]; then
+        systemctl stop swss
+    fi
 }
 
 function clear_warm_boot()


rameshraghupathy (approved):
LGTM

scripts/reboot_smartswitch_helper
```diff
@@ -251,13 +251,15 @@ function reboot_dpu()
         return ${EXIT_ERROR}
     fi
 
-    reboot_dpu_platform "${DPU_NAME}" "${REBOOT_TYPE}"
-    if [ $? -ne 0 ]; then
-        log_message "ERROR: Failed to reboot the module for ${DPU_NAME}"
-        if [[ "$REBOOT_TYPE" != $MODULE_REBOOT_SMARTSWITCH ]]; then
-            clear_module_state_transition_flag "${DPU_NAME}"
+    if [[ "$REBOOT_TYPE" != $MODULE_REBOOT_SMARTSWITCH ]]; then
```
vvolam: @SahilChaudhari This is changing behavior for all smart switches and not limited to Pensando.

@dgsudharsan @gpunathilell please review?

vvolam:
@SahilChaudhari Please fix DCO warning

gpunathilell:
@SahilChaudhari can you explain the change to scripts/reboot_smartswitch_helper, its not clear

gpunathilell (changes requested):
reboot_dpu_platform needs to be called even for smartswitch type reboots for our platform

scripts/reboot_smartswitch_helper
vvolam: @SahilChaudhari based on Gagan's response, this change needs to be limited to your platform, if it is needed.

vvolam (changes requested):

scripts/reboot_smartswitch_helper
SahilChaudhari: done

rameshraghupathy:
@SahilChaudhari  LGTM

rameshraghupathy (dismissed):

scripts/reboot_smartswitch_helper
```diff
@@ -251,13 +251,15 @@ function reboot_dpu()
         return ${EXIT_ERROR}
     fi
 
-    reboot_dpu_platform "${DPU_NAME}" "${REBOOT_TYPE}"
-    if [ $? -ne 0 ]; then
-        log_message "ERROR: Failed to reboot the module for ${DPU_NAME}"
-        if [[ "$REBOOT_TYPE" != $MODULE_REBOOT_SMARTSWITCH ]]; then
+    PLATFORM=$(sonic-cfggen -H -v DEVICE_METADATA.localhost.platform 2>/dev/null)
+
+    if [[ "$REBOOT_TYPE" != "$MODULE_REBOOT_SMARTSWITCH" || "$PLATFORM" != *8102* ]]; then
```
vvolam: @SahilChaudhari please use proper macro instead of checking for 8102 here. Also, this needs to be && condition? please fix it

scripts/reboot_smartswitch_helper
```diff
@@ -251,13 +251,15 @@ function reboot_dpu()
         return ${EXIT_ERROR}
     fi
 
-    reboot_dpu_platform "${DPU_NAME}" "${REBOOT_TYPE}"
-    if [ $? -ne 0 ]; then
-        log_message "ERROR: Failed to reboot the module for ${DPU_NAME}"
-        if [[ "$REBOOT_TYPE" != $MODULE_REBOOT_SMARTSWITCH ]]; then
+    PLATFORM=$(sonic-cfggen -H -v DEVICE_METADATA.localhost.platform 2>/dev/null)
```
gpunathilell: the PLATFORM is already defined in reboot, we are sourcing reboot_smartswitch_helper and calling the relevant functions, use `assign-if-unset-or-null` so that we dont increase delay unnecessarily

scripts/reboot_smartswitch_helper
SahilChaudhari: @vvolam this is 
```
negation of [ "$REBOOT_TYPE" == "$MODULE_REBOOT_SMARTSWITCH" && "$PLATFORM" == *8102* ]
```

I will use proper macro

KrisNey-MSFT:
Check this during PMON call tomorrow.  2nd part of file changes to ensure we are platform independent.

scripts/reboot_smartswitch_helper
```diff
@@ -251,13 +252,15 @@ function reboot_dpu()
         return ${EXIT_ERROR}
     fi
 
-    reboot_dpu_platform "${DPU_NAME}" "${REBOOT_TYPE}"
-    if [ $? -ne 0 ]; then
-        log_message "ERROR: Failed to reboot the module for ${DPU_NAME}"
-        if [[ "$REBOOT_TYPE" != $MODULE_REBOOT_SMARTSWITCH ]]; then
+    PLATFORM=${PLATFORM:=$(sonic-cfggen -H -v DEVICE_METADATA.localhost.platform 2>/dev/null)}
+
+    if [[ "$REBOOT_TYPE" != "$MODULE_REBOOT_SMARTSWITCH" ]] || [[ "$PLATFORM" != $PLATFORM_CISCO_PATTERN ]]; then
```
rameshraghupathy: @vvolam @gpunathilell  During a system-wide SmartSwitch reboot, the reboot helper should invoke only module_pre_shutdown to gracefully halt DPU services, and must avoid calling the per-DPU module reboot path. The subsequent PMON restart as part of the chassis reboot will automatically handle DPU restart.

Rationale:

1. Avoids unnecessary increase in overall reboot time.
2. The reboot_smartswitch_helper platform API cannot be guaranteed to behave reliably during this phase, since NPU/PMON container restarts may occur concurrently. While it may appear to work under current timing conditions, it introduces a latent race that could surface with future timing variations.
3. Exiting early in this code path is preferable, as it reduces exposure to race conditions and avoids redundant or potentially conflicting operations during system reboot.

rameshraghupathy (approved):

gpunathilell (approved):

scripts/reboot_smartswitch_helper
vvolam: @rameshraghupathy As discussed in yesterday's meeting, as per initial design discussion we will have platform API call with specific reboot type we are going through. This will make the platform specific way of rebooting DPUs and NPU inside the API and keep the smartswitch helper script platform agnostic. Thanks.

vvolam (approved):

PR #4481, opened 2026-04-21
[show/vnet]: Fix missing status column in short-endpoint tunnel routes

### LINKED ISSUE #4480: When `val_state` is missing (e.g., in unit-test multi-ASIC mocks), the row ends up with fewer columns than `tunnel_header`
Copilot just identify a valid issue in legacy codebase. Use this issue to track.

> In the short-endpoint branch, the code only appends the route status when `val_state` is present. When `val_state` is missing (e.g., in unit-test multi-ASIC mocks), the row ends up with fewer columns than `tunnel_header`, which can drop/misalign the `status` column in the printed table. Always append a status value (empty string) so every row matches the header length.
> ```suggestion
>             state = val_state.get('state') if val_state else ""
>             r.append(state)
> ``` 

 _Originally posted by @Copilot in [#4332](https://github.com/sonic-net/sonic-utilities/pull/4332/changes#r3068781513)_

### LINKED ISSUE #27148: Bug: `show vnet routes tunnel` row/column mismatch when STATE_DB entry is missing
### Is it platform specific

generic

### Importance or Severity

Low

### Description of the bug

In `show/vnet.py`, the `_show_tunnel_helper()` function defines a 7-column header:

    tunnel_header = ['vnet name', 'prefix', 'endpoint', 'mac address', 'vni', 'metric', 'status']

However, in the short-endpoint code path (when `len(epval) < 40`), the `status` column is only appended when `val_state` is truthy:

    r.append(val.get('metric'))
    if val_state:
        r.append(val_state.get('state'))
    table.append(r)

When `val_state` is falsy (empty dict from `get_all()` on a nonexistent STATE_DB key, or in multi-ASIC unit-test mocks), the row is appended with only 6 elements while the header expects 7. This causes column misalignment in the `tabulate` output.

The long-endpoint path (`pretty_print()`) already handles this correctly by receiving a pre-resolved `state` parameter.

This bug was identified by Copilot during PR #4332 review but was not addressed before merge. It was later separately raised in `sonic-utilities`: https://github.com/sonic-net/sonic-utilities/issues/4480, raising it in `sonic-buildimage` as well for issue triage.

### Steps to Reproduce

1. Set up a multi-ASIC environment (or use unit-test mocks) where STATE_DB does not have an entry for a given VNET_ROUTE_TUNNEL_TABLE key
2. Run `show vnet routes tunnel`
3. Observe the output table — rows without STATE_DB entries will have misaligned columns

### Actual Behavior and Expected Behavior

Actual: When val_state is falsy, the status column is omitted entirely from the row, producing a 6-element row against a 7-column header. tabulate may misalign or silently drop the last column.

Expected: Every row should always have 7 elements matching the header. When no state is available, an empty string should be appended as the status value.

### Relevant log output

```shell

```

### Output of `show version`, `show techsupport`

```shell

```

### Attach files (if any)

_No response_

Bojun-Feng:
#### Why I did it
In `_show_tunnel_helper`, the short-endpoint code path only appends the route state when `val_state` is truthy. When `val_state` is missing the row gets 6 columns instead of 7.

Fix #4480
Fix https://github.com/sonic-net/sonic-buildimage/issues/27148

#### How I did it
* Always append a state value in the short-endpoint path using the same ternary pattern as the long-endpoint path (`val_state.get('state') if val_state else ""`)
* Updated multi-ASIC test expected output to include the `status` column header

#### How to verify it
1. Populate APPL_DB with a tunnel route entry while leaving STATE_DB empty for that key
2. Run `show vnet routes tunnel` and verify `status` column exist

Fix missing status column in short-endpoint vnet tunnel route display when state_db entry is absent

When `val_state` is missing from all entries the column is dropped:
```
vnet name    prefix       endpoint       mac address          vni  metric
-----------  -----------  -------------  -----------------  -----  --------
Vnet_2000    20.0.0.0/24  192.168.1.200  00:aa:bb:cc:dd:01   5000
```
```
vnet name    prefix       endpoint       mac address          vni  metric    status
-----------  -----------  -------------  -----------------  -----  --------  --------
Vnet_2000    20.0.0.0/24  192.168.1.200  00:aa:bb:cc:dd:01   5000
```

### DIFF
--- a/show/vnet.py
+++ b/show/vnet.py
@@ -651,8 +651,8 @@ def _show_tunnel_helper(vnet_name=None, appl_db=None, state_db=None):
             r.append(val.get('mac_address'))
             r.append(val.get('vni'))
             r.append(val.get('metric'))
-            if val_state:
-                r.append(val_state.get('state'))
+            state = val_state.get('state') if val_state else ""
+            r.append(state)
             table.append(r)
             continue
         state = val_state.get('state') if val_state else ""
--- a/tests/multi_asic_vnet_test.py
+++ b/tests/multi_asic_vnet_test.py
@@ -162,8 +162,8 @@ vnet name    prefix       nexthop       interface
 -----------  -----------  ------------  -----------
 Vnet_2000    10.0.0.0/24  10.10.10.100  Ethernet16
 
-vnet name    prefix       endpoint       mac address          vni  metric
------------  -----------  -------------  -----------------  -----  --------
+vnet name    prefix       endpoint       mac address          vni  metric    status
+-----------  -----------  -------------  -----------------  -----  --------  --------
 Vnet_2000    20.0.0.0/24  192.168.1.200  00:aa:bb:cc:dd:01   5000
 Vnet_3000    30.0.0.0/24  192.168.1.201                      6000
 
@@ -172,8 +172,8 @@ vnet name    prefix       nexthop       interface
 -----------  -----------  ------------  -----------
 Vnet_4000    40.0.0.0/24  10.10.10.200  Ethernet64
 
-vnet name    prefix       endpoint       mac address          vni  metric
------------  -----------  -------------  -----------------  -----  --------
+vnet name    prefix       endpoint       mac address          vni  metric    status
+-----------  -----------  -------------  -----------------  -----  --------  --------
 Vnet_4000    50.0.0.0/24  192.168.2.200  00:aa:bb:cc:dd:02   4000
 """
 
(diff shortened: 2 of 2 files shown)


Bojun-Feng:
Hi @qiluo-msft — this PR addresses the column mismatch issue you tracked in #4480.
Would you be able to take a look when you have a chance? Happy to address any feedback. Thanks!

prgeor:
@qiluo-msft can you review or assign right person for this?

Bojun-Feng:
Closed in place of https://github.com/sonic-net/sonic-utilities/pull/4452

PR #4488, opened 2026-04-23
route_check error messages during MACSec testcases run

### LINKED ISSUE #4487: route_check error messages during MACSec and other interface flap testcases
The route_check.py was generating false-positive ERR monit: routeCheck syslog entries during test teardown, causing loganalyzer to fail tests that were otherwise correct. This affected tests that cause BGP session disruption (e.g. BGP peer shutdown, interface-level config changes that trigger BGP flaps).

```
>               pt_assert(False, failure_message)
E               Failed: Got matched syslog in processes "analyze_logs--<MultiAsicSonicHost humm168>" exit code:"1"
E               match: 1
E               expected_match: 0
E               expected_missing_match: 0
E               
E               Match Messages:
E               2026 Mar 16 09:05:50.104873 humm168 ERR monit[1231]: 'routeCheck' status failed (255) -- Failure results: {{#012    "": {#012        "missed_ROUTE_TABLE_routes": [#012            "100.1.0.3/32",#012            "2064:100::3/128"#012        ]#012    }#012}}#012Failed. Look at reported mismatches above#012add: {#012    "": [#012        "194.45.162.0/25",#012        "194.46.176.0/25",#012        "194.46.178.0/25",#012        "194.46.19.128/25",#012        "194.46.195.128/25",#012        "194.46.242.128/25",#012        "194.46.32.128/25",#012        "194.46.66.0/25",#012        "194.47.112.128/25",#012        "194.47.130.0/25",#012        "194.47.208.128/25",#012        "194.48.128.0/25",#012        "194.48.195.0/25",#012        "194.48.208.0/25",#012        "194.48.208.128/25",#012        "194.48.224.128/25",#012        "194.48.226.128/25",#012        "194.48.227.128/25",#012        "194.48.242.128/25",#012        "194.48.3.0/25"
```

selva-nexthop:
Description:
route_check.py was generating false-positive ERR monit: routeCheck syslog entries during test teardown, causing loganalyzer to fail tests that were otherwise correct. This affected tests that cause BGP session disruption (e.g. BGP peer shutdown, interface-level config changes that trigger BGP flaps).

Root cause:
When BGP sessions are disrupted and restored, routes can take up to 2 minutes to fully reconverge — the time needed for session re-establishment and route re-advertisement. route_check.py has no tolerance for this window and immediately reports missed_ROUTE_TABLE_routes (routes present in APPL_DB but not yet programmed into ASIC_DB) as a hard failure.

Resolution:
- Added ROUTE_CHECK_RETRIES = 8 and ROUTE_CHECK_WAIT_TIME = 15 (8 × 15s = 120s max wait, consistent with FRR_WAIT_TIME)
- For rt_appl_miss: reuse the existing ASIC_DB subscriber (selector/subs) opened at check start. Events buffered during the sleep are drained via get_subscribe_updates — O(new events only), no new DB connection or full table scan.
- For rt_asic_miss: re-read APPL_DB (lightweight getKeys()) only when rt_asic_miss is still non-empty, avoiding unnecessary scans in the common case where only rt_appl_miss is present.
- Genuine failures (routes never converging) are still reported after all retries are exhausted.

Resolves: https://github.com/sonic-net/sonic-utilities/issues/4487

### DIFF
--- a/scripts/route_check.py
+++ b/scripts/route_check.py
@@ -82,6 +82,11 @@ FRR_WAIT_TIME = 15
 
 REDIS_TIMEOUT_MSECS = 0
 
+# Retry constants for transient route mismatch recovery.
+# BGP reconvergence after session disruption can take up to 2 minutes,
+# so retries use full DB re-reads with a sleep aligned to FRR_WAIT_TIME.
+ROUTE_CHECK_RETRIES = 8
+ROUTE_CHECK_WAIT_TIME = 15
 
 class Level(Enum):
     ERR = 'ERR'
@@ -921,11 +926,6 @@ def check_routes_for_namespace(namespace):
         # Look for subscribe updates for a second
         adds, deletes = get_subscribe_updates(selector, subs)
 
-    # Release the subscriber. If we keep the subscriber open then route updates will accumulate in the subscriber queue
-    # causing high client memory usage in redis.
-    del subs
-    del selector
-
     # Drop all those for which SET received
     rt_appl_miss, _ = diff_sorted_lists(rt_appl_miss, adds)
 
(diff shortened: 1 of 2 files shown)


selva-nexthop:
Will this change increase the time taken for route_check.py significantly?  As an alternate, can we consider disabling the route_checker script in the test script where we expect things to fail?

At the scale these test environments actually run — hundreds to thousands of routes — the worst case (all 8 retries exhausting plus the FRR check also failing) comes to roughly 178s, leaving about 120s of breathing room before the 300s monit kill. The retry loop itself only processes new ASIC_DB events that arrived during the sleep, so the number of routes in the system doesn't affect it. At 1M routes the story changes. The initial get_asicdb_routes call has to pop every single ASIC_DB entry, and at that scale that read alone could take 60–120s, which would push the failure-path total uncomfortably close to 300s. That said, this isn't something we're introducing — the same risk exists today without this change. Anyone running route_check.py on a 1M route system already faces a slow initial read. Our change only adds time in the failure path; if everything is healthy and routes are consistent, the execution time is identical to before.

yes, we can have fixture to ignore these errors. There are many around 30 affected files (bgp + macsec). So, it is better to put these in bgp and macsec folder level fixtures - we might need to change the conftest.py in bgp and macsec folders.

On the two approaches:
the sonic-mgmt conftest route is safe from a timing perspective since it doesn't touch route_check.py at all, but it treats the symptom rather than the cause — the ERR still fires, the test just learns to ignore it. It would also need changes in at least two places (tests/bgp/conftest.py and tests/macsec/conftest.py) to cover the ~30 affected test files today, and any new test that causes BGP disruption in a different directory would need the same treatment again. There's also a small risk of accidentally silencing a genuine routeCheck failure in those directories.

The route_check.py fix goes after the root cause directly — by waiting for convergence, the ERR never gets logged in the first place and loganalyzer has nothing to react to. It works for all tests across the board without any per-directory changes, and real failures still surface after the retries run out.

scripts/route_check.py
```diff
@@ -82,6 +82,11 @@
 
 REDIS_TIMEOUT_MSECS = 0
 
+# Retry constants for transient route mismatch recovery.
+# BGP reconvergence after session disruption can take up to 2 minutes,
+# so retries use full DB re-reads with a sleep aligned to FRR_WAIT_TIME.
+ROUTE_CHECK_RETRIES = 8
+ROUTE_CHECK_WAIT_TIME = 15
```
anamehra, line 89: This will also increase route check failure detection for production. May we just disable route check for this test during teardown?

selva-nexthop: The concern is valid within a single script invocation — genuine failures now take up to 120s longer to produce a non-zero exit code. However, the monit policy already requires 3 consecutive failures across 3×5-minute cycles before alerting, so the end-to-end alert latency increases from ~15 min to ~21 min. Given that routeCheck is not a sub-minute detection mechanism, and the false-positive rate from transient BGP reconvergence was causing real test failures, this trade-off is intentional. If the community prefers a shorter window, ROUTE_CHECK_RETRIES and ROUTE_CHECK_WAIT_TIME can be tuned down — e.g., 4×15s = 60s adds only 3 minutes to the alert timeline while still covering most reconvergence windows.

Yes, we can disable route_check in sonic-mgmt tests.  It is safe from a timing perspective since it doesn't touch route_check.py at all, but it treats the symptom rather than the cause — the ERR still fires, the test just learns to ignore it. It would also need changes in at least two places (tests/bgp/conftest.py and tests/macsec/conftest.py) to cover the ~30 affected test files today, and any new test that causes BGP disruption in a different directory would need the same treatment again. There's also a small risk of accidentally silencing a genuine route_check failure in those directories.

This fix goes after the root cause directly — by waiting for convergence, the ERR never gets logged in the first place and loganalyzer has nothing to react to. It works for all tests across the board without any per-directory changes, and real failures still surface after the retries run out.

deepak-singhal0408:
The problem this PR addresses appears to have been resolved by a different approach: [sonic-mgmt #24642](https://github.com/sonic-net/sonic-mgmt/pull/24642) (merged Jun 10, 2026) adds `ERR monit.*routeCheck` to the global `loganalyzer_common_ignore.txt`, which prevents the loganalyzer teardown failure for all tests.

That said, I have some reservation about the blanket-ignore approach — it means loganalyzer will now also miss genuine persistent route programming failures (e.g., a real orchagent bug that leaves routes stuck). The retry approach here has the advantage that route_check's ERR would only fire after convergence retries are exhausted, making it a higher-confidence signal that loganalyzer *should* catch.

So this PR may still have value as a long-term improvement — making route_check smarter so we can eventually remove the blanket ignore and restore loganalyzer's ability to detect real issues. But as a fix for the immediate test-failure problem, it's now redundant.

A couple of notes if this is picked up again:
- `TIMEOUT_SECONDS = 120` on real chassis (non-VS) — the 8×15s=120s budget is very tight against SIGALRM. Consider reducing to 4×15s=60s (as you suggested in your earlier reply).
- PR [#4576](https://github.com/sonic-net/sonic-utilities/pull/4576) also adds retry logic in the same code area — would need coordination to avoid conflicts.

deepak-singhal0408:
@selva-nexthop please recheck if this PR is still needed? Thanks,

PR #4511, opened 2026-05-01
Sfputil: Add the optional use-lpmode-pin flag to lpmode 

abhi-nexthop:
#### What I did

Added the ability to use the --lpmode-pin flag. 

#### How I did it
This will pass the flag down to sfp.get_lpmode and set_lpmode, only when it is explicitly certified and throws an error if the flag is not supported by the sfp class.
The default functionality (without the new flag) will not change whatsoever. 

#### How to verify it

Replace your sfputil main.py and run as usual. The behavior, CMIS register values should remain exactly the same.
When this flag is provided, if your platform supports the software and hardware distinction, it should assert/deassert hardware lpmode on the transceiver.
This PR is gated on https://github.com/sonic-net/sonic-platform-common/pull/663 being submitted

One such implementation has been made in pddf sfp here: https://github.com/sonic-net/sonic-buildimage/pull/27278

When flag is not supported:
LPMode pin flag not implemented for this platform

When supported: It remains the same


sfputil/main.py
```diff
@@ -1133,7 +1133,8 @@ def error_status(port, fetch_from_hardware):
 # 'lpmode' subcommand
 @show.command()
 @click.option('-p', '--port', metavar='<port_name>', help="Display SFP low-power mode status for port <port_name> only")
-def lpmode(port):
+@click.option('--use-hardware-control', is_flag=True, default=False, help='Use Xcvr LPMode pin instead of EEPROM')
```
prgeor: rename to -`-use-lpmode-pin`?

sfputil/main.py
```diff
@@ -1253,15 +1263,26 @@ def set_lpmode(logical_port, enable):
...
                 click.echo(f"{logical_port}: module {physical_port} is not present, skipping")
                 continue
             click.echo("{} low-power mode for port {} ... ".format(
                 "Enabling" if enable else "Disabling",
-                get_physical_port_name(logical_port, i, ganged)), nl=False)
-            result = sfp.set_lpmode(enable)
+                port_name), nl=False)
+            if use_hardware_control:
+                try:
+                    result = sfp.set_lpmode(
```
prgeor: @abhi-nexthop the parameters does not match with https://github.com/sonic-net/sonic-platform-common/blob/64beade8cddecdbc154531bc84bed2fa86581ea8/sonic_platform_base/sonic_xcvr/sfp_optoe_base.py#L285

I think you need to add an optional argument as follows:
```
def SfpOptoeBase(self, lpmode, use_lpmode_pin=false):
      if use_lpmode_pin:
           # Invoke platform override API that uses LPMode pin
           return self.set_lpmode_via_pin(lpmode)
       
      # Now fall back to SW based
     
```

sfputil/main.py
```diff
@@ -1316,12 +1325,14 @@ def reset(port_name):
         ganged = True
 
     for physical_port in physical_port_list:
+        physical_port_name = get_physical_port_name(port_name, i, ganged)
+        i += 1
```
prgeor: @abhi-nexthop why this is moved ?

sfputil/main.py
abhi-nexthop: Hi @prgeor, thanks for reviewing the PR. Unfortunately, the approach of sending use_lpmode_pin kwarg won't be ideal because each platform overrides the the set/get_lpmode functions itself, so even when lpmode pin is used, existing platforms would use the default path instead of throwing an error.

Instead I think we should invoke set_lpmode_via_pin directly (without the base class kwarg)

Based on this approach,
I have the base class updated here:
https://github.com/sonic-net/sonic-platform-common/pull/663
pddf changes here:
https://github.com/sonic-net/sonic-buildimage/pull/27278

and I updated the current PR to use lpmode_via_pin directly in sfputil when the flag is passed

Let me know what you think of this approach.

sfputil/main.py
```diff
@@ -1253,15 +1257,20 @@ def set_lpmode(logical_port, enable):
         ganged = True
 
     for physical_port in physical_port_list:
+        port_name = get_physical_port_name(logical_port, i, ganged)
+        i += 1
```
prgeor: @abhi-nexthop Why this is moved up?

sfputil/main.py
```diff
@@ -1253,15 +1257,20 @@ def set_lpmode(logical_port, enable):
...
             if not sfp.get_presence():
                 click.echo(f"{logical_port}: module {physical_port} is not present, skipping")
                 continue
             click.echo("{} low-power mode for port {} ... ".format(
                 "Enabling" if enable else "Disabling",
-                get_physical_port_name(logical_port, i, ganged)), nl=False)
-            result = sfp.set_lpmode(enable)
+                port_name), nl=False)
+            if use_lpmode_pin:
+                result = sfp.set_lpmode_via_pin(enable)
```
prgeor, line 1278: @abhi-nexthop do you have the PR ready for this new API in platform-common repo? `set_lpmode_via_pin`?

abhi-nexthop: Yes, I just updated it.

https://github.com/sonic-net/sonic-platform-common/pull/663
pddf changes here:
https://github.com/sonic-net/sonic-buildimage/pull/27278

sfputil/main.py
abhi-nexthop: I forgot to mention but this is a bit of a logical bugfix.

get_physical_port_name(port_name, i, ganged), if we don't increment i each loop (which can happen if the NotImplementedError is thrown in the try-catch), then get_physical_port_name gets desynchronised and we are off by one port number. So the physical and logical ports printed in the error would not correspond to each other

sfputil/main.py
abhi-nexthop: Same reason as above. The call for get_physical_port_name will be misaligned with the logical port you are operating with because i+1 never happens

sfputil/main.py
```diff
@@ -1253,15 +1257,20 @@ def set_lpmode(logical_port, enable):
...
+            else:
+                result = sfp.set_lpmode(enable)
         except NotImplementedError:
             click.echo("This functionality is currently not implemented for this platform")
             sys.exit(ERROR_NOT_IMPLEMENTED)
```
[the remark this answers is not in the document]
prgeor: @abhi-nexthop ?

sfputil/main.py
```diff
@@ -1167,13 +1168,18 @@ def lpmode(port):
...
                     if not sfp.get_presence():
                         output_table.append([port_name, "Not Present"])
                         continue
-                    lpmode = sfp.get_lpmode()
+                    if use_lpmode_pin:
+                        lpmode = sfp.get_lpmode_via_pin()
+                    else:
+                        lpmode = sfp.get_lpmode()
+
                 except NotImplementedError:
```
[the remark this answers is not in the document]
prgeor: @abhi-nexthop ?

sfputil/main.py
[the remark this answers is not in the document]
abhi-nexthop: Fixed

sfputil/main.py
prgeor: @abhi-nexthop I am bit lost here. there is a sys.exit for "NotImplemented" so the program terminates immediately. If you are talking about the `continue` then there are platforms for which the i starts from zero. Here I think `i` represent the sfp object index from 0,1,..n-1

sfputil/main.py
```diff
@@ -1167,14 +1168,19 @@ def lpmode(port):
 
             for physical_port in physical_port_list:
                 port_name = get_physical_port_name(logical_port_name, i, ganged)
+                i += 1
 
+                sfp = platform_chassis.get_sfp(physical_port)
                 try:
-                    sfp = platform_chassis.get_sfp(physical_port)
```
mihirpat1, line 1180: @abhi-nexthop Why is this being moved out of the try/except block?

sfputil/main.py
```diff
@@ -1167,14 +1168,19 @@ def lpmode(port):
...
                         continue
-                    lpmode = sfp.get_lpmode()
-                except NotImplementedError:
+                    if use_lpmode_pin:
+                        lpmode = sfp.get_lpmode_via_pin()
+                    else:
+                        lpmode = sfp.get_lpmode()
+
+                except (NotImplementedError, AttributeError):
                     click.echo("This functionality is currently not implemented for this platform")
```
mihirpat1: @abhi-nexthop Can you modify the log message to have more details on the exception being caught?

sfputil/main.py
abhi-nexthop: You're right, I was thinking in terms of the continue. Fixed

mihirpat1 (dismissed):

prgeor (approved):

PR #4516, opened 2026-05-05
parallel test case execution fixes

bhouse-nexthop:
## Why I did it

Fix critical test isolation issues in the parallel test infrastructure introduced by #4439 (commit 604df54c). These issues only manifest when running tests in parallel with pytest-xdist and cause test failures due to cross-test pollution.

The issues are:
1. **Test isolation failures**: sonic_platform mocks from one test file polluted subsequent test files on the same xdist worker, causing import errors and test failures
2. **Mock lifecycle issues**: Module-level mock injection was cleared by conftest hooks before tests ran, causing import failures in chassis and syseeprom tests

## How I did it

**Commit 1: Fix test isolation - clear sonic_platform mocks between test files**
- Modified tests/conftest.py to clear sonic_platform modules from sys.modules in the _reset_between_files() hook
- Prevents cross-test pollution when xdist workers are reused with --dist loadfile

**Commit 2: Fix sonic_platform mock injection for chassis and syseeprom tests**
- Moved sonic_platform mock injection from module-level to per-test autouse fixtures in tests/chassis_modules_test.py and tests/decode_syseeprom_test.py
- Fixes import failures when conftest._reset_between_files() clears sys.modules before tests run
- Per-test fixtures re-inject the mock after the conftest hook runs, ensuring it's available for lazy imports

## How to verify it

Run the parallel test suite to verify all tests pass with zero failures:

python3 -m pytest tests/ -n auto --dist loadfile
python3 -m pytest tests/ -n 1 --dist loadfile
python3 -m pytest tests/ -p no:xdist

Run specific affected test files to verify fixes:

python3 -m pytest tests/chassis_modules_test.py -v
python3 -m pytest tests/decode_syseeprom_test.py -v

Expected: All tests pass in all three execution modes (parallel, single-worker xdist, pure serial).

## Which release branch to backport

Master only. These fixes are required for the parallel test infrastructure introduced in #4439 (commit 604df54c).

PR #4519, opened 2026-05-06
[ecnconfig]: Fix races under pytest-xdist parallel execution

rminnikanti:
#### What I did
When running tests multiple workers execute test files simultaneously in the same process tree. All workers were writing and reading the same hardcoded path /tmp/ecnconfig, causing three distinct failures:

tests/multi_asic_ecnconfig_test.py::TestEcnConfigMultiAsic::test_ecn_config_change_other_threshold_masic
- KeyError: 'asic1' FAILED tests/ecn_test.py::TestEcnConfig::test_ecn_queue_set_q_off - KeyError: ''
FAILED
tests/ecn_test.py::TestEcnConfig::test_ecn_queue_set_all_on_verbose - KeyError: ''
FAILED tests/ecn_test.py::TestEcnConfig::test_ecn_queue_set_lossy_q_on - fileNotFoundError: [Errno 2] No such file or directory: '/tmp/ecnconfig'

#### How I did it
Fix mirrors the pattern already applied to mmuconfig (commit 604df54c):

1. scripts/ecnconfig: read ECNCONFIG_FILE env var (falling back to /tmp/ecnconfig) instead of hardcoding the path.  This allows the subprocess spawned by get_result_and_return_code() to write to the same per-worker path that the parent test process expects to read.

2. tests/ecn_test.py (TestEcnConfigBase): in setup_class, derive the output path from the WORKER_TMP env var that conftest.py sets to a per-worker directory (/tmp/worker-gw0/, /tmp/worker-gw1/, ...). Export it as ECNCONFIG_FILE so the ecnconfig subprocess inherits it. teardown_class removes the per-worker file and clears the env var.

3. tests/multi_asic_ecnconfig_test.py (TestEcnConfigMultiAsic): add super().setup_class() so the base class initialises _ecnconfig_file before the subclass overrides the topology. Without this call _ecnconfig_file remained None, causing TypeError when the executor tried to open the file.

#### How to verify it
Build sonic_utilities-1.2-py3-none-any.whl or run the pytest

### DIFF
--- a/scripts/ecnconfig
+++ b/scripts/ecnconfig
@@ -392,7 +392,7 @@ def main(command, show_config, profile, green_min,
          red_drop_prob, namespace, verbose, queue):
     test_filename = None
     if os.environ.get("UTILITIES_UNIT_TESTING", "0") == "2":
-        test_filename = '/tmp/ecnconfig'
+        test_filename = os.environ.get("ECNCONFIG_FILE", "/tmp/ecnconfig")
 
     try:
         load_db_config()
--- a/tests/ecn_test.py
+++ b/tests/ecn_test.py
@@ -23,9 +23,15 @@ sys.path.insert(0, modules_path)
 
 
 class TestEcnConfigBase(object):
+    # Per-worker file path to avoid race conditions in parallel test runs.
+    _ecnconfig_file = None
+
     @classmethod
     def setup_class(cls):
         print("SETUP")
+        worker_tmp = os.environ.get('WORKER_TMP', '/tmp')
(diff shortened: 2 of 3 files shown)


rminnikanti:
@bhouse-nexthop @saiarcot895 please review

bhouse-nexthop:
Yeah, unfortunately I probably missed a couple.  I tried running under a variety of parallelism flags to stress it and change execution order, but it wasn't perfect when I did it.  

Can you also cross-review #4516 which has additional fixes I uncovered in my own testing?

bhouse-nexthop (approved):

PR #4528, opened 2026-05-07
[show/vnet] Add pretty print wrapping for local/non-tunnel routes

miatttao:
#### What I did
Extended the pretty-print wrapping logic (from #4452) to `show vnet routes local` and `show vnet routes all` (non-tunnel/local routes). When a local VNET route has many ECMP nexthops and interfaces (e.g. 15 endpoints), the `nexthop` and `interface` columns are now wrapped into multiple rows instead of printing one extremely long line.

Depends on: #4452.

#### How I did it
- Renamed `pretty_print` → `pretty_print_tunnel` for clarity.
- Added `pretty_print_local()` that applies the same row wrapping to the `nexthop` / `interface` columns of local routes.
- Updated `_show_local_helper()` to call `pretty_print_local()` instead of appending raw comma-separated values.
- Both `pretty_print_tunnel` and `pretty_print_local` wrap at a fixed 2 items per row. Terminal-width-adaptive sizing isn't feasible here because rows are built one key at a time, before column widths are known — so this replaces the previous `2 if max_len > 15 else 3` heuristic with a single consistent cap.
- `pretty_print_tunnel` also splits `mac_address` and `vni` into the same chunks as `endpoint` when they are per-endpoint lists, so all three columns stay row-aligned at any ECMP scale.
- Original value spacing from the DB is preserved verbatim (values are split and re-joined on `,`).
- Fixed `show vnet endpoint <ip>`: `endpoint` / `endpoint_monitor` values are now stripped when split, so the exact-match filter and `BFD_SESSION_TABLE` key lookups work whether the DB uses `,` or `, ` separators.
- Added mock data with 15 nexthops (`Vnet_7959668`), normalized vnet mock data to consistent `, ` separators, and added unit tests for the new wrapping behavior.

#### How to verify it
On a SONiC device with a VNET route that has over 16 ECMP nexthops:
```
show vnet routes local
show vnet routes all
```

```
vnet name     prefix           nexthop                                                                                                                                                                                        interface
------------  ---------------  ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------  --------------------------------------------------------------------------------------------------------------------------------------------------------------------
Vnet_7959668  10.32.0.0/17     10.33.254.1,10.33.254.3,10.33.254.5,10.33.254.7,10.33.254.9,10.33.254.11,10.33.254.13,10.33.254.15,10.33.254.17,10.33.254.19,10.33.254.23,10.33.254.25,10.33.254.27,10.33.254.29,10.33.254.31  Po1031.106,Po1032.106,Po1033.106,Po1034.106,Po1035.106,Po1036.106,Po1037.106,Po1038.106,Po1039.106,Po1040.106,Po1042.106,Po1043.106,Po1044.106,Po1045.106,Po1046.106
```
```
vnet name        prefix            nexthop                     interface
---------------  ----------------  --------------------------  ----------------------
Vnet_7959668     10.32.0.0/17      10.33.254.1, 10.33.254.3    Po1031.106, Po1032.106
                                   10.33.254.5, 10.33.254.7    Po1033.106, Po1034.106
                                   10.33.254.9, 10.33.254.11   Po1035.106, Po1036.106
                                   10.33.254.13, 10.33.254.15  Po1037.106, Po1038.106
                                   10.33.254.17, 10.33.254.19  Po1039.106, Po1040.106
                                   10.33.254.23, 10.33.254.25  Po1042.106, Po1043.106
                                   10.33.254.27, 10.33.254.29  Po1044.106, Po1045.106
                                   10.33.254.31                Po1046.106
test_v4_in_v4-0  160.162.191.1/32  100.100.4.1                 Ethernet1
test_v4_in_v4-0  160.163.191.1/32  100.101.4.1, 100.101.4.2    Ethernet1, Ethernet2
```
```
show vnet route all
show vnet route local
```
on `miatao@LVL02-0101-0705-19T0`
image

image


show/vnet.py
```diff
@@ -526,6 +526,29 @@ def routes():
...
+        r.append(",".join(nexthops[i:i + row_width]) if i < len(nexthops) else "")
+        if interfaces:
+            r.append(",".join(interfaces[i:i + row_width]) if i < len(interfaces) else "")
+        else:
+            r.append(ifname_val if i == 0 else "")
+        i += row_width
+        table.append(r)
+        r = ["", ""]
+
+
```
anish-n, line 571: Could we rename the below function to pretty_print_tunnel for clarity?

show/vnet.py
```diff
@@ -526,6 +526,29 @@ def routes():
...
+    max_len = max((len(item) for item in all_items), default=0)
+    row_width = 2 if max_len > 15 else 3
+
+    max_entries = max(len(nexthops), len(interfaces))
+    i = 0
+    while i < max_entries:
+        r.append(",".join(nexthops[i:i + row_width]) if i < len(nexthops) else "")
+        if interfaces:
+            r.append(",".join(interfaces[i:i + row_width]) if i < len(interfaces) else "")
+        else:
```
anish-n: seems like this else logic  is not needed, can you check if it can be removed, or it causes some problem upon removal. If it can be removed, please also remove the redundant else cases in pretty_print_tunnel as well

anish-n (changes requested):

show/vnet.py
miatttao: Updated to `pretty_print_tunnel`. Thanks for the suggestion.

show/vnet.py
miatttao: I checked this. When `interfaces` is empty, `ifname_val` is also empty because `interfaces` is derived directly from `ifname_val`. As a result, the `else` branch effectively appends an empty string and is not needed for column alignment. I've simplified this logic accordingly.

alawing (approved):

tests/show_vnet_test.py
```diff
@@ -1,17 +1,28 @@
 # flake8: noqa: E501
 import os
 import pytest
+from unittest.mock import patch
 from click.testing import CliRunner
 from utilities_common.db import Db
 import show.main as show
 import show.vnet as vnet
 from tests.mock_tables import dbconnector
```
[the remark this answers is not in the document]
miatttao: Fixed.

tests/show_vnet_test.py
```diff
@@ -130,19 +141,60 @@ def test_pretty_print_scale(self, N, vnet_name, prefix, metric):
...
+        table = []
+        row = ["TestVnet", "10.0.0.0/24"]
+        vnet.pretty_print_local(table, row, "192.168.1.1", "Ethernet1")
+        assert table == [["TestVnet", "10.0.0.0/24", "192.168.1.1", "Ethernet1"]]
+
+        # 3 short nexthops, row_width=3, all fit on one row
+        table = []
+        row = ["TestVnet", "10.0.0.0/24"]
+        vnet.pretty_print_local(table, row, "1.1.1.1,2.2.2.2,3.3.3.3", "Eth1,Eth2,Eth3")
+        assert table == [["TestVnet", "10.0.0.0/24", "1.1.1.1,2.2.2.2,3.3.3.3", "Eth1,Eth2,Eth3"]]
```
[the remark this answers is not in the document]
miatttao: Fixed.

tests/show_vnet_test.py
```diff
@@ -244,28 +316,29 @@ def test_show_vnet_routes_local_basic(self):
...
+test_v4_in_v4-1  160.165.191.1/32  100.103.4.1,100.103.4.2    Ethernet1,Ethernet2
+                                   100.103.4.3                Ethernet3
+
+vnet name        prefix            nexthop                  interface
+---------------  ----------------  -----------------------  -------------------
+test_v4_in_v4-0  160.162.191.1/32  100.100.4.1              Ethernet1
+test_v4_in_v4-0  160.163.191.1/32  100.101.4.1,100.101.4.2  Ethernet1,Ethernet2
+test_v4_in_v4-0  160.164.191.1/32  100.102.4.1,100.102.4.2  Ethernet1,Ethernet2
+                                   100.102.4.3              Ethernet3
 """
```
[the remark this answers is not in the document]
miatttao: Fixed

show/vnet.py
```diff
@@ -526,10 +529,45 @@ def routes():
...
+def pretty_print_local(table, r, nexthop_val, ifname_val):
+    nexthops = [nexthop.strip() for nexthop in nexthop_val.split(',')] if nexthop_val else [""]
+    interfaces = [interface.strip() for interface in ifname_val.split(',')] if ifname_val else []
+
+    all_items = list(nexthops) + list(interfaces)
+    max_len = max((len(item) for item in all_items), default=0)
+    # route_header: ['vnet name', 'prefix', 'nexthop', 'interface']
+    # fixed_cols_est = vnet_name(~15) + prefix(~18) + 4-col spacing(6) = 39
+    # cap at 2 so nexthop/interface columns stay readable on any terminal width
+    row_width = min(2, _ecmp_row_width(max_len, num_wrap_cols=2, fixed_cols_est=39))
```
anish-n: Do we need this min? Can we unify the approach used for both tunnel and local route in terms of how many items are displayed, and put the unified logic in _ecmp_row_width instead of having additional row width handling logic here

anish-n (changes requested):

show/vnet.py
miatttao: The min(2, ...) is needed to cap local routes at 2 entries per line (as previously requested). 
Yes, we can unify the approach. I've moved that cap into `_ecmp_row_width` via the `max_per_row=2` parameter, so both `pretty_print_local` and `pretty_print_tunnel` now just call `_ecmp_row_width` and use the result directly.

alawing (approved):

tests/show_vnet_test.py
```diff
@@ -244,28 +320,22 @@ def test_show_vnet_routes_local_basic(self):
...
-vnet name        prefix            nexthop                                interface
----------------  ----------------  -------------------------------------  -------------------------------
-test_v4_in_v4-0  160.162.191.1/32  100.100.4.1                            Ethernet1
-test_v4_in_v4-0  160.163.191.1/32  100.101.4.1, 100.101.4.2               Ethernet1, Ethernet2
-test_v4_in_v4-0  160.164.191.1/32  100.102.4.1, 100.102.4.2, 100.102.4.3  Ethernet1, Ethernet2, Ethernet3
-test_v4_in_v4-1  160.165.191.1/32  100.103.4.1, 100.103.4.2, 100.103.4.3  Ethernet1, Ethernet2, Ethernet3
-"""
-        assert result.output == expected_output
-
-    def test_show_vnet_routes_local_vnetname(self):
```
anish-n, line 256: This test should not be deleted, please add it back with the correct output checked against

tests/show_vnet_test.py
```diff
@@ -244,28 +320,22 @@ def test_show_vnet_routes_local_basic(self):
...
-        assert result.exit_code == 0
-        expected_output = """\
-vnet name        prefix            nexthop                                interface
----------------  ----------------  -------------------------------------  -------------------------------
-test_v4_in_v4-0  160.162.191.1/32  100.100.4.1                            Ethernet1
-test_v4_in_v4-0  160.163.191.1/32  100.101.4.1, 100.101.4.2               Ethernet1, Ethernet2
-test_v4_in_v4-0  160.164.191.1/32  100.102.4.1, 100.102.4.2, 100.102.4.3  Ethernet1, Ethernet2, Ethernet3
+vnet name        prefix            nexthop                    interface
+---------------  ----------------  -------------------------  ---------------------
+Vnet_7959668     10.32.0.0/17      10.33.254.1,10.33.254.3    Po1031.106,Po1032.106
```
anish-n, line 325: there is space stripping occurring here, can you please keep the space after the comma to maintain existing behavior and not change the output

show/vnet.py
```diff
@@ -526,10 +529,48 @@ def routes():
     pass
 
 
-def pretty_print(table, r, epval, mac_addr, vni, metric, state):
-    endpoints = epval.split(',') if epval else []
-    if not endpoints:
-        endpoints = [""]
+def _ecmp_row_width(max_item_len, num_wrap_cols, fixed_cols_est, max_per_row=None):
```
anish-n, line 532: fixed_cols_est  magic numbers are stale for realistic data. For local routes you use  fixed_cols_est=39  from  vnet_name(~15) + prefix(~18) + spacing(6) . But e.g.  Vnet_7959668  is 12 chars while an IPv6 prefix like  fddd:a156:a251::a6:1/128  is 24. In the new mock,  10.32.0.0/17  is only 12 chars. When the estimate is wrong the row-width calc is skewed — the  max_per_row=2  cap saves you, but it means the terminal-width branch is essentially cosmetic. Either derive  fixed_cols_est  from the actual max widths already available in the caller (walk  table  before wrapping to compute), or acknowledge that the cap is the real control and simplify and just use 2 all the time.

tests/show_vnet_test.py
```diff
@@ -303,22 +376,26 @@ def test_show_vnet_routes_tunnel_ecmp(self):
...
+            "Vnet_7127926  30.0.21.0/24  100.106.230.44,10.134.85.10      00:22:48:03:8c:f8,60:45:bd:a3:8d:ab  7127926,7127926  5         active\n"
+            "                            100.106.229.38,100.106.229.170   60:45:bd:a3:21:88,60:45:bd:a2:e4:39  7127926,7127926\n"
+            "                            100.106.228.160,10.134.84.24     7c:1e:52:06:89:0f,7c:1e:52:06:8b:cd  7127926,7127926\n"
+            "                            100.106.230.168,10.90.92.16      60:45:bd:a3:8f:ae,60:45:bd:a2:e8:f9  7127926,7127926\n"
+            "                            10.224.116.42,100.106.228.134    60:45:bd:a2:e5:ee,60:45:bd:a4:be:3e  7127926,7127926\n"
+            "                            100.106.229.171,100.106.228.161  60:45:bd:a3:8d:ac,7c:1e:52:06:89:10  7127926,7127926\n"
+        )
+        assert expected_21 in output
+
+        # 30.0.20.0/24: 287 entries. Content assertions only.
```
anish-n, line 392: Why do we change this?

show/vnet.py
miatttao: Agreed — the terminal-width branch was cosmetic given the `max_per_row=2` cap, so I simplified: removed `fixed_cols_est`/`_ecmp_row_width()` and hardcoded `row_width=2` for both local and tunnel wrapping. Tests updated to match.

tests/show_vnet_test.py
miatttao: This was used for a scale testing at 287 endpoints, which was not implemented in mock data. Reverted.

tests/show_vnet_test.py
miatttao: Fixed. Removed the `.strip()` in `pretty_print_local` so the DB value's spacing is preserved, matching the old behavior of printing `val.get(nexthop)` directly.

tests/show_vnet_test.py
miatttao: Added back with the output updated for the new wrapping; the 3-nexthop route now wraps to a second row, and the space after the comma is preserved.

PR #4536, opened 2026-05-11
[show, clear]: add 'orchagent tasks' commands to surface per-Executor timing + scheduling latency

venkit-nexthop:
#### What I did

Add two new CLI commands that surface orchagent's per-Executor execution-time
statistics so operators can answer "where is orchagent spending its time?"
without rebuilding it:

- `show orchagent tasks` — per-Executor run-time and scheduling-latency table.
- `sonic-clear orchagent tasks` — reset the counters in place.

This is the CLI half; the orchagent-side instrumentation that publishes the
reply lives in a companion `sonic-swss` change. The CLI degrades gracefully
(times out with a non-zero exit code) if the daemon side is not present.

#### How I did it

Two new click groups (`show.orchagent.tasks` and `clear.orchagent.tasks`)
wired into `show/main.py` and `clear/main.py`.

The CLI talks to orchagent over two APPL_DB notification channels
(`ORCH_TASK_STATS_QUERY` / `ORCH_TASK_STATS_REPLY`) and parses the
14-field pipe-separated reply produced by orchagent:

```
count | total_run_ns
  | median_run_ns | q1_run_ns | q3_run_ns | max_run_ns
  | high_outliers | low_outliers
  | sched_count   | total_sched_ns
  | median_sched_ns | q1_sched_ns | q3_sched_ns | max_sched_ns
```

Rendering:

- **RUN TIME** — P² `median/q1/q3/max` of per-invocation wall-clock duration.
  Median is the headline, Q1/Q3 give spread, max exposes the worst tail.
- **RUNS** — number of completed invocations.
- **OUTLIERS** — Tukey 1.5×IQR sum (high + low).
- **SCHED LATENCY** — P² `median/q1/q3/max` of the gap between when the task
  finished and when it was next scheduled. Exposes select-loop starvation.
- **TOTAL** — cumulative `<run>/<sched>` wall-clock spent inside the task vs
  waiting before it.

Rows are sorted by `total_run_ns` descending; ties break by name ascending.
Empty slots (count = 0) print `-` for quartet/total cells but keep integer
`RUNS` / `OUTLIERS`.

`sonic-clear orchagent tasks` sends `op="clear"` on the same query channel
and prints `OK` on a successful round-trip.

`doc/Command-Reference.md` is updated with a new **Orchagent** section
(TOC + body) per the contributing guideline for new `show` / `sonic-clear`
subcommands.

#### How to verify it

`tests/show_orchagent_tasks_test.py` and `tests/clear_orchagent_tasks_test.py`
mock `swsscommon.NotificationProducer` / `NotificationConsumer` / `Select`
so the round-trip is exercised end-to-end without Redis. Coverage:

- Header presence (TASK / RUN TIME / median/q1/q3/max / RUNS / OUTLIERS /
  SCHED LATENCY / TOTAL / run/sched / (in msec))
- Sort order (`total_run_ns` descending, name asc tiebreak)
- Quartet formatting (ms, two decimals, no per-cell unit suffix)
- Outlier counts visible (high + low summed)
- TOTAL column = `<run>/<sched>` with empty-row dashes
- Empty-reply case prints headers only
- Timeout and orchagent-error reply paths surface as exit-code-1
- Correct op string is sent (`show` / `clear`)

12 / 12 pytests pass against the mocked swsscommon round-trip.

Manual: with the companion `sonic-swss` change loaded, run the commands on
a live switch and observe per-Executor stats / reset.

```
admin@sonic:~$ show orchagent tasks
TASK         RUN TIME                          RUNS  OUTLIERS  SCHED LATENCY                    TOTAL
             median/q1/q3/max                                  median/q1/q3/max                 run/sched
             (in msec)                                         (in msec)                        (in msec)
ROUTE_TABLE  1745.53/1391.34/2242.07/3913.36     43         5  1.06/0.41/48.40/1436.01          77619.63/5.04

admin@sonic:~$ sonic-clear orchagent tasks
OK
```


venkit-nexthop:
@prsunny @prabhataravind @vivekrnv @deepak-singhal0408 
Could you take a look and approve this PR if you think it is good.
This PR implements the CLI required for the `show orchagent tasks` and `sonic-clear orchagent tasks`
(described in HLD https://github.com/sonic-net/SONiC/pull/2328)

venkit-nexthop:
@deepak-singhal0408 @prsunny 
CI has passed. Can you take a look and approve the PR?

show/orchagent.py
```diff
@@ -0,0 +1,177 @@
...
+            f"{_ms(max_ns):.2f}")
+
+
+def _query_orchagent(op):
+    """Send a query to orchagent over APPL_DB notification channels and
+    return (op_ret, data_ret, fvs) on success.
+
+    Raises RuntimeError on timeout or transport error.
+    """
+    db = swsscommon.DBConnector("APPL_DB", 0)
```
deepak-singhal0408: **Multi-ASIC: not handled **

`DBConnector("APPL_DB", 0)` is the global/default-namespace form. On a multi-ASIC linecard, orchagent runs **per-ASIC inside each `asicN` namespace** with its own APPL_DB, so the `ORCH_TASK_STATS_QUERY` / `_REPLY` channels are per-namespace. The host/default APPL_DB this connects to has no orchagent responder → the `Select` just hits the 10s timeout. Same issue in `clear/orchagent.py` (line 13).

Please make it namespace-aware:
- add `@multi_asic.multi_asic_click_options` (`-n/--namespace`, `-d/--display`)
- iterate `multi_asic.get_namespace_list()` and build the connector per-ns:
  `swsscommon.DBConnector("APPL_DB", REDIS_TIMEOUT_MSECS, True, namespace)`
- render with an ASIC/namespace column and aggregate across namespaces.

Pattern reference: `show/muxcable.py` (4-arg `DBConnector(db_name, REDIS_TIMEOUT_MSECS, True, namespace)`).

This single change also settles the chassis story cleanly, so no extra plumbing is needed:
- **Single-ASIC / pizza box** — `get_namespace_list()` → `['']`; behavior unchanged.
- **Supervisor** — `get_namespace_list()` returns only front-panel (`front_ns`) namespaces, which is empty on a sup (only fabric ASICs run there, `switch_type: fabric` — no route orchagent). The command becomes a graceful no-op instead of a blind 10s timeout, with no special `is_supervisor()` gate required.
- **Cross-linecard aggregation** — intentionally out of scope for this CLI; operators can fan out with `rexec all -c "show orchagent tasks"`, and each linecard then returns all of its ASICs via the per-ns iteration above.

deepak-singhal0408 (changes requested):
Please take care of multi-asic case as well.

show/orchagent.py
venkit-nexthop: done

venkit-nexthop:
> Please take care of multi-asic case as well.

can you take a look now.

doc/Command-Reference.md
```diff
@@ -11021,6 +11024,64 @@ This command is used to delete a configured NTP server IP address.
 
 Go Back To [Beginning of the document](#) or [Beginning of this section](#NTP)
 
+## Orchagent
```
deepak-singhal0408, line 11027: could you enhance this to cover multi-asic as well?

venkit-nexthop: done

deepak-singhal0408 (approved):

PR #4551, opened 2026-05-17
[tempershow] guard global DB config load behind multi-asic check

yijingyan2:
#### What I did
Fixed `scripts/tempershow` so it no longer unconditionally calls `SonicDBConfig.load_sonic_global_db_config()`. On single-ASIC platforms the global DB config file `/var/run/redis/sonic-db/database_global.json` does not exist, and the swss-common C++ layer logs an ERR every time `tempershow` runs:
```
ERR python3: :- initializeGlobalConfig: Sonic database config global file
doesn't exist at /var/run/redis/sonic-db/database_global.json
```
Because `show techsupport` invokes `tempershow`, this ERR is captured by LogAnalyzer and fails sonic-mgmt test `show_techsupport/test_auto_techsupport.py::TestAutoTechSupport::test_sanity` on every single-ASIC run (e.g. KVM t0 PR checks).
 
#### How I did it
- Detect multi-ASIC via `sonic_py_common.multi_asic.is_multi_asic()`.
- On multi-ASIC: call `SonicDBConfig.load_sonic_global_db_config()` only if `isGlobalInit()` is False (unchanged behavior).
- On single-ASIC: call `SonicDBConfig.load_sonic_db_config()` only if `isInit()` is False — matches the pattern used by other utilities (e.g. `psushow`, `show` CLI helpers).

### DIFF
--- a/scripts/tempershow
+++ b/scripts/tempershow
@@ -8,6 +8,7 @@ import json
 from datetime import datetime
 
 from tabulate import tabulate
+from sonic_py_common import multi_asic
 from swsscommon.swsscommon import SonicDBConfig, SonicV2Connector
 from natsort import natsorted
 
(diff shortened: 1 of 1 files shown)


vvolam (approved):
LGTM, Thank you!

PR #4556, opened 2026-05-19
Dualtor neighbor flush inconsistent neighbors

yyynini:
#### What I did
Made neighbor inconsistency mitigation the default behavior in dualtor_neighbor_check.py. When inconsistent neighbor entries are detected on mux ports, the script now flushes those entries once, waits 10 seconds, then reruns the consistency check once to verify recovery.
 
If inconsistency remains after the post-flush check, the script logs an ALERT and exits non-zero.

#### How I did it
1. Added helper logic to flush inconsistent neighbor entries with:
   - `sudo ip -4 neigh flush to <neighbor-ip>` for IPv4
   - `sudo ip -6 neigh flush to <neighbor-ip>` for IPv6
2. Iterates failed neighbors, deduplicates by IP, flushes all inconsistent entries (both host-route and prefix-route).
3. After flush, waits 10 seconds for kernel ARP/ND relearning and orchagent reconvergence, then reruns the consistency check once to allow orchagent to re-learn correct state.
4. Logs an ALERT if the post-flush check still finds inconsistent neighbors.

#### How to verify it
 Verified by:
 
 ```bash
yawenni@sonic-mgmt-yawenni_master:/var/src/sonic-utilities$ python -m py_compile scripts/dualtor_neighbor_check.py tests/dualtor_neighbor_check_test.py
yawenni@sonic-mgmt-yawenni_master:/var/src/sonic-utilities$ 
 ```
Mocked DB/ASIC state stayed inconsistent, and correctly preserved failure and printed post-flush check still found inconsistent neighbors.
 ```bash
admin@str2-8101c1-02:~$  dualtor_neighbor_check.py
================================================================================
Neighbors in PREFIX-ROUTE mode:
================================================================================
NEIGHBOR       MAC                PORT       MUX_STATE    IN_MUX_TOGGLE    NEIGHBOR_IN_ASIC    PREFIX_ROUTE    NEXTHOP_TYPE    HWSTATUS
-------------  -----------------  ---------  -----------  ---------------  ------------------  --------------  --------------  ------------
192.168.0.129  1a:91:34:a2:90:01  Ethernet8  active       no               no                  no              N/A             inconsistent
fc02:1000::81  1a:91:34:a2:90:01  Ethernet8  active       no               yes                 yes             NEIGHBOR        consistent

Found neighbors that are inconsistent with mux states: ['192.168.0.129']
Failed PREFIX-ROUTE neighbors:
NEIGHBOR       MAC                PORT       MUX_STATE    IN_MUX_TOGGLE    NEIGHBOR_IN_ASIC    PREFIX_ROUTE    NEXTHOP_TYPE    HWSTATUS
-------------  -----------------  ---------  -----------  ---------------  ------------------  --------------  --------------  ------------
192.168.0.129  1a:91:34:a2:90:01  Ethernet8  active       no               no                  no              N/A             inconsistent
Flushing inconsistent neighbor entry: 192.168.0.129
Flushed 1 inconsistent neighbor entries.
Waiting 10 seconds before rerunning dualtor neighbor check.
================================================================================
Neighbors in PREFIX-ROUTE mode:
================================================================================
NEIGHBOR       MAC                PORT       MUX_STATE    IN_MUX_TOGGLE    NEIGHBOR_IN_ASIC    PREFIX_ROUTE    NEXTHOP_TYPE    HWSTATUS
-------------  -----------------  ---------  -----------  ---------------  ------------------  --------------  --------------  ------------
192.168.0.129  1a:91:34:a2:90:01  Ethernet8  active       no               no                  no              N/A             inconsistent
fc02:1000::81  1a:91:34:a2:90:01  Ethernet8  active       no               yes                 yes             NEIGHBOR        consistent

Found neighbors that are inconsistent with mux states: ['192.168.0.129']
Failed PREFIX-ROUTE neighbors:
NEIGHBOR       MAC                PORT       MUX_STATE    IN_MUX_TOGGLE    NEIGHBOR_IN_ASIC    PREFIX_ROUTE    NEXTHOP_TYPE    HWSTATUS
-------------  -----------------  ---------  -----------  ---------------  ------------------  --------------  --------------  ------------
192.168.0.129  1a:91:34:a2:90:01  Ethernet8  active       no               no                  no              N/A             inconsistent
ALERT: post-flush dualtor neighbor check still found inconsistent neighbors.
 ```

The script reported inconsistent neighbors and exited non-zero. It did not attempt mitigation.

The script reports inconsistent neighbors, flushes each unique inconsistent neighbor IP once, waits 10 seconds, reruns the consistency check once, and exits based on the post-flush result.

### DIFF
--- a/scripts/dualtor_neighbor_check.py
+++ b/scripts/dualtor_neighbor_check.py
@@ -204,6 +204,8 @@ result['asic_nexthop_table'] = asic_nexthop_table
 return redis.status_reply(cjson.encode(result))
 """
 
+# Allow kernel ARP/ND and orchagent to converge before validating the
+# one-time automatic neighbor flush mitigation.
 POST_FLUSH_CHECK_DELAY_SEC = 10
 
 DB_READ_SCRIPT_CONFIG_DB_KEY = "_DUALTOR_NEIGHBOR_CHECK_SCRIPT_SHA1"
@@ -271,7 +273,7 @@ def parse_args():
     parser.add_argument(
         "--flush-inconsistent-neighbors",
         action="store_true",
-        help="flush inconsistent kernel neighbor entries and rerun the check"
+        help="deprecated: inconsistent kernel neighbor entries are flushed once by default"
     )
     args = parser.parse_args()
 
@@ -370,16 +372,22 @@ def flush_neighbor(neighbor_ip):
 
 def flush_inconsistent_neighbors(failed_neighbors):
     """Flush kernel neighbor entries for failed neighbors."""
+    attempted_neighbors = set()
     flushed_neighbors = set()
 
     for neighbor in failed_neighbors:
         # Inconsistency can affect both host-route and prefix-route neighbors.
         neighbor_ip = neighbor["NEIGHBOR"]
-        if neighbor_ip in flushed_neighbors:
+        if neighbor_ip in attempted_neighbors:
             continue
 
+        attempted_neighbors.add(neighbor_ip)
         WRITE_LOG_WARN("Flushing inconsistent neighbor entry: %s", neighbor_ip)
-        flush_neighbor(neighbor_ip)
+        try:
+            flush_neighbor(neighbor_ip)
+        except (RuntimeError, ValueError) as err:
+            WRITE_LOG_ERROR("Failed to flush inconsistent neighbor entry %s: %s", neighbor_ip, err)
+            continue
         flushed_neighbors.add(neighbor_ip)
 
     return len(flushed_neighbors)
@@ -759,7 +767,7 @@ def main():
     check_results = run_neighbor_check(appl_db, mux_server_to_port_map, if_oid_to_port_name_map)
     res, failed_neighbors = parse_check_results(check_results)
 
-    if not res and args.flush_inconsistent_neighbors:
+    if not res:
         flush_count = flush_inconsistent_neighbors(failed_neighbors)
         if flush_count:
             WRITE_LOG_WARN("Flushed %d inconsistent neighbor entries.", flush_count)
@@ -771,7 +779,9 @@ def main():
             check_results = run_neighbor_check(appl_db, mux_server_to_port_map, if_oid_to_port_name_map)
             res, _ = parse_check_results(check_results)
             if not res:
-                WRITE_LOG_WARN("Post-flush dualtor neighbor check still found inconsistent neighbors.")
+                WRITE_LOG_ERROR("ALERT: post-flush dualtor neighbor check still found inconsistent neighbors.")
+        else:
+            WRITE_LOG_ERROR("ALERT: no inconsistent neighbor entries were flushed.")
 
     return 0 if res else 1
 
--- a/scripts/dualtor_neighbor_check.py
+++ b/scripts/dualtor_neighbor_check.py
@@ -265,6 +265,11 @@ def parse_args():
         default=None,
         help="stdout log level"
     )
+    parser.add_argument(
+        "--flush-inconsistent-neighbors",
+        action="store_true",
+        help="flush inconsistent kernel neighbor entries and rerun the check"
+    )
     args = parser.parse_args()
 
     if args.log_output == LogOutput.STDOUT:
@@ -351,6 +356,29 @@ def redis_cli(redis_cmd):
     if "error" in result or "ERR" in result:
         raise RuntimeError("Redis command '%s' failed: %s" % (redis_cmd, result))
     return result
+
+
+def flush_neighbor(neighbor_ip):
+    """Flush a kernel neighbor entry by IP address."""
+    ip = ipaddress.ip_address(neighbor_ip)
+    ip_version = "4" if ip.version == 4 else "6"
+    run_command("sudo ip -%s neigh flush to %s" % (ip_version, ip.compressed))
+
+
+def flush_inconsistent_neighbors(failed_neighbors):
+    """Flush the kernel neighbor entries for failed neighbors."""
+    flushed_neighbors = set()
+
+    for neighbor in failed_neighbors:
+        neighbor_ip = neighbor["NEIGHBOR"]
+        if neighbor_ip in flushed_neighbors:
+            continue
+
+        WRITE_LOG_WARN("Flushing inconsistent neighbor entry: %s", neighbor_ip)
+        flush_neighbor(neighbor_ip)
+        flushed_neighbors.add(neighbor_ip)
+
+    return len(flushed_neighbors)
 
 
 def read_tables_from_db(appl_db):
@@ -590,7 +618,7 @@ def check_neighbor_consistency(neighbors, mux_states, hw_mux_states, mac_to_port
     return check_results
 
 
-def parse_check_results(check_results):
+def parse_check_results(check_results, return_failed_neighbors=False):
     """Parse the check results to see if there are neighbors that are inconsistent with mux state."""
     failed_neighbors = []
     bool_to_yes_no = ("no", "yes")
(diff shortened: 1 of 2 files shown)


lolyu (changes requested):
### 📊 Overview
**Files Changed**: 2 (`scripts/dualtor_neighbor_check.py`, `tests/dualtor_neighbor_check_test.py`)
**Lines**: +228 / -19

Adds an opt-in `--flush-inconsistent-neighbors` flag to `dualtor_neighbor_check.py`. When inconsistencies are detected outside a mux toggle window, the script flushes the offending kernel neighbor entries and reruns the consistency check to determine the post-remediation exit code. Default (check-only) behavior is preserved.

### ✅ Strengths

- **Opt-in by default** — `action="store_true"` with no default means existing cron/systemd invocations are unaffected. Good defensive design.
- **`main()` extraction** — moving the `if __name__ == "__main__"` body into `main()` and calling `sys.exit(main())` is the right pattern; it makes the entry point unit-testable without subprocess overhead and is a clear improvement.
- **`run_neighbor_check` helper** — cleanly factors out the read+check loop so the rerun after flushing reuses the same code path without duplication.
- **Deduplication in `flush_inconsistent_neighbors`** — the `flushed_neighbors` set avoids redundant `ip neigh flush` calls when the same IP appears in multiple failed-neighbor rows.
- **Test coverage** — `flush_neighbor` (IPv4 + IPv6), `flush_inconsistent_neighbors`, `parse_check_results` with `return_failed_neighbors`, `run_neighbor_check`, `main` skip-non-dualtor, and the full flush+rerun flow are all covered. The `test_parse_args_flush_inconsistent_neighbors` test is also present. Solid.

### 📝 Review Findings

#### ⚠️ Major Issues

**1. Race condition: rerun immediately follows flush, before ARP/NDP relearning**

After `flush_inconsistent_neighbors` calls `ip neigh flush`, the kernel neighbor entries are gone. The rerun reads APPL_DB (which still holds the last-known neighbor state) and then checks kernel neighbor state — but at this point the kernel entries may still be in `INCOMPLETE`/`FAILED`/missing state because ARP/NDP relearning is traffic-driven and takes time (typically 100 ms – 1 s depending on ARP response). The rerun result is therefore non-deterministic: on a quiet link the second check will likely still fail, and the script will exit 1 even though the flush succeeded.

```python
flush_count = flush_inconsistent_neighbors(failed_neighbors)
if flush_count:
    WRITE_LOG_WARN("Flushed %d inconsistent neighbor entries. Rerunning dualtor neighbor check.", flush_count)
    check_results = run_neighbor_check(...)   # ← no delay here
    res = parse_check_results(check_results)
```

If the intent is "flush and let the normal periodic invocation declare success on the next run", the rerun could simply be skipped (always exit 0 after a successful flush). If immediate verification is needed, a brief sleep + retry loop with a timeout is warranted. As written, the rerun adds latency without reliable signal.

**2. Final scope is wider than the last constraint commit documents**

The commit history tells this story:
- Commit 2: *"Constrain --flush-inconsistent-neighbors remediation to failed prefix-route mux-port neighbors only. Host-route failures and non-port neighbors remain check-only."*
- Final commit: *"Apply the remediation to every failed neighbor…"* (un-constrains it back)

The current `flush_inconsistent_neighbors` flushes **all** entries in `failed_neighbors` regardless of `_NEIGHBOR_MODE` — including `host-route` neighbors, confirmed by `test_flush_inconsistent_neighbors_flushes_all_failed_neighbors` which asserts `192.168.0.4` (`host-route`) is flushed. Host-route neighbors are for directly connected hosts; flushing them is higher-impact than flushing prefix-route entries. The rationale for this reversal deserves a comment or PR description update. If the intent is truly to flush everything, that should be stated explicitly.

#### 📝 Minor Issues

**3. `parse_check_results` return type changes based on a parameter**

```python
def parse_check_results(check_results, return_failed_neighbors=False):
    ...
    if return_failed_neighbors:
        return False, failed_neighbors
    return False
    if return_failed_neighbors:
        return True, failed_neighbors
    return True
```

This is a valid Python pattern but fragile — a future caller that forgets `return_failed_neighbors=True` will get a `bool` and silently fail to unpack the tuple (or vice versa). A cleaner API would always return `(bool, list)` and have callers ignore the second element when not needed, e.g.:

```python
def parse_check_results(check_results):
    ...
    return True, []   # or False, failed_neighbors
```

The second call in `main()` (`res = parse_check_results(check_results)`) would become `res, _ = parse_check_results(check_results)`, which is explicit and consistent.

**4. The "empty table" output is a behavior change, not a preservation**

The commit message says *"Keep the no-neighbor output compatible with deployed dualtor_neighbor_check.py by printing the empty prefix-route table header even when no rows are returned."* However, the original code path (no `prefix_route_results`, no `host_route_results`) simply returned `True` without printing anything — the empty-table output is new, not existing behavior. This is a harmless addition, but the commit message framing is misleading. A downstream log parser that keyed on the header appearing *only* when neighbors existed would be broken by this. Worth a clarifying note.

#### 💡 Suggestions

- Consider logging the post-flush result explicitly: if the rerun still fails, a `WRITE_LOG_WARN("Post-flush check still inconsistent for %d neighbors", ...)` helps operators distinguish "flush happened but convergence is slow" from "flush did not help at all".
- `NOT_AVAILABLE` as a `PORT` value in a `failed_neighbor` is passed to `flush_neighbor` — the IP address itself is still valid so the flush works, but it might be worth logging a note that the port association was unknown at flush time.

### 🧪 Testing

Coverage of the new helpers and the `main()` flow is good. One gap: there is no test for the case where `flush_count == 0` (i.e., `failed_neighbors` is empty or all IPs are duplicates) — `main()` should skip the rerun in that case, but this branch (`if flush_count:` is falsy) is not tested. Low risk given the deduplication logic is already covered, but worth adding.

### Status
⚠️ Changes requested — the race condition between flush and immediate rerun (issue 1) and the undocumented scope reversal on host-route neighbors (issue 2) should be addressed before merge.

lolyu (dismissed):
LGTM

prsunny:
Can the mitigation be default? like try once and if its repeated, then create log for alert?

yxieca:
Review note:

**Suggest addressing before merge:**

- `flush_inconsistent_neighbors` calls `flush_neighbor` → `run_command`, which raises `RuntimeError` on any non-zero exit from `sudo ip neigh flush`. A transient per-entry failure (entry already gone, interface flap race, IPv6 link-local needing a zone id, AppArmor edge cases) will skip all remaining flushes and propagate as an unhandled traceback out of `main()` instead of producing a clean exit code. Since this flag exists specifically for already-degraded states, those transient failures are most likely exactly when it's invoked. Suggest wrapping the per-neighbor flush in `try/except RuntimeError`, logging, and continuing so the rerun can still surface remaining inconsistencies and the script preserves its intended exit code.
- PR description vs. code: the description says "retries the consistency check up to 3 times (1s delay between retries)", but the code does a single rerun after `POST_FLUSH_CHECK_DELAY_SEC = 10` seconds. The single longer settle window looks like the better design for kernel ARP/ND + orchagent reconvergence — suggest updating the description (and adding a brief comment on the constant) to match the implementation rather than changing the code.

**Minor nits:**

- No test for `flush_neighbor` when `run_command` fails or when the IP string is malformed (`ipaddress.ip_address` raises `ValueError`); a small case would lock in the intended behavior after the try/except above.

Refactor itself (`main()` / `run_neighbor_check()` extraction, `parse_check_results` returning `(bool, list)`, test coverage of v4/v6 flush, dedup, skip-non-dualtor, success and still-inconsistent paths) looks clean.

Reviewed by AI agent on behalf of Ying.

prsunny (approved):

PR #4562, opened 2026-05-21
[route_check] exclude BMC interface from ROUTE_TABLE/ASIC_DB consistency check

gordon-nexthop:
#### What I did

- `routeCheck` was generating false-positive `missed_ROUTE_TABLE_routes` alarms for `169.254.100.0/30` (BMC subnet) on platforms with BMCs (bmc.json)
- The BMC interface connected route exists in `ROUTE_TABLE` (APPL_DB) but is never programmed into `ASIC_DB` — it is a management/OOB route, not a dataplane route
- This caused `monit routeCheck` to fire continuously after 8 retries, which loganalyzer caught as an ERR during teardown, failing unrelated tests across many packages

#### How I did it

`filter_out_local_interfaces()` excluded `eth0`, `eth1`, `docker0` but not the BMC interface (`bmc0`). The fix reads
`DEVICE_METADATA.bmc.bmc_if_name` from CONFIG_DB and adds it to the local interface exclusion set.

#### How to verify it

- Added test case `"35"` to `route_check_test_data.py`: ROUTE_TABLE has `169.254.100.0/30` with `ifname=bmc0`, CONFIG_DB has `DEVICE_METADATA.bmc.bmc_if_name=bmc0`, ASIC_DB has no entry for it — expects pass (RET=0)
- Tests which were failing due to loganaylzer errors concerning route inconsistencies fixed

### DIFF
--- a/scripts/route_check.py
+++ b/scripts/route_check.py
@@ -579,6 +579,13 @@ def filter_out_local_interfaces(namespace, keys):
     chassis_local_intfs = chassis.get_chassis_local_interfaces()
     local_if_lst.update(set(chassis_local_intfs))
 
+    # Exclude BMC interface (e.g. usb0) — connected route for BMC subnet is a management
+    # route that will never be programmed into the ASIC.
+    cfg_db = multi_asic.connect_config_db_for_ns(namespace)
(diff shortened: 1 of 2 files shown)


gordon-nexthop:
How BMC interface information is acquired was recently changed, will post a different PR to fix this issue.

PR #4566, opened 2026-05-25
[config] Remove redundant "Error: " prefix from ctx.fail messages

ganglyu:
#### What I did

Fixed duplicate `"Error:"` prefix in CLI error messages. `click`'s `ctx.fail()` automatically prepends `"Error: "` when printing, but the affected code also embedded `"Error: "` in the message string itself, resulting in output like:

```
Error: Error: Port mirror direction 'tx' is not supported by the ASIC
```

#### How I did it

Removed the redundant `"Error: "` string prefix from all `ctx.fail()` calls in `config/main.py`. Affected commands:
- `config mirror_session span add` / `erspan add` — validation of src/dst port, direction, ASIC capability
- `config synchronous-mode`
- `config yang-config-validation`
- `config portchannel del`
- `config interface sflow sample-direction` / `config sflow sample-direction`

No logic changes; error messages remain identical except the leading duplicate `"Error: "` is removed.

#### How to verify it

Run a command known to trigger a validation error, e.g.:
```
config mirror_session span add SPAN1 Ethernet496 Ethernet16 tx
```

Before: `Error: Error: Port mirror direction 'tx' is not supported by the ASIC`  
After:  `Error: Port mirror direction 'tx' is not supported by the ASIC`

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -2909,7 +2909,8 @@ def remove_portchannel(ctx, portchannel_name):
                 ctx.fail("{} has vlan {} configured, remove vlan membership to proceed".format(portchannel_name, str(k)))
 
         if len([(k, v) for k, v in db.get_table('PORTCHANNEL_MEMBER') if k == portchannel_name]) != 0: # TODO: MISSING CONSTRAINT IN YANG MODEL
-            ctx.fail("Portchannel {} contains members. Remove members before deleting Portchannel!".format(portchannel_name))
+            ctx.fail("Portchannel {} contains members. Remove members before deleting Portchannel!"
+                     .format(portchannel_name))
 
         # Dont proceed if the port channel is used in dhcpv4_relay
         try:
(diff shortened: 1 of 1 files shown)


stephenxs (approved):

keboliu (approved):

liat-grozovik:
@ganglyu i assume that there are not tests or log analyzer which depends on the exact string and will now fail. If that is not the case i will be happy to approve and merge it.

BTW, can you check (not part of this PR) if there are additional ctx.fail() usage with Error:

ganglyu:
> @ganglyu i assume that there are not tests or log analyzer which depends on the exact string and will now fail. If that is not the case i will be happy to approve and merge it.
> 
> BTW, can you check (not part of this PR) if there are additional ctx.fail() usage with Error:

I checked all tests and the log analyzer, none of them depend on the exact "Error: Error: ..." strings being changed.
After this PR there are zero remaining ctx.fail("Error: ...") occurrences across the whole sonic-utilities tree.

PR #4509, opened 2026-04-30
[202511] Use ijson streaming in route_check.py for large route tables

deepak-singhal0408:
## Description
Rewrite route_check.py's fetch_routes() to use ijson streaming JSON parsing instead of loading the entire FRR routing table into memory. This prevents OOM on devices with 500K+ routes.

This is the equivalent/backport of master PRs:
- sonic-net/sonic-utilities#4205 (Initial chunked route_check)
- sonic-net/sonic-utilities#4294 (Fix chunked route check)
- sonic-net/sonic-utilities#4256 (Parallel v4/v6 fetch)
- sonic-net/sonic-utilities#4436 (Fix failing route check)

## Motivation
On devices with large route tables (500K+ prefixes), the current route_check.py loads the entire `show ip route json` output (~1.2GB for 500K routes) into memory at once, causing OOM kills. The ijson streaming approach processes each prefix entry as it arrives from the pipe, keeping memory usage constant (~50MB regardless of table size).

## Changes
- `scripts/route_check.py`:
  - Added `import ijson`
  - Rewrote `fetch_routes()` to use `ijson.kvitems()` for streaming parsing
  - Changed return type to `(missing_routes, failing_routes)` tuple
  - Updated `get_frr_routes_parallel()` to use new return format
  - Simplified `check_frr_pending_routes()` (filtering now in fetch_routes)
- `setup.py`: Added `ijson>=3.2.3` dependency
- `tests/fetch_routes_chunk_test.py`: New unit test file (6 tests)

## Companion PR
- sonic-net/sonic-buildimage#27100: Install python3-ijson package

## Verification
- Unit tests: 6/6 passing
- VS KVM testbed: `sudo python3 route_check.py` runs cleanly
- Verified ijson.kvitems() correctly parses 34 route prefixes via streaming
- route_check.py -m DEBUG shows: FRR Missing Routes: [], FRR Failed Routes: []

PR #4119, opened 2025-11-05
Add FRR failed route check in route_check.py

ZhaohuiS:
#### What I did

It supports to detect the route with offload False or without offload, which can capture queued route, because queued route doesn’t have offload, but for rejected route, the offload is True. It can’t detect the rejected route.
So, we add a new detection check for the value of key failed for route entries, which can cover both rejected and queued routes.
It will help to detect rejected route and queued route on device.

#### How I did it
Append failed route prefix into failed list, if it's not empty, script will print error message into syslog
```
        # Check for failed state
        if entry.get('failed', False):
            failed_rt.append(route_prefix)
```

#### How to verify it

- **For rejected route:**

The output of route_check.py
```
Some routes have failed state in FRR : ['0.0.0.0/0', '192.168.128.0/25', '192.168.128.128/25', '192.168.136.0/25', '192.168.136.128/25', '192.168.144.0/25', '192.168.144.128/25', '192.168.152.0/25', '192.168.152.128/25', '192.168.160.0/25', '192.168.160.128/25', '192.168.168.0/25', '192.168.168.128/25', '192.168.176.0/25', '192.168.176.128/25', '192.168.184.0/25', '192.168.184.128/25', '192.168.192.0/25', '192.168.192.128/25', '192.168.200.0/25', '192.168.200.128/25', '192.168.208.0/25', '192.168.208.128/25', '192.168.216.0/25', '192.168.216.128/25', '192.168.224.0/25', '192.168.224.128/25', '192.168.232.0/25', '192.168.232.128/25', '192.168.240.0/25', '192.168.240.128/25', '192.168.248.0/25', '192.168.248.128/25', '192.169.0.0/25', '192.169.0.128/25', '192.169.104.0/25', '192.169.104.128/25', '192.169.112.0/25', '192.169.112.128/25', '192.169.120.0/25', '192.169.120.128/25', '192.169.128.0/25', '192.169.128.128/25', '192.169.136.0/25', '192.169.136.128/25', '192.169.144.0/25', '192.16
Failure results: {{
    "": {
        "failed_FRR_routes": [
            "0.0.0.0/0",
            "192.168.128.0/25",
            "192.168.128.128/25",
            "192.168.136.0/25",
            "192.168.136.128/25",
            "192.168.144.0/25",
            "192.168.144.128/25",
            "192.168.152.0/25",
            "192.168.152.128/25",
            "192.168.160.0/25",
            "192.168.160.128/25",
            "192.168.168.0/25",
[...]

PR #4576, opened 2026-05-29
Fix route_check.py: performance, dict bug, and intersection retry

mike-dubrovsky:
#### What I did

Four fixes to `route_check.py`:

1. **Bug fix** (`mitigate_installed_not_offloaded_frr_routes` crashes on plain
   strings): `fetch_routes()` appends bare prefix strings to `missing_routes`,
   but `mitigate_installed_not_offloaded_frr_routes()` was accessing
   `entry['prefix']` and `entry['protocol']` expecting dicts, causing a
   `TypeError: string indices must be integers` whenever suppress-fib-pending
   triggered mitigation.

2. **Performance** (replace ijson JSON scan with vtysh text streaming):
   Replace the ijson-based parser with a streaming line-by-line scan of
   `vtysh show ip/ipv6 route` plain-text output and a focused regex that
   matches only queued (`q`) or offload-failed (`o`) routes.

3. **Correctness** (intersection-based retry in `check_frr_pending_routes`):
   Replace the overwrite-on-each-iteration retry loop with an intersection
   accumulator so that only routes stuck in *every* poll window are mitigated,
   avoiding premature mitigation of routes that are still converging.

4. **Correctness** (snapshot ASIC/APPL DB after FRR polling completes):
   `check_frr_pending_routes()` takes up to ~30 s (2 × `FRR_WAIT_TIME` sleeps).
   Previously the ASIC/APPL DB snapshot was taken *before* FRR polling, so
   `rt_appl_miss` and `rt_asic_miss` could be up to 30 s stale when the
   mitigation cross-check fires (`rt_frr_miss and not rt_appl_miss and not
   rt_asic_miss`). Moving the snapshot to after FRR polling ensures both
   observations describe the same point in time, preventing false-negative
   mitigations when routes finish programming during the FRR retry window.

#### How I did it

**Bug fix**: `mitigate_installed_not_offloaded_frr_routes()` was fixed to treat
`missed_frr_rt` as a list of plain prefix strings and hardcode `'bgp'` as the
protocol, matching what `fetch_routes()` has always returned.  Changing the
mitigation caller is correct because `fetch_routes()` only ever collects BGP
routes (the regex anchors on the leading `B`), so carrying the protocol in each
list entry is unnecessary.

**Performance**: `vtysh show ip route json` for a large route table serialises
the entire in-memory RIB to JSON before sending a single byte, producing
~200 MB that must then be parsed by Python. The plain-text output is streamed
incrementally by vtysh and scanned line-by-line with a single regex — no JSON
heap allocation. On a healthy device (no stuck routes) the function returns
empty lists without any heap allocations at all.

FRR `zebra_vty.c` assembles each route line as three fixed-position characters
(`zebra_route_char` + `selected_char` + `re_status_output_char`):

```
B>* 10.0.0.0/24   selected, offloaded (healthy)
B>q 10.0.0.0/24   selected, queued    (suppress-fib-pending)
B>o 10.0.0.0/24   selected, offload failed
```

The regex `r'^B>([qo])\s+(\S+)\s'` matches only the two cases that require
action.

**Correctness (intersection retry)**: the previous loop overwrote `missed_rt` on every iteration,
so a route that appeared stuck only in the last iteration could be mitigated
even though it was still converging. The new logic seeds an accumulator on the
first iteration and intersects it with each subsequent result, returning only
prefixes that were stuck in every poll.

**Correctness (snapshot ordering)**: `check_frr_pending_routes()` sleeps up to
`(FRR_CHECK_RETRIES - 1) × FRR_WAIT_TIME` = 30 s between FRR polls. In the
previous order, ASIC/APPL DB snapshots were taken before FRR polling began, so
by the time the cross-check

```python
if rt_frr_miss and not rt_appl_miss and not rt_asic_miss:
```

ran, `rt_appl_miss` and `rt_asic_miss` could describe system state from up to
30 s earlier. A route being programmed at snapshot time (showing as a miss)
but finishing before FRR polling ended would produce a stale non-empty
`rt_appl_miss`, silently suppressing mitigation even though hardware was fully
in sync. Moving the ASIC/APPL snapshot to after `check_frr_pending_routes()`
returns ensures both observations are temporally aligned.

#### Why this matters — timeout risk on T2 chassis

See bug https://github.com/sonic-net/sonic-buildimage/issues/27612

In the latest sonic-mgmt, suppress-fib-pending is enabled on T2 chassis,
and `check_frr_pending_routes` runs unconditionally on every
`route_check.py` invocation.

`route_check.py` runs under two hard time limits:
- **SIGALRM at 2 minutes** (TIMEOUT_SECONDS timer set inside the script itself)
- **monit kill at 5 minutes** (`/etc/monit/conf.d/sonic-host`)

On a T2 chassis a line card can have **3 ASICs**, each running an independent
FRR instance. `sonic-mgmt` tests use ~51k IPv4 + ~51k IPv6 routes per ASIC.
Route download speed is ~1k prefixes/sec, so BGP converging during a test run are
common.

With the old JSON-based implementation, if BGP was flapping at the time
`route_check.py` ran, the script would enter the retry loop (up to 3
iterations × 15 s sleep + ~70 s JSON parse + vtysh fetch per ASIC), easily
exceeding the 2-minute and 5-minute thresholds and getting killed before it
could report or mitigate anything.

#### Measured results (88-LC0-36FH, 3 ASICs, 51k IPv4 + 51k IPv6 per ASIC)

| Scenario                     | Old wall time | New wall time |
|------------------------------|---------------|---------------|
| Healthy device (0 stuck)     | ~120 s        | ~31 s         |
| All routes stuck, 3 retries  | ~350–400 s    | ~85 s         |

#### Unit tests

`tests/fetch_routes_chunk_test.py` (28 tests, ijson/JSON-based) was deleted.
`tests/fetch_routes_test.py` (19 tests) replaces it:

- **`TestFetchRoutes`** (14 tests) — mocks `subprocess.Popen` to return an
  iterator of plain vtysh text lines and verifies the regex
  `r'^B>([qo])\s+(\S+)\s'`: flag chars `q`/`o`/`*`, non-BGP lines, non-selected
  routes, mixed output, 100-route scale, empty output, header/legend lines,
  IPv4/IPv6 command construction (no `json` suffix), IPv6 prefix extraction,
  non-zero exit code, and `FileNotFoundError`.

- **`TestCheckFrrPendingRoutes`** (5 tests) — patches `get_frr_routes_parallel()`
  directly and verifies the intersection accumulator: fast-exit when first poll is
  empty, all routes persistently stuck, converging route removed mid-intersection,
  early exit on mid-retry clear, and symmetry for `failing_routes`.

#### How to verify it

Deploy `route_check.py` to a T2 line card with suppress-fib-pending enabled
and a large BGP route table. Confirm that:

1. The script completes within the 2-minute window on a healthy device.
2. When routes are stuck (queued), the script detects them, waits through
   retries to confirm they are persistently stuck, and sends mitigation
   notifications — after which `vtysh show ip route` no longer shows `B>q`
   lines for those prefixes.
3. Routes that converge between retry iterations are not mitigated.
4. Routes that finish programming in ASIC DB *during* the FRR retry window
   are not falsely treated as APPL/ASIC mismatches — mitigation fires
   correctly when FRR is stuck but hardware is in sync at the time FRR
   polling completes.

N/A — no CLI changes.

N/A — no CLI changes.


mike-dubrovsky:
@venkit-nexthop @deepak-singhal0408 

mitigate_installed_not_offloaded_frr_routes() was introduced in 311add29 ("Add FRR failed route check", #4119) expecting missed_frr_rt to be a list of dicts — it accesses entry['prefix'] and entry['protocol'] to build the fpmsyncd response-channel notification. At that point the caller appended full FRR route dicts, so the contract held.

master a9543cba ("Fix route_check.py to not hog a lot of memory", #4205, author: [venkit@nexthop.ai](mailto:venkit@nexthop.ai)) refactored fetch_routes() to use ijson streaming and changed missed_rt.append(entry) (full dict) to missing_routes.append(prefix) (bare string). This silently broke the mitigate contract: whenever suppress-fib-pending fires and mitigation is attempted, entry['prefix'] on a plain string raises TypeError: string indices must be integers.

The 202511 backport (9d00815c) caught this and preserved the dict format ({'prefix': prefix, 'protocol': route_entry.get('protocol', '')}).


This PR follows the direction already taken in master: a9543cba implicitly scoped mitigation to BGP by filtering out connected/kernel/static and by dropping the protocol field from each list entry. This PR makes that scope explicit — the regex ^B> anchors on the BGP route character from the FRR CLI output format, so only BGP routes ever reach the mitigation path, and the mitigate caller can safely hardcode 'bgp' as the protocol and work with bare prefix strings. No new assumption is introduced; the implicit assumption already in master is just made visible.

The two approaches are from different angles, each correct in its own framing:

The 202511 approach ({'prefix': ..., 'protocol': ...}) is correct from the SONiC side:  SONiC suppress-fib-pending config knob is already protocol-agnostic.

The master approach is correct from the FRR CLI side:  "bgp suppress-fib-pending" hold-down and the resulting q/o status characters are a BGP-only feature.

venkit-nexthop:
> @venkit-nexthop @deepak-singhal0408
> 
> mitigate_installed_not_offloaded_frr_routes() was introduced in [311add2](https://github.com/sonic-net/sonic-utilities/commit/311add29b22cf96f5eca4bcdbfe2d4aff8155554) ("Add FRR failed route check", #4119) expecting missed_frr_rt to be a list of dicts — it accesses entry['prefix'] and entry['protocol'] to build the fpmsyncd response-channel notification. At that point the caller appended full FRR route dicts, so the contract held.
> 
> master [a9543cb](https://github.com/sonic-net/sonic-utilities/commit/a9543cbad9510b5a0301c12bd43cd5171911f79c) ("Fix route_check.py to not hog a lot of memory", #4205, author: [venkit@nexthop.ai](mailto:venkit@nexthop.ai)) refactored fetch_routes() to use ijson streaming and changed missed_rt.append(entry) (full dict) to missing_routes.append(prefix) (bare string). This silently broke the mitigate contract: whenever suppress-fib-pending fires and mitigation is attempted, entry['prefix'] on a plain string raises TypeError: string indices must be integers.
> 
> The 202511 backport ([9d00815](https://github.com/sonic-net/sonic-utilities/commit/9d00815cc431747d6b9f766f4547f082839be0f2)) caught this and preserved the dict format ({'prefix': prefix, 'protocol': route_entry.get('protocol', '')}).
> 
> This PR follows the direction already taken in master: [a9543cb](https://github.com/sonic-net/sonic-utilities/commit/a9543cbad9510b5a0301c12bd43cd5171911f79c) implicitly scoped mitigation to BGP by filtering out connected/kernel/static and by dropping the protocol field from each list entry. This PR makes that scope explicit — the regex ^B> anchors on the BGP route character from the FRR CLI output format, so only BGP routes ever reach the mitigation path, and the mitigate caller can safely hardcode 'bgp' as the protocol and work with bare prefix strings. No new assumption is introduced; the implicit assumption already in master is just made visible.
> 
> The two approaches are from different angles, each correct in its own framing:
> 
> The 202511 approach ({'prefix': ..., 'protocol': ...}) is correct from the SONiC side:  SONiC suppress-fib-pending config knob is already protocol-agnostic.
> 
> The master approach is correct from the FRR CLI side:  "bgp suppress-fib-pending" hold-down and the resulting q/o status characters are a BGP-only feature.

Hi Mike,
I dont understand why we need to go back to screen scraping to fix this issue. 
If you want to return a list of dict we could easily do that with the existing ijson mechanism. To me even returning a list of (prefix, protocol) tuple and handling it in the caller is also a fine option. The `mitigate_installed_not_offloaded_frr_routes()` could be called with this list of tuples.

deepak-singhal0408:
@mike-dubrovsky — Thanks for the thorough write-up. A few thoughts:

**The PR bundles 4 independent changes** — I'd suggest splitting into 3 buckets:

**Bucket A — Bug fix (land immediately):** Fix the dict crash in `mitigate_installed_not_offloaded_frr_routes`. The 202511 fix (commit `9d00815c`) preserves the dict format — adopting that approach on master would keep things simple and unblock quickly.

**Bucket B — Correctness (land after A):** Intersection-based retry + snapshot reorder. These are solid improvements — moving the ASIC/APPL snapshot after FRR polling is clearly correct, and the intersection logic prevents premature mitigation of converging routes.

**Bucket C — Performance (needs discussion):** Replacing ijson with text-streaming regex. @venkit-nexthop raised a valid question that hasn't been addressed yet. One additional data point:

Issue [#27612](https://github.com/sonic-net/sonic-buildimage/issues/27612) — the failure log is from Apr 8, but PR [#4509](https://github.com/sonic-net/sonic-utilities/pull/4509) (ijson yajl2_c C backend) landed on 202511 in early May. Prior measurements on similar scale (51K routes, 2 ASICs) showed yajl2_c brought route_check from ~21 min → ~1m12s. Has the timeout been re-validated on chassis-packet with the yajl2_c backend in place? If the C backend already resolves the timeout, the text-streaming rewrite becomes an optional optimization rather than a blocking fix.

Would be great to see Buckets A and B unblocked from the ijson-vs-text debate.

mike-dubrovsky:
Hi @venkit-nexthop and @deepak-singhal0408

The reason I moved away from JSON parsing is performance. vtysh show ip route json serializes the entire in-memory RIB to JSON before emitting a single byte , and Python then has to parse all of it. ijson streams the parse, but it can't stream the production — vtysh has already built the whole JSON blob first. The plain-text output, on the other hand, is emitted incrementally and scanned line-by-line with a single regex, so parsing finishes as soon as the last line arrives from vtysh — and it is significantly faster overall.

Scale: sonic-mgmt tests for 202511 on a T2 LC with 3 ASICs (each an independent FRR instance) use ~51k IPv4 and ~51k IPv6 routes, and check_frr_pending_routes runs across all of them on every invocation.

Single-pass measurements (3 ASICs in parallel):

**JSON approach (ijson):**

Total wall time: ~89 s
A jq test showed vtysh_jq ≈ 38–39 s per ASIC just to generate and stream the JSON — so roughly ~40 s for vtysh to produce it and ~47–49 s for ijson to parse it (per ASIC, limited by the GIL when run in parallel).

**Text approach (streaming):**

Total wall time: ~26–29 s
Per-ASIC: 5–9 s combined for vtysh to generate the text and Python to scan it (the regex scan itself is only ~0.6–1.2 s).
The retry loop then runs up to 3 times (15 s sleep between iterations) to confirm a route is persistently stuck before mitigating — which multiplies the per-pass cost.

This matters because route_check.py kills itself via SIGALRM at 2 min and is killed by monit at 5 min. With BGP still converging, the old single pass alone consumed most of the 2-min budget, and 3 retries blew well past 5 min, causing the script to be killed before it could mitigate anything.

saravanan-nexthop:
cc @selva-nexthop for viz

mike-dubrovsky:
@deepak-singhal0408
split  the change into 3 separate PRs as per review comment.
[#4649](https://github.com/sonic-net/sonic-utilities/pull/4649)
[#4650](https://github.com/sonic-net/sonic-utilities/pull/4650)
[#4651](https://github.com/sonic-net/sonic-utilities/pull/4651)

deepak-singhal0408:
Closing this as Separate PRs has been posted.

PR #4581, opened 2026-06-01
nokia armhf config reload with swss sync restart + time.monotonic

yanmarkman:
#### What I did
1.  Nokia-armhf config-reload with swss and sync restart
    Fix for issue reported by the: sonic-net/sonic-buildimage#24766
2. On chrony or "date -s SET" command the wall-time is changed
   and abrupts the wall/realtime taken by the time.time().

#### How I did it
1. Back-port from 202511 and master #4174
    Guaranty "config reload" by a swss.service restart "inside" the command implementation.
    The restart is done ONLY for the device Nokia-7215 and does not impact other devices and platforms
2. Use the time.monotonic() which is not changed un date/wall-time.

#### How to verify it
1. Loop with "config reload -y"
2. 2 loops --  for("config reload -y") + for('date -s "+15 min"' ; 'date -s -10 min"')

Check "docker ps -a"
swss and syncd (but could be other) containers/services are exited (failed)

Check "docker ps -a"
All containers are running

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -857,12 +857,12 @@ def storm_control_delete_entry(port_name, storm_type):
 def _wait_until_clear(tables, interval=0.5, timeout=30, verbose=False):
     if timeout == 0:
         return True
-    start = time.time()
+    start = time.monotonic()
     empty = False
     app_db = SonicV2Connector(host='127.0.0.1')
     app_db.connect(app_db.APPL_DB)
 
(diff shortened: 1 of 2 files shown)


Pavan-Nokia:
@yanmarkman  We do not seem to see this issue on the 2505 branch so this Fix is not needed for this branch .

Can you please close this PR

PR #4586, opened 2026-06-02
Add multi-asic support for Route flow counter cli command

saksarav-nokia:
#### What I did
Fix for https://github.com/sonic-net/sonic-buildimage/issues/27680 to address the wrong error message.  The test case should be updated since the  Route flow counter feature is not supported by BCM DNX platforms.
Added multi-asic support for Route flow counter cli command 
#### How I did it
Modified the Route flow counter cli code to check the capability in the respective namespace instead of checking in global name space's STATE_DB
#### How to verify it
Verified counterpoll flowcnt_route [-n namespace]  enable/disable commands in asic namespaces and also global name spaces.
Waiting for swss to initialize route flow counter capability, please try again later
The following  message is printed if the Route flow counter capability is not supported. It was existing message
Route flow counter is not supported on this platform

### DIFF
--- a/counterpoll/main.py
+++ b/counterpoll/main.py
@@ -491,7 +491,7 @@ def flowcnt_trap_disable(ctx):
 def flowcnt_route(ctx, namespace):
     """ Route flow counter commands """
     ctx.obj = connect_to_db(namespace)
-    exit_if_route_flow_counter_not_support()
+    exit_if_route_flow_counter_not_support(namespace)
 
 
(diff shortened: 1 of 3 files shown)


deepak-singhal0408:
Hi @abdosi , @arlakshm could you help review the PR. thanks,

saksarav-nokia:
Since https://github.com/sonic-net/sonic-utilities/pull/4604 addresses the same, closing this PR

PR #4587, opened 2026-06-03
[route_check]: Add truncation indicator and mismatch summary counts

### LINKED ISSUE #27509: route_check.py truncates large failure reports without showing mismatch counts
## Description

`route_check.py` can print an incomplete failure report when the mismatch list is large. The script builds a complete `results` object internally, but the CLI/syslog output path truncates the formatted JSON at `PRINT_MSG_LEN_MAX` characters. As a result, the printed `Failure results` block can be cut off in the middle of a route string and does not show how many entries were omitted.

This makes the output hard to diagnose: a user only sees the first portion of the mismatch list and has no indication of the actual number of missing/unaccounted routes.

Relevant code on `master`:

- `PRINT_MSG_LEN_MAX = 1000`: https://github.com/sonic-net/sonic-utilities/blob/master/scripts/route_check.py#L78
- `print_message()` stops appending once the accumulated message reaches that limit: https://github.com/sonic-net/sonic-utilities/blob/master/scripts/route_check.py#L126-L145
- `check_routes_for_namespace()` stores the complete missing route list in `results["missed_ROUTE_TABLE_routes"]`: https://github.com/sonic-net/sonic-utilities/blob/master/scripts/route_check.py#L939-L940
- `check_routes()` formats and prints the whole `results` object through `print_message()`: https://github.com/sonic-net/sonic-utilities/blob/master/scripts/route_check.py#L1002-L1007

## Example symptom

For a large mismatch set, the output can look like this:

```text
Failure results: {{
    "": {
        "missed_ROUTE_TABLE_routes": [
            "2001:db8:1001::/64",
            "2001:db8:1002::/64",
            "2001:db8:1003::/64",
            ...
            "2001:db8:10
Failed. Look at reported mismatches above
add: {
    "": []
}
del: {
    "": []
}
```

The output is not valid JSON, does not state that it was truncated, and does not provide a count such as `missed_ROUTE_TABLE_routes: <N>`.

## Expected behavior

The checker should make the size of each mismatch category visible even when the detailed route list is too long. For example:

1. Print a summary count per namespace/category before the detailed list.
2. If detailed output must remain capped, explicitly mark it as truncated.
3. Prefer printing the first N entries plus `... <remaining> more` instead of cutting the JSON mid-string.

## Impact

When route_check reports many mismatches, operators can underestimate the scope of the failure because only the first part of the list is visible. The complete data already exists in the returned `results` object; the issue is in the reporting path.

Bojun-Feng:
#### What I did

Fix https://github.com/sonic-net/sonic-buildimage/issues/27509

Added a truncation indicator and mismatch summary counts to `route_check.py` failure output.

When `print_message()` truncates output at `PRINT_MSG_LEN_MAX` (1000 chars), it now appends `... (truncated)` within the length budget so operators know data was omitted. Additionally, a compact per-namespace mismatch count summary line is emitted before the detailed (potentially truncated) JSON dump.

#### How I did it
* Add `PRINT_MSG_TRUNCATION_SUFFIX` module-level constant and truncation marker logic in `print_message()` that keeps total length ≤ `PRINT_MSG_LEN_MAX`
* Extract `summarize_results()` helper to produce `{namespace: {category: count}}` dict
* Emit compact summary via `json.dumps(..., separators=(",",":"))` before the detailed failure results in both `check_routes()` and `check_sids()`

#### How to verify it
1. Run existing `test_logging` — confirms backward compatibility (all length assertions pass unchanged)
2. Run new `test_logging_truncation_indicator` — confirms truncation suffix behavior at realistic `PRINT_MSG_LEN_MAX=50`, short message passthrough, and edge case where max < suffix length
3. Manually: inject 200+ route mismatches on a VS image, run `route_check.py -l INFO`, verify summary count line appears and truncated output ends with `... (truncated)`

```
Failure results: {
    "": {
        "missed_ROUTE_TABLE_routes": [
            "2001:db8:1001::/64",
            "2001:db8:1002::/64",
            "2001:db8:10
Failed. Look at reported mismatches above
```

```
Route mismatch counts: {"":{"missed_ROUTE_TABLE_routes":47,"missed_INTF_TABLE_entries":2}}
Failure results: {
    "": {
        "missed_ROUTE_TABLE_routes": [
            "2001:db8:1001::/64",
            "2001:db8:1002::/64",
            "2001:db8:10 ... (truncated)
Failed. Look at reported mismatches above
```

### DIFF
--- a/scripts/route_check.py
+++ b/scripts/route_check.py
@@ -1019,7 +1019,8 @@ def check_routes(namespace):
                 return -1, results
 
     if results:
-        print_message(syslog.LOG_WARNING, "Route mismatch counts: ", json.dumps(summarize_results(results), separators=(",", ":")))
+        print_message(syslog.LOG_WARNING, "Route mismatch counts: ",
+                     json.dumps(summarize_results(results), separators=(",", ":")))
         print_message(syslog.LOG_WARNING, "Failure results: {",  json.dumps(results, indent=4), "}")
         print_message(syslog.LOG_WARNING, "Failed. Look at reported mismatches above")
         print_message(syslog.LOG_WARNING, "add: ", json.dumps(all_adds, indent=4))
@@ -1090,7 +1091,8 @@ def check_sids(namespace):
                 return -1, results
 
     if results:
-        print_message(syslog.LOG_WARNING, "SID mismatch counts: ", json.dumps(summarize_results(results), separators=(",", ":")))
+        print_message(syslog.LOG_WARNING, "SID mismatch counts: ",
+                     json.dumps(summarize_results(results), separators=(",", ":")))
         print_message(syslog.LOG_WARNING, "SIDs Check Failure results: {",  json.dumps(results, indent=4), "}")
         return -1, results
     else:
--- a/scripts/route_check.py
+++ b/scripts/route_check.py
@@ -1020,7 +1020,7 @@ def check_routes(namespace):
 
     if results:
         print_message(syslog.LOG_WARNING, "Route mismatch counts: ",
-                     json.dumps(summarize_results(results), separators=(",", ":")))
+                      json.dumps(summarize_results(results), separators=(",", ":")))
         print_message(syslog.LOG_WARNING, "Failure results: {",  json.dumps(results, indent=4), "}")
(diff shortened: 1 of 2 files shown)


deepak-singhal0408 (approved):
LGTM. thanks,

PR #4591, opened 2026-06-04
[BMC] Skip swss check on reload for platforms where swss is disabled

roger-nexthop:
#### What I did
config reload fails on BMC due to check for swss container, which does not run.

#### How I did it
Add a check for BMC platforms and skip the swss check if so.

#### How to verify it
Verify config reload goes through on BMC.


judyjoseph:
@oleksandrivantsiv @YairRaviv f.y.i

judyjoseph (approved):

roger-nexthop:
This seems to be addressed by https://github.com/sonic-net/sonic-utilities/pull/4606 that was merged.

PR #4185, opened 2026-01-11
[SED] Change/Reset SED password CLI

benle7:
HLD: https://github.com/sonic-net/SONiC/pull/2171
This PR is dependent on: https://github.com/sonic-net/sonic-platform-common/pull/620

#### What I did
I added 2 CLIs to change the SED password and reset the SED password.

#### How I did it
I added 'config sed' group with the 2 relevant methods, using the platform chassis.

#### How to verify it
config sed change-password
config sed reset-password

PR #4583, opened 2026-06-02
Align SED tests to the new conftest code

benle7:
#### What I did
After https://github.com/sonic-net/sonic-utilities/pull/4516 landed in master
tests/conftest.py clears sonic_platform entries from sys.modules between test
files so pytest-xdist workers do not leak mocks across files. tests/sed_test.py used
@patch('sonic_platform.platform.Platform') without re-injecting a fake
sonic_platform module, so CI failed with ModuleNotFoundError: No module named 'sonic_platform' once the conftest reset ran before those tests. The same pattern was
already fixed for decode_syseeprom_test.py in #4516; this change applies the
equivalent fix to sed_test.py.

#### How I did it
- Added module-level `sys.modules['sonic_platform'] = MagicMock()` so
  `@patch('sonic_platform.platform.Platform')` can resolve its target.
- Added an autouse pytest fixture `_inject_sonic_platform` that re-injects the mock
  before every test, after `conftest._reset_between_files()` clears it between test
  files.

#### How to verify it
tests/sed_test.py

PR #4600, opened 2026-06-10
fix(tests): mock sonic_platform module in sed_test.py

rookie-who:
#### Why I did it

`tests/sed_test.py` (added in #4185) uses `@patch('sonic_platform.platform.Platform')`, but `sonic_platform` is a hardware-specific package not available in the build/test environment. The `@patch` decorator tries to import the module and fails with `ModuleNotFoundError`, causing all 8 SED tests to fail.

This breaks sonic-buildimage CI across **all** platforms (aspeed, broadcom, marvell, alpinevs, vpp, vs) — see [sonic-buildimage#27773](https://github.com/sonic-net/sonic-buildimage/pull/27773).

#### How I did it

Inject `MagicMock()` stubs into `sys.modules` for `sonic_platform` and `sonic_platform.platform` before importing `config.main`, following the established pattern used by `psuutil_test.py`, `ssdutil_test.py`, `fwutil_test.py`, and `bmc_test.py`.

Without these stubs, `@patch('sonic_platform.platform.Platform')` attempts a real import of the module, which does not exist in the build environment. The other test files that mock `sonic_platform` all pre-inject it into `sys.modules` at module level — `sed_test.py` was the only one missing this step.

#### How to verify it

All 8 `sed_test.py` tests should pass in CI instead of failing with `ModuleNotFoundError: No module named 'sonic_platform'`.

Fixes: sonic-net/sonic-buildimage#27773

### DIFF
--- a/tests/sed_test.py
+++ b/tests/sed_test.py
@@ -1,5 +1,13 @@
+import sys
 from click.testing import CliRunner
 from mock import patch, MagicMock
+
+# sonic_platform is a hardware-specific package not available in the
+# build/test environment.  Inject stub modules so @patch can resolve
+# the dotted path without raising ModuleNotFoundError.
+sys.modules['sonic_platform'] = MagicMock()
+sys.modules['sonic_platform.platform'] = MagicMock()
+
 import config.main as config
 
 


vaibhavhd (approved):

rookie-who:
Closing — this is a duplicate of #4596 (auto cherry-pick of #4583 to 202605). #4596 already covers the same fix.

PR #4604, opened 2026-06-11
[multi-ASIC] counterpoll flowcnt-route namespace-aware capability check

DavidZagury:
Pass the namespace to exit_if_route_flow_counter_not_support() so the STATE_DB capability check queries the correct per-namespace instance instead of the hardcoded 127.0.0.1 address.

#### What I did
Added multi-ASIC support for counterpoll flowcnt-route by making the route flow counter capability check namespace-aware.

#### How I did it

The flowcnt_route group in counterpoll/main.py already accepted a -n/--namespace option and connected to the correct per-namespace ConfigDB, but the capability check (exit_if_route_flow_counter_not_support) was still querying STATE_DB at the hardcoded address 127.0.0.1, ignoring the selected namespace.

Two changes were made:

- flow_counter_util/route.py: Added an optional namespace parameter to get_route_flow_counter_capability() and exit_if_route_flow_counter_not_support(). When a namespace is provided, the STATE_DB connection uses the unix socket path with the namespace (consistent with connect_to_db in counterpoll). Falls back to host="127.0.0.1" for single-ASIC, preserving backward compatibility for all existing callers.
- counterpoll/main.py: Pass namespace to exit_if_route_flow_counter_not_support(namespace) in the flowcnt_route group callback.

#### How to verify it

Should check capability against the correct namespace STATE_DB
counterpoll flowcnt-route -n asic0 enable
counterpoll flowcnt-route -n asic1 enable

Single-ASIC / no namespace — backward-compatible behavior unchanged
counterpoll flowcnt-route enable
counterpoll flowcnt-route disable
counterpoll flowcnt-route interval 5000

Confirm that on a namespace where route flow counters are not supported, the command correctly exits with "Route flow counter is not supported on this platform" rather than silently succeeding or querying the wrong STATE_DB instance.

### DIFF
--- a/counterpoll/main.py
+++ b/counterpoll/main.py
@@ -491,7 +491,7 @@ def flowcnt_trap_disable(ctx):
 def flowcnt_route(ctx, namespace):
     """ Route flow counter commands """
     ctx.obj = connect_to_db(namespace)
-    exit_if_route_flow_counter_not_support()
+    exit_if_route_flow_counter_not_support(namespace)
 
 
 @flowcnt_route.command(name='interval')
--- a/flow_counter_util/route.py
+++ b/flow_counter_util/route.py
@@ -1,6 +1,6 @@
 import os
 import sys
-from swsscommon.swsscommon import SonicV2Connector
+from swsscommon.swsscommon import SonicV2Connector, SonicDBConfig
 
 try:
     if os.environ["UTILITIES_UNIT_TESTING"] == "1" or os.environ["UTILITIES_UNIT_TESTING"] == "2":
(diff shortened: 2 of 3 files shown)


saksarav-nokia:
I already created PR  https://github.com/sonic-net/sonic-utilities/pull/4586 . Why do we need another PR?

oleksandrivantsiv (approved):

oleksandrivantsiv:
@saksarav-nokia, this is just a coincidence — looks like you both were trying to extend those commands at the same time.

I suggest keeping this PR, since #4586 doesn't add tests to cover the extension.

qiluo-msft (approved):

PR #4606, opened 2026-06-11
[config reload] Skip swss readiness check when swss is not present

william8545:
#### What I did
`config reload` (without `-f`) fails on platforms that do not ship the swss container (e.g. BMC) with `swss container is not ready. Retry later or use -f to avoid system checks`. The readiness gate waits on a `swss.service` that does not exist on such images, so reload can never proceed without `-f`. This makes the gate skip the wait when swss is not present.

#### How I did it
In `_per_namespace_swss_ready()`, check the service's systemd `LoadState` first. If it is `not-found` (service not built into the image) or `masked`, return ready — there is no swss to wait for. Platforms that ship swss read `LoadState=loaded`, so the existing active/120s-settle checks are unchanged. Added unit tests for the not-found/masked paths and regression guards for the loaded/inactive and active/settle behavior.

#### How to verify it
- `pytest tests/config_test.py::TestSwssReady`
- On a platform without swss (`swss.service` is `not-found`): `config reload -y` proceeds instead of aborting.
- On a switch that ships swss: behavior is unchanged (still waits until swss has been active for >120s).

```
root@sonic:~# config reload -y
Acquired lock on /etc/sonic/reload.lock
swss container is not ready. Retry later or use -f to avoid system checks
Released lock on /etc/sonic/reload.lock
```

```
root@sonic:~# config reload -y
Acquired lock on /etc/sonic/reload.lock
Running command: ...   # reload proceeds normally
Released lock on /etc/sonic/reload.lock
```

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -1206,7 +1206,8 @@ def _restart_services():
     reset_mgmt_interface_if_usb_not_running()
 
 def _per_namespace_swss_ready(service_name):
-    out, _ = clicommon.run_command(['systemctl', 'show', str(service_name), '--property', 'LoadState', '--value'], return_cmd=True)
+    out, _ = clicommon.run_command(
+        ['systemctl', 'show', str(service_name), '--property', 'LoadState', '--value'], return_cmd=True)
     if out.strip() in ("not-found", "masked"):
         # swss not present on this platform (e.g. BMC): nothing to wait for.
         return True
--- a/config/main.py
+++ b/config/main.py
@@ -1206,6 +1206,10 @@ def _restart_services():
     reset_mgmt_interface_if_usb_not_running()
 
 def _per_namespace_swss_ready(service_name):
+    out, _ = clicommon.run_command(['systemctl', 'show', str(service_name), '--property', 'LoadState', '--value'], return_cmd=True)
+    if out.strip() in ("not-found", "masked"):
+        # swss not present on this platform (e.g. BMC): nothing to wait for.
+        return True
     out, _ = clicommon.run_command(['systemctl', 'show', str(service_name), '--property', 'ActiveState', '--value'], return_cmd=True)
     if out.strip() != "active":
         return False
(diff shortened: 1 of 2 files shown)


oleksandrivantsiv (approved):

yxieca (approved):
Nice, well-targeted fix. `config reload` without `-f` was effectively deadlocked on platforms that don't ship the swss container (BMC), since the readiness gate waited on a `swss.service` that doesn't exist. Checking systemd `LoadState` first and treating `not-found`/`masked` as "nothing to wait for" is the right call, and it's at the correct abstraction layer — the function that probes the service decides from systemd ground truth, rather than threading a platform conditional through the reload path.

I verified the important properties:
- **Switch behavior is provably unchanged.** The change is purely additive (`+5/-0`, no existing line touched). A swss-shipping switch reports `LoadState=loaded`, so the new branch is skipped and the existing `ActiveState=="active"` + 120s-settle logic runs verbatim.
- **Failure modes fail safe.** Empty/garbage/failed-`systemctl` output (`out.strip()` not in the tuple) falls through to the existing ActiveState check → `!= active` → not ready. So there's no path where a switch that *should* have swss wrongly skips the wait. `not-found` also can't transiently fire on a real switch — an installed unit file reports `loaded` long before `config reload` runs.
- **Including `masked` is correct**, not scope creep — it's the same deadlock class (a masked unit can never become active).
- **Tests are great for a 5-line change**: both new paths (with `call_count == 1` asserting the short-circuit) plus regression guards for the `loaded`/`inactive`/`active`/settle behavior, and the settle-math mocks match the production formula.

Two small, non-blocking suggestions:

- **Optional multi-ASIC `not-found` test.** `_swss_ready()` already has a single-ASIC not-found end-to-end case; a `get_num_asics() == 2` case with both `swss@0/1` reporting `not-found` would round out coverage. The multi-ASIC iteration itself is unchanged code, so this is belt-and-suspenders rather than a gap.

- **Docstring on the overloaded return.** `_per_namespace_swss_ready()` now returns `True` for both "swss active and settled" and "swss absent." That's exactly right for this gate (both mean "safe to proceed"), but a one-line docstring noting the contract is "safe to proceed past the swss gate" (not "swss is active") would prevent a future caller from misreading it.

Neither blocks. Approving.

AI agent on behalf of Ying.

PR #4611, opened 2026-06-13
[BMC] Improvements in show chassis/platform leak_profile commands 

judyjoseph:
#### What I did

* Fix to read leak_profile table from STATE_DB
* default admin_status as down for switch-host
* add Power-On-Delay (sec)   Shutdown-Timeout (sec) to show chassis module status
* Remove the Physical-Slot column in show chassis module status

#### How to verify it
```
admin@host1-remote-bmc:~$ sudo config chassis modules startup SWITCH-HOST
Starting up chassis module SWITCH-HOST
admin@host1-remote-bmc:~$ show chassis modules status 
       Name                              Description    Physical-Slot    Oper-Status    Admin-Status    Serial
-----------  ---------------------------------------  ---------------  -------------  --------------  --------
SWITCH-HOST  Main x86 Switch Host CPU managed by BMC              N/A         ONLINE              up       N/A

admin@host1-remote-bmc:~$ sudo config chassis modules shutdown SWITCH-HOST
Module SWITCH-HOST is already in down state
admin@host1-remote-bmc:~$ show chassis modules status 
       Name                              Description    Physical-Slot    Oper-Status    Admin-Status    Serial
-----------  ---------------------------------------  ---------------  -------------  --------------  --------
SWITCH-HOST  Main x86 Switch Host CPU managed by BMC              N/A        OFFLINE            down       N/A

dmin@host1-remote-bmc:~$ show platform  leak profiles
Sensor-Type      Max-Minor-Duration-Sec
-------------  ------------------------
default                             300

<show chassis module status> o/p after removing the Physical Slot 

admin@host1-remote-bmc:~$ show chassis module status

Name         Description   Oper Status   Admin Status   Serial   Power-On-Delay (sec)   Shutdown-Timeout (sec)
--------------------------------------------------------------------------------------------------------------------------------
SWITCH-HOST  Switch Host   online        up             ABC123   300                    120

```

### DIFF
--- a/config/chassis_modules.py
+++ b/config/chassis_modules.py
@@ -252,6 +252,8 @@ if is_bmc():
         config_db = db.cfgdb
         fvs = config_db.get_entry('CHASSIS_MODULE', chassis_module_name) or {}
         fvs['power_on_delay'] = str(seconds)
+        # Seed admin_status of SWITCH-HOST with default if not already set
+        fvs.setdefault('admin_status', 'down')
         config_db.set_entry('CHASSIS_MODULE', chassis_module_name, fvs)
         click.echo(f"Power-on-delay for {chassis_module_name} set to {seconds} seconds")
 
@@ -272,5 +274,7 @@ if is_bmc():
         config_db = db.cfgdb
         fvs = config_db.get_entry('CHASSIS_MODULE', chassis_module_name) or {}
         fvs['graceful_shutdown_timeout'] = str(seconds)
+        # Seed admin_status of SWITCH-HOST with default if not already set
+        fvs.setdefault('admin_status', 'down')
         config_db.set_entry('CHASSIS_MODULE', chassis_module_name, fvs)
         click.echo(f"Shutdown-timeout for {chassis_module_name} set to {seconds} seconds")
--- a/show/chassis_modules.py
+++ b/show/chassis_modules.py
@@ -124,8 +124,8 @@ def status(db, chassis_module_name):
         # Determine admin_status
         if smartswitch:
             admin_status = 'down'
-        elif is_bmc():
-            # On BMC, modules default to 'down' (kept powered off on boot)
+        elif is_bmc() and key_list[1].startswith("SWITCH-HOST"):
+            # On BMC, SWITCH-HOST default is 'down' (kept powered off on boot)
             admin_status = 'down'
         else:
             admin_status = 'up'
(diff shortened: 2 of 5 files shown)


yxieca (approved):
## Review summary

LGTM on the code. I reviewed the three behavior changes and they look correct:

- `show platform leak profiles` reading `LEAK_PROFILE` from **STATE_DB** is correct. The actual writer is `thermalctld`'s `LiquidCoolingUpdater`, which writes `LEAK_PROFILE` to STATE_DB at startup (same pattern as the sibling leak tables), so the old CONFIG_DB reader was the bug.
- The BMC `show chassis modules status` column change is aligned: `Physical-Slot` is omitted on BMC and the two timing columns are appended; row/header ordering stays consistent.
- The `admin_status='down'` seed is narrowly scoped: BMC-only command registration, SWITCH-HOST-only validation, `setdefault` preserves existing values, and the show path already treats missing SWITCH-HOST admin-status as down.
- CI is green.

A few non-blocking asks before merge:

**1. Add/strengthen an exact-output test for the new BMC table shape.** Existing `tests/chassis_modules_test.py` has exact expected-output assertions for `show chassis modules status`. The new BMC tests mostly assert substrings/absence, which can miss table column-order or value-alignment regressions. Since this PR changes a tabular CLI contract, please add an exact-output assertion for the BMC `show chassis modules status SWITCH-HOST` output (or equivalent normalized full-line comparison).

**2. Please clean up the PR verification examples.** The PR body currently has inconsistent `show chassis modules status` examples: one still shows `Physical-Slot`, while the later intended BMC output removes it and adds `Power-On-Delay (sec)` / `Shutdown-Timeout (sec)`. There is also a `dmin@...` typo. Since this PR changes user-visible CLI output, the examples should match the final output.

**3. Please confirm the `admin_status='down'` seeding semantics.** The implementation looks safe, but the timing commands now persist an additional field when a fresh SWITCH-HOST entry is created. Please confirm no downstream consumer depends on distinguishing absent `admin_status` from explicit `admin_status=down`.

**4. Follow-up for stale LEAK_PROFILE docs/schema drift.** The code path is STATE_DB, but some stale design/schema material appears to describe `LEAK_PROFILE` as CONFIG_DB. This PR fixes the CLI reader correctly; it would be good to clean up the stale docs/schema in a follow-up so the reader does not regress back to CONFIG_DB later.

Minor optional nit: in `show/chassis_modules.py`, `bmc = is_bmc()` is already computed; the later `is_bmc()` calls inside `status()` could reuse the local for readability.

PR #4621, opened 2026-06-16
improve the performance of show ip interface

yanjundeng:
#### What I did
Improve the response performance of "show ip interface" and "show ipv6 interfaces" in large scale setup.

#### How I did it
Main improvement for show ip interface: major scalability/performance rewrite in ipintutil.

1. Net effect is a clear performance and scale improvement for show ip interface, especially on large systems with many interfaces/VRFs.
2. Architecture is cleaner: one kernel-state provider in multi_asic.py plus in-memory rendering in ipintutil.
3. Functional output compatibility was considered (IPv6 link-local formatting and BGP fallback retained).

Details:

1. interface/address collection is changed from per-interface calls to a bulk model:

- Old path: per interface netifaces.ifaddresses plus per interface admin/oper sysfs reads.
- New path: two bulk RTNETLINK dumps (get_links + get_addr), then in-memory assembly.
- The patch explicitly calls out old O(N^2)-like behavior from repeated getifaddrs walks, and moves to near O(N) processing for many interfaces/VRFs.

2. New shared multi-ASIC helper added

- adds multi_asic_get_kernel_intf_state(namespace, af) in utilities_common/multi_asic.py.
- It returns normalized plain dicts for links and addrs, with namespace push/pop handling.
- This isolates kernel fetch logic for reuse and unit-test monkey-patching.

3. Admin/oper state retrieval optimized and behavior cleaned up

- In the bulk path, admin comes from interface flags and oper comes from RTNETLINK operstate, removing per-interface sysfs/subprocess overhead.
- UNKNOWN operstate is treated as up to align with devices that do not expose carrier semantics (loopback/VRF/bridge-like behavior).

4. Correctness/format improvements included in the rewrite

- IPv6 link-local addresses now restore zone id form like fe80::...%Interface so output matches expected historical format
- Master/VRF resolution now uses link master index from netlink data, which is namespace-correct in this flow (instead of relying on separate sysfs path logic).
- Existing BGP-neighbor mapping behavior is preserved with interface-based fallback for IPv6 unnumbered peers.

#### How to verify it
Verified it on large scale setup.
In a system with about 800 interfaces, "show ip interface" and "show ipv6 interfaces" need **55s** to get a response.
with this change, it only takes about **3s**.

[show-ip-int-improve-ut.log](https://github.com/user-attachments/files/29013863/show-ip-int-improve-ut.log)

### DIFF
--- a/scripts/ipintutil
+++ b/scripts/ipintutil
@@ -78,17 +78,9 @@ def get_if_admin_state(iface, namespace):
     """
     Given an interface name, return its admin state reported by the kernel
     """
-    if namespace == constants.DEFAULT_NAMESPACE:
-        # Read sysfs directly — avoids spawning a subprocess per interface
-        try:
-            with open("/sys/class/net/{0}/flags".format(iface)) as f:
-                flags = int(f.read().strip(), 16)
-            return "up" if flags & 0x1 else "down"
-        except (OSError, ValueError):
-            return "error"
-
-    cmd = ["sudo", "ip", "netns", "exec", namespace,
-           "cat", "/sys/class/net/{0}/flags".format(iface)]
+    cmd = ["cat", "/sys/class/net/{0}/flags".format(iface)]
+    if namespace != constants.DEFAULT_NAMESPACE:
+        cmd = ["sudo", "ip", "netns", "exec", namespace] + cmd
     try:
         proc = subprocess.Popen(
             cmd,
@@ -108,23 +100,19 @@ def get_if_admin_state(iface, namespace):
     except ValueError:
         return "error"
 
-    return "up" if flags & 0x1 else "down"
+    if flags & 0x1:
+        return "up"
+    else:
+        return "down"
 
 
 def get_if_oper_state(iface, namespace):
     """
     Given an interface name, return its oper state reported by the kernel.
     """
-    if namespace == constants.DEFAULT_NAMESPACE:
-        # Read sysfs directly — avoids spawning a subprocess per interface
-        try:
-            with open("/sys/class/net/{0}/carrier".format(iface)) as f:
-                return "up" if f.read().strip() == "1" else "down"
-        except OSError:
-            return "error"
-
-    cmd = ["sudo", "ip", "netns", "exec", namespace,
-           "cat", "/sys/class/net/{0}/carrier".format(iface)]
+    cmd = ["cat", "/sys/class/net/{0}/carrier".format(iface)]
+    if namespace != constants.DEFAULT_NAMESPACE:
+        cmd = ["sudo", "ip", "netns", "exec", namespace] + cmd
     try:
         proc = subprocess.Popen(
             cmd,
@@ -139,7 +127,10 @@ def get_if_oper_state(iface, namespace):
         return "error"
 
     oper_state = state_file.rstrip()
-    return "up" if oper_state == "1" else "down"
+    if oper_state == "1":
+        return "up"
+    else:
+        return "down"
 
 
 def get_if_master(iface):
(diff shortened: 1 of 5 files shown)


yanjundeng:
@prsunny ,  pls help review.

scripts/ipintutil
```diff
@@ -147,47 +147,99 @@ def get_if_master(iface):
```
pbrisset, line 147: Dead code left behind (should remove). After the rewrite, [ipintutil:77](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), [ipintutil:109](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), and [ipintutil:136](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) have no remaining callers (I grepped the whole repo — only their old call sites at lines 186–188, which the diff replaces). They should be deleted. Similarly, [multi_asic_get_ip_intf_addr_from_ns](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) becomes production-unused (only referenced by tests/mocks/conftest now); [multi_asic_get_ip_intf_from_ns](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) is still used by storm-control in [main.py:518](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), so that one stays.

pbrisset (approved):
One more comment:
UNKNOWN oper-state is now treated as up.
Old path read [/sys/class/net/<iface>/carrier](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) (1 → up, otherwise down). New path uses operstate in ('UP', 'UNKNOWN') → 'up'. Mapping UNKNOWN to up is reasonable for carrier-less devices (loopback, VRF, bridge), and the comment explains it well. But for a physical port that the kernel reports as UNKNOWN, the displayed oper state will now differ from before. Is that acceptable for all interface types this command renders?

The PR otherwise looks good

scripts/ipintutil
yanjundeng: we may remove them later after merging this PR.

yanjundeng:
> One more comment: UNKNOWN oper-state is now treated as up. Old path read /sys/class/net//carrier (1 → up, otherwise down). New path uses operstate in ('UP', 'UNKNOWN') → 'up'. Mapping UNKNOWN to up is reasonable for carrier-less devices (loopback, VRF, bridge), and the comment explains it well. But for a physical port that the kernel reports as UNKNOWN, the displayed oper state will now differ from before. Is that acceptable for all interface types this command renders?
> 
> The PR otherwise looks good

"a physical port the kernel reports as UNKNOWN with the link actually down" — cannot occur. If it's UNKNOWN, carrier is 1, which the old code also rendered as up. No divergence.

How this maps to SONiC interface types
Front-panel EthernetN (routed): the SAI host-interface netdev gets its carrier driven by the SDK from real SAI port oper-status, so it reports UP/DOWN (carrier 1/0) → identical to old. If the SDK hasn't driven a carrier event yet, it's UNKNOWN+carrier=1 → old code also showed up.
PortChannel, Vlan, Loopback, VRF devices: carrier-less; report UNKNOWN → up, same as old (carrier=1).
Admin-down interfaces: the only real change. Old code's carrier read returns EINVAL → it printed error; new code prints down. That's a fix, not a regression.

Issue #4595, opened 2026-06-06
[sonic-clear] Missing clear commands for QoS counters: buffer_pool watermark, PFCWD counters


pinky-nexthop:
### Description

`sonic-clear` is missing commands to clear several QoS / PFC counter and watermark types that are routinely needed for troubleshooting. This issue tracks two related **missing commands**:

1. No CLI to clear **buffer_pool watermark**.
2. No CLI to clear **PFC watchdog (PFCWD) counters**.

In each case the data is shown by an existing `show` command, so the clear side is the gap.

---

### 1. No CLI to clear buffer_pool watermark

`show buffer_pool watermark` reports occupancy, but there is no `sonic-clear` command to reset it:

```
$ show buffer_pool watermark
Shared pool maximum occupancy:
                 Pool     Bytes
---------------------  --------
 egress_lossless_pool  75444604
ingress_lossless_pool  75447652

$ sudo sonic-clear
arp   counters   headroom-pool   priority-group   queue   ...
# no 'buffer-pool' entry
```

`sonic-clear counters` does not reset the buffer pool watermark either.

**Note:** the backend already supports this — `watermarkstat -c -t buffer_pool` and `watermarkstat -c -p -t buffer_pool` exist. Only the `sonic-clear` command group is missing. The neighboring `headroom-pool` clear group is a direct precedent.

**Expected:**
```
sonic-clear buffer-pool watermark
sonic-clear buffer-pool persistent-watermark
```

---

### 2. No CLI to clear PFC watchdog (PFCWD) counters

There is no way to clear PFC watchdog counters. Clearing PFCWD counters is important when troubleshooting PFC storms, where stale counts from a prior event make it hard to tell whether a storm is ongoing.

`show pfcwd stats` displays the counters, but no corresponding `sonic-clear`/clear command exists to reset them.

**Expected:** a command to clear PFCWD counters, e.g.
```
sonic-clear pfcwd counters
```
(or an equivalent `pfcwd <clear>` subcommand).

---

### Affected platforms / versions

- Both items are platform-independent — the commands simply don't exist.
- Observed on Broadcom platforms, SONiC 202505 / master.

PR #4634, opened 2026-06-22
[clear] Add 'sonic-clear pfcwdstats' to clear PFC watchdog counters

pinky-nexthop:
#### What I did

There is currently no CLI to clear PFC watchdog statistics. Clearing them is needed when troubleshooting PFC storms — stale counts from a prior event make it hard to tell whether a storm is currently ongoing. `show pfcwd stats` displays the counters, but there is no corresponding clear command. This PR adds `sonic-clear pfcwdstats`.

Addresses the PFC-watchdog-counter-clear item in #4595.

#### How I did it

- `scripts/pfcwdstat`: new tool that zeroes the `PFC_WD_QUEUE_STATS_*` fields in `COUNTERS_DB`, following the standard counter-clear pattern (`pfcstat -c`, `queuestat -c`, …). It enumerates queues via `COUNTERS_QUEUE_NAME_MAP` — the same source `show pfcwd stats` uses — so stale stats from queues that are no longer actively monitored (`STATUS=N/A`) are also cleared.
- Queues with an active storm (`PFC_WD_STATUS == 'stormed'`) are skipped, so the in-flight `DEADLOCK_DETECTED`/`DEADLOCK_RESTORED` pair is not split (which would otherwise leave `RESTORED > DETECTED` once the storm ends).
- `clear/main.py`: adds `sonic-clear pfcwdstats`, invoking `pfcwdstat -c`.
- `setup.py`: installs `scripts/pfcwdstat`.
- Unit tests for the clear command and the `pfcwdstat` clear logic (including the stale-stats and active-storm cases).

#### How to verify it

```
pytest tests/pfcwd_test.py::TestPfcwdstatClear tests/clear_test.py -k pfcwd
```

Manually: `show pfcwd stats` to see counts, `sonic-clear pfcwdstats` to clear them; an actively storming queue is left untouched (and reported as skipped) until the storm ends.

```
admin@sonic:~$ show pfcwd stats
        QUEUE       STATUS    STORM DETECTED/RESTORED    TX OK/DROP    RX OK/DROP    TX LAST OK/DROP    RX LAST OK/DROP
-------------  -----------  -------------------------  ------------  ------------  -----------------  -----------------
Ethernet192:3  operational                        1/1           0/0           0/0                0/0                0/0
```

```
admin@sonic:~$ sonic-clear pfcwdstats
PFC Watchdog counters cleared.

admin@sonic:~$ show pfcwd stats
  QUEUE    STATUS    STORM DETECTED/RESTORED    TX OK/DROP    RX OK/DROP    TX LAST OK/DROP    RX LAST OK/DROP
-------  --------  -------------------------  ------------  ------------  -----------------  -----------------
```

- [ ] 201811
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511
- [ ] 202605

### DIFF
--- a/clear/main.py
+++ b/clear/main.py
@@ -208,6 +208,14 @@ def pfccounters():
     command = ["pfcstat", "-c"]
     run_command(command)
 
+
+@cli.command()
+def pfcwdstats():
+    """Clear PFC Watchdog stats counters"""
+    command = ["pfcwdstat", "-c"]
+    run_command(command)
+
+
 @cli.command()
 def dropcounters():
     """Clear drop counters"""
--- a/scripts/pfcwdstat
+++ b/scripts/pfcwdstat
@@ -1,0 +1,83 @@
+#!/usr/bin/env python3
+
+"""
+pfcwdstat - PFC Watchdog statistics tool
+
+Usage:
+    pfcwdstat -c      Clear PFCWD stats
+"""
+
+import argparse
+import sys
+from swsscommon.swsscommon import SonicV2Connector
+
+PFC_WD_STATS_FIELDS = [
+    'PFC_WD_QUEUE_STATS_DEADLOCK_DETECTED',
+    'PFC_WD_QUEUE_STATS_DEADLOCK_RESTORED',
+    'PFC_WD_QUEUE_STATS_TX_PACKETS',
+    'PFC_WD_QUEUE_STATS_TX_DROPPED_PACKETS',
+    'PFC_WD_QUEUE_STATS_RX_PACKETS',
+    'PFC_WD_QUEUE_STATS_RX_DROPPED_PACKETS',
+    'PFC_WD_QUEUE_STATS_TX_PACKETS_LAST',
+    'PFC_WD_QUEUE_STATS_TX_DROPPED_PACKETS_LAST',
+    'PFC_WD_QUEUE_STATS_RX_PACKETS_LAST',
+    'PFC_WD_QUEUE_STATS_RX_DROPPED_PACKETS_LAST',
+]
+
+def clear_stats():
+    """Zero out PFCWD stats in COUNTERS_DB.
+
+    Enumerates all queues via COUNTERS_QUEUE_NAME_MAP — the same source
+    'show pfcwd stats' uses — so stale stats from queues that are no longer
+    actively monitored (PFC_WD_TABLE entries gone, STATUS=N/A) are also cleared.
+
+    Queues with an active storm (PFC_WD_STATUS == 'stormed') are skipped.
+    """
+    db = SonicV2Connector()
+    db.connect(db.COUNTERS_DB)
+
+    queue_name_map = db.get_all(db.COUNTERS_DB, 'COUNTERS_QUEUE_NAME_MAP')
+    if not queue_name_map:
+        print("No PFC Watchdog stats found.")
+        return
+
+    cleared = False
+    skipped = []
+    for queue_name, queue_oid in queue_name_map.items():
+        counter_key = 'COUNTERS:' + queue_oid
+        stats = db.get_all(db.COUNTERS_DB, counter_key)
+        if not stats:
+            continue
+        if not any(f in stats for f in PFC_WD_STATS_FIELDS):
+            continue
+        if stats.get('PFC_WD_STATUS') == 'stormed':
+            skipped.append(queue_name)
+            continue
+        for field in PFC_WD_STATS_FIELDS:
+            db.set(db.COUNTERS_DB, counter_key, field, '0')
+        cleared = True
+
+    if cleared:
+        print("PFC Watchdog counters cleared.")
+    elif not skipped:
+        print("No PFC Watchdog stats found.")
+
+    if skipped:
+        print("Skipped queue(s) with an active storm (not cleared): {}".format(
+            ', '.join(sorted(skipped))))
+
+def main():
+    parser = argparse.ArgumentParser(
+        description='PFC Watchdog statistics tool'
+    )
+    parser.add_argument('-c', '--clear', action='store_true',
+                        help='Clear PFCWD stats')
+    args = parser.parse_args()
+
+    if args.clear:
+        clear_stats()
+    else:
+        parser.print_help()
+
+if __name__ == '__main__':
+    main()
--- a/tests/pfcwd_test.py
+++ b/tests/pfcwd_test.py
@@ -1027,3 +1027,166 @@ class TestMultiAsicPfcwdShow(object):
         importlib.reload(mock_tables.mock_single_asic)
         import pfcwd.main
         importlib.reload(pfcwd.main)
+
+
+class TestPfcwdstatClear(object):
+    @classmethod
+    def setup_class(cls):
+        print("SETUP")
+        scripts_path = os.path.join(
+            os.path.dirname(os.path.dirname(os.path.abspath(__file__))),
+            'scripts'
+        )
+        import importlib.util
+        from importlib.machinery import SourceFileLoader
+        script_path = os.path.join(scripts_path, "pfcwdstat")
+        loader = SourceFileLoader("pfcwdstat", script_path)
+        spec = importlib.util.spec_from_loader("pfcwdstat", loader, origin=script_path)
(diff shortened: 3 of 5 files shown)


scripts/pfcwdstat
```diff
@@ -0,0 +1,83 @@
...
+def clear_stats():
+    """Zero out PFCWD stats in COUNTERS_DB.
+
+    Enumerates all queues via COUNTERS_QUEUE_NAME_MAP — the same source
+    'show pfcwd stats' uses — so stale stats from queues that are no longer
+    actively monitored (PFC_WD_TABLE entries gone, STATUS=N/A) are also cleared.
+
+    Queues with an active storm (PFC_WD_STATUS == 'stormed') are skipped.
+    """
+    db = SonicV2Connector()
```
saksarav-nokia, line 36: sonicV2connector() handles only global namespace and does not work in multi asic. Please add support for multi-asic also

scripts/pfcwdstat
```diff
@@ -0,0 +1,83 @@
...
+    db.connect(db.COUNTERS_DB)
+
+    queue_name_map = db.get_all(db.COUNTERS_DB, 'COUNTERS_QUEUE_NAME_MAP')
```
saksarav-nokia, line 39: Usually all the clear commands take the snapshot of the  stats from COUNTER_DB and saves it in /tmp/cache . Then show commands shows the diff between stats COUNTER_DB and /tmp/cache. Why are you clearing it from COUNTER_DB. Is this the requirement?

clear/main.py
```diff
@@ -208,6 +208,14 @@ def pfccounters():
     command = ["pfcstat", "-c"]
     run_command(command)
 
+
+@cli.command()
+def pfcwdstats():
+    """Clear PFC Watchdog stats counters"""
```
saksarav-nokia, line 214: Please  update doc/Command-Reference.md

scripts/pfcwdstat
```diff
@@ -0,0 +1,83 @@
...
+        stats = db.get_all(db.COUNTERS_DB, counter_key)
+        if not stats:
+            continue
+        if not any(f in stats for f in PFC_WD_STATS_FIELDS):
+            continue
+        if stats.get('PFC_WD_STATUS') == 'stormed':
+            skipped.append(queue_name)
+            continue
+        for field in PFC_WD_STATS_FIELDS:
+            db.set(db.COUNTERS_DB, counter_key, field, '0')
```
saksarav-nokia, line 57: Even if you have to clear from COUNTERS_DB, can't you use hmset and clear all stats at once instead of clearing 1 by 1

scripts/pfcwdstat
vmittal-msft, line 36: **Multi-ASIC: this won't clear anything on a multi-ASIC/VOQ chassis** (+1 to @saksarav-nokia, confirming it's a functional gap, not style).

`SonicV2Connector()` with no namespace binds only to the *global* DB. But the reader `show pfcwd stats` (`pfcwd/main.py`) walks each ASIC via the `@multi_asic_util.run_on_multi_asic` decorator, and the sibling clear tools (`pfcstat`, `queuestat`) and even `pg-drop -c clear` all iterate `multi_asic.get_namespace_list()`.

So on a multi-ASIC box, `COUNTERS_QUEUE_NAME_MAP` in the global DB is empty → this prints `No PFC Watchdog stats found.` and clears nothing, while `show pfcwd stats` still shows per-ASIC counts. Please wrap `clear_stats()` in a namespace loop (`MultiAsic` / `run_on_multi_asic`) or add `-n/--namespace` like `pg-drop`.

scripts/pfcwdstat
vmittal-msft, line 57: **Directly zeroing `COUNTERS_DB` diverges from the standard clear pattern** (this is the crux of @saksarav-nokia's question).

Every other stats-clear in this repo is non-destructive: `pfcstat -c`, `queuestat -c`, `portstat -c`, and `pg-drop -c clear` snapshot `COUNTERS_DB` into `UserCache` (`/tmp/...`) and the `show` command renders `current - cached`. They never mutate `COUNTERS_DB`.

Writing `'0'` straight into the shared `COUNTERS_DB` is a different contract: any other consumer of these counters (telemetry / gNMI streaming, rate calculators) will observe a reset / negative delta, and the clear is global + irreversible rather than per-user. It only works here because you skip stormed queues (no active producer to overwrite the zeros).

Preferred: the cache+diff approach (which also means teaching `show pfcwd stats` to subtract the baseline). If a hard reset is genuinely required, please call that out and justify it in the PR description.

Minor (also @saksarav-nokia): if you keep the DB write, one `hmset`/pipeline per queue avoids the per-field round-trips.

scripts/pfcwdstat
```diff
@@ -0,0 +1,83 @@
...
+
+"""
+pfcwdstat - PFC Watchdog statistics tool
+
+Usage:
+    pfcwdstat -c      Clear PFCWD stats
+"""
+
+import argparse
+import sys
```
vmittal-msft, line 11: Nit: `sys` is imported but unused (F401). Also `clear_stats` (L27), `main` (L69), and the `__main__` guard (L82) are separated by a single blank line — PEP8 wants two (E302/E305).

Heads-up: the repo's pre-commit runs `flake8==4.0.1` in `--diff` mode, which does **not** flag these on a brand-new file (I reproduced it locally — that's why the Static Analysis check is green), so they'll silently slip through unless cleaned up by hand.

clear/main.py
```diff
@@ -208,6 +208,14 @@ def pfccounters():
     command = ["pfcstat", "-c"]
     run_command(command)
 
+
+@cli.command()
+def pfcwdstats():
```
vmittal-msft, line 213: +1 to @saksarav-nokia on docs. Please add `sonic-clear pfcwdstats` to `doc/Command-Reference.md` — there's already a `show pfcwd stats` section and a `NOTE: PFC counters ... sonic-clear pfccounters` blurb, so a parallel note under the PFC watchdog section keeps the new command discoverable.

tests/pfcwd_test.py
```diff
@@ -1027,3 +1027,166 @@ def teardown_class(cls):
         importlib.reload(mock_tables.mock_single_asic)
         import pfcwd.main
         importlib.reload(pfcwd.main)
+
+
+class TestPfcwdstatClear(object):
```
vmittal-msft, line 1032: These tests fully mock `SonicV2Connector`, so they pass regardless of the multi-ASIC gap flagged in `scripts/pfcwdstat` — a single mocked DB hides the fact that only the global namespace is cleared.

Once namespace handling is added, please add a multi-ASIC case (e.g. `UTILITIES_UNIT_TESTING_TOPOLOGY=multi_asic` with per-namespace mock DBs) asserting each namespace's counters are cleared. Tests for the `-c` arg path in `main()` and for the `Skipped queue(s) ...` / `No PFC Watchdog stats found.` output would also lock in the behavior.

vmittal-msft (commented):
Thanks for the PR! I reviewed the change end-to-end against the current sources. It works on a single-ASIC box, but two items I'd treat as blocking:

1. **No multi-ASIC / chassis support** — `SonicV2Connector()` only touches the global DB, so nothing is cleared on multi-ASIC while `show pfcwd stats` still shows counts.
2. **Destructive `COUNTERS_DB` mutation** — every sibling clear command uses the non-destructive `UserCache` snapshot+diff pattern instead of writing into the shared counters DB.

Plus a `doc/Command-Reference.md` update. Details inline — these corroborate and extend @saksarav-nokia's review. Nice touch skipping actively-stormed queues to keep the `DEADLOCK_DETECTED`/`RESTORED` pair intact.

rookie-who:
Hi — this PR is flagged for cherry-pick to the **202605 branch**, but it appears to be a new feature/enhancement rather than a bug fix. The 202605 branch is in stabilization, so we're limiting cherry-picks to bug fixes and critical platform support.

Could you please retarget this to the next release branch? If there's a specific customer requirement driving this for 202605, please comment with the justification and we can reconsider.

*Comment by @vaibhavhd via automated tooling*

PR #4649, opened 2026-06-29
Fix route_check.py: return dict format from fetch_routes for mitigation

mike-dubrovsky:
## Summary

Split from #4576 per [@deepak-singhal0408's review](https://github.com/sonic-net/sonic-utilities/pull/4576#issuecomment-2976544324) — **Bucket A (bug fix, land immediately)**.

`mitigate_installed_not_offloaded_frr_routes()` expects `missed_frr_rt` entries with `prefix` and `protocol` keys, but `fetch_routes()` has been appending bare prefix strings since the ijson refactor (#4205). This causes `TypeError: string indices must be integers` whenever suppress-fib-pending triggers mitigation.

Adopts the 202511 approach (#4509): append `{'prefix': prefix, 'protocol': route_entry.get('protocol', '')}` so the mitigation contract is preserved. Keeps the existing ijson JSON parser unchanged.

## Test plan

- [ ] `tests/fetch_routes_chunk_test.py` expectations updated for dict format
- [ ] CI unit tests pass
- [ ] On a device with suppress-fib-pending enabled and stuck routes, mitigation no longer crashes

### DIFF
--- a/scripts/route_check.py
+++ b/scripts/route_check.py
@@ -417,7 +417,7 @@ def fetch_routes(ipv6=False, namespace=multi_asic.DEFAULT_NAMESPACE):
         if not route_entry.get('selected', False):
             return
         if not route_entry.get('offloaded', False):
-            missing_routes.append(prefix)
+            missing_routes.append({'prefix': prefix, 'protocol': route_entry.get('protocol', '')})
         if route_entry.get('failed', False):
             failing_routes.append(prefix)
 
--- a/tests/fetch_routes_chunk_test.py
+++ b/tests/fetch_routes_chunk_test.py
@@ -19,6 +19,11 @@ sys.path.append("scripts")
 import route_check  # noqa: E402
 
 
+def get_missing_prefixes(result):
+    """Extract prefix strings from missing_routes dicts for assertion comparison."""
+    return {entry['prefix'] for entry in result}
+
+
 class ChunkedBytesIO:
     def __init__(self, chunks):
         self.chunks = chunks
@@ -79,7 +84,7 @@ class TestFetchRoutes:
         with patch('route_check.subprocess.Popen', return_value=mock_proc):
             result = route_check.fetch_routes()
 
-        assert result == (["192.168.1.0/24"], [])
+        assert result == ([{"prefix": "192.168.1.0/24", "protocol": "bgp"}], [])
 
     def test_json_split_across_chunks(self):
         """Test parsing when JSON is split across multiple chunks"""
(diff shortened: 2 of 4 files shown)


deepak-singhal0408 (commented):
**Test coverage gap: the mitigation contract this PR fixes isn't exercised end-to-end**

The one-line producer fix looks right. One concern on the test side: [`route_check_test.py::run_test`](https://github.com/sonic-net/sonic-utilities/blob/master/tests/route_check_test.py) patches `check_frr_pending_routes` with `mock_fetch_routes`, and that mock still does `missed_route_list.append(r)` — a **bare prefix string**. The golden data in [`route_check_test_data.py`](https://github.com/sonic-net/sonic-utilities/blob/master/tests/route_check_test_data.py) also still expects `"missed_FRR_routes": ['10.10.196.12/31']` (strings).

So these end-to-end tests pass only because the mock is stale — they no longer mirror production, which now emits dicts, and nothing drives `rt_frr_miss` (dicts) → `mitigate_installed_not_offloaded_frr_routes()` through the real path. A future regression that reverts the producer back to strings would still pass `route_check_test.py`.

Could you update `mock_fetch_routes` and the `missed_FRR_routes` expectations in `route_check_test_data.py` to the dict format, and ideally add one case that drives the real mitigation path with a stuck route? That closes the loop on the exact contract this PR restores.

deepak-singhal0408 (approved):
Re-reviewed after the latest push (`62ca7d5d`) — the test-fidelity gap I raised is resolved, approving.

`tests/route_check_test.py::mock_fetch_routes` now returns `{'prefix': r, 'protocol': e.get('protocol', '')}`, which byte-matches the production `process_route_entry` shape (`missing_routes.append({'prefix': prefix, 'protocol': route_entry.get('protocol', '')})`), and the four `route_check_test_data.py` `missed_FRR_routes` expectations are updated to the dict form. The end-to-end path now exercises the dict/mitigation contract this PR fixes, so a regression of the producer back to bare strings would be caught. `scripts/route_check.py` is unchanged, and CI is green after the re-run. Nice work.

deepak-singhal0408 (approved):

prsunny:
how about 202511?

deepak-singhal0408:
> how about 202511?

202511 already has this change @prsunny. 
202511 was fixed as part of https://github.com/sonic-net/sonic-utilities/pull/4509(a few PR cherry-picks + this bug fix)

vaibhavhd:
@mike-dubrovsky please update testing results from 202605 branch

deepak-singhal0408:
> @mike-dubrovsky please update testing results from 202605 branch

@mike-dubrovsky, gentle reminder on this one. the issue is showing up frequently in our 202605 Nightly runs. could you help do a quick test on 202605 branch as that is a gating requirement for cherry-picks now. thank you.

mike-dubrovsky:
@vaibhavhd and @deepak-singhal0408

I did test 202605 branch and it looks good. test results attached.
[202605_testing_results.log](https://github.com/user-attachments/files/30073946/202605_testing_results.log)

PR #4650, opened 2026-06-29
Fix route_check.py: intersection retry and snapshot reorder

mike-dubrovsky:
## Summary

Split from #4576 per [@deepak-singhal0408's review](https://github.com/sonic-net/sonic-utilities/pull/4576#issuecomment-2976544324) — **Bucket B (correctness, land after Bucket A)**.

Depends on #4649 (dict-format bugfix). Includes two improvements:

1. **Intersection-based retry** in `check_frr_pending_routes()`: only routes stuck in *every* poll window are mitigated, preventing premature mitigation of converging routes.
2. **Snapshot reorder**: move ASIC/APPL DB snapshot to after FRR polling completes so `rt_appl_miss` / `rt_asic_miss` are temporally aligned with FRR results when the mitigation cross-check fires.

Keeps the ijson JSON parser (no text-streaming change).

## Test plan

- [ ] `tests/check_frr_pending_routes_test.py` — intersection accumulator cases
- [ ] CI unit tests pass
- [ ] Routes that converge between retry iterations are not mitigated
- [ ] Routes finishing ASIC programming during the FRR retry window do not suppress mitigation

### DIFF
--- a/scripts/route_check.py
+++ b/scripts/route_check.py
@@ -742,20 +742,45 @@ def is_feature_bgp_enabled(namespace):
 
 def check_frr_pending_routes(namespace):
     """
-    Check FRR routes for offload flag presence by executing "show ip route json"
-    Returns a list of routes that have no offload flag.
+    Check FRR routes for offload flag presence by executing "show ip route json".
+    Returns lists of routes that are persistently non-offloaded across all retry
+    iterations (intersection logic).
+
+    Intersection approach: only routes that appear as stuck in *every* poll
+    are returned for mitigation.  A route that clears between iterations is
+    converging normally and should not be mitigated.
     """
+    acc_miss = []
+    acc_fail = []
 
-    missed_rt = []
-    failed_rt = []
-    retries = FRR_CHECK_RETRIES
-    for i in range(retries):
-        missed_rt, failed_rt = get_frr_routes_parallel(namespace)
+    for i in range(FRR_CHECK_RETRIES):
+        curr_miss, curr_fail = get_frr_routes_parallel(namespace)
 
-        if not missed_rt and not failed_rt:
+        if i == 0:
+            acc_miss = curr_miss
+            acc_fail = curr_fail
+        else:
+            curr_miss_set = {entry['prefix'] for entry in curr_miss}
+            curr_fail_set = set(curr_fail)
+            acc_miss = [entry for entry in acc_miss if entry['prefix'] in curr_miss_set]
+            acc_fail = [prefix for prefix in acc_fail if prefix in curr_fail_set]
+
+        if not acc_miss and not acc_fail:
             break
 
-        time.sleep(FRR_WAIT_TIME)
+        if i < FRR_CHECK_RETRIES - 1:
+            time.sleep(FRR_WAIT_TIME)
+
+    missed_rt = acc_miss
+    failed_rt = acc_fail
+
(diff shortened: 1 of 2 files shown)


scripts/route_check.py
```diff
@@ -426,7 +417,7 @@ def process_route_entry(prefix, route_entry):
         if not route_entry.get('selected', False):
             return
         if not route_entry.get('offloaded', False):
-            missing_routes.append(prefix)
+            missing_routes.append({'prefix': prefix, 'protocol': route_entry.get('protocol', '')})
```
deepak-singhal0408, line 429: this change is already merged. please rebase your PR.

mike-dubrovsky: done

scripts/route_check.py
```diff
@@ -733,20 +733,37 @@ def is_feature_bgp_enabled(namespace):
...
+            acc_fail = curr_fail
+        else:
+            curr_miss_set = {entry['prefix'] for entry in curr_miss}
+            curr_fail_set = set(curr_fail)
+            acc_miss = [entry for entry in acc_miss if entry['prefix'] in curr_miss_set]
+            acc_fail = [prefix for prefix in acc_fail if prefix in curr_fail_set]
 
-        if not missed_rt and not failed_rt:
+        if not curr_miss and not curr_fail:
             break
```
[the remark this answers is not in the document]
mike-dubrovsky: fixed

scripts/route_check.py
```diff
@@ -742,20 +742,37 @@ def is_feature_bgp_enabled(namespace):
...
-        if not missed_rt and not failed_rt:
+        if not acc_miss and not acc_fail:
             break
 
-        time.sleep(FRR_WAIT_TIME)
+        if i < FRR_CHECK_RETRIES - 1:
+            time.sleep(FRR_WAIT_TIME)
+
+    missed_rt = acc_miss
+    failed_rt = acc_fail
```
deepak-singhal0408, line 775: Thanks for the rebase — diff is clean now, and nice touch moving the loop break onto the accumulator (`if not acc_miss and not acc_fail`) plus the final-iteration sleep guard.

**Suggestion:** this change shifts the mitigation set from *last poll* to *intersection of all polls*, so the set `(last poll) − (intersection)` — routes that would have been mitigated under the previous logic but now aren't (they converged during the retry window, or appeared after the first poll) — is now excluded without any log record. Since this is an intentional behavior change, a DEBUG line for the excluded set would make it diagnosable from syslog alone:

```python
missed_rt = acc_miss
failed_rt = acc_fail
excluded_miss = {e['prefix'] for e in curr_miss} - {e['prefix'] for e in acc_miss}
excluded_fail = set(curr_fail) - set(acc_fail)
if excluded_miss or excluded_fail:
    print_message(syslog.LOG_DEBUG,
        f"Routes pending in final poll but excluded by intersection "
        f"(not mitigated; converged during retry window): "
        f"missed={excluded_miss} failed={excluded_fail}")
```

DEBUG (not INFO) to avoid recurring noise on healthy scheduled runs, consistent with the existing `FRR missed/failed routes` debug lines. Purely observability — doesn't change mitigation behavior.

mike-dubrovsky: done

deepak-singhal0408 (approved):
All review comments addressed. Changes LGTM.

PR #4652, opened 2026-06-29
acl-loader: Skip adding `IP_TYPE` on ACLs that don't support it

arista-nwolfe:
In https://github.com/sonic-net/sonic-utilities/pull/4462 functionality was added to include a IP_TYPE qualifier on rules with IP_PROTOCOL.
This broke custom ACL types that don't support the IP_TYPE qualifier.
https://github.com/sonic-net/sonic-buildimage/issues/28028
This change will avoid adding IP_TYPE on those custom ACL types.

#### What I did
Skip adding `IP_TYPE` qualifier automatically on custom ACL types that don't support `IP_TYPE` qualifier

#### How I did it
In `acl-loader` check if `IP_TYPE` is in the `MATCHES` attribute of `ACL_TABLE_TYPE` before adding the `IP_TYPE` qualifier.

#### How to verify it
I executed the same steps described in
https://github.com/sonic-net/sonic-buildimage/issues/28028

Before this change I'd see the `IP_TYPE` added to the `RULE_1` entry in CONFIG_DB:
```
# sonic-db-cli CONFIG_DB hgetall "ACL_RULE|TEST_NOIPTYPE|RULE_1"
{'IP_PROTOCOL': '6', 'IP_TYPE': 'IP', 'PACKET_ACTION': 'FORWARD', 'PRIORITY': '9999', 'SRC_IP': '20.0.0.2/32'}
root@ctn106:~# show acl rule
Table          Rule    Priority    Action    Match                Status
-------------  ------  ----------  --------  -------------------  --------
TEST_NOIPTYPE  RULE_1  9999        FORWARD   IP_PROTOCOL: 6       Inactive
                                             IP_TYPE: IP
                                             SRC_IP: 20.0.0.2/32
```
And the error syslog in orchagent:
```
2026 Jun 29 18:22:34.681995 ctn106 ERR swss#orchagent: :- validateAclRuleMatch: Match SAI_ACL_ENTRY_ATTR_FIELD_ACL_IP_TYPE in rule RULE_1 is not supported by table TEST_NOIPTYPE             2026 Jun 29 18:22:34.681995 ctn106 ERR swss#orchagent: :- doAclRuleTask: Unknown or invalid rule attribute 'IP_TYPE : IP'
2026 Jun 29 18:22:34.682107 ctn106 ERR swss#orchagent: :- doAclRuleTask: Failed to create ACL rule. Rule configuration is invalid
```

After this change I see the `IP_TYPE` isn't added to the `RULE_1` entry in CONFIG_DB:
```
root@ctn106:~# sonic-db-cli CONFIG_DB hgetall "ACL_RULE|TEST_NOIPTYPE|RULE_1"
{'IP_PROTOCOL': '6', 'PACKET_ACTION': 'FORWARD', 'PRIORITY': '9999', 'SRC_IP': '20.0.0.2/32'}
root@ctn106:~# show acl rule
Table          Rule    Priority    Action    Match                Status
-------------  ------  ----------  --------  -------------------  --------
TEST_NOIPTYPE  RULE_1  9999        FORWARD   IP_PROTOCOL: 6       Active
                                             SRC_IP: 20.0.0.2/32
```
And the rule is successfully added:
```
2026 Jun 29 18:56:24.439088 ctn106 NOTICE swss#orchagent: :- add: Successfully created ACL rule RULE_1 in table TEST_NOIPTYPE
```

N/A

N/A

#### Casts
* 202511
* 202605

### DIFF
--- a/acl_loader/main.py
+++ b/acl_loader/main.py
@@ -72,6 +72,7 @@ class AclLoader(object):
 
     ACL_TABLE = "ACL_TABLE"
     ACL_RULE = "ACL_RULE"
+    ACL_TABLE_TYPE = "ACL_TABLE_TYPE"
     CFG_ACL_TABLE = "ACL_TABLE"
     APPL_ACL_TABLE = "ACL_TABLE_TABLE"
     STATE_ACL_TABLE = "ACL_TABLE_TABLE"
@@ -121,6 +122,7 @@ class AclLoader(object):
         self.mirror_stage = None
         self.current_table = None
         self.tables_db_info = {}
+        self.tables_type_info = {}
         self.rules_db_info = {}
         self.rules_info = {}
         self.tables_state_info = None
@@ -202,6 +204,17 @@ class AclLoader(object):
                     else:
                         self.tables_db_info[table]['ports'] += entry.get(
                             'ports', [])
+
+        # Update user defined acl table types
+        host_acl_table_type = self.configdb.get_table(self.ACL_TABLE_TYPE)
+        if self.per_npu_configdb:
+            for ns, config_db in self.per_npu_configdb.items():
+                acl_table_type = config_db.get_table(self.ACL_TABLE_TYPE)
+                for table_type, entry in acl_table_type.items():
+                    if table not in self.tables_type_info:
+                        self.tables_type_info[table_type] = entry
+        else:
+            self.tables_type_info.update(host_acl_table_type)
 
         if self.per_npu_configdb:
             # Note: Ability to read table information from APPL_DB is not yet supported for masic devices
@@ -455,6 +468,16 @@ class AclLoader(object):
         :return: True if table type is ACL_TABLE_TYPE_CTRLPLANE else False
         """
         return self.tables_db_info[tname]['type'].upper() == self.ACL_TABLE_TYPE_CTRLPLANE
+
+    def acl_table_has_match(self, tname, match):
+        """
+        """
+        table_type = self.tables_db_info[tname]["type"]
+        # Predefined table types aren't defined in the tables_type_info
+        if table_type not in self.tables_type_info:
+            return True
+        else:
+            return match in self.tables_type_info[table_type]["MATCHES"]
 
     @staticmethod
     def parse_acl_json(filename):
@@ -795,7 +818,7 @@ class AclLoader(object):
         deep_update(rule_props, self.convert_transport(table_name, rule_idx, rule))
         deep_update(rule_props, self.convert_input_interface(table_name, rule_idx, rule))
 
-        if "IP_PROTOCOL" in rule_props and "IP_TYPE" not in rule_props:
+        if "IP_PROTOCOL" in rule_props and "IP_TYPE" not in rule_props and self.acl_table_has_match(table_name, "IP_TYPE"):
             # If we don't include IP_TYPE as a qualifier in the IP_PROTOCOL rule
             # we could match on non-IP packets if the bits at the same offset match
             # https://github.com/sonic-net/sonic-mgmt/issues/23960
--- a/acl_loader/main.py
+++ b/acl_loader/main.py
@@ -818,7 +818,8 @@ class AclLoader(object):
         deep_update(rule_props, self.convert_transport(table_name, rule_idx, rule))
         deep_update(rule_props, self.convert_input_interface(table_name, rule_idx, rule))
 
-        if "IP_PROTOCOL" in rule_props and "IP_TYPE" not in rule_props and self.acl_table_has_match(table_name, "IP_TYPE"):
+        if ("IP_PROTOCOL" in rule_props and "IP_TYPE" not in rule_props
+            and self.acl_table_has_match(table_name, "IP_TYPE")):
             # If we don't include IP_TYPE as a qualifier in the IP_PROTOCOL rule
             # we could match on non-IP packets if the bits at the same offset match
             # https://github.com/sonic-net/sonic-mgmt/issues/23960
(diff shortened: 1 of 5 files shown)


AnantKishorSharma (commented):
The negative path — a custom `ACL_TABLE_TYPE` whose `MATCHES` omits `IP_TYPE` ⇒ no injection — is covered by the new `test_noiptype_custom_acl_table_type`, and the predefined-type path (`IP_TYPE` still injected) stays covered by the existing tests. One gap: the positive custom-type case — a custom type whose `MATCHES` *includes* `IP_TYPE` ⇒ `IP_TYPE` still injected — isn't tested; adding it would lock down that the new gate can't over-skip. Also, the multi-ASIC `tables_type_info` population flagged inline isn't asserted by any test (`TestMasicAclLoader` doesn't check it).

acl_loader/main.py
```diff
@@ -203,6 +205,17 @@ def read_tables_info(self):
                         self.tables_db_info[table]['ports'] += entry.get(
                             'ports', [])
 
+        # Update user defined acl table types
+        host_acl_table_type = self.configdb.get_table(self.ACL_TABLE_TYPE)
+        if self.per_npu_configdb:
+            for ns, config_db in self.per_npu_configdb.items():
+                acl_table_type = config_db.get_table(self.ACL_TABLE_TYPE)
+                for table_type, entry in acl_table_type.items():
+                    if table not in self.tables_type_info:
```
AnantKishorSharma: `table` here is the leftover loop variable from the `ACL_TABLE` loop just above (line 199) — this looks like it should be `table_type`. As written the guard keys off a stale table *name* (essentially never a key in `tables_type_info`, since that dict is keyed by table *type*), so the de-dup never takes effect: on multi-ASIC the last namespace's definition overwrites earlier ones instead of the first one winning.

Separately, `host_acl_table_type` (line 209) is computed but only used in the `else` branch — in this multi-ASIC path it's discarded. Is that intended, or should host-defined types be merged here too?

arista-nwolfe: Great catch! Thanks for catching this!

Regarding the `host_acl_table_type` I tried to replicate what is being done above with the handling of the non-control plane ACLs, which is use the per-namespace entries on multi-asic and use the global namespace entries on single-asic. However, I do agree it looks weird to define `host_acl_table_type` outside of the conditional given it's only used in the `else`. I'll move the definition to the `else` clause.

arista-nwolfe:
> The negative path — a custom `ACL_TABLE_TYPE` whose `MATCHES` omits `IP_TYPE` ⇒ no injection — is covered by the new `test_noiptype_custom_acl_table_type`, and the predefined-type path (`IP_TYPE` still injected) stays covered by the existing tests. One gap: the positive custom-type case — a custom type whose `MATCHES` _includes_ `IP_TYPE` ⇒ `IP_TYPE` still injected — isn't tested; adding it would lock down that the new gate can't over-skip. Also, the multi-ASIC `tables_type_info` population flagged inline isn't asserted by any test (`TestMasicAclLoader` doesn't check it).

The positive custom-type case is already covered by the following tests
test_icmp_translation_in_custom_acl_table_type
test_icmpv6_translation_in_custom_acl_table_type
https://github.com/sonic-net/sonic-utilities/blob/master/tests/acl_loader_test.py#L208-L232

AnantKishorSharma:
> > The negative path — a custom `ACL_TABLE_TYPE` whose `MATCHES` omits `IP_TYPE` ⇒ no injection — is covered by the new `test_noiptype_custom_acl_table_type`, and the predefined-type path (`IP_TYPE` still injected) stays covered by the existing tests. One gap: the positive custom-type case — a custom type whose `MATCHES` _includes_ `IP_TYPE` ⇒ `IP_TYPE` still injected — isn't tested; adding it would lock down that the new gate can't over-skip. Also, the multi-ASIC `tables_type_info` population flagged inline isn't asserted by any test (`TestMasicAclLoader` doesn't check it).
> 
> The positive custom-type case is already covered by the following tests test_icmp_translation_in_custom_acl_table_type test_icmpv6_translation_in_custom_acl_table_type https://github.com/sonic-net/sonic-utilities/blob/master/tests/acl_loader_test.py#L208-L232

You're right, `test_icmp[v6]_translation_in_custom_acl_table_type` already cover the positive custom-type case; the `c1af0f16` typo fix + cleanup look good, thanks!

AnantKishorSharma (approved):
`TestMasicAclLoader` doesn't assert the multi-asic `tables_type_info` first-ns-wins merge — and since the asic mocks don't define `ACL_TABLE_TYPE` it'd need seed data, so fine to defer to a follow-up.

yizhenzha:
@arista-nwolfe @AnantKishorSharma I have validated this patch and confirmed it solved the issue https://github.com/sonic-net/sonic-buildimage/issues/28028

bingwang-ms:
@arista-nwolfe Thanks for fixing the issue. PR test is failing, can you help take a look?

arista-nwolfe:
> @arista-nwolfe Thanks for fixing the issue. PR test is failing, can you help take a look?

Thanks for retriggering the checks @bingwang-ms looks like they are passing now if you're able to merge this.

bingwang-ms (approved):

rookie-who:
This PR has the "Request for 202605 Branch" label but is missing 202605-specific test results. Please:
1. Tick the **202605** checkbox in the "Tested branch" section of the PR description
2. Provide 202605 branch test evidence in the "Test result" section
3. Add the label **"Tested for 202605 branch"** once done

Master branch test results alone are not sufficient for 202605 cherry-pick approval.

*Comment on behalf of @vaibhavhd*

vmittal-msft:
@arista-nwolfe can we add label "tested on 202511" once done ?

PR #4660, opened 2026-07-01
[setup]: Pin docker-image-py to 0.1.13

lunyue-ms:
#### What I did

`docker-image-py` was unpinned (`>=0.1.10`). The 0.2.0 release (2026-06-29) breaks
`sonic-utilities`: its `split_docker_domain()` mis-parses `Azure/sonic` (failing
`test_get_registry_for` and corrupting registry paths), and it requires
`regex>=2026.6.28`, which breaks the wheel install against the apt `regex`
(`uninstall-no-record-file`). Pin to `==0.1.13` to keep the pre-0.2.0 behavior.

Minimal alternative to #4655 (which makes the code 0.2.0-compatible instead) —
merge only one.

#### How I did it

- `setup.py`: `'docker-image-py>=0.1.10'` -> `'docker-image-py==0.1.13'`.

#### How to verify it

```
pytest tests/sonic_package_manager/test_registry.py -v
```
passes; pip resolves `docker-image-py` to 0.1.13 (not 0.2.0).

N/A

N/A

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -249,7 +249,7 @@ setup(
         'urllib3>=2',
         'click-log>=0.3.2',
         'docker>=4.4.4',
-        'docker-image-py>=0.1.10',
+        'docker-image-py==0.1.13',
         'filelock>=3.0.12',
         'enlighten>=1.8.0',
         'ijson>=3.2.3',


setup.py
```diff
@@ -249,7 +249,7 @@
         'urllib3>=2',
         'click-log>=0.3.2',
         'docker>=4.4.4',
-        'docker-image-py>=0.1.10',
+        'docker-image-py>=0.1.10,<0.2.0',
```
roeebar-arista: Why not `docker-image-py==0.1.13`? That is how it is pinned in sonic-buildimage, and we don't know what other versions of docker-image-py might be released.

lunyue-ms: It seems that there is no version between 0.1.13 and 0.2.0.  Yes, docker-image-py==0.1.13 will be more accurate to pin the version.

roeebar-arista (approved):

yijingyan2 (approved):

roeebar-arista:
@yijingyan2 Can you please also add this to 202511?

justin-wong-ce:
@vaibhavhd @vmittal-msft could you help take a look and see if the cherry-pick job be triggered for this PR please?

lunyue-ms:
https://github.com/sonic-net/sonic-utilities/pull/4655 fixed the unit test issue when using 0.2.0 docker-image-py. It also needs to be added for 202511

PR #4663, opened 2026-07-02
[hft] Target per-key path in JSON Patch ops to avoid wiping the table

DavidZagury:
#### What I did
Fix `config hft add profile` / `config hft add group` so the emitted
JSON Patch no longer wipes pre-existing rows in
`HIGH_FREQUENCY_TELEMETRY_PROFILE` /
`HIGH_FREQUENCY_TELEMETRY_GROUP`.

#### How I did it
Changed _build_profile_patch and _build_group_patch in
config/hft.py to target the row's per-key path
(/HIGH_FREQUENCY_TELEMETRY_PROFILE/<name> or
/HIGH_FREQUENCY_TELEMETRY_GROUP/<profile>|<group>) with the row's
attributes as the value, instead of the table-level path with a
single-entry value dict. This matches what _build_stream_state_patch
and the _build_*_remove_patch helpers in the same file already do.

Updated the two CLI unit tests in tests/config_hft_test.py that
asserted the old patch shape.

No CLI surface or CONFIG_DB schema change.

#### How to verify it
- pytest tests/config_hft_test.py — both updated assertions pass.
- Manual smoke against a running SONiC switch:

config hft add profile P --poll_interval 10000
config hft add group P --group_type PORT  --object_names Ethernet0
config hft add group P --group_type QUEUE --object_names Ethernet0
sonic-cfggen -d -v 'HIGH_FREQUENCY_TELEMETRY_GROUP'

- After the fix both P|PORT and P|QUEUE are present; before the fix
only P|QUEUE remained.

N/A - no user-facing CLI output change.

N/A - no user-facing CLI output change.

### DIFF
--- a/config/hft.py
+++ b/config/hft.py
@@ -171,20 +171,34 @@ def _split_csv_items(value):
 
 
 def _build_profile_patch(op, profile_name, attributes):
-    """Construct a JSON Patch entry targeting the profile table."""
+    """Construct a JSON Patch entry targeting a single profile row.
+
+    The patch must target the per-key path (e.g. /HIGH_FREQUENCY_TELEMETRY_PROFILE/foo)
+    rather than the table-level path (/HIGH_FREQUENCY_TELEMETRY_PROFILE), because per
+    RFC 6902 an "add" at an existing object member replaces that member's value -
+    which at the table level would wipe every other row.
+    """
     return [
-        _build_patch_entry(op, PROFILE_TABLE_PATH, {
-            profile_name: attributes
-        })
+        _build_patch_entry(
+            op,
+            _join_pointer(PROFILE_TABLE_PATH, profile_name),
+            attributes,
+        )
     ]
 
 
 def _build_group_patch(op, profile_name, group_type, attributes):
-    """Construct a JSON Patch entry targeting the group table."""
+    """Construct a JSON Patch entry targeting a single group row.
+
+    See _build_profile_patch for why this targets the per-key path rather than
+    the table path.
+    """
     return [
-        _build_patch_entry(op, GROUP_TABLE_PATH, {
-            _compose_group_key(profile_name, group_type): attributes
-        })
+        _build_patch_entry(
+            op,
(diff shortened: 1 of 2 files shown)


Yakiv-Huryk (dismissed):

Pterosaur (changes requested):
The row-level JSON Patch change is correct, but it makes it possible to add or replace a group without removing sibling rows while the profile may still be running. We should not implicitly reconfigure a live HFT stream.

Please reject any operation that adds, updates, or deletes a group while the profile has an active runtime stream. The operator must explicitly run `config hft disable <profile>` and wait until the stream is fully stopped before changing the group configuration. Checking only `CONFIG_DB.stream_state` is not sufficient because it can become `disabled` before orchagent has completed the runtime transition; please validate the operational state (for example, the corresponding STATE_DB sessions) or provide an equivalent synchronization mechanism.

Please also add tests showing that group add/update/delete is rejected while the stream is active and succeeds only after it is stopped. This is especially important in MIXED mode, where multiple groups share the same tel_type/report resources.

DavidZagury:
> The row-level JSON Patch change is correct, but it makes it possible to add or replace a group without removing sibling rows while the profile may still be running. We should not implicitly reconfigure a live HFT stream.
> 
> Please reject any operation that adds, updates, or deletes a group while the profile has an active runtime stream. The operator must explicitly run `config hft disable <profile>` and wait until the stream is fully stopped before changing the group configuration. Checking only `CONFIG_DB.stream_state` is not sufficient because it can become `disabled` before orchagent has completed the runtime transition; please validate the operational state (for example, the corresponding STATE_DB sessions) or provide an equivalent synchronization mechanism.
> 
> Please also add tests showing that group add/update/delete is rejected while the stream is active and succeeds only after it is stopped. This is especially important in MIXED mode, where multiple groups share the same tel_type/report resources.

@Pterosaur please check the new commits

Yakiv-Huryk (approved):

PR #4671, opened 2026-07-09
Fix circular dependency when creating new MSTP instance via CLI

qianqian-33:
#### What I did
Fixed MSTP CLI circular dependency bug in config/stp.py.
The original CLI required MST instance to exist before setting priority/vlan mapping, but no create command was available, blocking users from creating new MST instances.
#### How I did it
Modified the pre-check logic for config spanning-tree mst instance priority and config spanning-tree mst instance vlan add;
Auto-create MST_INSTANCE entry in CONFIG_DB STP_MST_INST table when the target instance does not exist;
Keep invalid instance ID validation to prevent invalid DB records;
Maintain backward compatibility for existing configured MST instances.
#### How to verify it
Test new empty MST instance: run sudo config spanning-tree mst instance vlan add 1 10, command succeeds and auto-generates MST_INSTANCE DB record;
Test existing MST instance: modify its priority/vlan mapping, the command works normally without error;
Execute sonic-utilities unit tests to confirm no regression.
sudo config spanning-tree mst instance vlan add 1 10
Error: MST instance 1 does not exist. Please create it first.

sudo config spanning-tree mst instance priority 1 32768
Error: MST instance 1 does not exist. Please create it first.
sudo config spanning-tree mst instance vlan add 1 10
# Command executes successfully, MST_INSTANCE|1 entry auto created in CONFIG_DB

sudo config spanning-tree mst instance priority 1 32768
# Command executes successfully, updates priority of MST instance 1

### DIFF
--- a/clear/stp.py
+++ b/clear/stp.py
@@ -17,7 +17,7 @@ def spanning_tree(ctx):
 @click.pass_context
 def stp_clr_stats(ctx):
     if ctx.invoked_subcommand is None:
-        command = 'sudo stpctl clrstsall'
+        command = ["sudo","docker","exec","stp","stpctl","clrstsall"]
         clicommon.run_command(command)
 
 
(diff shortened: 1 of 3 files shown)


qianqian-33:
Restart based on master

Issue #4629, opened 2026-06-18
[STP/MSTP] Circular dependency validation bug in `config spanning-tree mst instance` soft-locks instance creation


mbze430:
Component:
sonic-utilities / config/stp.py

SONiC Version/Build info:

SONiC Software Version: SONiC.202511.0-dirty-20260506.192100

Description:
There is a hard-coded circular dependency bug in the Click CLI configuration wrapper for Multiple Spanning Tree Protocol (MSTP) settings. The validation logic inside config spanning-tree mst instance vlan add and config spanning-tree mst instance priority strictly checks if an MST instance exists in the configuration database (CONFIG_DB) before allowing modifications. However, the CLI entirely lacks a command block (such as an add or create verb) under the parent instance group to initialize a new instance key. As a result, users are completely soft-locked from deploying multi-instance STP layouts via the standard CLI.

Steps to Reproduce:
Attempt to add a VLAN to a new MST instance:

Bash
sudo config spanning-tree mst instance vlan add 1 10
Output: Error: MST instance 1 does not exist. Please create it first.

Attempt to initialize the instance by setting its bridge priority:

Bash
sudo config spanning-tree mst instance priority 1 32768
Output: Error: MST instance 1 does not exist. Please create it first.

Inspect the subcommands available under config spanning-tree mst instance:

Bash
sudo config spanning-tree mst instance
Output:

Commands:
  interface  Configure MSTP instance interface settings
  priority   Configure bridge priority for an MST instance.
  vlan       VLAN to instance mapping for MST.
(Note the complete absence of a create or add verb to initialize the container).

Root Cause Analysis:
In /usr/local/lib/python3.13/dist-packages/config/stp.py, the validation blocks for both the priority and vlan commands enforce an explicit entry check against the STP_MST_INST table:

Python
# Line 1786-1788 in config/stp.py
instance_key = f"MST_INSTANCE|{instance_id}"
if not db.get_entry('STP_MST_INST', instance_key):
    ctx.fail(f"MST instance {instance_id} does not exist. Please create it first.")
Because db.get_entry() looks for a key that cannot be generated anywhere else via the CLI, the execution path hits a dead end.

dlanov:
Update: I found the existing PR #4672 after a broader search. I will not open a competing PR for this issue.

During the investigation I confirmed the circular dependency and found several additional test and key-consistency gaps. I will share those findings on #4672 or track the separate key issue independently.

PR #3848, opened 2025-04-14
[MSTP] CLI Support

wajahatrazi:
#### What I did

- Implemented CLI utilities for MSTP (Multiple Spanning Tree Protocol) in SONiC

#### How I did it

- Introduced new configuration and show commands to manage MSTP at the global, region, instance, and interface levels.
- Updated the configuration database schema to support MSTP, including new tables and additions to the existing STP_GLOBAL_TABLE.

PR #4672, opened 2026-07-09
[utilities]: Fix circular dependency when creating new MSTP instance

qianqian-33:
…via CLI

Fix deadlock that users cannot create MST instances, since vlan subcmds require pre-existing instance while no create CLI is provided. The fix auto initializes MST_INSTANCE DB entry when needed, compatible with old configs.

#### What I did
Fixed MSTP CLI circular dependency bug in config/stp.py.
The original CLI required MST instance to exist before setting priority/vlan mapping, but no create command was available, blocking users from creating new MST instances.
#### How I did it
Modified the pre-check logic for config spanning-tree mst instance priority and config spanning-tree mst instance vlan add;
Auto-create MST_INSTANCE entry in CONFIG_DB STP_MST_INST table when the target instance does not exist;
Keep invalid instance ID validation to prevent invalid DB records;
Maintain backward compatibility for existing configured MST instances.
#### How to verify it
Test new empty MST instance: run sudo config spanning-tree mst instance vlan add 1 10, command succeeds and auto-generates MST_INSTANCE DB record;
Test existing MST instance: modify its priority/vlan mapping, the command works normally without error;
Execute sonic-utilities unit tests to confirm no regression.
sudo config spanning-tree mst instance vlan add 1 10
Error: MST instance 1 does not exist. Please create it first.

sudo config spanning-tree mst instance priority 1 32768
Error: MST instance 1 does not exist. Please create it first.
sudo config spanning-tree mst instance vlan add 1 10
Command executes successfully, MST_INSTANCE|1 entry auto created in CONFIG_DB
sudo config spanning-tree mst instance priority 1 32768
Command executes successfully, updates priority of MST instance 1

### DIFF
--- a/config/stp.py
+++ b/config/stp.py
@@ -1817,7 +1817,7 @@ def mst_instance_vlan_add(_db, instance_id, vlan_id):
     db = _db.cfgdb
 
     # Validate instance_id range
-    if not (0 < instance_id < MST_MAX_INSTANCES):
+    if not (0 <= instance_id < MST_MAX_INSTANCES):
         ctx.fail(f"Instance ID must be in range 0-{MST_MAX_INSTANCES - 1}")
     # Disallow VLAN configuration for MST instance 0
     if instance_id == 0:
@@ -1825,8 +1825,8 @@ def mst_instance_vlan_add(_db, instance_id, vlan_id):
 
     # Check if instance exists
     instance_key = f"MST_INSTANCE|{instance_id}"
-    if not db.get_entry('STP_MST_INST', instance_key):
-        ctx.fail(f"MST instance {instance_id} does not exist. Please create it first.")
+    # if not db.get_entry('STP_MST_INST', instance_key):
+    #     ctx.fail(f"MST instance {instance_id} does not exist. Please create it first.")
 
     # Validate VLAN ID range
     if not (1 <= vlan_id <= 4094):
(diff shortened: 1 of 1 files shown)


dlanov:
Thanks for working on #4629. I was investigating the same issue and found a few points that may help complete this PR:

- The current diff changes the VLAN add/delete paths, but the priority command still rejects a missing MST instance even though the PR description says that path is fixed.
- Focused regression tests do not appear to be included yet.
- Deleting the last VLAN removes the whole STP_MST_INST entry, which also discards any configured bridge_priority.
- Current master initializes instance 0 with MST_INSTANCE:INSTANCE0, while the priority and VLAN paths look up MST_INSTANCE|<id>; this appears to be a separate key-format inconsistency.

Happy to help with focused tests or prepare a separate follow-up for the instance-0 key mismatch.

mbze430:
@wajahatrazi Pinging you directly since you authored the original MSTP CLI implementation in config/stp.py (PR #3848). We are dead-in-the-water here. The circular dependency validation bug is actively soft-locking MST instance creation on our Tomahawk hardware (Celestica DX010). This PR (#4672) fixes it explicitly. Can you get eyes on this diff and stamp an approval so we can get this merged?

wajahatrazi:
Hi @mbze430 can you we schedule a call and discuss your issue in detail?

mbze430:
@wajahatrazi We need someone to review this fix so it can be merged, I pinged you because you are the inital dev that created stp.py  Otherwise, this PR fixes the problem.  If that is not enough information I am more than happy to voice chat

wajahatrazi:
Let me review it, and I will add comments if any. @mbze430 Thanks

PR #2807, opened 2023-04-24
sonic-utilities: WRED stats feature changes on sonic-utilities

rpmarvell:
#### What I did
CLI support for WRED and ECN statistics feature

#### How I did it
* New script for wredstat CLI commands
* portstat script updated to accomodate WRED port stats
* counterpoll script updated to support wredport and wredqueue counters
* CLi to script mapping changes
* UT for the new script changes

#### How to verify it
sonic-utilities UT and Marvell DUT manual testing

#### Previous command output (if the output of a command-line utility has changed)
root@sonic-dut:~# show interfaces counters detailed Ethernet8
Packets Received 64 Octets..................... 0
Packets Received 65-127 Octets................. 2
Packets Received 128-255 Octets................ 0
Packets Received 256-511 Octets................ 0
Packets Received 512-1023 Octets............... 0
Packets Received 1024-1518 Octets.............. 0
Packets Received 1519-2047 Octets.............. 0
Packets Received 2048-4095 Octets.............. 0
Packets Received 4096-9216 Octets.............. 0
Packets Received 9217-16383 Octets............. 0

Total Packets Received Without Errors.......... 2
Unicast Packets Received....................... 0
Multicast Packets Received..................... 2
Broadcast Packets Received..................... 0

Jabbers Received............................... N/A
Fragments Received............................. N/A
Undersize Received............................. 0
Overruns Received.............................. 0

Packets Transmitted 64 Octets.................. 32,893
Packets Transmitted 65-127 Octets.............. 16,449
Packets Transmitted 128-255 Octets............. 3
Packets Transmitted 256-511 Octets............. 2,387
Packets Transmitted 512-1023 Octets............ 0
Packets Transmitted 1024-1518 Octets........... 0
Packets Transmitted 1519-2047 Octets........... 0
Packets Transmitted 2048-4095 Octets........... 0
Packets Transmitted 4096-9216 Octets........... 0
Packets Transmitted 9217-16383 Octets.......... 0

Total Packets Transmitted Successfully......... 51,732
Unicast Packets Transmitted.................... 0
Multicast Packets Transmitted.................. 18,840
Broadcast Packets Transmitted.................. 32,892
Time Since Counters Last Cleared............... None

#### New command output (if the output of a command-line utility has changed)
root@sonic-dut:~# show interfaces counters detailed Ethernet8
Packets Received 64 Octets..................... 0
Packets Received 65-127 Octets................. 2
Packets Received 128-255 Octets................ 0
[...]

PR #4676, opened 2026-07-10
Remove unused STATE_DB connection from wredstat get_counters()

asechoud:
## Summary
- Remove per-queue `SonicV2Connector` + `STATE_DB` connect from `wredstat.get_counters()`.
- The connector was introduced in #2807 but never used: all WRED counter values are read from `COUNTERS_DB` via `self.db`.

## Problem
On multi-ASIC platforms, `wredstat -c` / show walks every queue and opened a new TCP Redis connection to `STATE_DB` per queue. With hundreds of VoQ ports and multiple namespaces this can exhaust ephemeral ports and fail with:

```
RuntimeError: Unable to connect to redis - Cannot assign requested address
```

## Note on portstat
PR #2807 also added legitimate `STATE_DB` reads in `portstat` (WRED capability flags via `self.db.get(STATE_DB, ...)` once per collect). That path is unchanged. `wredstat` does not perform those checks and does not need a separate connector per queue.

## Test plan
- [x] Existing `tests/wred_queue_counter_test.py` coverage unchanged (no `state_db` dependency)
- [ ] `wredstat` on multi-ASIC DUT: `wredstat -c` completes without Redis connection errors
- [ ] `wredstat` / `wredstat -j` output matches pre-change counter values

### DIFF
--- a/scripts/wredstat
+++ b/scripts/wredstat
@@ -193,8 +193,6 @@ class Wredstat(object):
             fields = ["0","0","0","0","0","0"]
             fields[0] = get_queue_index(table_id)
             fields[1] = get_queue_type(table_id)
-            self.state_db = SonicV2Connector(use_unix_socket_path=False)
-            self.state_db.connect(self.state_db.STATE_DB)     
 
             for counter_name, pos in counter_bucket_dict.items():
                 full_table_id = COUNTER_TABLE_PREFIX + table_id


scripts/wredstat
```diff
@@ -193,8 +193,6 @@ class Wredstat(object):
             fields = ["0","0","0","0","0","0"]
             fields[0] = get_queue_index(table_id)
             fields[1] = get_queue_type(table_id)
-            self.state_db = SonicV2Connector(use_unix_socket_path=False)
```
zhixzhu, line 196: Also delete the now-unused SonicV2Connector import.

asechoud:
Re-pushed commit as **asechoud only** (removed \`Co-authored-by: Cursor\` trailer) to satisfy EasyCLA.

Re: removing \`SonicV2Connector\` import — it is still used in \`Wredstat.__init__\` for \`self.db\` when connecting to COUNTERS_DB, so the import stays. Only the per-queue unused \`self.state_db\` connect inside \`get_counters()\` was removed.

saravanan-nexthop:
@venky-nexthop for viz

abdosi:
@vmittal-msft : please review this.

PR #4685, opened 2026-07-14
[console] Add new commands for Console Logging feature in CONSOLE_PORT

nats-nokia:
#### What I did
Add new commands to enable, disable console logging. optional command to modify the filename, logrotate parameters for the filename.

Related PR from other sub modules:
https://github.com/sonic-net/sonic-buildimage/pull/28411
https://github.com/sonic-net/sonic-host-services/pull/409

#### How I did it
Add new commands: config console logging <line> {enable/disable}. 
optional command: config console logging <line> filename <path> [--logrotate-size SIZE] [--logrotate-count COUNT]
If filename,logrorate parameters are not configured, default values (/var/log/console-<line>.log, 10M size, 10 count) will be used.

#### How to verify it
Manually tested on physical testbed.

### DIFF
--- a/config/console.py
+++ b/config/console.py
@@ -1,5 +1,6 @@
 import click
 import string
+import re
 import utilities_common.cli as clicommon
 from .validated_config_db_connector import ValidatedConfigDBConnector
 from jsonpatch import JsonPatchConflict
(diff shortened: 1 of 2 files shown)


config/console.py
```diff
@@ -310,6 +311,140 @@ def update_console_escape_char(db, linenum, escape):
...
+    _set_console_logging_state(linenum, db, "no")
+
+
+@console_logging.command('filename')
+@clicommon.pass_db
+@click.argument('filename', metavar='<file_name>', required=True)
+@click.argument('logrotate_keyword', metavar='logrotate', required=False)
+@click.argument('size', metavar='<size>', required=False)
+@click.argument('count', metavar='<count>', required=False, type=click.IntRange(1, 100))
+def set_console_logging_filename(db, filename, logrotate_keyword, size, count):
```
lucas-li-micas: Would it be possible to use click options for logrotate parameters instead of positional arguments?
Since logrotate has multiple optional fields, options may provide better extensibility and avoid ambiguous CLI parsing.

config/console.py
```diff
@@ -310,6 +311,139 @@ def update_console_escape_char(db, linenum, escape):
...
+
+
+@console_logging.command('filename')
+@clicommon.pass_db
+@click.argument('filename', metavar='<file_name>', required=True)
+@click.option('--logrotate-size', '-s', 'logrotate_size', metavar='<size>', default=None,
+              help='Logrotate size threshold (e.g. 10M, 100k). Default: 10M.')
+@click.option('--logrotate-count', '-c', 'logrotate_count', metavar='<count>', default=None,
+              type=click.IntRange(1, 100), help='Number of rotated log files to retain. Default: 10.')
+def set_console_logging_filename(db, filename, logrotate_size, logrotate_count):
```
ediwibowo-msft, line 398: `set_console_logging_filename` accepts any string for `filename`. A value containing whitespace, newlines, or `{/}` could inject arbitrary logrotate directives (e.g. `postrotate`/`firstaction`, which execute shell as root). The CLI should validate `filename`: require an absolute path, reject newlines/whitespace/control chars and `{}`.

config/console.py
```diff
@@ -310,6 +311,139 @@ def update_console_escape_char(db, linenum, escape):
         ctx.fail("Trying to update console port setting, which is not present.")
 
 
+#
+# 'console logging' group ('config console logging ...')
+#
+LOGROTATE_SIZE_PATTERN = re.compile(r'^[0-9]+[kKmMgG]?$')
```
ediwibowo-msft: `LOGROTATE_SIZE_PATTERN = r'^[0-9]+[kKmMgG]?$'` accepts e.g. `10m`, `10g`. logrotate's size directive expects M, G as per logic in `config.c`:
```
if (key[l] == 'k' || key[l] == 'K') { multiplier = 1024; }
else if (key[l] == 'M')            { multiplier = 1024*1024; }
else if (key[l] == 'G')            { multiplier = 1024*1024*1024; }
else if (!isdigit(key[l])) {
    message(MESS_ERROR, "%s:%d unknown unit '%c'\n", ...);   // rejects the config
} else { multiplier = 1; }
```

Therefore, lowercase `m` and `g` can be misinterpreted or rejected at rotate time.

The accurate fix is to drop lowercase m/g:
```
LOGROTATE_SIZE_PATTERN = re.compile(r'^[0-9]+[kKMG]?$')
```

nats-nokia: Removed 'm' and 'g' options form logrotate_size.

config/console.py
nats-nokia: Sure, absolute path is validated for logrotate filename. Thanks

ediwibowo-msft (approved):
LGTM

PR #4687, opened 2026-07-14
[show]: Fix 'show storm-control -n <namespace>' crash on multi-ASIC platforms

vvolam:
Related issue: sonic-net/sonic-buildimage#28419

#### What I did
Fixed `show storm-control -n <namespace>` on multi-ASIC platforms. There were
two related problems:

1. **Group path crashed.** `show storm-control -n <ns>` raised
   `AttributeError: module 'sonic_py_common.multi_asic' has no attribute 'multi_asic_get_ip_intf_from_ns'`,
   so storm-control could never be shown for a namespace.
2. **`interface` subcommand ignored the namespace.** `show storm-control -n <ns> interface <intf>`
   accepted `-n` but always connected to the host `CONFIG_DB`, so on multi-ASIC
   the per-ASIC storm-control configuration was never read (empty output).

#### How I did it
In `show/main.py`:

- Added a helper `connect_storm_control_config_db(namespace)` that connects to
  the host `CONFIG_DB` when `namespace is None`, and otherwise to the per-ASIC
  database via `multi_asic.connect_config_db_for_ns(namespace)`.
- Threaded `namespace` through `get_storm_interface(intf, body, namespace)` and
  `display_storm_interface(intf, namespace)`, both of which now obtain their
  connection from the helper instead of a hard-coded host `ConfigDBConnector()`.
- `storm_control()` now calls `multi_asic_util.multi_asic_get_ip_intf_from_ns()`
  (the helper lives in `utilities_common.multi_asic`, imported as
  `multi_asic_util`, not in `sonic_py_common.multi_asic`) and passes the
  namespace down.
- The `interface` subcommand passes its parent namespace to
  `display_storm_interface()`.

Routing every storm-control DB connection through the single helper keeps the
host and per-ASIC paths covered by tests and removes the dead code that failed
diff-coverage.

Tests in `tests/storm_control_test.py`:

- `test_show_storm_namespace` — exercises the group namespace branch and asserts
  the per-ASIC table renders instead of crashing.
- `test_show_storm_interface_namespace` — asserts
  `show storm-control -n asic0 interface Ethernet0` reads the per-ASIC
  `CONFIG_DB` (distinctive asic0 rates) instead of the host database.

#### How to verify it
- Unit tests:
  ```
  python3 -m pytest tests/storm_control_test.py -v
  ```
  (`test_show_storm`, `test_show_storm_interface`, `test_show_storm_namespace`,
  `test_show_storm_interface_namespace`).
- On a multi-ASIC switch, `show storm-control -n asic0 interface <intf>` prints
  the storm-control table for that ASIC, and `show storm-control -n asic0` no
  longer raises `AttributeError`.

#### Verification results
- **Unit tests:** all storm-control tests pass; static analysis (flake8) and
  diff-coverage checks pass in CI.
- **Real hardware (dual-ASIC platform):** validated end-to-end with
  storm-control configuration present only on `asic0`:
  - *Before* the fix, `show storm-control -n asic0 interface <intf>` printed
    nothing (it read the empty host `CONFIG_DB`).
  - *After* the fix, the same command printed the `asic0` rates, while
    `show storm-control -n asic1 interface <intf>` printed nothing — confirming
    the command reads the correct per-namespace database (per-ASIC isolation).
  - The group path `show storm-control -n asic0` no longer raises
    `AttributeError` and renders the table.

```
$ show storm-control -n asic0
  ...
  File ".../show/main.py", in storm_control
    interfaces = multi_asic.multi_asic_get_ip_intf_from_ns(namespace)
AttributeError: module 'sonic_py_common.multi_asic' has no attribute 'multi_asic_get_ip_intf_from_ns'

$ show storm-control -n asic0 interface Ethernet0
                      # nothing printed, even though asic0 has storm-control config
```

```
$ show storm-control -n asic0 interface Ethernet0
+------------------+-------------------+---------------+
| Interface Name   | Storm Type        |   Rate (kbps) |
+==================+===================+===============+
| Ethernet0        | broadcast         |        111000 |
+------------------+-------------------+---------------+
| Ethernet0        | unknown-unicast   |        222000 |
+------------------+-------------------+---------------+
| Ethernet0        | unknown-multicast |        333000 |
+------------------+-------------------+---------------+
```

### DIFF
--- a/show/main.py
+++ b/show/main.py
@@ -227,16 +227,8 @@ def connect_storm_control_config_db(namespace=None):
 #
 # Get storm-control configurations per interface append to body
 #
-def get_storm_interface(intf, body, namespace=None):
+def get_storm_interface(intf, body, config_db):
     storm_type_list = ['broadcast', 'unknown-unicast', 'unknown-multicast']
-
-    config_db = connect_storm_control_config_db(namespace)
-
-    table = config_db.get_table('PORT_STORM_CONTROL')
-
-    #To avoid further looping below
-    if not table:
-        return
 
     for storm_type in storm_type_list:
         storm_key = intf + '|' + storm_type
@@ -539,9 +531,13 @@ def storm_control(ctx, namespace, display):
         if namespace is None:
             display_storm_all()
         else:
-            interfaces = multi_asic_util.multi_asic_get_ip_intf_from_ns(namespace)
-            for intf in interfaces:
-                get_storm_interface(intf, body, namespace)
+            config_db = connect_storm_control_config_db(namespace)
+            table = config_db.get_table('PORT_STORM_CONTROL')
+            # To avoid further looping below
+            if table:
+                interfaces = multi_asic_util.multi_asic_get_ip_intf_from_ns(namespace)
+                for intf in interfaces:
+                    get_storm_interface(intf, body, config_db)
             click.echo(tabulate(body, header, tablefmt="grid"))
 
 @storm_control.command('interface')
--- a/show/main.py
+++ b/show/main.py
@@ -525,7 +525,7 @@ def storm_control(ctx, namespace, display):
         if namespace is None:
             display_storm_all()
         else:
-            interfaces = multi_asic.multi_asic_get_ip_intf_from_ns(namespace)
+            interfaces = multi_asic_util.multi_asic_get_ip_intf_from_ns(namespace)
             for intf in interfaces:
                 get_storm_interface(intf, body)
             click.echo(tabulate(body, header, tablefmt="grid"))
--- a/show/main.py
+++ b/show/main.py
@@ -209,20 +209,28 @@ def display_storm_all():
 
     click.echo(tabulate(body, header, tablefmt="grid"))
 
-#
-# Get storm-control configurations per interface append to body
-#
-def get_storm_interface(intf, body, namespace=None):
-    storm_type_list = ['broadcast','unknown-unicast','unknown-multicast']
 
+#
+# Connect to the CONFIG_DB that holds storm-control configuration. On
+# multi-ASIC platforms storm-control config lives in the per-ASIC CONFIG_DB,
+# so connect to the requested namespace instead of the default (host) database.
+#
+def connect_storm_control_config_db(namespace=None):
     if namespace is None:
         config_db = ConfigDBConnector()
         config_db.connect()
     else:
-        # On multi-ASIC platforms storm-control config lives in the per-ASIC
-        # CONFIG_DB, so read it from the requested namespace instead of the
-        # default (host) database.
         config_db = multi_asic.connect_config_db_for_ns(namespace)
+    return config_db
+
+
+#
+# Get storm-control configurations per interface append to body
+#
+def get_storm_interface(intf, body, namespace=None):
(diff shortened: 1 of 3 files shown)


show/main.py
```diff
@@ -523,7 +521,7 @@ def storm_control(ctx, namespace, display):
         if namespace is None:
             display_storm_all()
         else:
-            interfaces = multi_asic.multi_asic_get_ip_intf_from_ns(namespace)
+            interfaces = multi_asic_util.multi_asic_get_ip_intf_from_ns(namespace)
```
lucas-li-micas: Could we check whether other callers are still using multi_asic.multi_asic_get_ip_intf_from_ns()? If this API is deprecated, it would be better to migrate all usages to multi_asic_util consistently.

vvolam: Good question — I checked this.

`multi_asic_get_ip_intf_from_ns()` is defined **only** in `utilities_common/multi_asic.py` (line 219); it does not exist in `sonic_py_common.multi_asic` at all. So it isn't a deprecated API — the previous `multi_asic.multi_asic_get_ip_intf_from_ns(...)` call was just referencing the wrong module, which is exactly what raised the `AttributeError`.

Grepping the repo, this line (`show/main.py:526`) is the **only** caller in the codebase — there are no other usages to migrate. The tests already reference the correct `multi_asic_util.multi_asic_get_ip_intf_from_ns`, and the sibling helper `multi_asic_get_ip_intf_addr_from_ns` also lives only in `utilities_common`. So after this fix all usages are consistent on `multi_asic_util`.

tests/storm_control_test.py
```diff
@@ -143,6 +143,21 @@ def test_show_storm_interface(self):
...
+        # with AttributeError because show.main called
+        # multi_asic.multi_asic_get_ip_intf_from_ns (sonic_py_common) instead
+        # of multi_asic_util.multi_asic_get_ip_intf_from_ns
+        # (utilities_common). Ensure the namespace path renders correctly.
+        runner = CliRunner()
+        result = runner.invoke(show.cli.commands["storm-control"], ["-n", "asic0"])
+        print(result.exit_code)
+        print(result.output)
+        assert result.exit_code == 0
+        assert result.output == show_storm_interface_output
```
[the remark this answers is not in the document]
vvolam: Fixed it!

show/main.py
```diff
@@ -209,14 +209,28 @@ def display_storm_all():
 
     click.echo(tabulate(body, header, tablefmt="grid"))
 
+
+#
+# Connect to the CONFIG_DB that holds storm-control configuration. On
+# multi-ASIC platforms storm-control config lives in the per-ASIC CONFIG_DB,
+# so connect to the requested namespace instead of the default (host) database.
+#
+def connect_storm_control_config_db(namespace=None):
```
gechiang, line 218: This creates a new CONFIG_DB connection every time get_storm_interface() or display_storm_interface() is called. The namespace group path therefore reconnects once per interface, which can be expensive on systems with many ports and may leave connections open. Consider connecting once in storm_control() and passing the connection into the helper, or explicitly closing the connector after use

vvolam: Fixed this in latest commit. Thank you

gechiang (approved):
LGTM

PR #4691, opened 2026-07-15
Secure boot installation

elybarneams:


elybarneams:
Closing as an empty/stray duplicate. The Secure Boot db-certificate work is now tracked in #4700 (enroll) and #4701 (prune).

PR #4696, opened 2026-07-15
[Mellanox] Add `config sdk nvidia-bluefield sdk techsupport-ct-dump <enabled|disabled>` command.

judsonwilson-nvidia:
**Depends on:**

 [sonic-sairedis#1993](https://github.com/sonic-net/sonic-sairedis/pull/1993) (profile parsing), [sonic-sairedis#1994](https://github.com/sonic-net/sonic-sairedis/pull/1994) (sentinel cleanup at syncd startup).

#### Why I did it
SDK Connection Table dumps in techsupport on NVIDIA BlueField DPUs need a runtime enable/disable control. A sentinel file in the syncd container signals whether `generate_dump` should pass `CT_TABLE_DUMP_ENABLE=true` to `saisdkdump` during techsupport collection.

##### Work item tracking
- Microsoft ADO **(number only)**: N/A

#### How I did it
- Add `config platform nvidia-bluefield sdk techsupport-ct-dump {enabled|disabled}` to create/remove sentinel file `sdk-techsupport-ct-dump.enabled` in syncd.
- Update `collect_nvidia_bluefield()` in `generate_dump` to check the sentinel and run `saisdkdump -p` with `CT_TABLE_DUMP_ENABLE=true` when enabled.
- Add unit tests in `tests/config_nvidia_bluefield_test.py`.

#### How to verify it
```bash
pytest tests/config_nvidia_bluefield_test.py -k techsupport_ct_dump
config platform nvidia-bluefield sdk techsupport-ct-dump enabled
show techsupport   # on BlueField DPU; confirm CT dump in archive
config platform nvidia-bluefield sdk techsupport-ct-dump disabled
```

- [ ] 201811
- [ ] 201911
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

- [x] master (202605–202611)

Add config platform nvidia-bluefield sdk techsupport-ct-dump command.

### DIFF
--- a/scripts/generate_dump
+++ b/scripts/generate_dump
@@ -2153,9 +2153,41 @@ collect_nvidia_bluefield() {
     local timeout_cmd="timeout --foreground ${TIMEOUT_MIN}m"
     local sai_dump_folder="/root/saisdkdump"
     local sai_dump_filename="${sai_dump_folder}/sai_sdk_dump_$(date +"%m_%d_%Y_%I_%M_%p")"
+    # SDK techsupport CT dump sentinel path. Keep in sync across:
+    #   config/plugins/nvidia_bluefield.py (SDK_TECHSUPPORT_CT_DUMP_SENTINEL)
+    #   scripts/generate_dump (sdk_techsupport_ct_dump_sentinel)
+    #   sonic-sairedis/syncd/scripts/syncd_init_common.sh (config_syncd_nvidia_bluefield)
+    local sdk_techsupport_ct_dump_sentinel="/var/run/sonic-platform-nvidia-bluefield/sdk-techsupport-ct-dump.enabled"
+    local sdk_techsupport_saisdkdump_profile=""
 
     ${CMD_PREFIX}docker exec syncd mkdir -p $sai_dump_folder
-    ${CMD_PREFIX}docker exec syncd saisdkdump -f $sai_dump_filename
+
+    local prepare_ct_dump_profile_script='
+profile="$1"
+# Start from the running SAI profile if present, otherwise an empty file
+if [ -f /tmp/sai.profile ]; then cp /tmp/sai.profile "$profile"; else : > "$profile"; fi
+# Guard against a source profile with no trailing newline.
+if [ -s "$profile" ] && [ "$(tail -c1 "$profile")" != "" ]; then
+    echo >> "$profile"
+fi
+# Drop any existing CT_TABLE_DUMP_ENABLE=... lines, then append CT_TABLE_DUMP_ENABLE=true.
+sed -i "/^CT_TABLE_DUMP_ENABLE=/d" "$profile"
+echo CT_TABLE_DUMP_ENABLE=true >> "$profile"
+'
+
+    if docker exec syncd test -f $sdk_techsupport_ct_dump_sentinel; then
+        if $NOOP; then
+            echo "docker exec syncd mktemp /tmp/techsupport-saisdkdump.XXXXXX"
+            sdk_techsupport_saisdkdump_profile="/tmp/techsupport-saisdkdump.XXXXXX"
+        else
+            sdk_techsupport_saisdkdump_profile=$(docker exec syncd mktemp /tmp/techsupport-saisdkdump.XXXXXX)
+        fi
+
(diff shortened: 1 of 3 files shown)


croos12 (dismissed):

dgsudharsan (dismissed):

scripts/generate_dump
```diff
@@ -2153,9 +2153,41 @@ collect_nvidia_bluefield() {
...
+# Guard against a source profile with no trailing newline.
+if [ -s "$profile" ] && [ "$(tail -c1 "$profile")" != "" ]; then
+    echo >> "$profile"
+fi
+# Drop any existing CT_TABLE_DUMP_ENABLE=... lines, then append CT_TABLE_DUMP_ENABLE=true.
+sed -i "/^CT_TABLE_DUMP_ENABLE=/d" "$profile"
+echo CT_TABLE_DUMP_ENABLE=true >> "$profile"
+'
+
+    if docker exec syncd test -f $sdk_techsupport_ct_dump_sentinel; then
```
qiluo-msft: bare `docker` here, unlike every other call in this function (`${CMD_PREFIX}docker exec`). so in `--noop`/`-n` this branch probe runs for real and needs syncd up. it's read-only so no mutation, and you do need something to pick the branch — just flagging the dry-run deviation. also: if syncd is unreachable `test` returns non-zero and you silently fall through to the plain dump (error treated as disabled). the config plugin distinguishes those cases; here it's a silent degrade. fine, but worth a note.

scripts/generate_dump
```diff
@@ -2153,9 +2153,41 @@ collect_nvidia_bluefield() {
...
+        if $NOOP; then
+            echo "docker exec syncd mktemp /tmp/techsupport-saisdkdump.XXXXXX"
+            sdk_techsupport_saisdkdump_profile="/tmp/techsupport-saisdkdump.XXXXXX"
+        else
+            sdk_techsupport_saisdkdump_profile=$(docker exec syncd mktemp /tmp/techsupport-saisdkdump.XXXXXX)
+        fi
+
+        ${CMD_PREFIX}docker exec syncd sh -c "$prepare_ct_dump_profile_script" _ "$sdk_techsupport_saisdkdump_profile"
+        ${CMD_PREFIX}docker exec syncd saisdkdump -f $sai_dump_filename -p $sdk_techsupport_saisdkdump_profile
```
qiluo-msft: only saisdkdump's rc feeds the existing `if [ $? != 0 ]` below. if the profile-prep `sh -c` on line 2186 fails, saisdkdump still runs with a partial/empty profile and the failure is invisible. a `|| { echo ...; }` on the prep step would surface it. wdyt?

scripts/generate_dump
judsonwilson-nvidia: > bare docker here, unlike every other call in this function (${CMD_PREFIX}docker exec). so in --noop/-n this branch probe runs for real and needs syncd up. bare docker here, unlike every other call in this function (${CMD_PREFIX}docker exec). so in --noop/-n this branch probe runs for real and needs syncd up.

This is intentional. I believe this is used for dry run, and I think it's better to have the dry run have conditional behavior that matches a real-run, in this case.  This check is relatively short-running compared to the other things in this dump, so it makes sense to do it in dry run.

>  also: if syncd is unreachable test returns non-zero and you silently fall through to the plain dump (error treated as disabled). the config plugin distinguishes those cases; here it's a silent degrade. fine, but worth a note.

If syncd is not up, there is no CT table to dump.  An error message could be nice, but I opted to keep this code simple, partially to reduce the odds that complexity introduces bugs.

scripts/generate_dump
judsonwilson-nvidia: Good catch. Fixed it to run without the profile file in such a case, with an error message.

scripts/generate_dump
judsonwilson-nvidia: Regarding the second one: added an error message in the failure case.

dgsudharsan (approved):

qiluo-msft (approved):

PR #4700, opened 2026-07-17
[sonic-installer] Enroll image db certificate before Secure Boot verification

elybarneams:
#### What I did
Enroll the image's UEFI Secure Boot **db** certificate into the firmware db immediately before `sonic-installer` runs CMS signature verification, so an image signed by a rotated, KEK-authorized db key verifies on install.

- Microsoft ADO (number only): 38692567

#### How I did it
- Add `scripts/secure_boot_enroll_db.sh`, which extracts `boot/DB.auth` from the installer image and appends it to the UEFI db (`efi-updatevar -a`), persisting each shipped cert under `/host/db-auth/DB-<sha256>.auth`.
- Invoke it from the grub bootloader (`enroll_image_secure_boot_keys`) right before `verify_image_sign`; base `Bootloader` keeps a no-op default for other bootloaders.
- Best-effort; firmware still validates the db update against the enrolled KEK, so verification is not weakened.

#### How to verify it
`pytest tests/installer_bootloader_grub_test.py`. On a UEFI DUT, install an image signed by a rotated db key -> enrollment + verification pass; `/host/db-auth` receives `DB-<sha256>.auth`.

### DIFF
--- a/scripts/secure_boot_enroll_db.sh
+++ b/scripts/secure_boot_enroll_db.sh
@@ -1,0 +1,120 @@
+#!/bin/bash
+# Enroll the UEFI Secure Boot db certificate bundled inside a SONiC installer image
+# (boot/DB.auth) into the running system's UEFI db variable.
+#
+# This must run BEFORE the image's CMS signature is verified (verify_image_sign.sh reads the
+# verification certificate from the UEFI db via "efi-readvar -v db"). When an image is signed
+# with a rotated db key, the new db certificate is not yet present in the firmware, so the
+# verification would fail. Enrolling the bundled, pre-signed db update here lets the new image
+# be verified and trusted without a manual key-enrollment step.
+#
+# Security: DB.auth is a TIME_BASED_AUTHENTICATED_WRITE update that was signed (by the KEK) at
+# build time. UEFI firmware validates that signature against the already-enrolled KEK before
+# applying the update, so a db certificate can only be added when it is authorized by a trusted
+# KEK. An image carrying a db certificate signed by an untrusted KEK is rejected by firmware,
+# the db is left unchanged, and the subsequent CMS verification still fails. This therefore
+# does not bypass the secure-upgrade trust model; it implements KEK-authorized db key rotation.
+#
+# On any failure this logs the cause and exits non-zero; the calling script (sonic_installer)
+# ignores the exit status, so a failure here still never blocks an install.
+
+set -euo pipefail
+
+image_file="${1:-}"
+DB_GUID="d719b2cb-3d3a-4596-a3bc-dad00e67656f"
+EFIVARS_DIR="/sys/firmware/efi/efivars"
+DB_VAR="${EFIVARS_DIR}/db-${DB_GUID}"
+
+log() {
+    echo "secure_boot_enroll_db: $*"
+}
+
+if [ -z "$image_file" ] || [ ! -f "$image_file" ]; then
+    log "ERROR: image file '${image_file}' not found"
+    exit 1
+fi
+
+if [ ! -d "$EFIVARS_DIR" ]; then
+    log "ERROR: efivars not available at $EFIVARS_DIR, cannot enroll db certificate"
+    exit 1
+fi
+if ! mountpoint -q "$EFIVARS_DIR" 2>/dev/null; then
+    if ! mount -t efivarfs efivarfs "$EFIVARS_DIR" 2>/dev/null; then
+        log "ERROR: failed to mount efivarfs at $EFIVARS_DIR, cannot enroll db certificate"
+        exit 1
+    fi
+fi
+
+TMP_DIR=$(mktemp -d)
+trap 'rm -rf "$TMP_DIR"' EXIT
+FS_ZIP="${TMP_DIR}/fs.zip"
+DB_AUTH="${TMP_DIR}/DB.auth"
+
+# The .bin is a self-extracting archive: a shell header up to the "exit_marker" line, followed
+# by a tar of the installer/ directory (which contains fs.zip), then an optional appended CMS
+# signature. Strip the header, then pull installer/fs.zip out of the tar; --occurrence=1 makes
+# tar stop after the first match so the trailing signature is ignored.
+SHARCH_SIZE=$(sed '/^exit_marker$/q' "$image_file" | wc -c)
+tail -c +$((SHARCH_SIZE + 1)) "$image_file" | tar --occurrence=1 -xO installer/fs.zip 2>/dev/null > "$FS_ZIP" || true
+if [ ! -s "$FS_ZIP" ]; then
+    log "image does not contain installer/fs.zip, no db certificate to enroll"
+    exit 0
+fi
+
+unzip -p "$FS_ZIP" boot/DB.auth 2>/dev/null > "$DB_AUTH" || true
+if [ ! -s "$DB_AUTH" ]; then
+    log "image does not bundle boot/DB.auth, no db certificate to enroll"
+    exit 0
+fi
+
+# Persist the db certificate shipped with the image under /host/db-auth using a unique,
+# content-based name, so certificates from successive installations accumulate instead of
+# overwriting each other. The file name is the certificate's sha256, so a matching name implies
+# identical content; a name clash with differing content would be a hash collision and is
+# treated as an error rather than silently trusting the wrong certificate.
+DB_AUTH_DIR=/host/db-auth
+if ! mkdir -p "$DB_AUTH_DIR"; then
+    log "ERROR: failed to create $DB_AUTH_DIR"
+    exit 1
+fi
+
+db_hash=$(sha256sum "$DB_AUTH" | cut -d' ' -f1)
+db_dest="$DB_AUTH_DIR/DB-${db_hash}.auth"
+
+if [ -e "$db_dest" ]; then
+    if cmp -s "$DB_AUTH" "$db_dest"; then
+        log "DB.auth already present at $db_dest, not copying"
+    else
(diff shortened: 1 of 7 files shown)


scripts/secure_boot_enroll_db.sh
```diff
@@ -0,0 +1,122 @@
...
+    done
+    if [ -n "$found_auth" ]; then
+        log "DB.auth already present at $found_auth, not copying"
+    else
+        db_hash=$(sha256sum "$DB_AUTH" 2>/dev/null | cut -d' ' -f1)
+        if [ -n "$db_hash" ]; then
+            db_dest="$DB_AUTH_DIR/DB-${db_hash}.auth"
+        else
+            db_dest="$DB_AUTH_DIR/DB-$(date -u +%Y%m%d%H%M%S).auth"
+        fi
```
saiarcot895: OpenSSL can calculate a fingerprint using any desired hashing algorithm. For example:

```
$ openssl x509 -in /usr/share/ca-certificates/mozilla/DigiCert_Assured_ID_Root_G3.crt -noout -SHA256 -fingerprint SHA256
Fingerprint=7E:37:CB:8B:4C:47:09:0C:AB:36:55:1B:A6:F4:5D:B8:40:68:0F:BA:16:6A:95:2D:B1:00:71:7F:43:05:3F:C2
```

saiarcot895: Ah, never mind, this is the auth file.

scripts/secure_boot_enroll_db.sh
```diff
@@ -0,0 +1,122 @@
...
+    done
+    if [ -n "$found_auth" ]; then
+        log "DB.auth already present at $found_auth, not copying"
+    else
+        db_hash=$(sha256sum "$DB_AUTH" 2>/dev/null | cut -d' ' -f1)
+        if [ -n "$db_hash" ]; then
+            db_dest="$DB_AUTH_DIR/DB-${db_hash}.auth"
+        else
+            db_dest="$DB_AUTH_DIR/DB-$(date -u +%Y%m%d%H%M%S).auth"
```
saiarcot895: Why would the hash be empty?

scripts/secure_boot_enroll_db.sh
```diff
@@ -0,0 +1,122 @@
...
+DB_AUTH_DIR=/host/db-auth
+if mkdir -p "$DB_AUTH_DIR" 2>/dev/null; then
+    found_auth=""
+    for existing_auth in "$DB_AUTH_DIR"/*.auth; do
+        [ -e "$existing_auth" ] || continue
+        if cmp -s "$DB_AUTH" "$existing_auth"; then
+            found_auth="$existing_auth"
+            break
+        fi
+    done
```
saiarcot895: Instead of checking each and every file, use the hash to only check the matching file.

saiarcot895: Additionally, if it exists, but the contents are different, then raise an error. That means there's a hash collision.

scripts/secure_boot_enroll_db.sh
```diff
@@ -0,0 +1,122 @@
...
+# be verified and trusted without a manual key-enrollment step.
+#
+# Security: DB.auth is a TIME_BASED_AUTHENTICATED_WRITE update that was signed (by the KEK) at
+# build time. UEFI firmware validates that signature against the already-enrolled KEK before
+# applying the update, so a db certificate can only be added when it is authorized by a trusted
+# KEK. An image carrying a db certificate signed by an untrusted KEK is rejected by firmware,
+# the db is left unchanged, and the subsequent CMS verification still fails. This therefore
+# does not bypass the secure-upgrade trust model; it implements KEK-authorized db key rotation.
+#
+# Best-effort: every failure is logged and the script exits 0 so it never blocks an install.
```
saiarcot895: Instead of this, I would recommend failing with a non-zero exit code, and ignoring the exit code in the calling script.

scripts/secure_boot_enroll_db.sh
```diff
@@ -0,0 +1,122 @@
...
+
```
saiarcot895, line 20: `set -euo pipefail`

scripts/secure_boot_enroll_db.sh
```diff
@@ -0,0 +1,122 @@
...
+    log "image file '${image_file}' not found, skipping db enrollment"
+    exit 0
+fi
+
+if [ ! -d "$EFIVARS_DIR" ]; then
+    log "efivars not available, skipping db enrollment"
+    exit 0
+fi
+if ! mountpoint -q "$EFIVARS_DIR" 2>/dev/null; then
+    mount -t efivarfs efivarfs "$EFIVARS_DIR" 2>/dev/null || true
```
saiarcot895: This really shouldn't be `|| true`. If the EFI variables can't be mounted, then this script can't do anything useful.

scripts/secure_boot_enroll_db.sh
```diff
@@ -0,0 +1,122 @@
...
+    fi
+else
+    log "WARNING: failed to create $DB_AUTH_DIR"
+fi
+
+# Clear the immutable (i) attribute the kernel sets on existing Secure Boot variables so the
+# variable can be written. Best-effort; efi-updatevar also manages the flag itself.
+clear_db_immutable() {
+    [ -e "$DB_VAR" ] || return 0
+    if command -v chattr >/dev/null 2>&1; then
```
saiarcot895: `chattr` is going to exist, it's a basic command.

scripts/secure_boot_enroll_db.sh
```diff
@@ -0,0 +1,122 @@
...
+        else
+            log "WARNING: failed to copy DB.auth to $db_dest"
+        fi
+    fi
+else
+    log "WARNING: failed to create $DB_AUTH_DIR"
+fi
+
+# Clear the immutable (i) attribute the kernel sets on existing Secure Boot variables so the
+# variable can be written. Best-effort; efi-updatevar also manages the flag itself.
```
saiarcot895: `efi-updatevar` does *not* manage the immutable flag itself.

scripts/secure_boot_enroll_db.sh
```diff
@@ -0,0 +1,122 @@
...
+    log "image does not bundle boot/DB.auth, skipping db enrollment"
+    exit 0
+fi
+
+# Persist the db certificate shipped with the image under /host/db-auth using a unique,
+# content-based name, so certificates from successive installations accumulate instead of
+# overwriting each other. If a byte-identical DB.auth is already stored, it is not copied
+# again. Best-effort.
+DB_AUTH_DIR=/host/db-auth
+if mkdir -p "$DB_AUTH_DIR" 2>/dev/null; then
```
saiarcot895: Invert the if-condition so that visually, failure conditions cause it to bail early, and less code is indented.

```bash
if ! mkdir -p "$DB_AUTH_DIR"; then
    log "ERROR: failed to create $DB_AUTH_DIR"
fi
```

scripts/verify_image_sign.sh
```diff
@@ -39,22 +39,29 @@ sig-list-to-certs $EFI_CERTS_DIR/db_efi $EFI_CERTS_DIR/db >/dev/null||
...
-    LOG=$(verify_image_sign_common $image_file $DATA_FILE $CMS_SIG_FILE)
-    VALIDATION_RES=$?
-    if [ $VALIDATION_RES -eq 0 ]; then
-        RESULT="CMS Verified OK using efi keys"
-        echo "verification ok:$RESULT"
-        # No need to continue.
-        # Exit without error if any success signature verification.
-        clean_up 0
-    fi
 done
```
saiarcot895: A more concise and resilient command (please test this first):

```bash
LOG=$(find $EFI_CERTS_DIR -type f -name 'db-*' -execdir openssl x509 -in {} -inform der \; 2>&1 > $EFI_CERTS_DIR/cert.pem)
if [ $? -ne 0 ]; then
    logger "cms_validation: $LOG"
fi
```

scripts/secure_boot_enroll_db.sh
elybarneams: Tool issue. If the device doesn't support sha256sum.
Do you think we can ignore this case?

saiarcot895: sha256sum will be present. This script will be present only on future images, so any tools that are being installed in `build_debian.sh` or `sonic_debian_extension.j2` in sonic-buildimage will be present.

saiarcot895 (approved):

yxieca (approved):
Approving. Enrolling the image's bundled db cert (KEK-authorized) right before CMS signature verification is a clean way to support rotated db keys without weakening the trust model: enroll is best-effort (its return is ignored) and `verify_image_sign` remains the real gate — an untrusted-KEK cert is rejected by firmware, so CMS verify still fails. The base `Bootloader` no-op default keeps other bootloaders unaffected.

The `verify_image_sign.sh` per-cert-loop -> combined-trust-store refactor is behavior-equivalent: the `find -execdir openssl x509 ... > cert.pem` redirect is opened once, so all db-* certs concatenate into a single multi-cert PEM, which the consumer accepts as -CAfile/-certfile (chain verify + `-nointern -noverify` variant). Faster on failing images, not a regression. The enroll script is robust (`set -euo pipefail`, `tar --occurrence=1` sharch parse, content-addressed /host/db-auth with sha256 collision guard, `efi-updatevar -a` append), and the new grub path has tests.

Non-blocking:
- The combined-store refactor of verify_image_sign.sh has no direct unit test (shell-level, verified by inspection) — worth adding if practical.
- verify_image_sign.sh loses its trailing newline (cosmetic).

PR #4709, opened 2026-07-21
Fix KeyError in 'show bmp tables' when BMP is not or partially configured

asechoud:
#### What I did
Fix KeyError exception in `show bmp tables` command when BMP feature is not configured or only partially configured.

#### How I did it
- Use `bmp_keys.get('table', {})` to safely handle missing 'table' key
- Use `table_config.get()` with 'false' default for each table field
- Always display all three table statuses with defaults when not configured

#### How to verify it
1. On a device without BMP configured, run `show bmp tables` - should display all tables as 'false' instead of crashing
2. Enable only one BMP table and run `show bmp tables` - should show mixed values without crashing

#### Previous command output (failing)
```
$ show bmp tables
BMP tables: 
  ...
KeyError: 'table'
```

#### New command output (fixed)
```
$ show bmp tables
BMP tables: 
Table_Name          Enabled
------------------  ---------
bgp_neighbor_table  false
bgp_rib_in_table    false
bgp_rib_out_table   false
```

### DIFF
--- a/show/main.py
+++ b/show/main.py
@@ -2705,10 +2705,10 @@ def tables(db):
     bmp_body = []
     click.echo("BMP tables: ")
     bmp_keys = db.cfgdb.get_table('BMP')
-    if bmp_keys['table']:
-        bmp_body.append(['bgp_neighbor_table', bmp_keys['table']['bgp_neighbor_table']])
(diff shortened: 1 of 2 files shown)


m11dedhia (approved):
lgtm

PR #4718, opened 2026-07-27
Persist a warm-reboot flag on disk in shutdown path

ryanzhu706:
#### What I did
When a warm/fast/express reboot coincides with a hardware fault, determine-reboot-cause makes the hardware cause the primary cause and demotes the reboot type to an unparsed free-text comment — so show reboot-cause history can't reliably be filtered by reboot type in exactly the case we need to catch. The dedicated marker records the user-invoked reboot type explicitly and persistently.
Persist a best-effort marker recording the reboot type/time whenever a
warm-/fast-/express-reboot is issued, so that on the next boot
`postupgrade_actions` can detect a hardware-induced fault that coincided with a
user-issued warm/fast reboot.

`scripts/fast-reboot` is the mainline reboot script (`warm-reboot` and
`express-reboot` are symlinks to it), so this gives incoming images the behavior
natively. It mirrors the equivalent change already made to the per-image
`fast_reboot_*` data scripts (which patch already-released images).
| User runs | REBOOT_SCRIPT_NAME (persisted as `reboot_type`) | REBOOT_TYPE (persisted as `reboot_type_effective`) |
|---|---|---|
| warm-reboot (Mellanox) | warm-reboot | fastfast-reboot &larr; misleading if used as reboot_type |
| warm-reboot (non-Mellanox) | warm-reboot | warm-reboot |
| fast-reboot | fast-reboot | fast-reboot |
| express-reboot | express-reboot | express-reboot |

#### How I did it

- Added a dedicated `persist_reboot_flag()` function. The flag path is a
  function-`local` (kept out of global script scope).
- It clears any previous marker first (`rm -f`) so a failed write can never leave
  a stale flag behind, then `mkdir -p` the directory and writes the marker.
- The whole operation is best-effort (`|| true`) so it never blocks the reboot
  (the script runs under `set -e`).
- The function is invoked right after the existing reboot-cause file is written.


scripts/fast-reboot
```diff
@@ -63,6 +63,24 @@ MIN_HD_SPACE_NEEDED=$((100000 * $NUM_ASIC)) # 100M * NUM_ASIC
...
+    # NOTE: This file persists across reboots, so readers MUST treat it as potentially stale
+    # unless it is explicitly cleared after processing (or cleared by non-warm reboot paths).
+    # Best-effort: never block the reboot if writing the flag fails (script runs with -e).
+    local WARM_REBOOT_FLAG_FILE="/host/reboot-cause/warm-reboot-flag.txt"
+    rm -f "${WARM_REBOOT_FLAG_FILE}" || true
+    mkdir -p "$(dirname "${WARM_REBOOT_FLAG_FILE}")" || true
+    {
+        echo "reboot_type=${REBOOT_TYPE}"
+        echo "reboot_time=${REBOOT_TIME}"
+    } > "${WARM_REBOOT_FLAG_FILE}" || true
```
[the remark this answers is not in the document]
hdwhdw: This is valid!

hdwhdw (dismissed):

scripts/fast-reboot
```diff
@@ -63,6 +63,25 @@ MIN_HD_SPACE_NEEDED=$((100000 * $NUM_ASIC)) # 100M * NUM_ASIC
 # List of ASICs that perform the reboot
 ASIC_LIST=($(seq 0 $(( NUM_ASIC - 1))))
 
+function persist_reboot_flag()
```
hdwhdw: btw would you consider record_reboot_marker? (and the filename as well)

hdwhdw: actually copilot already mention it.

scripts/fast-reboot
```diff
@@ -63,6 +63,25 @@ MIN_HD_SPACE_NEEDED=$((100000 * $NUM_ASIC)) # 100M * NUM_ASIC
...
+    # unless it is explicitly cleared after processing (or cleared by non-warm reboot paths).
+    # Best-effort: never block the reboot if writing the flag fails (script runs with -e).
+    local WARM_REBOOT_FLAG_FILE="/host/reboot-cause/warm-reboot-flag.txt"
+    rm -f "${WARM_REBOOT_FLAG_FILE}" || true
+    mkdir -p "$(dirname "${WARM_REBOOT_FLAG_FILE}")" || true
+    {
+        echo "reboot_script=\"${REBOOT_SCRIPT_NAME}\""
+        echo "reboot_type=\"${REBOOT_TYPE}\""
+        echo "reboot_time=\"${REBOOT_TIME}\""
+    } > "${WARM_REBOOT_FLAG_FILE}" || true
```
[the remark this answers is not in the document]
hdwhdw: This one is valid as well. We can also update the function name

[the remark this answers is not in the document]
ryanzhu706: Done.

hdwhdw (commented):
Can you take a look at my comment at copilot review thread? Suggest also changing the name of the function. What do you think of record_reboot_marker?

ryanzhu706:
> Can you take a look at my comment at copilot review thread? Suggest also changing the name of the function. What do you think of record_reboot_marker?

Done.

scripts/fast-reboot
```diff
@@ -63,6 +63,25 @@ MIN_HD_SPACE_NEEDED=$((100000 * $NUM_ASIC)) # 100M * NUM_ASIC
 # List of ASICs that perform the reboot
 ASIC_LIST=($(seq 0 $(( NUM_ASIC - 1))))
 
+function reboot_record_marker()
```
hdwhdw: record_reboot_marker?

hdwhdw: verbs followed by an object?

ryanzhu706: Thought reboot and record all can be verbs or objects but maybe record_reboot_marker is better.

Ryangwaite:
The PR does not explain why we cannot use `show reboot-cause history` for this task.

ryanzhu706:
> The PR does not explain why we cannot use `show reboot-cause history` for this task.

Added in description, thanks!

hdwhdw (approved):

PR #4723, opened 2026-07-29
Fix empty leaf-list regression in GCU sorter

rimunagala:
## Summary
Prevent BulkLeafListMoveGenerator from emitting a bulk `REPLACE ... []` for empty target leaf-lists, which could serialize to `""` in CONFIG_DB.

## Regression coverage
- Added a unit test to ensure empty target leaf-lists do not produce a bulk move.
- Added a sorter regression for removing the last BGP allowed prefix and verifying the patch sorts to a remove-only flow.
- Kept the non-empty leaf-list bulk-replace path covered.

## Tracking
- Fixes sonic-buildimage issue #27818

### DIFF
--- a/generic_config_updater/gu_common.py
+++ b/generic_config_updater/gu_common.py
@@ -9,8 +9,10 @@ import yang as ly
 import copy
 import re
 import os
+import hashlib
 from sonic_py_common import logger, multi_asic
 from enum import Enum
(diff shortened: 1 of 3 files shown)


rimunagala:
Replaced by #4724 with a clean branch and the same regression fix.

PR #4726, opened 2026-07-29
Fix empty leaf-list regression in GCU sorter

rimunagala:
## Summary
Prevent BulkLeafListMoveGenerator from emitting a bulk `REPLACE ... []` for empty target leaf-lists, which could serialize to `""` in CONFIG_DB.

## Regression coverage
- Added a unit test to ensure empty target leaf-lists do not produce a bulk move.
- Added a sorter regression for removing the last BGP allowed prefix and verifying the patch sorts to a remove-only flow.
- Kept the non-empty leaf-list bulk-replace path covered.

## Tracking
- Fixes sonic-buildimage issue #27818

### DIFF
--- a/generic_config_updater/patch_sorter.py
+++ b/generic_config_updater/patch_sorter.py
@@ -1355,6 +1355,7 @@ class BulkLeafListMoveGenerator:
             if isinstance(current_val, list) and isinstance(target_val, list):
                 # Only handle leaf-lists (lists of scalars)
                 if (current_val != target_val and
+                        target_val and
                         self._is_leaf_list(current_val) and
                         self._is_leaf_list(target_val)):
                     yield JsonMoveGroup(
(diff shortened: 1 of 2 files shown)


vaibhavhd:
Please provide the results after applying the repro steps mentioned on the issue - [27818](https://github.com/sonic-net/sonic-buildimage/issues/27818)

rimunagala:
Superseded by #4731.

This PR's head branch lives in the upstream repo, where branch protection blocks any update (normal push and force-push are both declined), so the review feedback could not be pushed here. #4731 is the same change opened from my fork, with both Copilot review comments addressed:

- The empty-target guard is now explicit (\len(target_val) > 0\) with a comment explaining why the bulk REPLACE path is skipped.
- \	est_generate__leaf_list_all_items_removed__single_replace_move\ renamed to \..._no_bulk_replace_move\ to match its assertion.

@vaibhavhd - continuing the repro discussion on #4731.

PR #4730, opened 2026-07-29
Preserve IPv6 connected routes during fast-reboot

YairRaviv:
#### What I did

Preserved IPv6 connected routes during fast reboot, matching the existing behavior for IPv4 connected routes.
This prevents IPv6 connected routes from being removed from APP_DB and delayed until fpmsyncd reconciliation.

Fixes this issue - [Bug: [fast-reboot] IPv6 connected routes are not preserved during fast-reboot and remain unavailable until route reconciliation](https://github.com/sonic-net/sonic-buildimage/issues/28603)

#### How I did it
Updated `fast-reboot-filter-routes.py` to collect connected routes from both address families:

1. show ip route connected json
2. show ipv6 route connected json

#### How to verify it

1. Configure IPv6 addresses on routed interfaces.
2. Confirm the connected routes exist:
    show ipv6 route connected
3. Run fast reboot with continuous IPv6 traffic.
4. Verify:
- IPv6 connected routes remain preserved in APP_DB.
- fpmsyncd does not report them as introducing new entry after the 120-second timer.
- Traffic loss remains within the fast-reboot limit.
- Existing IPv4 route-preservation behavior is unchanged.

### DIFF
--- a/scripts/fast-reboot-filter-routes.py
+++ b/scripts/fast-reboot-filter-routes.py
@@ -21,6 +21,14 @@ def get_connected_routes():
         for route in route_info.keys():
             connected_routes.append(route)
 
+    cmd = ['sudo', 'vtysh', '-c', "show ipv6 route connected json"]
+    output, ret = clicommon.run_command(cmd, return_cmd=True)
+    if ret != 0:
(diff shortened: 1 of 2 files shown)


prabhataravind:
@deepak-singhal0408 @vaibhavhd for viz.

abelamit (dismissed):

PR #4732, opened 2026-07-30
Fix migrate_sonic_packages() resolv.conf injection for relative symlinks

william8545:
#### What I did

Fixed `sonic-installer install` failing in `migrate_sonic_packages()` with `Temporary failure in name resolution` when `sonic-package-manager migrate` has to fetch a package manifest from a container registry inside the new image's chroot.

#4365 made the DNS injection symlink-aware, but its path math only handles absolute symlink targets: `lstrip("/")` is a no-op on a relative target such as `../run/resolvconf/resolv.conf` (the form the Debian resolvconf package ships in the image), so `os.path.join()` produced a path whose `..` escaped the chroot mount. The host's DNS configuration was written to the host filesystem (`/tmp/run/...`) instead of into the chroot overlay, leaving the chroot with whatever stale `resolv.conf` the image's squashfs happened to carry from its build server — so package migration worked or failed depending on which build machine produced the image.

#### How I did it

- Resolve a relative symlink target against the symlink's own directory (`etc/`), per POSIX symlink semantics.
- `os.path.normpath()` on the anchored absolute path collapses `..` and clamps it at the chroot root, mirroring how the kernel resolves the link inside the chroot — the computed path can never escape the image mount.
- `cp -L --remove-destination`: if the computed target already exists as a symlink, replace it instead of writing through it.
- Absolute targets pass through unchanged (`normpath` is the identity on them), so the case #4365 fixed behaves exactly as before; the regular-file branch is untouched.

#### How to verify it

- `tests/test_sonic_installer.py`: `test_install` is parametrized over the new image's `/etc/resolv.conf` layouts — absolute symlink, relative symlink, a target with excess `..`, and a regular file. The relative-symlink case fails against the previous code; the expected command list is asserted with strict order/argv equality.
- Verified end to end on a Mellanox MSN4700: installing an image whose squashfs ships the relative symlink and stale, unreachable build-time nameservers previously failed at the `sonic-package-manager migrate` step with the DNS error above (and left a stray copy of the host's DNS configuration under `/tmp/run/` on the host). With this fix the same install completes: the host DNS lands inside the chroot overlay (`rw/run/resolvconf/resolv.conf`) and no stray host file is created.

### DIFF
--- a/sonic_installer/main.py
+++ b/sonic_installer/main.py
@@ -387,13 +387,20 @@ def migrate_sonic_packages(bootloader, binary_image_version):
             # Inject host DNS into chroot for sonic-package-manager to resolve hostnames.
             chroot_resolv = os.path.join(new_image_mount, RESOLV_CONF_FILE)
             if os.path.islink(chroot_resolv):
-                # Symlink: populate the target inside the chroot so the symlink resolves.
-                # Cannot cp over the symlink because the absolute target path escapes
-                # the overlay mount to the host filesystem ("are the same file" error).
+                # Symlink: populate its target inside the chroot ("cp" over the link
+                # would write through it, outside the overlay mount).
                 resolv_target = os.readlink(chroot_resolv)
+                if not os.path.isabs(resolv_target):
+                    # Relative target resolves against the symlink's directory
+                    resolv_target = os.path.join("/", os.path.dirname(RESOLV_CONF_FILE), resolv_target)
+                # Leading "/" makes normpath clamp ".." at the chroot root
+                resolv_target = os.path.normpath(resolv_target)
                 chroot_target = os.path.join(new_image_mount, resolv_target.lstrip("/"))
                 run_command_or_raise(["mkdir", "-p", os.path.dirname(chroot_target)])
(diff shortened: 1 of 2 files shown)


dgsudharsan (approved):

yxieca (approved):
Reviewed and approving — correct, minimal security hardening for chroot resolv.conf symlink handling.

Verified by executing the exact path arithmetic on the parametrized test targets AND on hostile inputs (`../../../../etc/passwd`, `/run/../../../etc/x`, deep `../` to `/root/.ssh/authorized_keys`): every resolved target stays clamped under the chroot mount — no host escape via the root-privileged cp. The relative-symlink resolution (resolvconf's `../run/resolvconf/resolv.conf`) is handled correctly, and `cp -L --remove-destination` avoids writing through an existing symlink target. Regular-file branch is untouched (no regression).

Non-blocking: worth confirming whether any maintained release branch runs the old absolute-only logic and needs this backported — the prior code mishandles resolvconf's relative symlink, so migrate could fail/mis-populate there.

rookie-who:
Approved for 202605 cherry-pick.

Has the test gap been identified for the resolv.conf symlink issue in migrate_sonic_packages? Is there a test case that could catch relative symlink handling failures during image install to prevent this class of regression?

PR #4775, opened 2026-08-13
[pfcwd] Scale start_default timers off PFC-enabled ports only

arawat-nexthop:
#### What I did

Fixed `pfcwd start_default` so its port-count scaling counts only the ports the watchdog is actually configured on.

`start_default` scales detection, restoration and poll times by the number of ports, but it counted **every** entry in the `PORT` table while only configuring a subset. The ports it configures are the `DEVICE_NEIGHBOR` and backplane ports that have `pfc_enable` set in `PORT_QOS_MAP`; the rest are skipped by `verify_pfc_enable_status_per_port` with a "PFC is not enabled on port" warning.

These two have been inconsistent since 99d251f2d ("Enable PFCWD only on ports where PFC is enabled", #1508) introduced that gate without adjusting the scaling input. Ports that never receive a watchdog inflate the timers for the ports that do.

Concretely, on a system with 64 PFC-enabled front-panel ports plus 3 ports that are not PFC-capable (recycle ports, for example) present in `PORT`:

```
port_num = 67  ->  multiply = (67-1)//32+1 = 3  ->  detection_time = 600ms   (before)
port_num = 64  ->  multiply = (64-1)//32+1 = 2  ->  detection_time = 400ms   (after)
```

Only 64 ports ever get a watchdog, so 400ms is the intended value. Systems where every port in `PORT` is PFC-enabled are unaffected.

#### How I did it

Replaced the `PORT`-table count with a count of the ports that will actually be configured, applying the same `pfc_enable` test that `verify_pfc_enable_status_per_port` uses:

```python
port_num = sum(
    1 for port in active_ports
    if self.config_db.get_entry(PORT_QOS_MAP, port).get('pfc_enable') is not None
)
```

Two deliberate choices:

- The configuration loop still walks every entry in `active_ports`, so the per-port "PFC is not enabled on port" warnings are unchanged. Pre-filtering the loop would have been a smaller diff but would have dropped that diagnostic.
- `max(1, ...)` already handles an empty eligible set, so a system with no PFC-enabled port keeps the 200ms baseline rather than producing a zero or negative multiplier.

Updated `doc/Command-Reference.md`, which previously said the times scale with "number of ports in the system", to state that only ports the watchdog is configured on are counted.

**Test changes.** `test_pfcwd_start_default_512_ports` drove its port count by mocking the `PORT` table, which no longer feeds the scaling, so it would have silently stopped testing anything. It now supplies 512 PFC-enabled ports through `DEVICE_NEIGHBOR`/`PORT_QOS_MAP` via a shared `_run_start_default_with_ports()` helper and asserts on CONFIG_DB instead of a rendered table that would be 512 rows long; its now-unused expected-output vector is removed. Added `test_pfcwd_start_default_ignores_pfc_ineligible_ports` covering the case above.

`test_pfcwd_start_default` and `test_pfcwd_start_default_32_ports` are unchanged and unaffected: the mock CONFIG_DB has 2 PFC-enabled ports, so `multiply` is 1 either way and their 200ms expectations still hold.

#### How to verify it

```
python -m pytest tests/pfcwd_test.py -k start_default -v
```

Scaling arithmetic, for `DEFAULT_PORT_NUM=32`, `DEFAULT_*_TIME=200`, `MAX_POLL_INTERVAL_TIME=1000`:

| PFC-enabled ports | multiply | detection/restoration | poll |
|---|---|---|---|
| 0 | 1 | 200ms | 200ms |
| 32 | 1 | 200ms | 200ms |
| 33 | 2 | 400ms | 400ms |
| 64 | 2 | 400ms | 400ms |
| 512 | 16 | 3200ms | 1000ms (clamped) |

On a switch, with 64 PFC-enabled ports and some non-PFC ports present in `PORT`:

```
sudo pfcwd start_default
show pfcwd config
```

On a system with 64 PFC-enabled ports plus 3 non-PFC ports in `PORT`:

```
admin@sonic:~$ sudo pfcwd start_default
admin@sonic:~$ show pfcwd config
Changed polling interval to 600ms
       PORT    ACTION    DETECTION TIME    RESTORATION TIME    HISTORY
-----------  --------  ----------------  ------------------  ---------
  Ethernet0      drop               600                 600    disable
  Ethernet4      drop               600                 600    disable
        ...       ...               ...                 ...        ...
Ethernet252      drop               600                 600    disable
```

```
admin@sonic:~$ sudo pfcwd start_default
admin@sonic:~$ show pfcwd config
Changed polling interval to 400ms
       PORT    ACTION    DETECTION TIME    RESTORATION TIME    HISTORY
-----------  --------  ----------------  ------------------  ---------
  Ethernet0      drop               400                 400    disable
  Ethernet4      drop               400                 400    disable
        ...       ...               ...                 ...        ...
Ethernet252      drop               400                 400    disable
```

The set of configured ports is unchanged (64 in both cases); only the timer values differ.

### DIFF
--- a/pfcwd/main.py
+++ b/pfcwd/main.py
@@ -418,7 +418,14 @@ class PfcwdCli(object):
         if not enable or enable.lower() != "enable":
             return
 
-        port_num = len(list(self.config_db.get_table('PORT').keys()))
+        # Count only the ports that will actually be configured below. Ports without a
+        # PORT_QOS_MAP 'pfc_enable' are skipped by verify_pfc_enable_status_per_port, so
+        # counting every PORT entry inflates the timers on behalf of ports that never get
+        # a watchdog.
+        port_num = sum(
+            1 for port in active_ports
+            if self.config_db.get_entry(PORT_QOS_MAP, port).get('pfc_enable') is not None
+        )
 
         # Parameter values positively correlate to the number of ports.
         multiply = max(1, (port_num-1)//DEFAULT_PORT_NUM+1)
--- a/pfcwd/main.py
+++ b/pfcwd/main.py
@@ -401,6 +401,26 @@ class PfcwdCli(object):
                 continue
             self.config_db.mod_entry(CONFIG_DB_PFC_WD_TABLE_NAME, port, None)
 
+    def count_pfc_enabled_ports(self, ports):
+        """Count the ports that PFC watchdog will actually be configured on.
+
+        verify_pfc_enable_status_per_port() skips any port that has no 'pfc_enable' in
+        PORT_QOS_MAP, so these are the ports that receive a PFC_WD entry and contribute
+        watchdog polling load. Counting every PORT entry instead would inflate the timers
+        on behalf of ports that never get a watchdog.
+
+        Args:
+            ports (list): Candidate ports, i.e. the external and backplane ports.
+
+        Returns:
+            int: Number of the given ports that have PFC enabled.
+        """
+        port_qos_map = self.config_db.get_table(PORT_QOS_MAP)
+        return sum(
+            1 for port in ports
+            if port_qos_map.get(port, {}).get('pfc_enable') is not None
+        )
+
     @multi_asic_util.run_on_multi_asic
     def start_default(self):
         if os.geteuid() != 0:
@@ -418,14 +438,7 @@ class PfcwdCli(object):
         if not enable or enable.lower() != "enable":
             return
 
-        # Count only the ports that will actually be configured below. Ports without a
-        # PORT_QOS_MAP 'pfc_enable' are skipped by verify_pfc_enable_status_per_port, so
-        # counting every PORT entry inflates the timers on behalf of ports that never get
-        # a watchdog.
-        port_num = sum(
-            1 for port in active_ports
-            if self.config_db.get_entry(PORT_QOS_MAP, port).get('pfc_enable') is not None
-        )
+        port_num = self.count_pfc_enabled_ports(active_ports)
 
         # Parameter values positively correlate to the number of ports.
         multiply = max(1, (port_num-1)//DEFAULT_PORT_NUM+1)
--- a/tests/pfcwd_test.py
+++ b/tests/pfcwd_test.py
@@ -276,36 +276,87 @@ class TestPfcwd(object):
         assert result.exit_code == 0
         assert result.output == test_vectors.pfcwd_show_start_default_32_ports
 
+    @staticmethod
+    def _run_start_default_with_ports(db, pfc_ports, extra_ports=()):
+        """Run start_default with a synthetic port layout.
+
+        start_default scales its timers off the ports it actually configures, which are the
+        DEVICE_NEIGHBOR ports carrying a PORT_QOS_MAP 'pfc_enable'. extra_ports are present only
+        in PORT, standing in for ports that are not PFC-capable.
+
+        Args:
+            db (obj): Db instance to run against.
+            pfc_ports (list): Ports in DEVICE_NEIGHBOR with pfc_enable set.
+            extra_ports (list): Ports present only in PORT, without pfc_enable.
+        """
+        import pfcwd.main as pfcwd
+
+        original_get_table = db.cfgdb.get_table
+        original_get_entry = db.cfgdb.get_entry
+
+        def mock_get_table(table):
+            if table == 'PORT':
+                return {port: {} for port in list(pfc_ports) + list(extra_ports)}
+            if table == 'DEVICE_NEIGHBOR':
+                return {port: {} for port in pfc_ports}
+            return original_get_table(table)
+
+        def mock_get_entry(table, key):
+            if table == 'PORT_QOS_MAP':
+                if key in pfc_ports:
+                    return {'pfc_enable': '3,4'}
+                if key in extra_ports:
+                    return {}
+            return original_get_entry(table, key)
+
+        with patch.object(db.cfgdb, 'get_table', side_effect=mock_get_table), \
+                patch.object(db.cfgdb, 'get_entry', side_effect=mock_get_entry):
+            return CliRunner().invoke(pfcwd.cli.commands["start_default"], [], obj=db)
+
     @patch('pfcwd.main.os')
     def test_pfcwd_start_default_512_ports(self, mock_os):
         """Test start_default on 512-port system: multiply=16, detection/restoration=3200, poll=1000ms"""
-        import pfcwd.main as pfcwd
-        runner = CliRunner()
         db = Db()
-
-        # Patch PORT table to have 512 ports so multiply = (512-1)//32+1 = 16
-        original_get_table = db.cfgdb.get_table
-
-        def mock_get_table_512_ports(table):
-            if table == 'PORT':
-                return {'Ethernet%d' % i: {} for i in range(512)}
-            return original_get_table(table)
-
         mock_os.geteuid.return_value = 0
-        with patch.object(db.cfgdb, 'get_table', side_effect=mock_get_table_512_ports):
-            result = runner.invoke(
-                pfcwd.cli.commands["start_default"],
-                [],
-                obj=db
-            )
+
+        # 512 PFC-enabled ports so multiply = (512-1)//32+1 = 16. Asserted on CONFIG_DB rather
+        # than rendered output, which would be 512 rows.
+        ports = ['Ethernet%d' % i for i in range(512)]
+        result = self._run_start_default_with_ports(db, ports)
         assert result.exit_code == 0
 
-        result = runner.invoke(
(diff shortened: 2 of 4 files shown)


pinky-nexthop:
@vmittal-msft @kperumalbfn @bingwang-ms Kindly review.

arawat-nexthop:
The Azure failure here looks unrelated to this change. It's 4 tests in `tests/pfcstat_test.py::TestMultiAsicPfcstat`, where `show pfc counters` comes back without the backplane ports it expects. The same 4 fail the same way on #4773, which only touches sonic_installer, so it seems to be something on master rather than either PR.

The pfcwd tests all pass in that run:

```
PASSED tests/pfcwd_test.py::TestPfcwd::test_pfcwd_start_default_32_ports
PASSED tests/pfcwd_test.py::TestPfcwd::test_pfcwd_start_default_512_ports
PASSED tests/pfcwd_test.py::TestPfcwd::test_pfcwd_start_default_ignores_pfc_ineligible_ports
```

Happy to rebase once master is green if that's easier for review.

pfcwd/main.py
```diff
@@ -418,7 +418,14 @@ def start_default(self):
         if not enable or enable.lower() != "enable":
             return
 
-        port_num = len(list(self.config_db.get_table('PORT').keys()))
+        # Count only the ports that will actually be configured below. Ports without a
+        # PORT_QOS_MAP 'pfc_enable' are skipped by verify_pfc_enable_status_per_port, so
+        # counting every PORT entry inflates the timers on behalf of ports that never get
+        # a watchdog.
+        port_num = sum(
```
vmittal-msft: Good — this predicate is exactly the one `verify_pfc_enable_status_per_port` applies, so the count and the configuration loop can't drift out of sync.

Minor / non-blocking: `PORT_QOS_MAP` is now read twice per active port (once here for the count, once inside `verify_pfc_enable_status_per_port`), i.e. `2N` `hgetall`s at config time on large fabrics. If you'd like to avoid that, you could read `get_table(PORT_QOS_MAP)` once and count locally with the same `pfc_enable is not None` predicate. `start_default` is admin-invoked and rare, so this is purely optional.

tests/pfcwd_test.py
```diff
@@ -276,36 +276,87 @@ def mock_get_table_32_ports(table):
...
+        They are skipped by verify_pfc_enable_status_per_port, so counting them makes the
+        timers larger on behalf of ports that never receive a watchdog: 64 PFC-enabled ports
+        alongside 3 ineligible ones scaled off all 67 PORT entries, giving multiply=3 (600ms)
+        instead of the correct multiply=2 (400ms).
+        """
+        db = Db()
+        mock_os.geteuid.return_value = 0
+
+        pfc_ports = ['Ethernet%d' % (i * 4) for i in range(64)]
+        ineligible = ['Ethernet256', 'Ethernet260', 'Ethernet-Rec0']
```
vmittal-msft, line 341: Coverage gap on the actual predicate: these ports are excluded because they're **only in `PORT`** (not in `DEVICE_NEIGHBOR`, so not in `active_ports`), which means every port that reaches the `sum()` still has `pfc_enable`. The `is not None` **False** branch is never taken here — yet that branch (the parity with `verify_pfc_enable_status_per_port`) is the crux of this change.

Consider adding a case with a `DEVICE_NEIGHBOR` (or backplane) port that lacks `pfc_enable`, asserting it is excluded from the count, still logs the "PFC is not enabled on port" warning, and gets no `PFC_WD` entry.

vmittal-msft (commented):
Thanks for the careful writeup and the before/after arithmetic — this is a clean, well-motivated fix.

The new count mirrors the exact gate in `verify_pfc_enable_status_per_port` (same `PORT_QOS_MAP` / `pfc_enable is not None` predicate) over the same `active_ports` the config loop walks, so `port_num` now provably equals the number of ports that actually receive a `PFC_WD` entry. Scaling the poll interval off that count is also *more* correct, since only PFC-enabled ports get watchdogs and thus contribute polling load, and `max(1, ...)` keeps the 200 ms baseline for the empty case. LGTM overall — a few non-blocking notes.

**1. `test_pfcwd_start_default_32_ports` is now effectively vacuous.** Its `mock_get_table_32_ports` patches the `PORT` table, which no longer feeds the scaling, so it now just re-exercises the 2-PFC-port default CONFIG_DB (multiply=1) — the same thing `test_pfcwd_start_default` covers. Nothing now pins the 32→33 multiply boundary (added coverage is 0 / 64 / 512). Consider converting it to the new `_run_start_default_with_ports` helper with 32 (and 33) PFC-enabled ports.

**2. Behavior-change awareness for the 202511 backport.** On upgrade, systems that have non-PFC ports present in `PORT` (recycle ports, etc.) will see their default PFCWD timers shrink. This is the intended fix and is well documented here — just flagging it for release-note visibility given the backport label.

Two inline notes below. (Reviewed statically — I couldn't run the suite in my environment as the `swsscommon` bindings were absent — but I confirmed there are no dangling refs to the removed `pfcwd_show_start_default_512_ports` vector, the doc edit is the only relevant doc change, and the mock CONFIG_DB has no pre-seeded `PFC_WD` entries for the ineligible ports so the `== {}` assertions hold.)

pfcwd/main.py
kperumalbfn: @arawat-nexthop could we create a separate method to count active PFC enabled ports. Much easier to change the logic later if required.

kperumalbfn:
@arawat-nexthop Could you validate this in 202605 branch and add the results. We can take this to 202605 branch

rookie-who:
Hi — this PR is flagged for cherry-pick to the **202605 branch**, but it appears to be a new feature/enhancement rather than a bug fix. The 202605 branch is in stabilization, so we're limiting cherry-picks to bug fixes and critical platform support.

Could you please retarget this to the next release branch? If there's a specific customer requirement driving this for 202605, please comment with the justification and we can reconsider.

*Comment by @vaibhavhd via automated tooling*

arawat-nexthop:
Thanks both — pushed in 06dd6d9.

@vmittal-msft you were right about the coverage gap, and it was worse than it looked. The ineligible ports in that test are only in `PORT`, never in `DEVICE_NEIGHBOR`, so they never reach the count at all. I checked by deleting the predicate outright and leaving `sum(1 for port in active_ports)` — the whole suite still passed. Nothing was guarding it.

The new test puts a port in `DEVICE_NEIGHBOR` without `pfc_enable` and sits it on the 32/33 boundary, so counting it would move the timer from 200ms to 400ms and the assertion fails. That same mutation now gets caught. It also asserts the port still logs the "PFC is not enabled on port" warning and still gets no `PFC_WD` entry.

`test_pfcwd_start_default_32_ports` was vacuous as you said — it was passing off the two PFC-enabled ports in the mock CONFIG_DB rather than anything it mocked. Replaced with `test_pfcwd_start_default_multiply_boundary`, which pins both sides of the 32→33 step, and dropped its now-unused vector.

@kperumalbfn pulled the counting out into `count_pfc_enabled_ports()`. That also folds in Vineet's note about the double read — it does a single `get_table(PORT_QOS_MAP)` now instead of a `get_entry` per active port.

arawat-nexthop:
@kperumalbfn Tested the fix on 202605
```
PORT entries            : 65
DEVICE_NEIGHBOR entries : 64
ports with pfc_enable   : 64
non-Ext role ports      : Ethernet-Rec0  role=Rec
```
With Fix
```
admin@dut:~$ sudo pfcwd start_default
admin@dut:~$ show pfcwd config
Changed polling interval to 400ms
       PORT    ACTION    DETECTION TIME    RESTORATION TIME    HISTORY
-----------  --------  ----------------  ------------------  ---------
  Ethernet0      drop               400                 400    disable
  Ethernet4      drop               400                 400    disable
  Ethernet8      drop               400                 400    disable
...
```
Without fix:

```
ports configured : 64
distinct timers  : 400`
admin@dut:~$ sudo pfcwd start_default
admin@dut:~$ show pfcwd config
Changed polling interval to 600ms
       PORT    ACTION    DETECTION TIME    RESTORATION TIME    HISTORY
-----------  --------  ----------------  ------------------  ---------
  Ethernet0      drop               600                 600    disable
  Ethernet4      drop               600                 600    disable
  Ethernet8      drop               600                 600    disable
...
ports configured : 64
```

pfcwd/main.py
arawat-nexthop: addressed

tests/pfcwd_test.py
arawat-nexthop: addressed

kperumalbfn (approved):

kperumalbfn:
@vaibhavhd This is a bugfix and required for 202605, could you please add to the branch.

kperumalbfn:
@vmittal-msft please add to 202511

PR #4783, opened 2026-08-14
[route_check] Close the vtysh read end before waiting on a parse failure

ganglyu:
#### What I did

Made `fetch_routes()` terminate the entire `vtysh` process group when the streaming JSON parse fails, instead of logging a warning and calling `proc.wait()` on a child it never drained.

`fetch_routes` runs `sudo vtysh -c "show ip route json"`. On a SONiC host `/usr/bin/vtysh` is a wrapper script that runs `docker exec -i bgp vtysh ...`, so the process actually holding the exec session into the bgp container sits two levels below the child `Popen` returns:

```
sudo                              <- Popen's direct child
  └─ /bin/bash /usr/bin/vtysh     <- wrapper
       └─ docker exec -i bgp vtysh -c "show ip route json"
            ┊ containerd-shim
            └─ vtysh (in container)   <- blocked in write()
```

When the parse raised, the read was abandoned while `vtysh` still had megabytes queued. On a route table larger than the 64KB pipe buffer the in-container `vtysh` stays blocked in `write()`, so `proc.wait()` never returns and the exec session is never released. From that point the bgp container can no longer be stopped, and every later `docker exec` into it fails in `setns`:

```
OCI runtime exec failed: exec failed: unable to start container process:
error starting setns process: fork/exec /proc/self/fd/6: no such file or directory
```

`docker ps` still reports the container as `Up`, so this presents as a healthy container that nothing can talk to.

The exposure was introduced when `fetch_routes` moved from a single full read to a streaming parse. Until then it was `subprocess.check_output(cmd, text=True)` followed by `json.loads(output)`: stdout is drained to EOF before anything is parsed, so a parse error is raised after the child has already exited, and there is no state in which the reader stops while the writer still has data queued. Streaming with `ijson` made that state reachable — any exception leaving the `ijson.kvitems` loop ends the read for good, while `vtysh` may still have megabytes to write into a 64KB pipe, so it blocks deterministically rather than as a race.

#### How I did it

- `subprocess.Popen(..., start_new_session=True)`, so the child leads its own process group and the wrapper plus the `docker exec` client inherit it. This also isolates the concurrent IPv4 and IPv6 fetches from each other and from route_check's own process group.
- New `terminate_vtysh_process_group()` helper that does `os.killpg(proc.pid, SIGKILL)` and reaps with a bounded `wait()`. Signalling is best effort: a cleanup failure must never mask the error that got us there. It is reported at `WARNING` rather than `DEBUG` because a failed `killpg` is the only early signal that this cleanup did not take effect, and the default report level hides `DEBUG`.
- On any parse failure the helper runs before `proc.wait()`, which is what releases the exec session.

Error handling is otherwise untouched: parse failures, a missing command and unexpected errors are still logged and swallowed, and `fetch_routes` still returns whatever it managed to parse. This PR changes cleanup only.

One thing worth calling out for review: the helper deliberately passes `proc.pid` rather than `os.getpgid(proc.pid)`. It is only correct because `start_new_session=True` makes pgid == pid; deriving the group would target route_check's *own* group if the session was ever not created. There is a unit test guarding this.

Killing only the direct child — `proc.kill()` — does not fix this. `SIGKILL` cannot be propagated by `sudo`, so the wrapper and the `docker exec` client are orphaned and keep the exec session open. Measured on a switch with a 6.7MB route table, six concurrent abandoned parses, with the container-recreate workaround removed so the container ID is reused on restart:

| route_check behaviour | sudo | wrapper | docker exec client | in-container vtysh | container stop | result |
|---|---|---|---|---|---|---|
| no kill (before this change) | alive | alive | alive | alive | 15s + 15s timeouts | every later `docker exec` hangs |
| `proc.kill()` | dead | **alive** | **alive** | **alive** | 15s + 15s timeouts | every later `docker exec` hangs |
| `os.killpg(SIGKILL)` (this change) | dead | dead | dead | **dead** | **clean, no timeouts** | healthy |

Note that the in-container `vtysh` is reaped indirectly: killing the host-side `docker exec` client drops the attach, the daemon closes the exec stdout stream, and the in-container process gets `EPIPE`. The container-side process is not in the signalled group.

#### How to verify it

Unit tests:

```
pytest tests/route_check_test.py -k TestFetchRoutes -v
```

`test_terminates_whole_process_group_on_parse_error` is the regression guard. It builds a three level process chain whose *grandchild* floods the pipe — the stand-in for the `docker exec` client — feeds `fetch_routes` truncated JSON, and asserts the call still returns `([], [])` while the grandchild is gone and the group no longer exists. Against the previous code it does not fail fast, it deadlocks: the call is made on a worker thread so the hang is reported as a failure after 60s instead of stalling the run.

On a switch, with the container-recreate workaround removed from `bgp.sh` so the container ID is reused:

1. Wait for BGP to converge to a route table well over 64KB (a few MB makes it deterministic).
2. Drive six concurrent `fetch_routes()` calls with `ijson.kvitems` patched to consume a chunk, wait long enough for the pipe to fill, then raise.
3. Before the kill, confirm the chain really formed: six `docker exec -i bgp vtysh` clients on the host and six in-container `vtysh` processes with `wchan == pipe_write`.
4. After the kill, all of those counts must be zero.
5. `systemctl restart bgp`, then confirm the journal has no `Container failed to exit within ...` lines and that a burst of parallel `docker exec bgp true` all succeed.

Result of that run: preconditions met (6/6 and 6/6 before the kill), everything reaped afterwards, container stop clean with zero timeout messages even though the container ID was reused, and 240/240 parallel `docker exec` returned 0. A normal `route_check.py -m DEBUG` run against the same 6.7MB table completed with exit 0 and empty missing/failing lists.

### DIFF
--- a/scripts/route_check.py
+++ b/scripts/route_check.py
@@ -81,8 +81,11 @@ PRINT_MSG_TRUNCATION_SUFFIX = " ... (truncated)"
 FRR_CHECK_RETRIES = 3
 FRR_WAIT_TIME = 15
 
-# How long to wait for the vtysh process group to be reaped after SIGKILL
-KILL_WAIT_SECONDS = 10
+# How long to wait for vtysh to exit once we have closed the read end
+VTYSH_EXIT_TIMEOUT_SECONDS = 10
+
+# How long to wait for the direct child to be reaped after we SIGKILL it
+SIGKILL_REAP_TIMEOUT_SECONDS = 3
 
 REDIS_TIMEOUT_MSECS = 0
 
@@ -401,48 +404,6 @@ def is_suppress_fib_pending_enabled(namespace):
     return state == 'enabled'
 
 
-def terminate_vtysh_process_group(proc):
-    """
-    SIGKILL the whole process group started for the vtysh command and reap it.
-
-    fetch_routes runs "sudo vtysh", and on a SONiC host /usr/bin/vtysh is a
-    wrapper script that runs "docker exec -i bgp vtysh ...". proc.kill() only
-    signals sudo: the wrapper and the docker exec client survive it, the vtysh
-    inside the container stays blocked writing into a pipe nobody reads any
-    more, and from then on the bgp container cannot be stopped - every later
-    docker exec into it hangs. Signalling the whole group is what actually
-    releases the exec session.
-
-    This relies on Popen(start_new_session=True) making the child a process
-    group leader, so its pgid equals its pid. Do not derive the group with
-    os.getpgid(): if the new session was not created that returns route_check's
-    own group and we would kill ourselves.
-
-    It also assumes sudo leaves the command in that group. That holds for the
-    sudo we ship, but not under "Defaults use_pty" (the default from sudo
-    1.9.14), where sudo runs the command in its own session behind a pty and
-    only the sudo front-end would be signalled.
-
-    :return True if the process was reaped, False if it outlived the SIGKILL.
-    """
-    try:
-        os.killpg(proc.pid, signal.SIGKILL)
-    except (ProcessLookupError, PermissionError, OSError) as e:
-        # Best effort: a cleanup failure must never mask the error that got us here.
-        # Reported at WARNING because this is the only early signal that the
-        # cleanup did not take effect; the default report level hides DEBUG.
-        # Deliberately not "except Exception": a bad pid is a bug, not a
-        # cleanup failure, and must not be swallowed.
-        print_message(syslog.LOG_WARNING, f"Could not signal process group {proc.pid}: {e}")
-
-    try:
-        proc.wait(timeout=KILL_WAIT_SECONDS)
-        return True
-    except subprocess.TimeoutExpired:
-        print_message(syslog.LOG_ERR, f"vtysh (pid {proc.pid}) still alive after SIGKILL")
-        return False
-
-
 def fetch_routes(ipv6=False, namespace=multi_asic.DEFAULT_NAMESPACE):
     """
     Fetch routes using the given command.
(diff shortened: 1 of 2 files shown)


keboliu (dismissed):

dgsudharsan:
@deepak-singhal0408 @eddieruan-alibaba @venkit-nexthop @mike-dubrovsky @saiarcot895   FYI. The fix for the issue we discussed last week in routing subgroup

saiarcot895 (changes requested):
Manually killing the whole process feels like a bit of a hack. Additionally, `Popen`'s `__exit__` closes the stdout and stderr pipes, which should result in a `EPIPE`, cause the caller to stop writing to stdout, and gracefully exit.

Please test with the old code, but with the `proc.wait()` removed, or with `proc.stdout.close()` called before `proc.wait()`. If there's still a hang, try doing a read after `proc.stdout.close()` is called.

ganglyu:
@saiarcot895 You were right — closing the read end is enough.

Ran all three variants on a switch (40.9MB `show ip route json`, 64KB pipe), parse aborted mid-stream, as root:

| cleanup on parse failure | returns? | child rc | left behind |
|---|---|---|---|
| old code (`wait()` inside the `with`) | **no, deadlock** | — | **4** |
| (1) drop the explicit `wait()`, let `__exit__` close | 0.21s | 141 (SIGPIPE) | 0 |
| (2) `proc.stdout.close()` then `wait()` | 0.18s | 141 (SIGPIPE) | 0 |

(3) isn't needed — a read after `close()` just raises `ValueError: I/O operation on closed file`.

Root cause is exactly what you pointed at: `Popen.__exit__` *does* close `proc.stdout`, but the explicit `proc.wait()` sat **inside** the `with` block and blocked before `__exit__` could run. An ordering bug.

The PR now closes the read end in a `finally` before any `wait()`, and the remaining `wait()` is bounded. The `SIGPIPE` exit (141 through `sudo`) is no longer reported as a `vtysh` failure, since it is the expected outcome of closing rather than a real failure. Two guards added, both of which fail against the previous revision.

saiarcot895 (approved):

keboliu (approved):

deepak-singhal0408 (approved):

PR #4784, opened 2026-08-14
Implement BMC Firmware Update in sonic-utilities

brcarr-nv:
#### What I did

Fixed `fwutil` incorrectly treating a non-modular chassis as modular when the platform exposes a module that has **no firmware components**.

On the NVIDIA AST2700 BMC, `chassis.get_all_modules()` returns a single component-less `SWITCH-HOST` module. `PlatformDataProvider.is_modular_chassis()` only checked whether *any* module existed (`len(self.module_component_map) > 0`), so it concluded the BMC was a modular chassis. `fwutil` then required a `module` section in `platform_components.json`; since the BMC (correctly) has none, every `fwutil` command that instantiates a `ComponentUpdateProvider` (e.g. `fwutil show status`, `fwutil update ...`) aborted while parsing the platform schema.

Also adds a `--force-update` flag to `fwutil install/update fw` for platforms (e.g. PLDM-based BMC firmware update) that support forced reinstall or downgrade, and captures BMC firmware versions via `fwutil` in `generate_dump`.

#### How I did it

Changed `is_modular_chassis()` to base modularity on whether any module actually exposes firmware components, rather than on the mere presence of modules:

```python
# fwutil/lib.py
def is_modular_chassis(self):
    return any(self.module_component_map.values())
```

`module_component_map` maps each module name to its `{component: ...}` dict, so a component-less module contributes an empty dict and no longer forces the chassis to be considered modular. Genuine modular chassis (modules that expose components) are unaffected, and SmartSwitch handling — which is explicitly forced non-modular for parsing via `is_modular_chassis() and not is_smart_switch()` — is unchanged.

Added `--force-update` to `fwutil install/update fw` commands. Components that support it (detected via signature inspection) receive `force_update=True`; unsupported components log a warning and proceed normally.

Added `fwutil show status` and `fwutil show version` captures to `collect_nvidia_bmc_dump` in `generate_dump` so BMC firmware versions are collected even if the hw-management dump script is absent or times out.

Tests (`tests/fwutil_test.py`):
- Updated `test_regular_modular_chassis_parsing` to model a module that actually exposes a component, so it still verifies a genuine modular chassis is parsed as modular.
- Added `test_module_without_components_is_non_modular`, mirroring the BMC's component-less `SWITCH-HOST` module and asserting the chassis is parsed as non-modular.
- Added tests for `--force-update` behavior including backend support detection, warning/fallback paths, and keyword-only enforcement.
- Added tests ensuring techsupport dumps include firmware status/version information even when BMC dump scripts fail.

#### How to verify it

Run the fwutil unit tests:

```
pytest tests/fwutil_test.py -k "modular or module_without or smartswitch or force"
```

All modular/SmartSwitch/force-update cases pass, including the new component-less-module test.

On a platform whose `get_all_modules()` returns only component-less modules with a `platform_components.json` that has **no** `module` section, run `fwutil show status`. It now succeeds instead of aborting with a platform-schema parse error.

```
admin@sonic:~$ fwutil show status
Error: Failed to parse "platform_components.json": invalid platform schema: "module" key hasn't been found. Aborting...
```

```
admin@sonic:~$ fwutil show status
Chassis    Module    Component    Version (Current/Available)    Status
---------  --------  -----------  -----------------------------  --------------
P4102-A01  N/A       BMC          88.0060.2212 / N/A             up-to-date
```

### DIFF
--- a/fwutil/lib.py
+++ b/fwutil/lib.py
@@ -6,6 +6,7 @@
 
 try:
     import glob
+    import inspect
     import os
     import json
     import shutil
@@ -152,6 +153,30 @@ class URL(object):
     url = property(fget=get_url)
 
 
+def supports_force_update(update_callable):
+    """Return True if ``update_callable`` accepts a ``force_update`` argument.
+
+    Support is determined from the callable's signature *before* invocation, so
+    a ``TypeError`` raised from inside a firmware install/update (a genuine bug,
+    or an error after the transfer has already started) is never misread as
+    "force-update unsupported" and silently retried without it. The keyword is
+    omitted only when it is definitively unsupported; otherwise the caller
+    invokes with ``force_update`` and lets any exception propagate.
+    """
+    try:
+        sig = inspect.signature(update_callable)
+    except (TypeError, ValueError):
+        # C-implemented / builtin callables may not expose a signature;
+        # treat as not supporting the explicit force_update kwarg.
+        return False
+    for param in sig.parameters.values():
+        if param.name == "force_update":
+            return True
+        if param.kind == inspect.Parameter.VAR_KEYWORD:
+            return True
+    return False
+
+
 class PlatformDataProvider(object):
     """
     PlatformDataProvider
@@ -201,7 +226,13 @@ class PlatformDataProvider(object):
         return self.__chassis.is_smartswitch()
 
     def is_modular_chassis(self):
-        return len(self.module_component_map) > 0
+        # A chassis is modular (for firmware purposes) only if at least one
+        # module actually exposes firmware components. Merely having modules in
+        # the map is not enough: platforms may expose non-firmware modules
+        # (e.g. the NVIDIA AST2700 BMC exposes a component-less SWITCH-HOST
+        # module), and those must not force a "module" section in
+        # platform_components.json on an otherwise non-modular device.
+        return any(self.module_component_map.values())
 
     def is_chassis_has_components(self):
         return self.__chassis.get_num_components() > 0
@@ -593,11 +624,19 @@ class ComponentUpdateProvider(PlatformDataProvider):
             pcp.chassis_component_map
         )
 
-        self.__validate_component_map(
-            self.SECTION_MODULE,
-            self.module_component_map,
-            pcp.module_component_map
-        )
+        # For a non-modular chassis the parser does not read a "module" section
+        # (see parse_platform_components), so pcp.module_component_map is empty,
+        # while the data provider may still report component-less modules (e.g.
+        # the NVIDIA AST2700 BMC's SWITCH-HOST). Comparing the two would raise a
+        # spurious "module names mismatch", so only validate the module map when
+        # the chassis is actually modular. SmartSwitch keeps its existing
+        # per-mismatch handling inside __validate_component_map.
+        if self.is_modular_chassis():
+            self.__validate_component_map(
+                self.SECTION_MODULE,
+                self.module_component_map,
+                pcp.module_component_map
+            )
 
     def get_updates_status(self):
         status_table = [ ]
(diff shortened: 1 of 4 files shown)


oleksandrivantsiv (dismissed):

oleksandrivantsiv (dismissed):

liat-grozovik:
@judyjoseph @yxieca kindly reminder.

judyjoseph:
@brcarr-nv  is this fix to get the firmware upgrade work from BMC, In case where there is no components defined for Switch-Host ? A quick question : is whether we need this force update if we really define components for switch-host module?

brcarr-nv:
@judyjoseph This PR makes 3 changes:

1. Modular-chassis detection fix — is_modular_chassis() now checks whether any module actually has firmware components (any(self.module_component_map.values())), instead of just whether any module exists at all. This stops platforms with component-less modules (e.g. the NVIDIA AST2700 BMC's SWITCH-HOST module) from being misclassified as modular, which previously caused fwutil to abort requiring a "module" section in platform_components.json that legitimately shouldn't exist. Schema validation was also updated to skip the module-map check entirely for non-modular chassis.
2. --force-update flag — Added to fwutil install fw and fwutil update fw, letting a component backend perform a forced reinstall/downgrade (e.g. PLDM Force Update). Support is detected from the backend's method signature before invoking it; if the target component's backend doesn't support it, the command now aborts with a clear error (this is the change we just made and pushed) rather than silently continuing with a normal update.
3. generate_dump BMC firmware capture — collect_nvidia_bmc_dump() now runs and saves fwutil show status / fwutil show version output, so BMC firmware versions are included in tech-support dumps even if the hw-management dump script is missing or times out.


I also reworked how the --force-update flag is set up so that it cannot be used with any commands that don't explicitly support it.

fwutil/lib.py
```diff
@@ -201,7 +226,13 @@ def is_smart_switch(self):
         return self.__chassis.is_smartswitch()
 
     def is_modular_chassis(self):
-        return len(self.module_component_map) > 0
+        # A chassis is modular (for firmware purposes) only if at least one
```
vivekverma-arista, line 229: You already have an object to Chassis class, you can simply call the platform API `is_modular_chassis` instead of resorting to this method.

vivekverma-arista (changes requested):

fwutil/main.py
```diff
@@ -271,9 +288,13 @@ def validate_fw(ctx, param, value):
 # 'fw' subcommand
 @component_install.command(name='fw')
 @click.option('-y', '--yes', 'yes', is_flag=True, show_default=True, help="Assume \"yes\" as answer to all prompts and run non-interactively")
+@click.option('--force-update', 'force_update', is_flag=True, show_default=True,
```
vivekverma-arista, line 291: Can you also update https://github.com/sonic-net/SONiC/blob/master/doc/fwutil/fwutil.md to document the newly introduced argument `--force-update` and a clarification on how this differs from the existing `-f` argument.

judyjoseph (approved):
LGTM .. please address @vivekverma-arista comment

fwutil/lib.py
brcarr-nv: This is a false-friend name collision: ChassisBase.is_modular_chassis() (sonic-platform-common) answers "is this SONiC instance a supervisor/line-card in a physically modular chassis" (defaults False, only meaningful for real chassis platforms), while PlatformDataProvider.is_modular_chassis() in fwutil/lib.py answers "does any module expose firmware components, so platform_components.json needs a module section." Swapping to the platform API would silently break fwutil on any modular chassis whose platform doesn't correctly override is_modular_chassis(), and wouldn't even address the bug this PR fixes (the BMC case). I made a targeted fix: added a code comment explaining why the two aren't interchangeable, so the confusion doesn't resurface in future review or refactors.

fwutil/main.py
brcarr-nv: Sure. This will need to be done in a separate PR though as it is in a different repo. So I created https://github.com/sonic-net/SONiC/pull/2549

fwutil/lib.py
vivekverma-arista: In that case I would recommend changing the name of this function altogether to something closer to its true meaning. Merely documenting is not enough. I believe fwutil is the only user of the function so it shouldn't break anything.

PR #4789, opened 2026-08-18
generate_dump: gzip sai failure tar dump and clean up leftover tar file

jianyuewu:
#### What I did
Fix save_sai_failure_dump in generate_dump so that a SAI failure tar dump (*.tar) is gzip-compressed before it is collected into the techsupport tarball, and both the produced *.tar.gz and the original *.tar are removed during cleanup so no uncompressed tar file is left behind under /var/log/sai_failure_dump/.

#### How I did it
In save_sai_failure_dump():
- Before collecting a file, if it ends in .tar, run gzip -f on it and switch to the .tar.gz path. This keeps the existing save_file/save_symlink logic (which passes do_gzip=false for already-gz files) working correctly.
- In the cleanup step, rm -f now removes both file (the .gz) and {file%.gz} (the original tar), so nothing is left behind.

#### How to verify it
Run the bash test directly:
     bash tests/save_sai_failure_dump_test.sh
     On a device, trigger a SAI failure dump under /var/log/sai_failure_dump/, run generate_dump, and confirm the SAI dump is included as a .tar.gz in the techsupport tarball and that no *.tar / *.tar.gz is left behind in /var/log/sai_failure_dump/.

#### Tested branch
- [x] master

#### Test result
master: PASS.
**1. Unit test — `tests/save_sai_failure_dump_test.sh` (8 scenarios, 17 assertions)**
Results: 17 passed, 0 failed
**2. On-device functional test**
Created a fake SAI failure dump, ran `generate_dump`, and compared behavior before/after the fix.
Result: the SAI dump is collected into techsupport as `.tar.gz` and no uncompressed `.tar` is left behind under `/var/log/sai_failure_dump/`. The DUT's `generate_dump` was restored to the shipped version after testing.

### DIFF
--- a/scripts/generate_dump
+++ b/scripts/generate_dump
@@ -2470,6 +2470,34 @@ save_sai_failure_dump(){
         return
     fi
     for file in $(find_files "/var/log/sai_failure_dump/"); do
+        # A tar dump (*.tar) takes much more space than a gzipped one,
+        # compress it first to save space before collecting it.  Skip the
+        # in-place gzip if a "$file.gz" already exists, so we never clobber a
+        # pre-existing, independent .gz dump (gzip's overwrite would replace
+        # its content with this tar's).  In that case the .tar is collected
+        # as-is (save_file gzips it into the tarball without touching the
+        # sibling on disk).
+        #
+        # If gzip itself fails, that usually means a broken environment (e.g. a
+        # full filesystem) in which the subsequent save_file/tar-append would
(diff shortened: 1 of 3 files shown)


Junchao-Mellanox (approved):

keboliu (approved):

keboliu:
@qiluo-msft  would you please help to review?

qiluo-msft (approved):

PR #4793, opened 2026-08-22
[chassis][show_techsupport] Enhance generate_dump to save CHASSIS_APP_DB, CHASSIS_STATE_DB and Fabric info

mlok-nokia:
#### What I did
Enhance the show_techsupport to includes the database-chassis info (CHASSIS_APP_DB and CHASSIS_STATE_DB) on chassis platform. Also includes all outputs of  "show fabric" commands

#### How I did it
In generate_dump script,  add function save_chassis_dattatbase_info() to the generate_dump script to call the save_redis()  with CHASSIS_APP_DB and CHASSIS_STATE_DB from database-chassis.  Also, add function save_fabric_info() to save all outputs of show fabric commands.  Add logic if only IS_CHASSIS_MODULE is true, then execute these functions

#### How to verify it
Run the "show techsupport" command on a chassis (either LC or Sup", untar the techSuppport files and verify that CHASSIS_APP_DB.json and CHASSIS_STATE_DB.json should exist in the dump directory.

N/A
N/A

### DIFF
--- a/scripts/generate_dump
+++ b/scripts/generate_dump
@@ -875,7 +875,7 @@ save_chassis_database_info() {
         save_redis "CHASSIS_STATE_DB" "CHASSIS_STATE_DB" true
     else
         # In LC, in case database-chassis is not reachable, ping CHASSIS_APP_DB first
-        ret=$(timeout 5 sonic-db-cli CHASSIS_APP_DB PING | grep -c True)
+        ret=$(timeout 5 sonic-db-cli CHASSIS_APP_DB PING 2>/dev/null | grep -c True)
         if [ $ret -gt 0 ]; then
             save_redis "CHASSIS_APP_DB" "CHASSIS_APP_DB"  true
             save_redis "CHASSIS_STATE_DB" "CHASSIS_STATE_DB" true
@@ -884,7 +884,7 @@ save_chassis_database_info() {
 }
 
 ###############################################################################
-# Save fabirc info
+# Save fabric info
 # Globals:
 #  None
 # Arguments:
--- a/scripts/generate_dump
+++ b/scripts/generate_dump
@@ -52,6 +52,7 @@ RETURN_CODE=$EXT_SUCCESS
 DEBUG_DUMP=false
 ROUTE_TAB_LIMIT_DIRECT_ITERATION=24000
 IS_SUPERVISOR=false
+IS_CHASSIS_MODULE=false
 # True on a SONiC Switch-BMC (set below from switch_bmc=1 in platform_env.conf).
 # Switch data-plane collectors (ASIC/SDK, ports, FRR/BGP, STP, SAI) self-skip.
 # Distinct from is_bmc_supported() (Redfish debug-dump availability).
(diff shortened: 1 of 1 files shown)


mlok-nokia:
@abdosi and @judyjoseph This PR is toy enhance the show_techsupport to include database-chassis info and the "show fabric" outputs. Please help to review it.  Thanks

scripts/generate_dump
```diff
@@ -854,6 +855,59 @@ save_frr_info() {
...
+        ret=$(timeout 5 sonic-db-cli CHASSIS_APP_DB PING | grep -c True)
+        if [ $ret -gt 0 ]; then
+            save_redis "CHASSIS_APP_DB" "CHASSIS_APP_DB"  true
+            save_redis "CHASSIS_STATE_DB" "CHASSIS_STATE_DB" true
+        fi
+    fi
+}
+
+###############################################################################
+# Save fabirc info
```
saksarav-nokia: fix the typo in the comment "fabirc"

scripts/generate_dump
```diff
@@ -854,6 +855,59 @@ save_frr_info() {
...
+        return
+    fi
+    if $IS_SUPERVISOR; then
+        # Supervisor, allways dump both database
+        save_redis "CHASSIS_APP_DB" "CHASSIS_APP_DB"  true
+        save_redis "CHASSIS_STATE_DB" "CHASSIS_STATE_DB" true
+    else
+        # In LC, in case database-chassis is not reachable, ping CHASSIS_APP_DB first
+        ret=$(timeout 5 sonic-db-cli CHASSIS_APP_DB PING | grep -c True)
+        if [ $ret -gt 0 ]; then
```
saksarav-nokia, line 879: If ret is empty → broken [ $ret -gt 0 ] . Could happen when timeout kills sonic-db-cli after 5s or Connection error before PING completes ,then: ret=          # empty
[ $ret -gt 0 ]   # bash: integer expression expected

You may want to change it to 
ret=$(timeout 5 sonic-db-cli CHASSIS_APP_DB PING 2>/dev/null | grep -c True || true)

saksarav-nokia (approved):

scripts/generate_dump
```diff
@@ -854,6 +855,59 @@ save_frr_info() {
...
+    if $IS_SUPERVISOR; then
+        # Supervisor, allways dump both database
+        save_redis "CHASSIS_APP_DB" "CHASSIS_APP_DB"  true
+        save_redis "CHASSIS_STATE_DB" "CHASSIS_STATE_DB" true
+    else
+        # In LC, in case database-chassis is not reachable, ping CHASSIS_APP_DB first
+        ret=$(timeout 5 sonic-db-cli CHASSIS_APP_DB PING 2>/dev/null | grep -c True)
+        if [ $ret -gt 0 ]; then
+            save_redis "CHASSIS_APP_DB" "CHASSIS_APP_DB"  true
+            save_redis "CHASSIS_STATE_DB" "CHASSIS_STATE_DB" true
```
anarasimhan-upscale, line 881: the CHASSIS_APP_DB is from the database-chassis on the sup, why we have collect this info from every namespace?

liamkearney-msft: i dont think it is. "true" makes it only execute on the host/default namespace.

liamkearney-msft (approved):

scripts/generate_dump
mlok-nokia: This function is only executed on host.  Not namespace.   CHASSIS_APP_DB data are pushed by LC. It is good to have these data in LC showTech.

PR #4800, opened 2026-08-25
[generic_config_updater] Validate VLAN interface data before neighbor cleanup

ashutosh-agrawal:
#### What I did

Validate VLAN interface data before processing removed VLAN interfaces.

#### How I did it

- Check interface names against kernel interface-name constraints.
- Parse interface addresses before invoking neighbor cleanup.
- Ignore malformed entries and log a generic validation error.
- Add unit coverage for accepted and rejected VLAN interface data.

#### How to verify it

- [x] Added unit coverage for valid and invalid VLAN interface keys
- [x] Upstream CI

N/A

N/A

### DIFF
--- a/generic_config_updater/services_validator.py
+++ b/generic_config_updater/services_validator.py
@@ -1,8 +1,12 @@
+import ipaddress
 import os
-import subprocess
+import re
 import shlex
+import subprocess
 import time
 from .gu_common import genericUpdaterLogging
+
+IFNAME_RE = re.compile(r"[A-Za-z0-9_.-]{1,15}")
 
 logger = genericUpdaterLogging.get_logger(title="Service Validator")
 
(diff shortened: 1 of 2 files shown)


generic_config_updater/services_validator.py
```diff
@@ -200,6 +204,18 @@ def vlanintf_validator(old_config, upd_config, keys):
...
+                       print_to_console)
+            continue
+        try:
+            ipaddress.ip_interface(iface_ip)
+        except ValueError:
+            logger.log(logger.LOG_PRIORITY_ERROR,
+                       "vlanintf_validator: Invalid VLAN interface address",
+                       print_to_console)
+            continue
         rc = command_wrapper(f"ip neigh flush dev {iface} {iface_ip}")
```
qiluo-msft, line 219: nit (non-blocking): `command_wrapper` builds a shell-like string via f-string then re-tokenizes it with `shlex.split()`. The new validation here closes the immediate risk, but the pattern itself is fragile for future callers that interpolate less-constrained values.

The repo already has a safer alternative pattern elsewhere — `utilities_common.cli.run_command`, `sonic_installer.common.run_command`, and `sonic_package_manager.service_creator.creator.run_command` all take `command` as a pre-built `List[str]` passed straight to `subprocess`, with no string-build-then-reparse step at all.

Not asking to pull `utilities_common` into GCU here (there's a comment in `field_operation_validators.py` about keeping GCU's dependency footprint narrow). But `command_wrapper` itself could take `List[str]` instead of a format string, e.g. `command_wrapper(["ip", "neigh", "flush", "dev", iface, iface_ip])`, with the `nsenter` prefix built as list-concat instead of string interpolation. Bigger change than this PR needs to make, so a good follow-up rather than a blocker here.

qiluo-msft (approved):

generic_config_updater/services_validator.py
xq9mend: Good point. Opened #4805 as a follow-up: refactors `command_wrapper` to accept `List[str]` directly, removes the shlex-split round-trip, and updates all 7 existing callers. The `nsenter` prefix is now built as list concatenation too.

prhoskot:
@ashutosh-agrawal @qiluo-msft Can we please backport this fix to 202605. Thanks!

vrajeshe:
@qiluo-msft this needs to be ported to 202605

vrajeshe:
@bpar9 FYI

PR #4801, opened 2026-08-25
[202605] 7060CX device state migration via upgrade

justin-wong-ce:
#### What I did
Some 7060 devices need to have the ability to migrate its device state to make it compatible with future Arista driver code.

This migration only applies to 7060CX devices that has warm-upgraded from the 201811 release.

#### How I did it
The migration ensures the file /host/.system-prefdl exists on the machine. This is done by either copying from a cache file, or reading directly from EEPROM.

#### How to verify it

The expected behavior occurs on each of the 3 below scenarios:
(1) /host/.system-prefdl exists
- no action

(2) /host/.system-prefdl does not exist, /etc/sonic/.syseeprom exists
- Copy /etc/sonic/.syseeprom to /host/.system-prefdl, it has been tested
  that it is ok to direct copy even if the contents are slighlty
  different.

(3) /host/.sytem-prefdl and /etc/sonic/.syseeprom does not exist:
- Generate a /host/.system-prefdl by reading directly from EEPROM

This logic was tested on the following images:
- 20260510.10 -> newer Arista internal 202605 build

### DIFF
--- a/scripts/arista_7060_prefdl_gen.py
+++ b/scripts/arista_7060_prefdl_gen.py
@@ -1,0 +1,11 @@
+from arista.components.eeprom import I2cEeprom
+from arista.core.platform import loadPrerequisites
+from arista.core.types import I2cBus
+
+IDENT_BUS_NAME = 'SMBus PIIX4 adapter port 1 at 0b20'
+
+loadPrerequisites()
+eeprom = I2cEeprom(addr=I2cBus(IDENT_BUS_NAME).i2cAddr(0x52))
+eeprom.setup()
+pfdl = eeprom.readPrefdl()
+pfdl.writeToFile("/host/.system-prefdl")
--- a/scripts/arista_7060_prefdl_gen.py
+++ b/scripts/arista_7060_prefdl_gen.py
@@ -2,6 +2,15 @@ from arista.components.eeprom import I2cEeprom
 from arista.core.platform import loadPrerequisites
 from arista.core.types import I2cBus
 
+"""Read system information from EEPROM via I2C
+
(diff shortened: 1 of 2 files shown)


scripts/fast-reboot
```diff
@@ -620,7 +621,6 @@ function is_secureboot() {
 function setup_reboot_variables()
 {
     # Kernel and initrd image
-    HWSKU=$(show platform summary --json | python -c 'import sys, json; print(json.load(sys.stdin)["hwsku"])')
```
Ryangwaite, line 623: Can we leave the definition of this where it was? It is a global var that is referred to in other functions. I don't want to risk introducing any subtle hard to detect bugs due to this

scripts/arista_7060_prefdl_gen.py
Ryangwaite, line 1: Add a docstring explaining what/where/why/how of this script

Ryangwaite (changes requested):

scripts/fast-reboot
justin-wong-ce: `setup_reboot_variables()` are run at a much later stage than where this was moved to.

I can leave this assignment in and run duplicate logic if that is preferred, though this query takes a couple seconds to run.

justin-wong-ce: As per discussion offline, refactored the new logic into a separate function and restored the original HWSKU defintition

scripts/arista_7060_prefdl_gen.py
justin-wong-ce: Done

PR #4803, opened 2026-08-25
[counterpoll] Fix bash tab-completion by not requiring CONFIG_DB at import

yizhenzha:
#### Why I did it

`counterpoll <Tab>` does not auto-complete on the switch (e.g. `counterpoll water<Tab>` should complete to `watermark`), while `config`/`show` complete fine. Root cause: `register_dynamic_commands()` in `counterpoll/main.py` runs at **module-import time** and calls `ConfigDBConnector().connect()`.

The bash-completion generator (`sonic-utilities-data/generate_completions.py`) imports every CLI module in a **redis-less build container** to emit the completion wrappers. The `counterpoll` import raises there (no CONFIG_DB), `generate_completions` swallows the exception (`Cannot generate completion for counterpoll.main:cli!`), and **no `/etc/bash_completion.d/counterpoll` file is shipped** — so bash has nothing to complete against.

#### How I did it

Guard the import-time DB access. swsscommon raises `RuntimeError` when CONFIG_DB is unreachable or the DB config is missing, so catch only `RuntimeError` (unrelated errors still propagate) and log a warning naming the skipped DPU-only commands. Completion is dynamic (the wrapper re-invokes `counterpoll` at Tab time), so on a DPU with CONFIG_DB up the `eni`/`ha_set` commands are still registered and still complete; the guard only affects the redis-less build / non-DPU path where those DPU-only commands don't apply.

Also adds regression tests: `register_dynamic_commands()` swallows a `RuntimeError` connect failure but still propagates unrelated exceptions.

#### How to verify it

1. **Reproduce the bug**
   On a switch, `/etc/bash_completion.d/counterpoll` is missing, and typing `counterpoll water<Tab>` does not complete to `watermark`.

2. **Verify the fix**
   With the patch, the completion generator runs without a redis connection and installs the `counterpoll` completion file, `counterpoll water<Tab>` completes to `watermark`, and the skip path is logged to syslog (if log is kept from build time)
   ```
   WARNING counterpoll: CONFIG_DB not available at import, skipping DPU dynamic commands (eni/ha_set): Unable to connect to redis
   ```

### DIFF
--- a/counterpoll/main.py
+++ b/counterpoll/main.py
@@ -1,12 +1,14 @@
 import click
 import os
-from sonic_py_common import device_info, multi_asic
+from sonic_py_common import device_info, multi_asic, logger
 import utilities_common.multi_asic as multi_asic_util
 from utilities_common.llr import is_llr_capable
 from tabulate import tabulate
 from flow_counter_util.route import exit_if_route_flow_counter_not_support
 from swsscommon.swsscommon import ConfigDBConnector, SonicDBConfig
 from swsscommon.swsscommon import CFG_FLEX_COUNTER_TABLE_NAME as CFG_FLEX_COUNTER_TABLE
+
+log = logger.Logger("counterpoll")
 
 BUFFER_POOL_WATERMARK = "BUFFER_POOL_WATERMARK"
 PORT_BUFFER_DROP = "PORT_BUFFER_DROP"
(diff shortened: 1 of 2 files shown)


dgsudharsan (approved):

Yakiv-Huryk (approved):

qiluo-msft (approved):

yizhenzha:
The cherry pick 202605 PR was already created for branch 202605 @vaibhavhd can you approve? https://github.com/sonic-net/sonic-utilities/pull/4825

PR #4805, opened 2026-08-25
refactor: command_wrapper takes List[str] instead of format string

xq9mend:
## Why I did it

Follow-up to #4800, addressing the nit in the review: `command_wrapper` builds a shell-like string via f-string then re-tokenizes it with `shlex.split()`. The pattern is fragile for future callers that interpolate less-constrained values.

## How I did it

- `command_wrapper` now accepts `List[str]` instead of a format string
- The `nsenter` prefix is built as list concatenation, not string interpolation
- `shlex` import removed (no longer needed)
- All existing callers updated to pass pre-built lists

## How to verify it

Existing unit tests for `command_wrapper` and its callers should continue to pass without modification.

## Changelog description

Refactor `command_wrapper` in `generic_config_updater/services_validator.py` to accept `List[str]` instead of a format string, eliminating the shlex-split round-trip.

### DIFF
--- a/generic_config_updater/services_validator.py
+++ b/generic_config_updater/services_validator.py
@@ -1,7 +1,6 @@
 import ipaddress
 import os
 import re
-import shlex
 import subprocess
 import time
 from .gu_common import genericUpdaterLogging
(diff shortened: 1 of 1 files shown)


xq9mend:
@ashutosh-agrawal @maipbui Could you please review this PR? Thank you!

xq9mend:
@sneelam20 @kwangsuk Could you please review this PR? Thank you!

ashutosh-agrawal (approved):
LGTM

maipbui (approved):
LGTM

kwangsuk (approved):

PR #4809, opened 2026-08-26
Allow ENOENT in vlanintf_validator ip neigh flush

### LINKED ISSUE #4836: vlanintf_validator logs benign ENOENT errors when neighbors are already flushed
Summary
When a VLAN prefix is deleted from CONFIG_DB by Generic Config Updater (GCU), intfmgrd asynchronously removes the address and its neighbors from the kernel. The subsequent ip neigh flush command executed by vlanintf_validator can race with this removal and fail with ENOENT ("No such file or directory"), even though this is not an error condition since the neighbors are already gone.

Problem
Currently, vlanintf_validator logs these benign ENOENT failures as ERROR level syslog messages, which causes downstream test suite failures in LogAnalyzer that expects no error logs.

Impact
LogAnalyzer teardown failures in test suites (e.g., generic_config_updater/test_vlan_interface.py::test_vlan_interface_tc1_suite)
False error reports in monitoring/logging systems
Expected Behavior
ENOENT errors should be treated as success (not an error condition) since the intended outcome—removing the neighbors—is already achieved. Other flush failures (device not found, permission denied, etc.) should continue to be logged as errors.

How to Reproduce
Run the sonic-mgmt test:

Code
generic_config_updater/test_vlan_interface.py::test_vlan_interface_tc1_suite

divyaraj-arista:
When GCU deletes a VLAN prefix from CONFIG_DB, intfmgrd asynchronously removes the address and its neighbors from the kernel. The subsequent ip neigh flush in vlanintf_validator can race with this removal and fail with ENOENT ("No such file or directory"), which is benign since the neighbors are already gone.

Treat this specific failure as success instead of logging ERR records that cause LogAnalyzer teardown failures in test suites. Other flush failures (device not found, etc.) still fail as before.

The change can be verified by reproducing the scenario by running generic_config_updater/test_vlan_interface.py::test_vlan_interface_tc1_suite sonic-mgmt test. 

#### What I did

Given the scenario not being an ERR state, change the vlanintf_validator behavior to not log benign messages as ERRs

#### How I did it

By NOT logging benign ENOENT in vlanintf_validator as syslog ERR

#### How to verify it

The change can be verified by reproducing the scenario by running generic_config_updater/test_vlan_interface.py::test_vlan_interface_tc1_suite sonic-mgmt test. 

closes #4836

### DIFF
--- a/generic_config_updater/services_validator.py
+++ b/generic_config_updater/services_validator.py
@@ -175,10 +175,20 @@ def vlanintf_validator(old_config, upd_config, keys):
     deleted_keys = list(set(old_keys) - set(upd_keys))
     for key in deleted_keys:
         iface, iface_ip = key
-        rc = command_wrapper(f"ip neigh flush dev {iface} {iface_ip}")
-        if rc:
+        cmd = f"ip neigh flush dev {iface} {iface_ip}"
+        result = subprocess.run(
+            shlex.split(cmd), capture_output=True, check=False, text=True
+        )
+        if result.returncode != 0:
+            if "No such file or directory" in result.stderr:
+                continue
             logger.log(logger.LOG_PRIORITY_ERROR,
-                       f"vlanintf_validator: Failed to flush neighbors for {iface} {iface_ip}, returncode={rc}",
+                       f"vlanintf_validator: Failed to flush neighbors for "
+                       f"{iface} {iface_ip}, returncode={result.returncode}",
                        print_to_console)
+            if result.stderr:
+                logger.log(logger.LOG_PRIORITY_ERROR,
+                           f"stderr: {result.stderr}",
+                           print_to_console)
             return False
     return True


lolyu:
Hi @divyaraj-arista, please fix your PR commit signature

lolyu (approved):
LGTM

lolyu:
Hi @divyaraj-arista, the build is failing.

divyaraj-arista:
@lolyu  Build is passing now. It was failing earlier due to unrelated issue.

lolyu:
Hi @divyaraj-arista, should we open a PR to have this fix in master? could you please do that?

rajkumar1-arista:
> Hi @divyaraj-arista, should we open a PR to have this fix in master? could you please do that?

Cherry-picked the change for master: https://github.com/sonic-net/sonic-utilities/pull/4858

Please take a look.

PR #4810, opened 2026-08-27
[GCU] Performance rework of the patch sorter (#3831) [202511]

rimunagala:
> **PR 1 of 8** in a `202511` GCU backport stack. **Merge order is mandatory** - the later cherry-picks assume this one. Tracking issue: #4823 - full stack, merge order and evidence.

#### Why I did it

Cherry-pick of #3831 (`bd3de9da`, merged as `69b26608`), the foundational GCU performance rework. Everything else in this stack builds on it.

#### How I did it

Commit 1 is the cherry-pick. **Commit 2 is specific to `202511` and is not upstream** — please read this part carefully.

#3831 changes `PathAddressing.find_ref_paths` to:

```python
def find_ref_paths(self, paths, config, reload_config: bool = True):
    if reload_config:
        sy.loadData(config)
```

That gate was authored against master's tree, where it is safe **by construction**: PR **#4118** ("Remove direct dependency on libyang", 2026-04-27) had already deleted both `import yang as ly` and `_get_inner_leaf_xpaths()`, replacing them with `sy.find_data_dependencies()`. No `sy.root` dereference remains in that function, so skipping the load cannot fail there.

**#4118 is not an ancestor of `202511`** (confirmed via `merge-base`), so `_get_inner_leaf_xpaths()` is still present at `gu_common.py:649`, along with `import yang as ly` at line 8. It does:

```python
nodes = sy.root.find_path(xpath).data()
```

On this branch, therefore, `find_ref_paths()` requires YANG data to already be loaded. That requirement is satisfied today **only by statement ordering**: all three call sites (`patch_sorter.py:757`, `991`, `1687`) seed `reload_config=True` and flip to `False` only after the first call has loaded, and `sy` is a process-lifetime singleton (`create_sonic_yang_with_loaded_models()` calls `loadYangModel()` once and never `loadData()`), so `sy.root` is `None` only before the first-ever load in the process. Reaching a `reload_config=False` site first would raise:

```
AttributeError: 'NoneType' object has no attribute 'find_path'
```

So the helper was never *introduced* by this backport — it is ancestral code that master removed and `202511` kept. What the backport does is make that implicit ordering invariant load-bearing on a branch master has no reason to keep compatible with it.

Commit 2 is therefore one line:

```python
# 202511 note: unlike master, this branch's _get_inner_leaf_xpaths()
# dereferences sy.root, so the data must be loaded even when the
# caller passes reload_config=False (e.g. the recursive
# __remove_dependents path). Load whenever nothing is loaded yet.
if reload_config or sy.root is None:
```

**Why not backport #4118 instead?** It is 32 insertions / 31 deletions across 7 files (4 of them CI and docs) and reaches outside GCU into `config/config_mgmt.py` and `sonic_package_manager/manager.py`, changes the libyang usage model on a stable branch, and adds a semgrep CI gate. That is a larger risk surface than one line inside GCU on a branch that is not moving to libyang3.

#### How to verify it

**This change is defensive, and I want to be explicit about that: no production failure has been reproduced.**

- An instrumented end-to-end `sort()` — a create-only `PORT`/`lanes` change (100G breakout) on a port referenced by an `ACL_TABLE` `ports` leaf-list and by `VLAN_MEMBER` — logs every `find_ref_paths` call. Across 14 calls the sort completes in 6 moves and **no call reaches a `reload_config=False` site with `sy.root` still unloaded**. Behaviour is byte-identical with and without this commit.
- A *direct* call — `find_ref_paths(path, config, reload_config=False)` on a freshly constructed `PathAddressing` — does raise the `AttributeError` above, on every commit of this stack from #3831 onward. But no natural `sort()` flow makes that call.

So the value here is not a crash fixed; it is removing a correctness property's dependence on statement order, before a future reorder, new caller, or new move generator turns it into a hard failure on this branch only.

Cost is nil. Full suite: **460 passed, 81 subtests**. On hardware, with and without this commit the patched arm is identical (0.73 s vs 0.74 s; 24 moves vs 24 moves).

#### Backport notes

- Manual conflict: `tests/generic_config_updater/patch_sorter_test.py`, third-argument signature adjustment.
- Commit 2 is `202511`-only and intentionally absent from master.
- Upstream `202511` should be aware of this hazard independently of this backport.

---

#### Stack-level verification

| Check | Result |
|---|---|
| Diff coverage, this PR | **91.3% (the 202511-only fix commit itself: 100%)** |

### DIFF
--- a/generic_config_updater/change_applier.py
+++ b/generic_config_updater/change_applier.py
@@ -5,11 +5,12 @@ import jsondiff
 import importlib
 import os
 import tempfile
+import time
 from collections import defaultdict
 from swsscommon.swsscommon import ConfigDBConnector
 from sonic_py_common import multi_asic
 from .gu_common import GenericConfigUpdaterError, genericUpdaterLogging
-from .gu_common import get_config_db_as_json
+from .gu_common import JsonChange
 
 SCRIPT_DIR = os.path.dirname(os.path.realpath(__file__))
 UPDATER_CONF_FILE = f"{SCRIPT_DIR}/gcu_services_validator.conf.json"
@@ -64,8 +65,8 @@ class DryRunChangeApplier:
     def __init__(self, config_wrapper):
         self.config_wrapper = config_wrapper
 
-    def apply(self, change):
-        self.config_wrapper.apply_change_to_config_db(change)
+    def apply(self, current_configdb: dict, change: JsonChange) -> dict:
+        return self.config_wrapper.apply_change_to_config_db(current_configdb, change)
 
     def remove_backend_tables_from_config(self, data):
         return data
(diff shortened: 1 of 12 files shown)


vaibhavhd (approved):

PR #4815, opened 2026-08-27
[generic_config_updater] Cache loadData() calls to reduce redundant YANG parsing (#4476) [202511]

rimunagala:
> **PR 6 of 8** in a `202511` GCU backport stack. **Draft until the PRs ahead of it merge.**
>
>

#### Why I did it

Cherry-pick of #4476. Every move validation re-parses the entire config through libyang. On `202511` this shows up as the exact invariant `loadData calls == 2 × moves`, measured on real hardware (38 = 2×19, 20 = 2×10, 114 = 2×57). At 512 ports that is **1020** parses; upstream reported ~1027.

#### How I did it

Cherry-pick plus one manual conflict resolution — **please review this hunk specifically.**

#4476 adds md5-based caching *inside* the `if reload_config:` block introduced by #3831. On this branch that condition was widened to `if reload_config or sy.root is None:` (see PR 1). The resolution keeps the widened condition and takes #4476's caching body:

```python
if reload_config or sy.root is None:
    <#4476 md5 hash-caching body>
```

The optimisation is fully preserved — the hash check still skips redundant loads. Only the entry condition differs from master, and only because `202511` lacks #4118.

#### How to verify it

`loadData` count collapses to **2**, independent of scale (512-port case: 1020 → 2). Confirmed on hardware in all four leaf-list scenarios. Full suite green: 460 passed, 81 subtests.

#### Backport notes

- Manual conflict as described above. This is the one hunk in the stack that intentionally differs from master.

---

#### Stack-level verification

| Check | Result |
|---|---|
| Diff coverage, this PR | **100%** |

### DIFF
--- a/generic_config_updater/gu_common.py
+++ b/generic_config_updater/gu_common.py
@@ -9,6 +9,7 @@ import yang as ly
 import copy
 import re
 import os
+import hashlib
 from sonic_py_common import logger, multi_asic
 from enum import Enum
 from functools import cmp_to_key
@@ -81,6 +82,8 @@ class ConfigWrapper:
         self.scope = scope
         self.yang_dir = YANG_DIR
         self.sonic_yang_with_loaded_models = None
+        self._validate_config_cache = {}
+        self._currently_loaded_hash = None
 
     def get_config_db_as_json(self):
         return get_config_db_as_json(self.scope)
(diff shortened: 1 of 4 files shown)


vaibhavhd (approved):

PR #4816, opened 2026-08-27
Fix GCU patch sorter crash on create-only field change referenced by a leaf-list (#4668) [202511]

rimunagala:
> **PR 7 of 8** in a `202511` GCU backport stack. **Draft until the PRs ahead of it merge.**
>
>

#### Why I did it

Cherry-pick of #4668 (Brad House, Nexthop). A create-only PORT field change (e.g. `lanes` during a breakout) transiently removes the port from a multi-member leaf-list such as `ACL_TABLE.ports`, leaving a dangling leafref. `list.index` then raises, and the exception escapes `RemoveCreateOnlyDependencyMoveValidator._validate_member`, aborting the entire sort:

```
ValueError: 'Ethernet312' is not in list
```

Any GCU / `config replace` based breakout on a realistic config fails and auto-rolls-back. (The native `config interface breakout` path bypasses GCU and is unaffected.)

#### How I did it

Clean cherry-pick. Catches `ValueError` / `KeyError` at the reference-resolution call and returns `False`, so the DFS backtracks to a valid ordering rather than aborting.

#### How to verify it

**Mock-free reproduction of the exact failure.** The trigger is a *mismatch*: the libyang singleton still holds a config in which the port is a leaf-list member, while the simulated config passed for xpath→configdb-path conversion has already had it removed. libyang therefore yields a leafref xpath `ports[.='Ethernet0']`, and resolving that against the simulated config performs `list.index('Ethernet0')`. That mismatch is precisely what `reload_config=False` produces — the mode `RemoveCreateOnlyDependencyMoveGenerator.__remove_dependents` uses.

Priming the singleton with the config where the port *is* a member, then calling with the config where it is *not*, using no `unittest.mock` at all:

```
part 1 — the real PathAddressing.find_ref_paths
  before : ValueError: 'Ethernet0' is not in list
  after  : ValueError: 'Ethernet0' is not in list      (library behaviour, unchanged)

part 2 — RemoveCreateOnlyDependencyMoveValidator._validate_member
  before : PROPAGATED ValueError  -> aborts the entire sort
  after  : RETURNED False         -> move rejected, DFS backtracks
```

So the exception is genuine, not injected, and this commit is exactly what converts "abort the sort" into "reject this ordering".

The upstream regression test also passes/fails as its PR claims, run against this branch's stack (test pinned at `90927acd`, implementation varied):

```
before (impl @ b60cd69f) : 1 FAILED — ValueError: 'Ethernet312' is not in list
after  (impl @ 90927acd) : 1 passed
```

A broader revert test on the assembled stack (revert only this source hunk, keep its tests) gives **4 passed → 4 FAILED**.

#### Caveats — please read

- The reproduction above drives `_validate_member` directly at its documented precondition. I did **not** find a small end-to-end `PatchSorter.sort()` config that reaches it: six scenarios (2-member and 4-member leaf-lists, multi-port lane changes, two ACL tables sharing a port, breakout-shaped changes, single-member list) produced **zero** internal raises, at three different stack positions. On those the DFS finds a valid ordering without ever exploring the dangling-leafref intermediate.
- I had hypothesised that #4478's leaf-list batching was masking the trigger. **That is disproven** — the counts are zero with and without #4478 present.
- No hardware A/B. Its only device scenario is the create-only/breakout path, which times out (>1500 s) on **both** stock and patched at 56-port scale, so no completed comparison exists.
- Net: the failure mode and the fix are proven on `202511`; how easily a real switch config reaches it is not established. Reviewer judgement on whether that is sufficient.

#### Backport notes

Clean; no conflicts.

---

#### Stack-level verification

| Check | Result |
|---|---|
| Diff coverage, this PR | **100%** |

### DIFF
--- a/generic_config_updater/patch_sorter.py
+++ b/generic_config_updater/patch_sorter.py
@@ -748,6 +748,7 @@ class RemoveCreateOnlyDependencyMoveValidator:
     def __init__(self, path_addressing):
         self.path_addressing = path_addressing
         self.create_only_filter = CreateOnlyFilter(path_addressing).get_filter()
+        self.logger = genericUpdaterLogging.get_logger(title="Patch Sorter - RemoveCreateOnly")
 
     def validate(self, group: JsonMoveGroup, diff, simulated_config) -> Tuple[bool, Optional[str]]:
         # Note: group is not used by this validator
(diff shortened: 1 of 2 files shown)


vaibhavhd (approved):

PR #4817, opened 2026-08-27
[generic_config_updater]: Remove empty leaf-lists from the simulated ConfigDB target (#4731) [202511]

rimunagala:
> **PR 8 of 8** in a `202511` GCU backport stack. **Draft until the PRs ahead of it merge.**
>
>

#### Why I did it

Cherry-pick of #4731. **The upstream rationale does not apply to `202511`, but the fix is still needed — for a different reason.**

Upstream describes a CONFIG_DB corruption that only reproduces on libyang3 (`import libyang as ly`). `202511` uses libyang 1.0.73 via SWIG (`import yang as ly`) — verified against the installed `202511` wheel (`sonic_yang_ext.py:5`). **So the corruption is not reachable here.**

However, emptying a leaf-list on `202511` fails hard regardless:

```
GenericConfigUpdaterError: There is no possible sorting
```

#### How I did it

Clean cherry-pick.

#### How to verify it

Isolated revert test — remove only this commit from the assembled stack:

```
clean 202511      : EXCEPTION "There is no possible sorting"
full stack        : OK — 2 moves
stack minus #4731 : EXCEPTION "There is no possible sorting"
```

So this commit is precisely what makes the operation work. On hardware, emptying a leaf-list went from **TIMEOUT >300 s** on stock to **14.09 s / 24 moves** patched.

#### Backport notes

Clean. Worth flagging to the branch manager that this is being taken for a different reason than its upstream title suggests.

---

#### Stack-level verification

| Check | Result |
|---|---|
| Diff coverage, this PR | **94.4%** |

### DIFF
--- a/generic_config_updater/generic_updater.py
+++ b/generic_config_updater/generic_updater.py
@@ -113,7 +113,7 @@ class PatchApplier:
 
         # Generate target config
         self.logger.log_notice(f"{scope}: simulating the target full config after applying the patch.")
-        target_config = self.patch_wrapper.simulate_patch(patch, old_config)
+        target_config = self.patch_wrapper.simulate_config_db_patch(patch, old_config)
 
         # Validate all JsonPatch operations on specified fields
         self.logger.log_notice(f"{scope}: validating all JsonPatch operations are permitted on the specified fields")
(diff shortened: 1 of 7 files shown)


vaibhavhd (approved):

PR #4820, opened 2026-08-27
[202511] Generic Configuration Updater (GCU) performance enhancements (backport of #3831)

bhouse-nexthop:
#### What I did

Manual cherry-pick of #3831 ("Generic Configuration Updater (GCU) performance enhancements") to the **202511** branch, as requested by @vmittal-msft — the automated cherry-pick could not be created because of merge conflicts.

Original PR: https://github.com/sonic-net/sonic-utilities/pull/3831
Original commit: bd3de9da6394e8cfd3379a144a70b70e411cd71c

Recap of the change being backported:

Generic Configuration Updater is extremely slow, using the python profiler it was possible to determine the worst offenders where changes could be made without affecting the overall algorithm and HLD design documentation.

* Prevent `copy.deepcopy()` calls where possible
* Don't run validation twice back to back
* Move configdb path <> xpath conversion logic to sonic-yang-mgmt where it belongs and enhance it to support schema conversion (not just data) and add caching.
* Sort table keys by the number of schema backlinks and must statements for the node to try better guess the right order of the patches to generate rather than doing it in alphabetical order which is likely to cause validation failures.
* Add ability to Group patches together in some commits where its known they will not cause issues, these are things like grouping parameter updates under the same key.
* When applying changes, do not re-read the configuration from redis twice between each applied patch (this is **extremely** slow, and actually hid a race condition). We are mutating the configuration and a lock is held so we know the expected before and after. There is still a final validation to ensure something didn't go sideways.

Dependencies (all already required by the original PR, `sonic-yang-mgmt` side):
 * sonic-yang-mgmt enhancements: https://github.com/sonic-net/sonic-buildimage/pull/22254
 * sonic-yang-mgmt parse uses/grouping: https://github.com/sonic-net/sonic-buildimage/pull/21907

Fixes https://github.com/sonic-net/sonic-buildimage/issues/22372

#### How I did it

`git cherry-pick -x -m 1 bd3de9da` onto `202511`, then resolved the conflicts against the three GCU changes that are present in 202511 but were not in the original commit's base. Each resolution keeps the 202511 behaviour **and** the change from #3831:

1. `generic_config_updater/gu_common.py` — `validate_config_db_config()`
   Conflicts with #4484 ("GCU: silence transient YANG leafref ERR during patch-sort search"). Resolution keeps 202511's `quiet=True` argument (and its explanatory comment) while applying this PR's removal of the redundant `copy.deepcopy()` and of the second `validate_data_tree()` pass:
   ```python
   sy.loadData(config_db_as_json, quiet=True)
   ```
   This matches the state of `master` today, which carries both changes.

2. `generic_config_updater/patch_sorter.py` — `LowLevelMoveGenerator._traverse_current_list()`
   Conflicts with #4486 ("[GCU][MGGMT_INTERFACE] Fix the remove forced_mgmt_routes unexpected value"). Resolution keeps 202511's `len(ptr) <= 1` condition with this PR's `JsonMoveGroup()` wrapping of the yielded `JsonMove`:
   ```python
   if len(ptr) <= 1:
       yield JsonMoveGroup(JsonMove(self.diff, OperationType.REMOVE, current_tokens))
   ```
   This also matches `master` today.

3. `tests/generic_config_updater/patch_sorter_test.py` — `test_validate__passes_quiet_true_to_config_wrapper`
   This test was added by #4484 and so does not exist in the original commit; git merged it in cleanly but it was left calling the old two-argument `FullConfigMoveValidator.validate(move, diff)`, which this PR changes to `validate(move, diff, simulated_config)`. Updated the call to the new signature; the assertion the test guards (`validate_config_db_config` called with the simulated config and no `quiet` kwarg) is unchanged, so the regression guard still does its job.

Everything else applied cleanly. Post-base 202511 GCU changes that did not conflict were verified to still be present after the cherry-pick: the per-VRF `BGP_PEER_RANGE` duplicate check (#4215), the telemetry restart on port speed change (#4429), the buffer profile object-level validator (#4585), and the various HWSKU/validator config additions.

#### How to verify it

Run the sonic-utilities unit tests, in particular `tests/generic_config_updater/`.

The change was verified on hardware on the master version of this PR (see #3831 for the profiling numbers: ~3x faster `config replace -d`, ~7.5x faster `config replace`); the 202511 content is identical apart from the three conflict resolutions listed above.

N/A

N/A

### DIFF
--- a/generic_config_updater/change_applier.py
+++ b/generic_config_updater/change_applier.py
@@ -5,11 +5,12 @@ import jsondiff
 import importlib
 import os
 import tempfile
+import time
 from collections import defaultdict
 from swsscommon.swsscommon import ConfigDBConnector
 from sonic_py_common import multi_asic
 from .gu_common import GenericConfigUpdaterError, genericUpdaterLogging
-from .gu_common import get_config_db_as_json
+from .gu_common import JsonChange
 
 SCRIPT_DIR = os.path.dirname(os.path.realpath(__file__))
 UPDATER_CONF_FILE = f"{SCRIPT_DIR}/gcu_services_validator.conf.json"
@@ -64,8 +65,8 @@ class DryRunChangeApplier:
     def __init__(self, config_wrapper):
         self.config_wrapper = config_wrapper
 
-    def apply(self, change):
-        self.config_wrapper.apply_change_to_config_db(change)
+    def apply(self, current_configdb: dict, change: JsonChange) -> dict:
+        return self.config_wrapper.apply_change_to_config_db(current_configdb, change)
 
     def remove_backend_tables_from_config(self, data):
         return data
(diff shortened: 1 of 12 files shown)


bhouse-nexthop:
@vmittal-msft here is the manual 202511 cherry-pick you asked for on #3831.

There were three spots where 202511 has GCU changes that were not in the original commit's base — `quiet=True` in `validate_config_db_config()` (#4484), `len(ptr) <= 1` in `_traverse_current_list()` (#4486), and the regression test #4484 added, which was still calling the old two-argument `FullConfigMoveValidator.validate()`. All three are resolved to keep both the 202511 behaviour and this change; the first two now match what `master` looks like today. Details are in the PR description and in the commit message.

PTAL, thanks!

vmittal-msft:
@bhouse-nexthop please add label "Tested on 202511" once done.

PR #4832, opened 2026-09-01
syslog: Replace shell-based container lookup with exact matching

ashutosh-agrawal:
#### What I did

Updated the syslog rate-limit enable and disable commands to check running containers by exact name.

The existing implementation inserted each configured feature name into a shell-based `grep` expression. Container discovery only requires exact name matching, so interpreting the name as part of a shell command is unnecessary and can produce unexpected behavior for unusual input.

#### How I did it

Run `docker ps` with structured arguments, parse the returned container names into a set, and compare each configured feature name directly against that set.

This removes the shell and regular-expression dependency while preserving the existing behavior for valid feature names and containers that are not running.

#### How to verify it

`python3 -m pytest tests/syslog_test.py -q`

Result: 41 passed.

N/A

N/A

### DIFF
--- a/config/syslog.py
+++ b/config/syslog.py
@@ -561,6 +561,15 @@ def get_feature_names_to_proceed(db, service_name, namespace):
     return feature_list
 
 
+def get_running_container_names():
+    """Return the exact names of currently running containers."""
+    output, _ = clicommon.run_command(
+        ['docker', 'ps', '-f', 'status=running', '--format', '{{.Names}}'],
+        return_cmd=True
+    )
+    return set(output.splitlines()) if output else set()
+
+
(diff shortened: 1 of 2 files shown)


Junchao-Mellanox:
Does this new approach work for multi-ASIC which has container name like "swss0, swss1"?

ashutosh-agrawal:
> Does this new approach work for multi-ASIC which has container name like "swss0, swss1"?

@Junchao-Mellanox Yes, the approach works for multi-ASIC because it checks the already-expanded names swss0, swss1, etc., not the logical CONFIG_DB name swss. The new test now explicitly protects that behavior.

Junchao-Mellanox (approved):

PR #4833, opened 2026-09-01
[fdb]: Add CLI to select the MAC synchronization mode

tahmed-dev:
#### Why I did it

MAC (FDB) state can now be synchronized with FRR either through the Linux kernel
bridge FDB or directly over the FPM channel. Operators need a supported way to
select which transport is in use, and a way to roll back without reimaging.

#### How I did it

Add `config fdb mac-sync-mode <kernel|fpm>`, which writes
`FDB_SYNC|global mac_sync_mode`. The value is constrained to the two modes by
click, and the write goes through `ValidatedConfigDBConnector` so it is checked
against the YANG model.

The default is `kernel`, so behaviour is unchanged unless explicitly set.

#### How to verify it

```
make target/python-wheels/trixie/sonic_utilities-1.2-py3-none-any.whl
# 5409 passed
```

On a device:

```
config fdb mac-sync-mode fpm
config save -y
redis-cli -n 4 hgetall "FDB_SYNC|global"
```

Note that `config save -y` is required; without it the setting is lost on
reboot.

#### Dependency / merge order

Requires the `sonic-swss-common` FDB_SYNC schema PR to merge **first**, for
`CFG_FDB_SYNC_TABLE_NAME`.

#### Which release branch to backport

None requested at this time.

#### Tested branch

Built and tested against `sonic-net/sonic-buildimage` master.

Add `config fdb mac-sync-mode` to select how MAC state is synchronized with FRR.

### DIFF
--- a/config/main.py
+++ b/config/main.py
@@ -10667,6 +10667,31 @@ def motd(message):
 
 
 #
+# 'fdb' group ('config fdb ...')
+#
+
+@config.group()
+def fdb():
+    """FDB (MAC) configuration tasks"""
+    pass
+
+
+@fdb.command('mac-sync-mode')
+@click.argument('mode', metavar='<kernel|fpm>', required=True, type=click.Choice(['kernel', 'fpm']))
+@clicommon.pass_db
+@click.pass_context
+def mac_sync_mode(ctx, db, mode):
+    """Select how MAC (FDB) state is synchronized with FRR"""
+
+    config_db = ValidatedConfigDBConnector(db.cfgdb)
+    try:
+        config_db.mod_entry(swsscommon.CFG_FDB_SYNC_TABLE_NAME, 'global',
+                            {'mac_sync_mode': mode})
+    except (ValueError, JsonPatchConflict) as e:
+        ctx.fail("Failed to save to ConfigDB. Error: {}".format(e))
+
+
+#
 # 'vnet' group ('config vnet ...')
 #
 
--- a/scripts/fast-reboot
+++ b/scripts/fast-reboot
@@ -518,6 +518,7 @@ function backup_database()
                                           and not string.match(k, 'VXLAN_TUNNEL_TABLE|') \
                                           and not string.match(k, 'BUFFER_MAX_PARAM_TABLE|') \
                                           and not string.match(k, 'STORAGE_INFO|') \
+                                          and not string.match(k, 'FDB_SYNC_TABLE|') \
                                           and not string.match(k, 'FAST_RESTART_ENABLE_TABLE|') then
                 redis.call('del', k)
             end
--- a/show/main.py
+++ b/show/main.py
@@ -12,6 +12,7 @@ import utilities_common.multi_asic as multi_asic_util
 from importlib import reload
 from natsort import natsorted
 from sonic_py_common import device_info
+from swsscommon import swsscommon
 from swsscommon.swsscommon import SonicV2Connector, ConfigDBConnector
 from tabulate import tabulate
 from utilities_common import util_base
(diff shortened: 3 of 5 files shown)


config/main.py
```diff
@@ -10663,6 +10663,31 @@ def motd(message):
...
+@fdb.command('mac-sync-mode')
+@click.argument('mode', metavar='<kernel|fpm>', required=True, type=click.Choice(['kernel', 'fpm']))
+@clicommon.pass_db
+@click.pass_context
+def mac_sync_mode(ctx, db, mode):
+    """Select how MAC (FDB) state is synchronized with FRR"""
+
+    config_db = ValidatedConfigDBConnector(db.cfgdb)
+    try:
+        config_db.mod_entry(swsscommon.CFG_FDB_SYNC_TABLE_NAME, 'global',
```
pbrisset, line 10685: This command writes FDB_SYNC through ValidatedConfigDBConnector, but this PR does not add the corresponding YANG/schema support.

On a real device with config validation enabled, mod_entry(CFG_FDB_SYNC_TABLE_NAME, "global", ...) may fail if the sonic-fdb-sync/FDB_SYNC schema is not already present in the image. The unit tests use the mock DB path, so they may not catch that integration failure.

Can we either add the required YANG/schema change in this PR, or document/chain the required dependency so this command cannot merge ahead of schema support?

scripts/fast-reboot
```diff
@@ -510,6 +510,7 @@ function backup_database()
                                           and not string.match(k, 'VXLAN_TUNNEL_TABLE|') \
                                           and not string.match(k, 'BUFFER_MAX_PARAM_TABLE|') \
                                           and not string.match(k, 'STORAGE_INFO|') \
+                                          and not string.match(k, 'FDB_SYNC_TABLE|') \
```
pbrisset, line 513: This adds FDB_SYNC_TABLE to the STATE_DB keys preserved across fast reboot, which is important because fpmsyncd uses it to persist the MAC replay generation.

Can we add a small test or explicit coverage for this backup filter? Otherwise a future rename/regression could silently drop the generation key and allow replay generation reuse across restart.

show/main.py
```diff
@@ -1283,6 +1284,27 @@ def aging_time(ctx):
...
         else:
             click.echo("Aging time not configured for the {}".format(key.split(':')[-1]))
+
+
+#
+# 'fdb' group ("show fdb ...")
+#
+
+@cli.group(cls=clicommon.AliasedGroup)
+def fdb():
```
pbrisset, line 1294: Existing FDB operational commands live under "show mac", including "show mac aging-time", while this PR adds a separate "show fdb mac-sync-mode" tree.

If "show fdb" is intentional to match the HLD/config command, this is fine, but it creates two user-facing command families for FDB/MAC state. Can we align this CLI? and consider adding this under the existing "show mac" group as well?

tests/fdb_mac_sync_mode_test.py
```diff
@@ -0,0 +1,127 @@
...
+
+@pytest.fixture
+def cli_db_connection(enable_click_ut_mode):
+    db = Db()
+    return CliRunner(), db
+
+
+class TestFdbMacSyncModeConfig:
+    @pytest.mark.parametrize("mac_sync_mode", ["kernel", "fpm"])
+    def test_mac_sync_mode_config(self, cli_db_connection, mac_sync_mode):
```
pbrisset, line 34: The config tests verify the mock DB write, but they do not exercise YANG/config validation for the new FDB_SYNC table.

Please add coverage that would fail if FDB_SYNC is missing from the schema, or make the PR dependency on the schema update explicit.

tests/fdb_mac_sync_mode_test.py
```diff
@@ -0,0 +1,127 @@
...
+
+        result = runner.invoke(config.commands["fdb"].commands["mac-sync-mode"], ["kernel"], obj=db)
+        assert result.exit_code == 0, (
+            f"Got exit code {result.exit_code} - {result.output}, expected 0"
+        )
+        assert db.cfgdb.get_table(FDB_SYNC_TABLE)[FDB_SYNC_KEY][MAC_SYNC_MODE_FIELD] == "kernel"
+
+
+class TestFdbMacSyncModeShow:
+    def test_mac_sync_mode_show_default(self, cli_db_connection):
```
pbrisset, line 95: show defaults to "kernel" when FDB_SYNC is absent. It would also be useful to cover FDB_SYNC|global existing without mac_sync_mode, since partial config can happen during manual edits or migrations.

Expected output should still be "kernel".

pbrisset (changes requested):

PR #4834, opened 2026-09-03
7060CX device state migration via upgrade

justin-wong-ce:
#### What I did
Some 7060 devices need to have the ability to migrate its device state to make it compatible with future Arista driver code.

This migration only applies to 7060CX devices that has warm-upgraded from the 201811 release.

#### How I did it
The migration ensures the file /host/.system-prefdl exists on the machine. This is done by either copying from a cache file, or reading directly from EEPROM.

#### How to verify it

The expected behavior occurs on each of the 3 below scenarios:
(1) /host/.system-prefdl exists
- no action

(2) /host/.system-prefdl does not exist, /etc/sonic/.syseeprom exists
- Copy /etc/sonic/.syseeprom to /host/.system-prefdl, it has been tested
  that it is ok to direct copy even if the contents are slighlty
  different.

(3) /host/.sytem-prefdl and /etc/sonic/.syseeprom does not exist:
- Generate a /host/.system-prefdl by reading directly from EEPROM

This logic was tested on the following images:
- 20260510.10 -> 20260510.13

#### Backport

Needs backport to 202605.

### DIFF
--- a/scripts/arista_7060_prefdl_gen.py
+++ b/scripts/arista_7060_prefdl_gen.py
@@ -1,0 +1,11 @@
+from arista.components.eeprom import I2cEeprom
+from arista.core.platform import loadPrerequisites
+from arista.core.types import I2cBus
+
+IDENT_BUS_NAME = 'SMBus PIIX4 adapter port 1 at 0b20'
+
+loadPrerequisites()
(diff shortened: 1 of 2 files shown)


justin-wong-ce:
FYI @Ryangwaite , @vaibhavhd

PR #4840, opened 2026-09-09
[show] Use 'comm' in 'processes summary' to avoid leaking process args

lakki12233:
### What
`show process summary` runs `ps -eo pid,ppid,cmd,%mem,%cpu`. The `cmd` field prints
each process's full command line, which can include secrets passed as arguments
(tokens, passwords, API keys).

### Fix
Use `comm` instead of `cmd` — it shows only the program name, never arguments, so
credentials are no longer exposed. Same columns; `show process cpu` / `memory` are
unaffected. Unit test updated to match.

Note: `comm` is capped at 15 chars, so long names appear truncated (cosmetic only).


lakki12233:
@vvolam @qiluo-msft @jleveque could you please help review this security fix when you get a chance?  Thanks!

PR #4776, opened 2026-08-13
Speed up techsupport collection

dovsianko-nvda:
generate_dump collects FEC histogram counters separately for every interface in two counter snapshots. On full-split systems, executing these commands sequentially significantly increases techsupport time.

Run the per-interface commands in bounded batches of eight and wait only for their recorded process IDs. Preserve the existing per-interface output files and unmatched-glob handling.

#### What I did

Reduced `show techsupport` execution time on full-split systems by
collecting per-interface FEC histogram counters in parallel.

The existing per-interface files and two counter snapshots are preserved.

Before, for a full-split 512 port config:

   ```bash
   time show techsupport
   ...
   real    9m48.540s
   user    0m0.327s
   sys     0m0.216s
   ```
After:
   ```bash
   time show techsupport
   ...
   real    4m31.055s
   user    0m0.318s
   sys     0m0.227s
   ```
#### How I did it

Updated `save_counter_snapshot()` to run FEC histogram commands in
bounded batches of eight.

Each background process ID is recorded and waited for before starting
the next batch. The final partial batch is also waited for, ensuring all
FEC files are complete before queue counters and `COUNTERS_DB` are
collected.

#### How to verify it

1. Validate the script syntax:

   ```bash
   bash -n /usr/local/bin/generate_dump
2. Run techsupport on a healthy full-split system:

   ```bash
    time show techsupport
 
3. Confirm the archive is generated successfully.

4.  Confirm that both snapshots contain one FEC histogram file per
interface:

   ```bash
    tar -tf <techsupport.tar.gz> |
    grep -E 'interface\.counters\.fec-histogram\.Ethernet.*\.[12]$' |
    wc -l
   ```
   For a system exposing 514 Ethernet interfaces, the expected total is
   1,028 files: 514 files for each snapshot.

5. Compare the execution time with the sequential implementation.

#### Tested branch
 
- [x]  202511
- [x]  202605

## Summary by CodeRabbit

* **Performance Improvements**
  * FEC histogram data is now collected more efficiently through concurrent processing, reducing snapshot generation time.

PR #4845, opened 2026-09-10
[Cherry-pick][202511]Speed up techsupport collection

dovsianko-nvda:
generate_dump collects FEC histogram counters separately for every interface in two counter snapshots. On full-split systems, executing these commands sequentially significantly increases techsupport time.

Run the per-interface commands in bounded batches of eight and wait only for their recorded process IDs. Preserve the existing per-interface output files and unmatched-glob handling.

#### What I did

Cherry-pick of PR #4776


dovsianko-nvda:
Original PR https://github.com/sonic-net/sonic-utilities/pull/4776

dovsianko-nvda:
Hi @vaibhavhd, I’ve created this PR to cherry-pick the change into the `202511` branch. Could you please review and approve it? Thanks!

PR #4598, opened 2026-06-09
Add reusable disk-space validation helper for image installation and reboot

rameshraghupathy:
Add reusable disk-space validation helper for image installation and reboot

Fixes:  #4661
Fixes: https://github.com/sonic-net/sonic-buildimage/issues/28734

#### What this PR does

This PR introduces a reusable disk-space validation helper that validates sufficient free disk space before critical operations begin and prevents the operation from proceeding when the required free space is not available.

The helper is designed to be reusable across multiple workflows, avoiding duplicated disk-space validation logic.

The current implementation supports:

Regular switch image installation
SmartSwitch switch image installation
Local DPU image installation
Remote single-DPU image installation
Remote multi-DPU image installation
Switch reboot
DPU reboot (reboot -d)

The helper automatically detects whether it is running on a switch or DPU and applies the appropriate disk-space threshold.

##### Platform configuration

Disk-space validation is opt-in. The corresponding check is performed only when the appropriate threshold is configured in platform.json.
```
{
    "min_free_disk_in_gb_for_switch_image": 16,
    "min_free_disk_in_gb_for_dpu_image": 12,
    "min_free_disk_in_gb_for_switch_reboot": 4,
    "min_free_disk_in_gb_for_dpu_reboot": 4
}
```
Skip disk-space validation when the corresponding threshold is not configured or is invalid, preserving the existing
behavior. Log an error when an invalid configuration is encountered.

##### Implementation

Added a reusable helper:

utilities_common/image_disk_space.py

The module provides reusable APIs for:

1. Local image installation validation
2. Remote DPU image installation validation
3. Local reboot validation
4. Remote DPU reboot validation
5. Automatic runtime selection based on the current execution environment

The implementation includes:

1. Runtime platform detection
2. Platform-specific threshold lookup from platform.json
3. Safe default values for image installation
4. Optional reboot thresholds
5. Local filesystem validation
6. Remote DPU validation over SSH
7. Configurable SSH options
8. Robust parsing of remote df output
9. Fail-fast behavior
10. Clear logging for configuration, connectivity, parsing, and insufficient disk-space failures
##### Integration

The helper is integrated into:

sonic-installer install
reboot
reboot -d

Each workflow performs disk-space validation before beginning any operation that could leave the filesystem or boot state in an inconsistent state.

The implementation is prevention-only. It does not perform automatic cleanup, recovery, or file deletion.

##### Example
```
if not check_image_install_free_disk_space():
    raise click.ClickException(
[...]

PR #4850, opened 2026-09-15
Read SWAP size and total memory threshold from platform.json

SahilChaudhari:
Read SWAP size and total memory threshold from platform.json

#### What I did

`sonic-installer install` sets up temporary SWAP memory for the duration of the installation. Its size comes from `--swap-mem-size` (default 1024 MiB) and the allocation is gated on `--total-mem-threshold` (default 2048 MiB): `SWAPAllocator` only creates a swapfile when `MemTotal + SwapTotal` is below that threshold.

Neither default suits every platform, and today the only remedy is for the operator to pass both options by hand on every install.

The gate is the more serious of the two. A platform whose system memory sits above the 2048 MiB default never satisfies the condition, so **no swapfile is created at all**, regardless of the size requested. On such a platform `--swap-mem-size` alone has no effect; `--total-mem-threshold` must also be raised, which is neither obvious nor discoverable.

This PR lets a platform declare its own minimums in `platform.json`, so the correct values are used automatically while remaining fully opt-in.

##### Platform configuration

Both keys are optional. A missing or invalid key preserves the existing behaviour exactly.

```
{
    "min_swap_mem_size_in_mb": 3072,
    "min_total_mem_threshold_in_mb": 8192
}
```

The value requested on the command line is used as-is, and **only the platform minimum can raise it**:

1. the value requested on the command line, or the built-in default if none was requested,
2. raised to the platform minimum from `platform.json`, if one is configured.

So a platform that needs more than the default gets it without operator intervention, while an operator asking for more than the platform minimum is still honoured — as is an operator deliberately asking for less than the built-in default. The keys are floors, not overrides, and the built-in default is not itself a floor over an explicit request.

#### How I did it

Added a helper:

`utilities_common/swap_mem.py`

providing:

1. `get_min_swap_mem_size_in_mb()` — optional platform minimum SWAP size
2. `get_min_total_mem_threshold_in_mb()` — optional platform minimum threshold
3. `resolve_swap_mem_size()` — effective SWAP size
4. `resolve_total_mem_threshold()` — effective total memory threshold

Both resolvers share a single `_resolve()` implementation, so the two keys cannot diverge in behaviour.

The lookup reuses the `platform.json` parsing and positive-integer validation introduced in #4598 rather than adding a second reader. The two helpers there are exposed under public names (`load_platform_json`, `get_optional_positive_int`) for that purpose; the private names are unchanged.

The SWAP defaults are lifted to module level (`SWAP_MEM_SIZE_DEFAULT`, `TOTAL_MEM_THRESHOLD_DEFAULT`) with the `SWAPAllocator` class attributes referencing them. This is required because `tests/test_sonic_installer.py` patches out `SWAPAllocator`; reading `SWAPAllocator.SWAP_MEM_SIZE` at the call site would yield a `MagicMock` and fail the `max()` comparison with `'>' not supported between instances of 'int' and 'MagicMock'`.

The implementation is resolution-only. It does not change how SWAP is allocated, when it is allocated, or the behaviour of `--skip-setup-swap`.

##### Integration

The helper is integrated into `sonic-installer install`. Both values are resolved immediately before `SWAPAllocator` is constructed, and only when swap setup is enabled, so `--skip-setup-swap` never reads `platform.json`. A message is logged only when the platform actually raises a value above what would otherwise have been used:

```
Using SWAP memory size of 3072 MiB required by the platform
Using total memory threshold of 8192 MiB required by the platform
```

##### Unit tests

Added `tests/swap_mem_test.py` — 30 tests covering, for both keys:

- key present, key missing, and invalid values (`0`, `-1`, `"abc"`, `null`)
- missing `platform.json` file, and a `platform.json` whose top level is not an object
- each branch of the resolution: the platform minimum raising the requested value, a larger requested value being preserved, and an explicitly requested value below the built-in default being honoured
- no platform key present (requested value preserved)
- no requested value provided (built-in default used)
- the two keys are independent and do not read each other's value

Added two tests to `tests/test_sonic_installer.py` covering the integration itself: that `--skip-setup-swap` resolves nothing, and that the resolved values are what `SWAPAllocator` is constructed with. Existing installer tests are otherwise unaffected.

##### Functional testing

Validated on an AMD Pensando Elba DPU (`arm64-elba-asic-flash128-r0`), which has `min_swap_mem_size_in_mb: 3072` and `min_total_mem_threshold_in_mb: 8192`:

| Case | Command | Result |
| --- | --- | --- |
| Platform value used | `sonic-installer install  -y` | 3 GiB swapfile |
| Requested below platform minimum | `... --swap-mem-size 2048` | 3 GiB swapfile |
| Requested above platform minimum | `... --swap-mem-size 4096` | 4 GiB swapfile |
| Platform value below default | platform key `512` | 1 GiB swapfile (request unchanged) |
| CLI threshold above platform | `... --total-mem-threshold 16384` | 16384 MiB used |
| No keys configured | platform without either key | unchanged behaviour |

Before this change, every one of the above required `--total-mem-threshold 8192` on the command line or no swapfile was created.

The resolution rule was revised during review so that an explicitly requested value below the built-in default is honoured rather than being raised to it (`--swap-mem-size 512` with no platform key now yields 512 MiB, not 1024 MiB). Every case in the table above produces the same result as before that revision; the revised rule itself is covered by unit tests rather than re-run on hardware.

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511
- [x] 202605

Both options had to be supplied manually, and omitting `--total-mem-threshold` silently produced no swapfile at all:

```
admin@sonic:~$ sudo sonic-installer install sonic-<platform>.bin -y --total-mem-threshold 8192 --swap-mem-size 3072
Installing image SONiC-OS-... and setting it as default...
Setup SWAP memory
Command: mkswap /host/swapfile
Setting up swapspace version 1, size = 3 GiB (3221221376 bytes)
```

The platform minimums are applied automatically, and the source of each value is reported:

```
admin@sonic:~$ sudo sonic-installer install sonic-<platform>.bin -y
Installing image SONiC-OS-... and setting it as default...
Using SWAP memory size of 3072 MiB required by the platform
Using total memory threshold of 8192 MiB required by the platform
Setup SWAP memory
Command: mkswap /host/swapfile
Setting up swapspace version 1, size = 3 GiB (3221221376 bytes)
```

Platforms that do not configure either key see no change in output or behaviour.


seveerap79:
@vvolam  and @rameshraghupathy : Please can you review this PR.

utilities_common/swap_mem.py
```diff
@@ -0,0 +1,105 @@
...
+    candidates = [default]
+
+    if requested is not None:
+        candidates.append(requested)
+
+    platform_value = _get_platform_minimum(key, platform_json_path)
+    if platform_value is not None:
+        candidates.append(platform_value)
+
+    resolved = max(candidates)
```
rameshraghupathy: @SahilChaudhari To maintain 100% backward compatibility, consider the following change: resolved = requested if requested is not None else default

platform_value = _get_platform_minimum(...)

if platform_value is not None:
    resolved = max(resolved, platform_value)

return resolved

rameshraghupathy (commented):
@SahilChaudhari Mostly looks good. Please take care of the comments.

rameshraghupathy (approved):
@SahilChaudhari LGTM

vvolam:
@SahilChaudhari what happens if the existing platform.json does not have updated values? does the installation fail if the image requires higher swap size?

PR #4854, opened 2026-09-16
[202511] Create GCU wheel (backport of #4310)

rimunagala:
#### Why I did it

The standalone GCU wheel exists so the Generic Config Updater can be installed and run independently of the SONiC image, typically from a container using the host's Python environment.

That separation is worth having because the two move at very different speeds. Replacing an image on deployed devices is a scheduled, staged, validated operation with a long lead time; a GCU change is small, self-contained, and confined to one Python package. Coupled, every config-updater improvement — a bug fix, a performance regression, new behaviour — waits for the next image cycle. Decoupled, it ships on its own.

**202511 cannot currently produce a usable wheel.** Its packaging predates #4310 (`gcu/setup.py` + `scripts/gcu.py`) and differs from master in three ways, each of which independently breaks consumption:

1. **No entry point.** master's wheel installs a `gcu-standalone` console script (`gcu-standalone = generic_config_updater.main:main`). The 202511 wheel ships `scripts/gcu.py` instead, so anything that invokes `gcu-standalone` — or imports `generic_config_updater.main` — cannot run against a 202511-built wheel.
2. **It will not build in a clean environment.** `gcu/setup.py` calls `pkg_resources.get_distribution()` on `sonic-config-engine`, `sonic-py-common` and `sonic-yang-mgmt` and `exit(1)`s if any is absent — at *build* time, not install time. A minimal builder without SONiC packages already installed cannot produce the wheel at all. #4310 drops `install_requires` entirely; the virtualenv is created `--system-site-packages` and the wheel installed `--no-deps`, so runtime dependencies come from the host.
3. **It mutates the source tree while building.** `gcu/setup.py` `shutil.copytree()`s `../generic_config_updater` into the build directory. #4310 uses `package_dir` and leaves the tree untouched.

The net effect is that a branch which is actually deployed is the one branch whose GCU component cannot be built and delivered on its own.

**What this does not cover.** Only the `generic_config_updater` package ships in the wheel. Everything it depends on from the base image — the YANG models, `sonic-yang-mgmt`/libyang, and the wider config stack — stays where it is and still requires an image upgrade. This decouples updater logic from those components; it does not deliver them.

master already carries #4310, and 202511 is the only active branch that does not, so packaging behaviour currently differs per release branch for no functional reason.

#### How I did it

Straight cherry-pick of #4310 — no manual conflict resolution.

To be clear about the scope, this is **not** a pure file move. #4310 extracts the `config apply-patch` implementation out of `config/main.py` into a new `generic_config_updater/main.py` so the wheel can expose it standalone; `config/main.py` then delegates to it via `apply_patch_from_file`.

- `gcu/.coveragerc` → `generic_config_updater/.coveragerc` (100% rename); `gcu/pytest.ini` → `generic_config_updater/pytest.ini` (99% rename)
- delete `gcu/setup.py` (−99) and `scripts/gcu.py` (−568)
- add `generic_config_updater/setup.py` (+54) and `generic_config_updater/main.py` (+822)
- add `tests/generic_config_updater/main_test.py` (+965); `tests/config_test.py` updated (+77/−58)
- `config/main.py` (+25/−260): `get_all_running_config`, `append_emptytables_if_required`, `filter_duplicate_patch_operations`, `validate_patch` and `apply_patch_for_scope` move to the new module
- `azure-pipelines.yml`: build/publish from `generic_config_updater/` instead of `gcu/` (artifact name `gcu_wheels` unchanged)
- `generic_updater.py`: `+import jsonpatch`, `+JsonChange` import
- `field_operation_validators.py`: inline `DEFAULT_SUPPORTED_FECS_LIST` so the wheel does not need `utilities_common`

13 files, +1958/−989. No change to `gu_common.py`, `patch_sorter.py` or `change_applier.py`.

**Two behaviour changes carried over from #4310, called out explicitly since this is a release branch.** Both already exist on master; this PR does not introduce them uniquely to 202511.

1. `validate_patch()` now imports `sonic_yang_cfg_generator` lazily and returns `True` — skipping YANG validation — if it is unavailable, where previously a missing module was a hard import failure. This is required so the standalone wheel does not hard-depend on `sonic-config-engine`. On a switch the module is present, so there is no practical change there, but the new path is fail-open.
2. apply-patch progress/result logging moves from the SONiC syslog logger (`log.log_notice` / `log.log_error`) to a stdlib `logging.getLogger(__name__)`. No handler is configured, so the previous syslog breadcrumbs for apply-patch start/success are no longer emitted on the `config apply-patch` path.

I verified that this cherry-pick does not silently revert any 202511-specific work: it applied with zero conflicts, meaning the removed region of `config/main.py` matched master pre-#4310 exactly, and I AST-compared each moved function between the old and new locations to confirm the differences are limited to the two items above plus naming and formatting.

#### How to verify it

- `python3 -m build -n` inside `generic_config_updater/` produces `sonic_gcu-1.0.0-py3-none-any.whl` on a stock Python 3.11 with only `build` + `wheel` installed and no SONiC packages present. The same command inside `gcu/` on current 202511 fails with `sonic-config-engine is not found!`.
- The built wheel's `dist-info/entry_points.txt` contains `gcu-standalone = generic_config_updater.main:main`.
- Files inside the wheel are byte-identical to the branch — `gu_common.py` blob `63c8201be3ce3a5fae53710c6c5b94e43f742fd5`, `patch_sorter.py` blob `f7791db873b7af7cd78db18d07679d0fb7a6c67f` — i.e. the wheel carries the branch's current GCU sources byte-for-byte, with no divergence introduced by packaging.
- I also ran an A/B container build that builds the wheel from a branch checkout and installs it: the build fails on stock 202511 at the wheel step (`no pyproject.toml or setup.py`) and succeeds with this change applied, with `gcu-standalone` resolving correctly in the resulting image.
- `pytest tests/generic_config_updater/` (including the new `main_test.py`) and `pytest tests/config_test.py` run in this pipeline.

#### Which release branch to port

- [ ] 202605
- [x] 202511
- [ ] 202505


rimunagala:
## On-device validation (202511)

Validated on a real 202511 switch in addition to CI.

**Device**

| | |
|---|---|
| Testbed | `vms91-t0-7060x6-512-3` |
| DUT | `str5-7060x6-512-3` |
| Image | `202511` / `20251110.51` |
| HwSku | `Arista-7060X6-64PE-B-C512S2` (512 ports) |
| OS | Debian 13, Python 3.13.5 |

**Method** — wheel built from this PR's HEAD (`14734be`) and installed side-by-side via `pip3 install --no-deps --target`. No system packages touched, no image swap, every run `--dry-run`.

**Results**

1. **Entry point resolves** — `gcu-standalone --help` lists all 7 subcommands (`create-checkpoint`, `delete-checkpoint`, `list-checkpoints`, `apply-patch`, `replace`, `save`, `rollback`).
2. **Empty patch `[]`** — full applier pipeline runs, including `Patch Sorter - Strict: Validating target config according to YANG models`. 0 changes, exit 0, 3s.
3. **Real patch** `[{"op":"add","path":"/SYSLOG_SERVER/10.20.30.40","value":{}}]` — converted into 1 change, reported `** DryRun: Would apply ...`, exit 0, 3s.
4. **No config mutation** — `redis-cli -n 4 KEYS "SYSLOG_SERVER|*"` byte-identical before and after.

**On the fail-open YANG import disclosed above:** `from sonic_yang_cfg_generator import SonicYangCfgDbGenerator` imports successfully on this image, so the `ImportError` fallback in `validate_patch()` is never reached on a switch. The relaxation only applies where `sonic-config-engine` is genuinely absent — the packaging case it exists for.

vaibhavhd (approved):

PR #4856, opened 2026-09-17
[generic_config_updater]: Rewrite key-level removes that empty a ConfigDB table

### LINKED ISSUE #4852: [GCU] Need enhancement to support deleting keys one by one without empty-table restriction

Verma-Anukul:
#### What I did

Fixes #4852

`config apply-patch` now accepts per-key removes that empty a ConfigDB table (for example `/VLAN/Vlan3911`) and rewrites them to a table-level remove. Automation can delete keys one by one without inspecting running config or special-casing the empty-table restriction.

#### How I did it

After field-operation validation on the original simulated target, `PatchApplier.apply` calls `rewrite_patch_emptying_tables()`. Only `op=remove` with path `/TABLE/KEY` is rewritten, and only when every operation on that emptied table is of that form. Other patch shapes are unchanged. Protected paths such as `/LOOPBACK_INTERFACE/Loopback0` are still validated on the original transition.

#### How to verify it

Unit tests:
- `tests/generic_config_updater/gu_common_test.py::TestRewritePatchEmptyingTables`
- `tests/generic_config_updater/generic_updater_test.py` rewrite and Loopback0 tests

Device verification (VS):

```
admin@sonic:~$ show vlan config
Name        VID  Member    Mode
--------  -----  --------  ------
Vlan3911   3911
Vlan3912   3912
admin@sonic:~$ redis-cli -n 4 keys "VLAN*"
1) "VLAN|Vlan3911"
2) "VLAN|Vlan3912"
admin@sonic:~$ sudo config apply-patch vlan.json
Patch Applier: localhost: Patch: [{"op": "remove", "path": "/VLAN/Vlan3911"}, {"op": "remove", "path": "/VLAN/Vlan3912"}]
Patch Applier: localhost: key-level removes would empty table VLAN; rewriting to table-level remove. Rewritten patch: [{"op": "remove", "path": "/VLAN"}]
Patch applied successfully.
admin@sonic:~$ redis-cli -n 4 keys "VLAN*"
(empty array)
admin@sonic:~$ show vlan config
Name    VID    Member    Mode
------  -----  --------  ------
```

`vlan.json`:

```
[
  {"op": "remove", "path": "/VLAN/Vlan3911"},
  {"op": "remove", "path": "/VLAN/Vlan3912"}
]
```

```
given patch is not valid because it will result in empty tables which is not allowed in ConfigDb. Table: VLAN
```

```
Patch Applier: localhost: key-level removes would empty table VLAN; rewriting to table-level remove. Rewritten patch: [{"op": "remove", "path": "/VLAN"}]
Patch applied successfully.
```


Verma-Anukul:
@bhouse-nexthop @saiarcot895 @rimunagala @vaibhavhd @bingwang-ms @qiluo-msft 

Please review this change

generic_config_updater/generic_updater.py
```diff
@@ -116,10 +117,30 @@ def apply(self, patch, sort=True, trace_io: Optional[IO] = None):
...
+        # remove first would drop protected child paths (for example
+        # /LOOPBACK_INTERFACE/Loopback0) from the diff used by this check.
         self.logger.log_notice(f"{scope}: validating all JsonPatch operations are permitted on the specified fields")
         self.config_wrapper.validate_field_operation(old_config, target_config)
 
+        # ConfigDB cannot store empty tables. Automation scripts may emit per-key
+        # removes (for example /VLAN/Vlan10) without knowing they delete the last
+        # remaining entries. Only those key-level removes are rewritten to a
+        # table-level remove; other operations are left unchanged.
+        empty_tables = self.config_wrapper.get_empty_tables(target_config)
```
rimunagala, line 131: `PatchApplier` is also used by `ConfigReplacer` (`config replace`). If a replacement target contains `"VLAN": {}`, `generate_patch()` produces per-key removes; this branch rewrites and applies them as `/VLAN`, but `ConfigReplacer` then compares the absent table with the original empty-table target and raises after ConfigDB has already been modified. Before this change, `EmptyTableError` occurred before any writes. Please restrict this rewrite to the explicit `apply-patch` flow, or reject/canonicalize empty replacement targets before applying, and add a `config replace` regression test.

generic_config_updater/gu_common.py
```diff
@@ -435,6 +435,62 @@ def remove_empty_leaf_lists(config):
...
+        return patch
+
+    if path_addressing is None:
+        path_addressing = PathAddressing()
+
+    empty_tables = set(empty_tables)
+    parsed_ops = []
+    ops_by_table = {table: [] for table in empty_tables}
+
+    for operation in patch:
```
rimunagala, line 460: Non-blocking edge: eligibility considers only `path`, but RFC 6902 `move`/`copy` can reference this table through `from`. For example, `remove /VLAN/B`, `copy /VLAN/A -> /OTHER/Y`, then `remove /VLAN/A` is rewritten to remove `/VLAN` before the copy, causing re-simulation to fail. Since GCU accepts `move`/`copy` and the PR states other patch shapes remain unchanged, please skip rewriting tables referenced through `from`, or otherwise preserve those dependencies.

rimunagala (changes requested):
The intended pure key-removal behavior and protected-field validation look correct, and all CI checks pass. Requesting changes for the config replace regression noted inline; the RFC 6902 from-path case is non-blocking.

PR #4863, opened 2026-09-21
nbrshow: fix CLI argument injection via shell=True

pkawatkar14:
## Why I did it
Security review (finding F046, Medium, ADO 39463389) identified that `nbrshow` builds its underlying `arp`/`ip neigh show` command as a **shell string** and executes it with `subprocess.Popen(cmd, shell=True)`. The `--ipaddr`, `--iface`, and `--namespace` CLI arguments are concatenated directly into that string with no sanitization, so an operator (or any automation invoking `nbrshow`) can inject shell metacharacters to run arbitrary commands as the `nbrshow` process user.

Confirmed PoC:
```
nbrshow -4 -ip '127.0.0.1; echo VULN_PROBE_F046 > /tmp/probe_f046 #'
```
This creates `/tmp/probe_f046`, proving arbitrary command execution.

Affected code:
- `ArpShow.__init__` / `NeighShow.__init__`: append `ipaddr`/`iface` directly into a `CMD` string.
- `NbrBase.fetch_nbr_data()`: `subprocess.Popen(cmd, shell=True, ...)`, and wraps the command in `"ip netns exec {} {}".format(namespace, cmd)` when a namespace is given — `namespace` is also unsanitized.

## How I did it
Replaced shell-string command construction with **argv lists**, and dropped `shell=True` from the `subprocess.Popen` call entirely (defaults to `shell=False`):
- `ArpShow.CMD` / `NeighShow.CMD` are now list literals (e.g. `["/usr/sbin/arp", "-n"]`, `["/bin/ip", "-6", "neigh", "show"]`) instead of strings.
- `__init__` builds a local `cmd` list via `.append()` / list concatenation instead of string concatenation.
- `fetch_nbr_data()`'s namespace wrapping is now `["ip", "netns", "exec", namespace] + cmd` instead of a formatted string.
- `subprocess.Popen` is called with the argv list and no `shell` kwarg.

## Justification of the fix
With `shell=False` (the new default) and an argv list, each argument — including `ipaddr`, `iface`, and `namespace` — is passed to `execve()` as a single, discrete argument. Shell metacharacters (`;`, `|`, `` ` ``, `$()`, etc.) in any of these values can never be interpreted as shell syntax, because no shell is invoked to parse the command at all. This closes the injection unconditionally for any input content, without relying on input validation, allowlisting, or escaping (which are easier to get wrong or bypass). Verified no other code in the repo references `ArpShow`, `NeighShow`, or `NbrBase` externally, so changing `CMD` from a string class attribute to a list is safe and has no other call sites to update.

## How to verify it
- Ran the new regression tests locally (`tests/nbrshow_test.py`) — all pass against the fixed code.
- Confirmed the same tests **fail** against the pre-fix code (reverted locally), proving they actually catch the vulnerability and aren't vacuous.
- `flake8` clean on both changed files.
- Manual review: with clean/valid inputs (e.g. `10.0.0.1`, `Ethernet0`), the resulting argv list is identical in content/order to the previous shell string equivalent — no behavior change for well-formed input.

### Test cases added (`tests/nbrshow_test.py`)
- `ArpShow`/`NeighShow` build an argv **list** (not a shell string) for both malicious payloads and clean inputs; the injection payload is preserved as a single literal argv element rather than being split/interpreted.
- `ArpShow` with omitted `ipaddr`/`iface` produces the base command only (no stray empty args).
- `fetch_nbr_data()` calls `subprocess.Popen` with the list unchanged and never passes `shell=True`.
- `fetch_nbr_data()`'s namespace-wrapping branch produces `["ip", "netns", "exec", <namespace>, ...cmd]` even when `namespace` itself contains a malicious payload.

## Which release branch to backport
None — targeting master only unless maintainers request a backport.

## Description for the changelog
Fix CLI argument injection in `nbrshow` by replacing shell-string command execution (`shell=True`) with argv-list execution (`shell=False`).


scripts/nbrshow
```diff
@@ -184,17 +186,18 @@ class NbrBase(object):
...
     HEADER = ['Address', 'MacAddress', 'Iface', 'Vlan']
-    CMD = "/usr/sbin/arp -n "
+    CMD = ["/usr/sbin/arp", "-n"]
 
     def __init__(self, ipaddr, iface, namespace=None):
 
+        cmd = list(self.CMD)
         if ipaddr is not None:
-            self.CMD += ipaddr
+            cmd.append(ipaddr)
```
qiluo-msft: Non-blocking follow-up: `nbrshow -4 --ipaddr=--file` builds `arp -n --file`, selecting ARP import mode instead of display. Confirmed the parser/argv path and the real `arp` file-open attempt locally without modifying neighbors; actual updates require root/CAP_NET_ADMIN and a populated default ethers file. Put legitimate options (including `-i <iface>`) before `--`, then append `ipaddr`, and add a regression case? Separate option-injection issue—the F046 shell fix is sound.

pkawatkar14: Sounds good

Fixed in  ee35cd0 .

Reordered  ArpShow.__init__  so trusted options come before  -- , and the untrusted  ipaddr  comes after:

`cmd += ['-i', iface]   # trusted option, parsed first`
`cmd += ['--', ipaddr]  # '--' ends option parsing; ipaddr is now always positional`

With  --  in place, arp can never reinterpret ipaddr as a flag — confirmed against the real binary:

arp -n -i lo -- --file  → treated as hostname ("Unknown server error")
arp -n --file           → old behavior, tries to open /etc/ethers

 NeighShow / ip neigh show  isn't affected — iproute2 doesn't do getopt-style option/positional permutation, so a dash-prefixed  ipaddr  there just fails prefix parsing instead of switching modes.

Added `test_arpshow_ipaddr_looking_like_an_option_cannot_change_arp_mode` and updated the two existing ArpShow tests for the new argv order. Full tests/nbrshow_test.py (8/8) passes, flake8 clean.

qiluo-msft (approved):

